PR #11, opened 2018-03-01
Adding a "futures" document

arschles:
cc/ @bketelsen - ~this is pretty much a copy/paste job from the [gist](https://gist.github.com/arschles/2d5eb22b04e4056cbbe8aa4f58786808) let's work off this from now on~

edit: I simplified this a lot from the above gist. trying to keep the initial scope smaller

### DIFF
--- a/DETAILS.md
+++ b/DETAILS.md
@@ -1,0 +1,115 @@
+# Details on Athens functionality.
+
+This document describes the various use cases that Athens supports or will
+support. Not everything here is implemented.
+
+[@bketelsen](https://github.com/bketelsen) and I spoke at length on Slack about
+how the Athens registry should work. I'm putting our discussion down on "paper"
+and adding some of my own detail (and some opinions) so we can grow this into
+a definitive document.
+
+# Where Will the Code Come From?
+
+Here's what was discussed on slack + various other forums:
+
+## Travis or Other CI systems (Webhook)
+
+After code passes tests, the publisher can hit a webhook to tell the registry to fetch
+code for the given tag from the given repository. The webhook will then do the following:
+
+* Check if the repository is already "known" (the owner of the repository will need to register it prior) - Fail if it isn't
+* Check if the given tag already exists (we want versions to be immutable) - Fail if it does
+* Download the code from the given tag and store it in a CDN
+
+We may need to require users to be authorized to hit the webhooks.
+
+## Manual Uploads
+
+This works similarly to the webhook flow, except a user would upload their module --
+including code -- manually to the registry. When they upload, they'll need to
+specify the module name and version, and obviously include the zip file with their
+source code.
+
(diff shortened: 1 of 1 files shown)


DETAILS.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Manual Uploads
+
+This works similarly to the webhook flow, except a user would upload their module --
+including code -- manually to the registry. When they upload, they'll need to
+specify the module name and version, and obviously include the zip file with their
+source code.
+
+The zip file will need to have the `go.mod` file in it, and the server will
+parse it out.
```
bketelsen: this may need to be two separate files uploaded - need to verify spec.

DETAILS.md
```diff
@@ -0,0 +1,115 @@
...
+local filesystem. The workflow will be the following:
+
+```console
+$ athens run local-proxy --root /someplace --vcs git
+$ GOPROXY=localhost:1234 go get my/module
+```
+
+The `go get` command executes the download protocol on the local proxy, which simply mirrors
+to the files in `/someplace`. In the case of `my/module`, it mirrors to `/someplace/my/module`, requires the `go.mod` to be there, and uses `git` (specified by `--vcs`) to list
+tags, etc...
```
bketelsen: vcs might not be necessary if we're just caching locally fetched artifacts.  let's discuss more.

DETAILS.md
```diff
@@ -0,0 +1,115 @@
...
+
+We've been able to specify "vanity" names (names that don't match the repository the
+code lives in) for our packages for a while not. https://gopkg.in was one of the earliest
+systems that let us do this. It "redirected" (using the `<meta>` tag) the `go get`
+requests against it to the appropriate repository in Github. The `go get` tool would
+then follow the redirect and use `git` to fetch the code.
+
+Now that there's a custom HTTP-based download protocol, we can use the same rename-by-redirecting mechanism, but with the following additions:
+
+* Allow users to specify their own "vanity" module names for a given codebase - We should probably allow 0 or more vanity names attached to an existing repository - We should allow users to turn off the repository-based proxy name (i.e. the `github.com`
```
bketelsen: s/0 or more/0 or 1
unless people are masochistic enough to want to publish the same package with different import names.

DETAILS.md
arschles: I'm starting to think that the local proxy isn't even valuable, if you can set `GOPROXY` to a `file:///` URL. need to verify

DETAILS.md
arschles: so this would only be for an upload protocol. for the download protocol, the `go.mod` definitely needs to be available outside of the zip

DETAILS.md
arschles: yea, good point. let's limit to 1

PR #13, opened 2018-03-01
athens: update to buffalo 0.11.0, add cors and csrf support

bketelsen:
* Update buffalo to 0.11.0
* Add cors support
* Enable csrf support


bketelsen:
fixes #12

Issue #17, opened 2018-03-01
docs: add documentation endpoint


bketelsen:
add documentation endpoint serving nicely formatted docs in markdown

PR #18, opened 2018-03-01
docs: hugo site generator for docs endpoint

bketelsen:
This PR adds a docs folder with content processed by hugo.
Future PR will move that into a place that it can be served by buffalo (or we serve it from a caddy endpoint, which may be my pref)


bketelsen:
fixes #17

PR #19, opened 2018-03-01
Adding note on immutable module names

arschles:
@bketelsen I added this rule to reduce the chance of a left-pad like incident with Athens. Not sure if it's too strict. My rationale:

1. I don't want the same code to be available with these two commands to download the same code, if someone adds a vanity name of `my/vanity/name` to their `github.com/my/thing` repository:

```
$ GOPROXY=gomods.io go get github.com/my/thing
$ go get gomods.io/my/vanity/name
```

2. But I also don't want this (or the opposite) to fail ever:

```
$ GOPROXY=gomods.io go get github.com/my/thing
$ # user changes to vanity name my/vanity/name
$ go get my/vanity/name
$ # user changes back to repo name
$ go get my/vanity/name # this fails!
```


bketelsen:
the go tool should protect us here.  If a package declares an import directive (gomods.io/my/package) then it won't fetch github.com/my/package because of the import declaration.

arschles:
but what if it doesn't declare an import directive? I'd like to protect against the case where `github.com/my/module` and `my/module` are the same code. now, the versions can skew in transitive dependencies and it's just confusing to users

bketelsen:
i think the answer is that if they register a module with the registry under a vanity domain, then we have to verify the go.mod file contains the vanity domain.

bketelsen:
or more clearly, if they register any package with any import path, we verify the import path in the mod file matches what they registered.

arschles:
100% agree. I was totally missing the point here. Will update this PR

arschles:
This is gonna conflict with #27 because the `DETAILS.md` is removed. Rebasing won't be a big deal, and I'll handle it

arschles:
But I'd like to get #27 in first

arschles:
Now that #27 is in and conflicts are resolved, I am merging this

PR #22, opened 2018-03-02
Adding a Netlify config file

arschles:
See https://deploy-preview-22--gomods-athens.netlify.com/ for an example


bketelsen:
niiice! lgtm

PR #23, opened 2018-03-02
athens: basic github oauth

bketelsen:
* adds github oauth
* nothing done with the oauth yet.


actions/goget.go
```diff
@@ -0,0 +1,30 @@
...
+
+// stubbed for now, look up package stuff
+func goGetMeta(c buffalo.Context) error {
+	sp, err := getStandardParams(c)
+	if err != nil {
+		return err
+	}
+	meta := fmt.Sprintf("<!DOCTYPE html><meta name='go-import' content='%s/%s mod https://gomods.io/%s/%s", sp.baseURL, sp.module, sp.baseURL, sp.module)
+	_, err = c.Response().Write([]byte(meta))
+	return err
```
arschles: this could be a little bit cleaner to use a template: 

```go
c.Set("baseURL", sp.baseURL)
c.Set("module", sp.module)
tpl := `<!DOCTYPE html>
<meta name="go-import" content=<%= baseURL %>/%<= module %> mod https://gomods.io/<%= baseURL %>/%<= module %>>`
c.Render(200, r.String(tpl))
```

Might even be worth moving the HTML out to a template file.

bketelsen: agreed - this was my little test stub.

PR #25, opened 2018-03-02
Adding an API for fetching module data from a CDN

arschles:
This is a proposal for the CDN API for doing all read operations on packages.

The write API can be an implementation of `(./pkg/storage).Saver`, so I haven't specified it here


pkg/cdn/getter.go
```diff
@@ -0,0 +1,12 @@
+package cdn
+
+// Getter gets the details about a given baseURL/module at vsn and returns
+// the base URL of the module metadata & content:
+//
+// - {baseURL}/{module}/@v/list
```
michalpristas: thinking about "{baseURL}/{module}/@v/" having list as default behavior without explicitly specifying

bketelsen: can't - that's not the spec from Russ's document

michalpristas: need to go through the rest of them finally. thanks

PR #34, opened 2018-03-07
athens: add userStore interface and mongoDB implementation

bketelsen:


actions/app.go
```diff
@@ -44,13 +47,19 @@ func App() *buffalo.App {
 	mem := memory.New()
 	storageReader := mem
 	storageWriter := mem
+	mgoStore := mongo.NewMongoUserStore("127.0.0.1:27017")
+	err := mgoStore.Connect()
+	if err != nil {
+		panic(err)
```
carolynvs: Are we doing any error handling yet elsewhere? Or is panic the norm?

actions/auth.go
```diff
@@ -5,24 +5,39 @@ import (
...
+func authCallback(store user.Store) func(c buffalo.Context) error {
+	return func(c buffalo.Context) error {
+		usr, err := gothic.CompleteUserAuth(c.Response(), c.Request())
+		if err != nil {
+			return c.Error(401, err)
+		}
+		// Do something with the user, maybe register them/sign them in
+		var u *user.User
+		u, err = store.Get(usr.UserID, usr.Provider)
+		if err != nil {
```
carolynvs: It doesn't look like we are handling any other error here than ErrNotFound, no?

pkg/user/mongo/user.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Save adds a user to the Mongo Store
+func (m *MongoUserStore) Save(u *user.User) error {
+	c := m.s.DB(m.d).C(m.c)
+	return c.Insert(u)
+}
+
+// Update updates a user in the Mongo Store
+func (m *MongoUserStore) Update(*user.User) error {
+	return nil
```
carolynvs: I suggest implementing something here, even if it's just returning "not implemented". That way when we wonder why it's not working, it will be a bit more obvious. 😀

.travis.yml
```diff
@@ -1,6 +1,8 @@
...
-go:
-  - "1.9"
-  - "1.10.x"
-script: go test ./...
+services:
+  - mongodb
+language: go
+install: false
+go:
+  - "1.9"
```
carolynvs: go 1.10 or bust! 😀

pkg/user/mongo/user_test.go
```diff
@@ -0,0 +1,57 @@
...
+			t.FailNow()
+		}
+		_, err := m.s.DB(m.d).C(m.c).RemoveAll(nil)
+		if err != nil {
+			t.Error(err)
+			t.FailNow()
+		}
+		return m
+	}
+	return nil
```
carolynvs: Should this return `m`?

actions/app.go
```diff
@@ -44,13 +47,19 @@ func App() *buffalo.App {
 	mem := memory.New()
 	storageReader := mem
 	storageWriter := mem
+	mgoStore := mongo.NewMongoUserStore("127.0.0.1:27017")
```
michalpristas: configurable?

pkg/user/mongo/user_test.go
```diff
@@ -0,0 +1,57 @@
...
+	g.Email = "goth@goth.user"
+	g.ExpiresAt = time.Now()
+	g.FirstName = "Goth"
+	g.LastName = "User"
+	g.Location = "Buffalo, NY"
+	g.Name = "Goth User"
+	g.NickName = "guser"
+	g.Provider = "github"
+	g.RefreshToken = "refreshtoken"
+	g.UserID = "012345678"
```
michalpristas: are all those field needed, don't know about github but microsoftonline does not provide them all

actions/app.go
bketelsen: in cases where the application can't start (like because the DB is unavailable) a panic is appropriate.

actions/auth.go
bketelsen: good catch!

pkg/user/mongo/user.go
bketelsen: done!

.travis.yml
bketelsen: agreed, I'll remove 1.9

pkg/user/mongo/user_test.go
bketelsen: yes - refactored to make more sense

pkg/user/mongo/user_test.go
bketelsen: technically no, but these are the base sets of fields from goth.

actions/app.go
bketelsen: eventually but I'd like to add that in a later PR - setting up all the different storage options

bketelsen:
PTAL

PR #37, opened 2018-03-08
Adding note on test dependency

arschles:


arschles:
This is important for all Athens developers. Merging

PR #53, opened 2018-03-13
[WIP] storage changed to server readerCloser

### LINKED ISSUE #38: Disk storage: Do not read entire source zipfiles into memory
After #20, there will be a disk storage system, which is likely to store large modules. The storage interface's `Get` function returns `[]byte`s for both the `go.mod` file and the zipped-up module source code. Of course, that's not ideal because module zips would need to be read into memory.

The storage interface `Get` method should be changed to return `io.ReadCloser`s instead of `[]byte`s for the zip. I don't think we'll need to do the same for `go.mod` files for now, since those will be relatively small.

michalpristas:
Fixes #38

### DIFF
--- a/pkg/storage/disk/disk_test.go
+++ b/pkg/storage/disk/disk_test.go
@@ -1,6 +1,7 @@
 package disk
 
 import (
+	"io/ioutil"
 	"path/filepath"
 )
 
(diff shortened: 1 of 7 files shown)


michalpristas:
I think I caught all occurrences and am closing reader everywhere, but the second pair of eyes would be nice

pkg/storage/disk/disk_test.go
```diff
@@ -22,11 +23,14 @@ func (d *DiskTests) TestGetSaveListRoundTrip() {
 	retVersion := listedVersions[0]
 	r.Equal(version, retVersion)
 	gotten, err := d.storage.Get(baseURL, module, version)
+	defer gotten.Zip.Close()
```
bketelsen: if gotten is nil, this will panic - defer after checking the error

PR #55, opened 2018-03-15
remove on err

michalpristas:


actions/fetch.go
```diff
@@ -30,6 +31,7 @@ func fetchHandler(store storage.Saver) func(c buffalo.Context) error {
 
 		path, err := git.DownloadRepo()
 		if err != nil {
+			os.Remove(path)
```
arschles: the most microscopic of nits: could you log here if removing failed?

Issue #46, opened 2018-03-09
Split proxy storage, CDN storage, and actions between proxy and registry


arschles:
The proxy and the registry are different enough that much of the packages and codebase needs to be split apart. For example:

- The registry needs to create a CDN getter
- The registry should not declare handlers for the metadata/code download paths
    - It should instead use a meta redirect to a CDN location
- The proxy needs to create storage
- The proxy needs to declare handlers for metadata/code download paths

Ref #44

PR #56, opened 2018-03-15
Splitting actions into registry and proxy

arschles:
This patch logically splits the registry and proxy actions apart, and forces you to configure Athens as a proxy or a registry, but not both.

Fixes https://github.com/gomods/athens/issues/46

### DIFF
--- a/actions/app.go
+++ b/actions/app.go
@@ -12,17 +12,21 @@ import (
 	"github.com/gobuffalo/envy"
 	"github.com/gobuffalo/packr"
 	"github.com/gomods/athens/pkg/user"
-	"github.com/gomods/athens/pkg/user/mongo"
 	"github.com/rs/cors"
 	"github.com/unrolled/secure"
-
-	"github.com/markbates/goth/gothic"
 )
 
 // ENV is used to help switch settings based on where the
 // application is being run. Default is "development".
 var ENV = envy.Get("GO_ENV", "development")
+
+// MODE identifies whether athens is running in proxy or registry mode.
+//
+// valid values are "proxy" or "registry"
+var MODE = envy.Get("ATHENS_MODE", "proxy")
 var app *buffalo.App
+
+// T is the translator to use
 var T *i18n.Translator
 
 var gopath string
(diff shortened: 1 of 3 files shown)


actions/app.go
```diff
@@ -88,25 +79,22 @@ func App() *buffalo.App {
...
-		app.GET("/{base_url:.+}/{module}/@v/{version}.zip", versionZipHandler(storage))
-		app.POST("/admin/upload/{base_url:[a-zA-Z./]+}/{module}/{version}", uploadHandler(storage))
-		app.POST("/admin/fetch/{base_url:[a-zA-Z./]+}/{owner}/{repo}/{ref}/{version}", fetchHandler(storage))
-
-		auth := app.Group("/auth")
-		auth.GET("/{provider}", buffalo.WrapHandlerFunc(gothic.BeginAuthHandler))
-		auth.GET("/{provider}/callback", authCallback(mgoStore))
-		//	app.GET("/{base_url:.+}/{module}", homeHandler)
-		// serve files from the public directory:
+		if MODE == "proxy" {
```
michalpristas: we have default set to proxy proxy should be also when invalid option is passed.
i would do explicit check for registry if(registry) else ( is proxy does not matter on value)

actions/app_registry.go
```diff
@@ -0,0 +1,22 @@
...
+func addRegistryRoutes(app *buffalo.App) error {
+	cdnGetter := newCDNGetter()
+	mgoStore := mongo.NewMongoUserStore("127.0.0.1:27017")
+	if err := mgoStore.Connect(); err != nil {
+		return err
+	}
+	// serve go-get requests
+	app.Use(GoGet(cdnGetter))
+	auth := app.Group("/auth")
+	auth.GET("/{provider}", buffalo.WrapHandlerFunc(gothic.BeginAuthHandler))
```
michalpristas: shouldnt this be in opposite order?
auth + store (*Handler) - for registry
goGet - for proxy (without auth) 
or is auth needed for getting sources? please ping me to explain. my brain is in friday mode

arschles: not sure what you mean - the middleware intercepts all requests that came from `vgo get` - those can't require authentication

actions/app.go
arschles: done - I added a check for unsupported mode

arschles:
I am going to merge this now and make further modifications in follow-ups

PR #60, opened 2018-03-16
Remove the base URL

arschles:
~**Note: wait for #64 to go in before merging this. It'll need a rebase before the merge**~ (#64 is in and this has been rebased)

I originally added a `baseURL` to our paths to follow exactly the download protocol in https://research.swtch.com/vgo-module. In reality, there's no need for it because modules are always referred to by their fully qualified names - like `swtch.com/testmod` - and their version.

- Code before this patch would parse `swtch.com` into baseURL and `testmod` into module, and then immediately concatenate them.
- This patch captures the entire `swtch.com/testmod` in one path parameter and then works with that as the module name in all further processing.

TODO:

- [x] Change the `build` target of the `Makefile` to run `buffalo build` instead of just `go build`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,9 @@
+build:
+	go build .
+
+run: build
+	./athens
+	
 cli: 
 	go build -o athens ./cmd/cli
 
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,10 @@
 build:
-	go build .
+	buffalo build
 
 run: build
 	./athens
 	
-cli: 
+cli:
 	go build -o athens ./cmd/cli
 
 docs:
--- a/actions/app.go
+++ b/actions/app.go
@@ -79,6 +79,8 @@ func App() *buffalo.App {
 		}
 		app.Use(T.Middleware())
 
+		app.GET("/", homeHandler)
+
 		if MODE == "proxy" {
 			log.Printf("starting athens in proxy mode")
 			if err := addProxyRoutes(app); err != nil {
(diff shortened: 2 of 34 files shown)


michalpristas:
if #64 goes in first there will be conflicts/build not passing. 
if this goes in first it's vice-versa.

michalpristas:
one more thought on naming. what about changing module to modulePath as it is a combination of basepath and moduleName.
Just a very minor and highly ignorable thing but it can help understanding things here for somebody who just read whitepapers and is new to project.

Makefile
```diff
@@ -1,3 +1,9 @@
+build:
+	go build .
```
makhov: It builds only go binary but not assets. Web UI will not works fine

arschles: Good catch - thanks @makhov. I'll need to change this to a `buffalo build`

arschles:
@michalpristas I'm totally ok rebasing this. not a single problem getting #64 in before this

Makefile
arschles: done in https://github.com/gomods/athens/pull/60/commits/74e74ef65fa374ddc82204ca016803772add0300

arschles:
This is a pretty important foundational change to get in as soon as we can, so it doesn't keep conflicting and/or cause conflicts in other PRs down the line.

Right now, it'll cause conflicts in #74 (cc/ @lcd1232 - I'll help work those out) and #62 (that's for me to work out 😄) and that's minimal damage, so I'm going to merge this now

Issue #50, opened 2018-03-09
Implement a CDN saver


arschles:
This would complement the CDN getter. The getter just gets a redirect location for a module; the setter should upload/update:

- The versions list
- The go.mod file
- The version metadata
- The source zip

It should also enter a redirect location (that the getter can fetch) into the CDN metadata store.

Make sure to put the entire operation (enter into CDN metadata store + upload) in a critical section. Ideas:

- Lock/lease the directory on the CDN
    - Azure blob storage gives you versioned blobs and also lock semantics for blobs (see [here](https://azure.microsoft.com/en-us/blog/managing-concurrency-in-microsoft-azure-storage-2/))
    - Google Cloud Storage gives you generational versioning and preconditions (see [here](https://cloud.google.com/storage/docs/generations-preconditions) and [here](https://cloud.google.com/storage/docs/gsutil/addlhelp/ObjectVersioningandConcurrencyControl))
    - Quick research on S3 indicates that blobs (buckets?) have versions. Need more to research whether the API includes preconditions for generations

arschles:
the [storage saver](https://github.com/gomods/athens/blob/master/pkg/storage/saver.go) interface exists and it's implemented in all our storage drivers. closing

PR #61, opened 2018-03-16
Adding a CDN saver interface

arschles:
Ref https://github.com/gomods/athens/issues/50

Does not implement one


arschles:
Merging this, it's not controversial

PR #64, opened 2018-03-17
Implemented backing mongo storage for modules

### LINKED ISSUE #41: Implement a MongoDB Storage Driver
The storage drivers at ./pkg/storage are used by the proxy to store and return metadata and source for a given module. We should implement a driver for MongoDB. By using the MongoDB API, we can connect the proxy to CosmosDB

michalpristas:
Fixes #41


arschles:
@michalpristas I'm not ignoring this, I swear 😄 

I should be able to finish the review tomorrow. Hope you had a great weekend!

pkg/storage/mongo/all_test.go
```diff
@@ -0,0 +1,43 @@
...
+	zip = []byte("456")
+)
+
+type MongoTests struct {
+	suite.Suite
+	storage storage.StorageConnector
+}
+
+func (d *MongoTests) SetupTest() {
+	store := NewMongoStorage("mongodb://127.0.0.1:27017")
```
bketelsen: Is this worth configuring, or is there a presumption that mongo will be available everywhere that wants to run unit tests?
Perhaps guard with a build flag, if that's the case?

bketelsen: also understand the irony of this comment, given it was probably copy/paste from my mongo impl :)

pkg/storage/not_found_err.go
```diff
@@ -24,7 +24,7 @@ func (e ErrVersionNotFound) Error() string {
 }
 
 func IsNotFoundError(err error) bool {
```
bketelsen: Nice!

pkg/storage/mongo/all_test.go
bketelsen: This is not a blocking change - so consider this a LGTM.

bketelsen:
LGTM after rebase.

pkg/storage/mongo/all_test.go
arschles: fwiw, I'm ok with keeping this for now since we have a docker-compose file in place to run mongo locally - we should consider adding a build flag when our corpus of tests grows

pkg/storage/mongo/mongo.go
```diff
@@ -0,0 +1,41 @@
...
+type MongoModuleStore struct {
+	s   *mgo.Session
+	d   string // database
+	c   string // collection
+	url string
+}
+
+// NewMongoStorage  returns an unconnected Mongo Module Storage
+// that satisfies the Storage interface.  You must call
+// Connect() on the returned store before using it.
```
arschles: we should indicate that this is an implementation of `StorageConnector`

I can do this in a follow-up 😄

PR #67, opened 2018-03-18
Fix assets manifest path

makhov:
Buffalo tries to find `manifest.json` in assetBox path:
```
// github.com/gobuffalo/buffalo/render/template.assetPath
manifest, err := s.AssetsBox.MustString("manifest.json")

// actions/render.go
var assetsBox = packr.NewBox("../public")
```

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -58,7 +58,7 @@ module.exports = {
       debug: false
     }),
     new ManifestPlugin({
-      fileName: "manifest.json"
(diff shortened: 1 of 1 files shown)


arschles:
@makhov can you let me know what this fixes, specifically?

makhov:
Before:
before

After:
after

arschles:
Ah, I see. Thanks for setting me straight @makhov!

PR #70, opened 2018-03-19
Use afero for as abstraction layer for FS

### LINKED ISSUE #65: Use spf13/afero for InMemory and Disk Storage
It would be possible to use the same code for both backends and automatically enable others i.e. sftp
https://github.com/spf13/afero

marpio:
Not sure why spf13/cobra has been added...

Fixes #65


marpio:
@michalpristas Thanks for the description edit!

PR #71, opened 2018-03-20
Added tests for parsers

adelowo:
I had to move `re = regexp.MustCompile(moduleRegexp)` to a package level variable so as to prevent `MustCompile` from being called all the time.  

Fixes #52

PR #72, opened 2018-03-20
gomod parsers tests

### LINKED ISSUE #52: Tests for gomod parsers
After #2 is done
Tests for parsers need to be implemented

makhov:
Some tests added.
Fixes #52


adelowo:
Oh... I previously PR'd #71 earlier anyways

bketelsen:
oops - do you want to take a look and see if there's value in combining the two test PR's into one?

adelowo:
I am currently on mobile and can't even view the changes conveniently...  Would take a look as soon as I get home - if this isn't merged by then.

makhov:
Wow.
Let's combine the two PR's.
@adelowo found a lot of good negative cases, noticed unnecessary regex compilation and so on, but I think my tests a little bit simpler to read and understand.

I'll commit some changes soon (and think about zip files).

makhov:
Hope I made it better (but not sure)

adelowo:
Looks good enough.... I think this is much more better than #71 .
It is just left for @arschles and @michalpristas to make a choice

arschles:
I'll need to take a look at this and #71 tomorrow before I can throw in my 2 cents 😄

pkg/gomod/bytes/parser_test.go
```diff
@@ -0,0 +1,33 @@
...
+	a := assert.New(t)
+
+	var testCases = []struct {
+		content     []byte
+		expected    string
+		expectedErr error
+	}{
+		{[]byte(`module "my/thing"`), "my/thing", nil},
+		{[]byte(`module "my/thing2`), "", parser.ErrNotFound},
+		{[]byte(`foobar`), "", parser.ErrNotFound},
```
michalpristas: try moultiple modules in one file, first should be selected, don't know how vgo behaves in this case

pkg/gomod/file/parser_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+func TestFileParser_ModuleName(t *testing.T) {
+	a := assert.New(t)
+
+	var testCases = []struct {
+		file        string
+		expected    string
+		expectedErr error
+	}{
+		{"testdata/go.0.mod", "my/thing", nil},
```
michalpristas: 💯 for testdata approach

pkg/gomod/file/testdata/go.0.mod
```diff
@@ -0,0 +1,5 @@
+module "my/thing"
```
michalpristas: maybe one more with different order, parser should look it up

pkg/gomod/parser.go
```diff
@@ -1,47 +1,51 @@
...
+)
+
+const (
+	moduleRegexp = `module "([\w\.@:%_\+-.~#?&]+/.+)"`
+)
+
+var (
+	re = regexp.MustCompile(moduleRegexp)
+	// ErrNotFound describe an error when can't find module name
+	ErrNotFound = errors.New("name not found")
```
michalpristas: 👍

pkg/gomod/file/testdata/go.0.mod
makhov: Fixed in https://github.com/gomods/athens/pull/72/files#diff-ae9bcd418668c90e559d1c8331670d4c

pkg/gomod/bytes/parser_test.go
makhov: Thanks! Added in https://github.com/gomods/athens/pull/72/files#diff-6d0b6ed72e589e371bc00e27c79d5d9cR28

michalpristas:
after last commits, perfect

makhov:
And thanks to @adelowo for his work!

Issue #84, opened 2018-03-22
Proposal: build separate binaries for proxy & registry


arschles:
We have a growing set of logic that's basically `if proxy { ... } else { ... }` which will be growing over time. I believe we're better served building separate binaries (i.e. having two `main()` functions), one for the proxy and one for the registry (and others for components to come if necessary). We already have a `cmd` directory, so my proposal is to create a `cmd/proxy` and a `cmd/registry`, and to house buffalo app inside of each directory. Common code should still be shared via the packages inside `pkg/`. Finally, I want to keep proxy, registry, and other code inside of a single repository.

bketelsen:
how could we keep the existing "buffalo dev" workflow?

bketelsen:
we can experiment and see how much "buffalo dev" is tied to the layout of the project.

arschles:
My original thought was to put a full buffalo application into each of `cmd/proxy` and `cmd/registry`, so you would run `buffalo dev` inside of either directory. This fits nicely into the standard buffalo tooling without any modifications.

The only thing that I'm sure of with `buffalo dev` itself is that it needs the `.buffalo.dev.yml` to run, and I think it needs a `main`. I know you and @markbates figured some things out and I'm not caught up on that.

I'd just advocate for keeping plain `buffalo dev` without any additional setup (besides DB setup) on the machine. In other words, I'd like to keep the workflow this:


```console
# do this once
$ git clone git@github.com:gomods/athens.git
$ cd athens
$ docker-compose up -d
$ buffalo db migrate up

# now do this as many times as you want
$ buffalo dev
```

(note - our README describes almost all of this flow, except for the `buffalo db migrate up` part)

markbates:
I agree with Aaron. Also, you could make a grift task in Athens that did all the docker, migration, seed, etc… then no matter happens the docks say run `buffalo t setup` the first tome. :)

arschles:
ooooh, the grift for all the migration etc... is an _awesome_ idea. thanks @markbates!

arschles:
I'll take this one

bketelsen:
See #87, close together

michalpristas:
#91 fixes this, closing

Issue #87, opened 2018-03-23
break out different targets into folders in the monorepo


bketelsen:
Add Olympus Buffalo App
Add Athens Buffalo App
Move Athens CLI App
Docs (Athens)
Docs (Olympus)
Docs (cli)

michalpristas:
#91 fixes this, closing

PR #91, opened 2018-03-24
Proposal: move buffalo app to cmd

michalpristas:
Just to keep root of the repo clear, executables in cmd (cli, webs) shared code in /pkg

If this is not ok, i will abandon, i don't need to have it there, but it might be ok if we go with one repo for multiple buffalo apps (either olympus or registry)

Fixes: #108


michalpristas:
added olympus app. on top of that

arschles:
Ref https://github.com/gomods/athens/issues/87

@michalpristas this fixes https://github.com/gomods/athens/issues/84, right?

bketelsen:
merging this so we can move ahead without ugly conflicts.

PR #109, opened 2018-03-25
fixed git fetcher

michalpristas:
pkg/repo/github

succesfuly uses github web API to download specified tag of owner/repo, stores it into /tmp
then module zip generation flow can proceed with this tmp folder


pkg/repo/github/github.go
```diff
@@ -21,82 +21,90 @@ const (
...
 	tr := tar.NewReader(gzr)
+	var dirName string
 
+fileLoop:
 	for {
 		hdr, err := tr.Next()
 		switch {
 		case err == io.EOF:
-			break
+			break fileLoop
```
arschles: can you try to refactor this to avoid the goto? they always lead to go code that's tough to follow

pkg/repo/github/github_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"os"
+	"testing"
+)
+
+func Test_Download(t *testing.T) {
+	o := "bketelsen"
+	r := "captainhook"
+	v := "v0.1.8"
```
arschles: could you rename these to more descriptive? maybe `org`, `repo` and `version`

michalpristas:
updates @arschles

michalpristas:
merging this so i can do something today based on that

PR #110, opened 2018-03-25
fix golint issues

### LINKED ISSUE #90: Fix golint issues
At the moment, golint $(go list ./...) reports quite a few issues (missing  comments and stutters).
We should run golint with -set_exit_status flag after they are fixed - see #85

marpio:
Fixes #90

### DIFF
--- a/actions/storage.go
+++ b/actions/storage.go
@@ -31,7 +31,7 @@ func newStorage() (storage.Backend, error) {
 		if err != nil {
 			return nil, fmt.Errorf("missing mongo URL (%s)", err)
 		}
-		return mongo.NewMongoStorage(mongoURI), nil
+		return mongo.NewStorage(mongoURI), nil
 	default:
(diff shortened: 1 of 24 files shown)


actions/storage.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/spf13/afero"
 )
 
-func newStorage() (storage.Storage, error) {
+func newStorage() (storage.Backend, error) {
```
michalpristas: minor: i understand that linter will complain about package and interface name being the same, but backend does not make sense to me. Client, GetterSaver, Provider non of them are good tbh but i don;t like backend for some reason

marpio: I wasn't sure myself. Service maybe? but it's very general...

michalpristas: what about driver? makes sense?

arschles: Could we move this decision to a follow-up?

michalpristas:
naming was a minor thing, not blocker

Issue #88, opened 2018-03-23
Add Open Tracing support


bketelsen:
[ ] to all binaries
[ ] add OT Collector to the docker-compose file.

arschles:
Closed by https://github.com/gomods/athens/pull/112

robjloranger:
Can we close this issue?

michalpristas:
@arschles closing this as per previous comment mentioning PR

PR #112, opened 2018-03-26
Feature: Add OpenTracing

bketelsen:
This change adds global OpenTracing support.  See the home.go HomeHandler for an example of how to use it.


.drone.yml
```diff
@@ -0,0 +1,18 @@
...
+      - sleep 15
+  build:
+    image: golang:1.10
+    commands:
+      - go test ./...
+
+services:
+  mongo:
+    image: mongo:3.0
+    command: [ --smallfiles ]
```
arschles: can you add a badge or link to where the drone page is? even better would be to add PR statuses from drone if possible

Gopkg.lock
```diff
@@ -498,6 +571,37 @@
   revision = "12b6f73e6084dad08a7c6e575284b177ecafbc71"
   version = "v1.2.1"
 
+[[projects]]
+<<<<<<< HEAD
```
arschles: can you fix these merge conflicts?

docker-compose.yml
```diff
@@ -1,15 +1,27 @@
...
+  mysql:
+    image: bitnami/mysql:5.7.21-r7
+    ports:
+      - "3306:3306"
+    environment:
+      - "ALLOW_EMPTY_PASSWORD=yes"
+      - "MYSQL_USER=vgp"
+      - "MYSQL_PASSWORD=vgp"
+      - "MYSQL_DATABASE=vgoprox"
+  jaeger:
```
arschles: 🎉

.drone.yml
bketelsen: Yes but it has to be enabled before I can add them so separate PR.

Gopkg.lock
bketelsen: fixed - d'oh!

bketelsen:
PTAL

bketelsen:
Fixes #88

arschles:
@bketelsen FYI if you put the `Fixes #123` into the OP it'll auto-close the issue on merge

PR #114, opened 2018-03-30
Generic vgo fetcher

michalpristas:
Generic (service invariant) fetcher of go sources.
As vgo tool prepares zip info and mod we take advantage of this and return path which has those file already prepared. 
~~On top of that, we need to make github fetcher to prepare those files. Will be solved in separate PR, as this is not used anywhere.~~

the test will fail, I modified my local .netrc file to contain login and token so i'm not throttled by api.github.com rates.

Followup after this PR: agree how to approach rate limiting with github #119

### DIFF
--- a/pkg/repo/fetcher.go
+++ b/pkg/repo/fetcher.go
@@ -1,0 +1,7 @@
+package repo
+
+// Fetcher is interface used for fetching sources from VCS or Services
+type Fetcher interface {
+	Fetch() (string, error)
+	Clear() error
+}
--- a/pkg/repo/generic/goget.go
+++ b/pkg/repo/generic/goget.go
@@ -1,0 +1,143 @@
+package generics
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/gomods/athens/pkg/repo"
+)
+
+const (
+	tmpRepoDir = "%s-%s" // owner-repo-ref
+)
+
+type genericFetcher struct {
+	repoURI string
+	version string
+	dirName string
+}
+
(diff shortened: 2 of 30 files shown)


michalpristas:
for now generic fetcher is only correct solution. 
for usecase with downloading zip we need to (re-)consider 
- multiple go.mod files in one repo e.g (owner/repo/cli/go.mod owner/repo/web/go.mod) 
- v2 (or any other vX) folder
- something else which I cannot think of right now because I had too many cakes today

pkg/repo/fetcher.go
```diff
@@ -0,0 +1,7 @@
+package repo
+
+// Fetcher is interface used for fetching sources from VCS or Services
+type Fetcher interface {
+	Fetch() (string, error)
+	Clear() error
+}
```
arschles: :+1:

pkg/repo/github/github.go
```diff
@@ -1,60 +1,137 @@
+// Package github is working version only
+// We need to (re-)consider
+// - multiple go.mod files in one repo e.g (owner/repo/cli/go.mod owner/repo/web/go.mod)
+// - v2 (or any other vX) folder
+// - something else which I cannot think of right now because I had too many cakes today
```
arschles: 😆

pkg/repo/generic/goget.go
```diff
@@ -0,0 +1,162 @@
+package generics
```
arschles: do we need this `generics` package for a specific reason? if not, I think it would be a better idea to put `func NewGenericFetcher` directly in the `repo` package

michalpristas: My idea was to have this generic one and then few specific like github which is unusable right now :D then in repo package i would like to have factory which based on module name tries specific.

But i think neither of this is a dis for moving it one level up

PR #121, opened 2018-04-05
added base-url option for cli

chrispickard:


cmd/cli/upload.go
```diff
@@ -58,8 +60,9 @@ func upload(c *uploadCmd) func(*cobra.Command, []string) error {
 			return fmt.Errorf("couldn't make zip (%s)", err)
 		}
 
+		uploadEndpoint := c.baseUrl + "/%s/%s"
```
michalpristas: consider using path.Joint to avoid having double slash

cmd/cli/upload.go
```diff
@@ -18,6 +18,7 @@ import (
 type uploadCmd struct {
 	moduleName string
 	version    string
+	baseUrl    string
```
michalpristas: rename to baseURL to get rid of golint complains

PR #123, opened 2018-04-05
replace fs operations with afero

### LINKED ISSUE #118: Replace Filesystem operations with Afero
Right now we use standard library for walking through filesystem as well as for modifications/reads/writes... 
As Karrick pointed out the standard library has some flaws when walk fn is used on Windows. 
Afero v0.10.0 provides full Windows compatibility

Goal is to replace all occurrences of Walk, Remove, Mkdir, Tmp.... operations.

marpio:
Fixes #70 
Fixes #118


pkg/gomod/file/parser_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"os"
 	"testing"
 
+	"github.com/spf13/afero"
```
michalpristas: can you group these together to be more compact?

pkg/gomod/zip/parser_test.go
```diff
@@ -47,16 +47,16 @@ func TestZipParser_ModuleName(t *testing.T) {
 	}
 }
 
-func zipTestMod(t *testing.T, src string, dstFileName string) (target string) {
+func zipTestMod(t *testing.T, fs afero.Fs, src string, dstFileName string) (target string) {
 	r := require.New(t)
-	zipfile, err := ioutil.TempFile("", "")
+	zipfile, err := afero.TempFile(fs, "", "")
```
michalpristas: 👍

pkg/repo/goget_test.go
```diff
@@ -1,16 +1,17 @@
...
+
+	"github.com/spf13/afero"
 )
 
 func Test_Download(t *testing.T) {
 	version := "v0.1.8"
 	gitURI := "github.com/bketelsen/captainhook"
-
-	fetcher, err := NewGenericFetcher(gitURI, version)
+	fs := afero.NewMemMapFs()
```
michalpristas: try running with osFs, not sure how this will cooperate with vgo

marpio: Sure... vgo knows nothing about afero.thanks!

pkg/gomod/file/parser_test.go
marpio: @michalpristas  it's the vs code go-extension. it uses goimports and it's separating stdlib packages from non-stdlib packages

michalpristas: Yes that's how I imagined it grouped by std/github

marpio: Ups... now i see - the empty lines... I thought you don't want the separation between std and non std pkgs. should be fixed now.

michalpristas: 👍 no problem @marpio, thank you very much, I would address it in separate PR otherwise.
DO you plan any other commits here?

marpio: @michalpristas no, I hope that's all. Did you find anything else?

michalpristas: nono, i think it looks good.
I would like to have at least one more eye on this though

michalpristas:
@bketelsen @arschles no thoughts?

arschles:
No thoughts - this looks good to me. @marpio can you resolve conflicts and ping me in the #athens room on slack? I'll merge right away

arschles:
also, regarding your comment in https://github.com/gomods/athens/pull/123/commits/749bd88ce2e1f6bd2e8574e05fad65a65566d918 - yes, this is a temporary problem, but it could recur. We should package golint (and other tools that we are using `go get` to fetch) into a docker image, or store them as pre-compiled binaries somewhere, so that we don't have to rely on `go get`.

Ironically, this project is in part trying to solve these kinds of temporary problems 😆

PR #129, opened 2018-05-02
Added support for eventlog

michalpristas:
Eventlog is append-only log. 
Will be used in
- Olympus as a log of backed modules
- Proxy as a log of backed modules
- Proxy as a log of registered cache misses (clearable after read)

Multiple backing storages should be supported, for now, mongo is implemented. 
Preparation for Olympus reader, which is remote reader of Olympus event log, is in place without implementation, this implementation is for proxies to read logs from Olympus, and Olympus instances between each other (Pull synchronization)


pkg/eventlog/multireader.go
```diff
@@ -0,0 +1,107 @@
+package eventlog
+
+import "github.com/gomods/athens/pkg/storage"
+
+type multiReader struct {
+	logs    []PointedLog
+	checker storage.Checker
+}
+
+// PointedLog is collection of event logs with specified starting pointers used by ReadFrom function.
```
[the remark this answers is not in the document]
michalpristas: I like it better than what i came up with

pkg/eventlog/eventlog.go
```diff
@@ -0,0 +1,29 @@
...
+
+	// ReadFrom reads all events from the log starting at event with specified id (excluded).
+	// If id is not found behaves like Read().
+	ReadFrom(id string) ([]Event, error)
+}
+
+// Appender is writer to append only event log.
+type Appender interface {
+	// Write appends Event to event log and returns its ID.
+	Write(event Event) (string, error)
```
arschles: nit: can we call this `Append` instead? that name seems to more accurately describe what it's doing

pkg/eventlog/mongo/mongo.go
```diff
@@ -0,0 +1,84 @@
...
+	c   string // collection
+	url string
+}
+
+// NewLog creates log reader from remote log
+func NewLog(url string) (*Log, error) {
+	return NewLogWithCollection(url, "eventlog")
+}
+
+// NewLogWithCollection creates log reader from remote log
```
arschles: seems to return more than a reader - a writer and an appender, right?

pkg/eventlog/mongo/mongo.go
```diff
@@ -0,0 +1,84 @@
...
+		d:   "athens",
+	}
+	return m, m.Connect()
+}
+
+// Connect establishes a session to the mongo cluster.
+func (m *Log) Connect() error {
+	s, err := mgo.Dial(m.url)
+	if err != nil {
+		panic(err)
```
arschles: can you return the error here? would rather do that than panic in a non-root stack frame

PR #144, opened 2018-05-31
Adding a Pointer registry for Athens

arschles:
- Adds an interface for looking up event log pointers
    - This'll be used in Athens to fetch log deltas from Olympus
- Remove `./pkg/cdn/cdn.go` - it wasn't used anywhere
- Add godoc to `./pkg/cdn/getter.go`
- Add an operation to events in the log
- Add a storage deleter (for use when modules need to be deleted in exceptional cases)


pkg/eventlog/pointer_registry.go
```diff
@@ -0,0 +1,8 @@
+package eventlog
+
+// PointerRegistry is a key/value store that stores an event log pointer for one
+// or more Olympus deployments. It is used in proxies (Athens) and Olympus
+// deployments as part of the event log sync process
+type PointerRegistry interface {
+	LookupPointer(deploymentID string) (string, error)
```
michalpristas: 👍 i wil use this in proxy

pkg/eventlog/merger.go
```diff
@@ -0,0 +1,22 @@
...
+// - Appending the log entry into the event log
+// - Modifying the module database accordingly
+type Merger struct {
+	getter   storage.Getter
+	deleter  storage.Deleter
+	setter   storage.Saver
+	appender Appender
+}
+
+func (m Merger) Merge(evt Event) error {
```
michalpristas: comment

pkg/eventlog/pointer_registry.go
michalpristas: also comment

pkg/eventlog/event.go
```diff
@@ -8,6 +8,9 @@ import (
 type Event struct {
 	// ID is identifier, also used as a pointer reference target.
 	ID string `json:"_id" bson:"_id,omitempty"`
+	// Op is the operation on the event log. Valid values are "ADD"
+	// TODO: add support for "DEL"
+	Op string `json:"op" bson:"op"`
```
michalpristas: Let's have a discussion about DEL, i have feedback from people that they would like to keep modules present even though they were deleted in VCS

Issue #146, opened 2018-05-31
[Meta] Implement drivers for the pointer registry


arschles:
https://github.com/gomods/athens/pull/144 added a "pointer registry" interface for use with the proxy (confusingly called "Athens" at the moment). Athens uses the pointers to store the last event log entry that it has fetched from Olympus, and of course it needs to persist them somewhere.

Right now there is only an interface, and we need implementations:

- ~~RDBMS's (the ones that pop supports)~~
- Mongo (done)
- File (done)
- Memory (for dev purposes)
    - need issue for this

This issue is "meta", meaning nobody needs to submit one PR that implements all the drivers.

robjloranger:
Looking at the current interface, there is only a look up method. Should there be a set or update method as well?

robjloranger:
I'm going to start work on the in memory implementation.

robjloranger:
Mongo is complete

arschles:
@robjloranger you are rad 🙌 

checked off mongo in the OP

robjloranger:
I am starting on the FS driver.

lcd1232:
I am starting RDBMS driver

arschles:
👍

michalpristas:
just pushed FS by @robjloranger

lcd1232:
@robjloranger 1-2 days

robjloranger:
Awesome, no rush @lcd1232. just wanted to touch base and see if you were still working on it.

sadlil:
@lcd1232 Is the RDBMS driver still open? Or I can try to create one? if yes, @robjloranger @arschles any concern using gorm to implement a generic driver? So caller needs to import the DB specific driver and pass as configurations? Or implementations needs to be specifics? 

@arschles any other driver in mind? may be redis or else?

michalpristas:
@sadlil  or we let RDBMS go. and wont support that - perf

lcd1232:
@sadlil yes you can do it

arschles:
This is highly related to the registry but as of #772, we're not going to try and build a registry for the time being, so I'm closing this issue

PR #149, opened 2018-06-01
WIP: add in memory pointer registry

robjloranger:
Just using a map.

Not sure about my current placement, created a new package for in memory registry. Declared an error for when a deployment is not present in the registry inside the eventlog pkg.

### DIFF
--- a/pkg/eventlog/memory/pointer_registry.go
+++ b/pkg/eventlog/memory/pointer_registry.go
@@ -1,0 +1,25 @@
+package memory
+
+import "github.com/gomods/athens/pkg/eventlog"
+
+// Registry is a pointer registry for olympus server event logs
+type Registry struct {
+	registry map[string]string
+}
+
+// NewRegistry creates a pointer registry
+func NewRegistry() *Registry {
+	memRegistry := make(map[string]string)
+	return &Registry{memRegistry}
+}
+
+// LookupPointer returns the event log pointer for a given deployment ID
+func (r *Registry) LookupPointer(deploymentID string) (string, error) {
+	var p string
+	if p, ok := r.registry[deploymentID]; !ok {
+		return p, eventlog.ErrDeploymentNotFound
+	}
+	return p, nil
+}
+
+// TODO: SetPointer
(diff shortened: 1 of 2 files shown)


marpio:
I'm not sure this one is needed. We get this one for free if we implement the file system driver just like for the proxy storage. What do you think @arschles @michalpristas

robjloranger:
Ok, no problem. So we only need the FS implementation, nothing specific to in memory?

marpio:
@robjloranger the FS driver will accept `afero.Fs` as a parameter and we can provide in memory file system like this `afero.NewMemMapFs()`

michalpristas:
for reference

https://github.com/gomods/athens/blob/8ceda38ad702b86ae9cb45090273e2e22fb14cbe/pkg/gomod/file/parser.go#L9-L12

robjloranger:
I'm not sure I'm following. I was going off #146 which mentioned both fs and in memory. Just a couple questions to clear me up :smile: 

Will there still be two separate implementations needed? (file, memory)

If so, are you saying for the in memory implementation to make use of the fs implementation internally?

arschles:
@robjloranger when we chatted in Slack, I left out an important detail! If you use `afero.Fs` in your pointer implementation, you get two 🐦s with one one stone (memory and on-disk).

Answers to your questions below

>Will there still be two separate implementations needed? (file, memory)

The implementations are provided by [afero](https://godoc.org/github.com/spf13/afero). They include memory and a bunch of filesystem ones. Check out [here](https://godoc.org/github.com/spf13/afero#Fs) for the implementations.

>If so, are you saying for the in memory implementation to make use of the fs implementation internally?

afero's [memory mapped "filesystem" implementation](https://godoc.org/github.com/spf13/afero#NewMemMapFs) of `afero.Fs` gives you a "filesystem" backed by memory, so that'll do the trick. We also use this as one of the storage drivers in the proxy storage backend (it's used basically for tests only)

robjloranger:
Awesome, I'll work on this later this weekend.

robjloranger:
So I'm going to close this and start on the FS driver, which will in turn be usable as our in memory driver as well. :beers:

arschles:
Rad, thanks!

PR #150, opened 2018-06-01
add basic mongo implementation of pointer registry

robjloranger:
Includes LookupPointer and SetPointer

SetPointer uses UpsertId so if you set an existing deployemnt's pointer it just updates.

Includes full coverage tests.

Partial implementation for #146


robjloranger:
Yep I think it's a good start, and working.

Issue #147, opened 2018-05-31
[Meta] Implement CDN Metadata Drivers


arschles:
There are two pieces of CDN storage:

- Functionality that uploads stuff to the CDN itself ("storage drivers")
- Functionality that saves and fetches metadata about the CDN ("metadata drivers")

Storage drivers and metadata drivers can be mixed and matched. For example, you can use Azure CDN storage drivers with MySQL metadata drivers, Google Cloud Storage with Mongo, and so on...

The issue for adding more storage drivers is https://github.com/gomods/athens/issues/83. This issue is about adding more metadata drivers for CDNs. Including but not limited to:

- Mongo (done)
- The RDBMS's that pop supports (done)
- File (done)
- Memory (#334)

This issue is marked "meta" so one PR doesn't need to implement all of these

marpio:
I will start with the rdbms if no one else is woking on it.

marwan-at-work:
@arschles @marpio I think this issue is no longer relevant because we won't be storing Metadata HTML anymore right?

michalpristas:
yep this is a CDN which is getting removed

PR #151, opened 2018-06-04
add rdbms support for cdn metadata storage

marpio:
Implements rdbms support and Mongo Saver for #147


pkg/cdn/metadata/rdbms/models/cdn_metadata_entry.go
```diff
@@ -0,0 +1,50 @@
...
+	"time"
+
+	"github.com/gobuffalo/pop"
+	"github.com/gobuffalo/uuid"
+	"github.com/gobuffalo/validate"
+	"github.com/gobuffalo/validate/validators"
+)
+
+// CdnMetadataEntry stores the module name and cdn URL.
+type CdnMetadataEntry struct {
```
michalpristas: please, extract this into cdn/metadata and use in both mongo and rdbms

pkg/cdn/metadata/rdbms/getter.go
```diff
@@ -0,0 +1,13 @@
...
+import "github.com/gomods/athens/pkg/cdn/metadata/rdbms/models"
+
+// Get retrieves the cdn base URL for a module
+func (s *MetadataStore) Get(module string) (string, error) {
+	result := models.CdnMetadataEntry{}
+	query := s.conn.Where("module = ?", module)
+	if err := query.First(&result); err != nil {
+		return "", err
+	}
+	return result.RedirectURL, nil
```
michalpristas: possible to return result.RedirectURL, err? it should not be nil right

pkg/cdn/metadata/rdbms/models/cdn_metadata_entry.go
```diff
@@ -0,0 +1,50 @@
...
+	ID          uuid.UUID `json:"id" db:"id"`
+	CreatedAt   time.Time `json:"created_at" db:"created_at"`
+	UpdatedAt   time.Time `json:"updated_at" db:"updated_at"`
+	Module      string    `json:"module" db:"module"`
+	RedirectURL string    `json:"redirect_url" db:"redirect_url"`
+}
+
+// String is not required by pop and may be deleted
+func (e CdnMetadataEntry) String() string {
+	je, _ := json.Marshal(e)
```
michalpristas: m: maybe just log the error? it does not feel right to ignore

pkg/cdn/metadata/rdbms/getter.go
marpio: Do you mean checkin if result == nil? If there is no row matching then First returns an error. Or did you mean cheching RedirectURL?

pkg/cdn/metadata/rdbms/models/cdn_metadata_entry.go
marpio: It's the model code auto-generated  by pop. Should I log this in the storage pkg as well?

pkg/cdn/metadata/rdbms/getter.go
michalpristas: i meant replacing it with single return saying

```go
err := query.First(&result); err != nil {
return result.RedirectURL, nil
```
result should not be nil so result.redirect uri should at least be empty string

pkg/cdn/metadata/rdbms/models/cdn_metadata_entry.go
michalpristas: guess not, i did not realized that it is autogenerated

pkg/cdn/metadata/rdbms/models/cdn_metadata_entry.go
michalpristas: will you update mongo in this pack as well?

michalpristas:
thanks @marpio , it looks good to me, you addressed all things I had in mind very promptly. I will not complete it just yet and let @arschles  review the pack as well

pkg/cdn/metadata/cdn_metadata_entry.go
```diff
@@ -0,0 +1,50 @@
...
+	"time"
+
+	"github.com/gobuffalo/pop"
+	"github.com/gobuffalo/uuid"
+	"github.com/gobuffalo/validate"
+	"github.com/gobuffalo/validate/validators"
+)
+
+// CdnMetadataEntry stores the module name and cdn URL.
+type CdnMetadataEntry struct {
```
arschles: @marpio can you name this `CDNMetadataEntry`?

pkg/cdn/metadata/mongo/mongo.go
```diff
@@ -0,0 +1,40 @@
...
+	session *mgo.Session
+	db      string
+	col     string
+	url     string
+}
+
+// NewMongoStorage returns an unconnected Mongo backed storage
+// that satisfies the Storage interface.  You must call
+// Connect() on the returned store before using it.
+func NewMongoStorage(url, dbName string) *MetadataStore {
```
arschles: nit: this name "stutters" because it will be called with `mongo.NewMongoStorage`. can you call it `NewStorage` instead?

pkg/cdn/metadata/mongo/mongo.go
```diff
@@ -0,0 +1,40 @@
...
+
+// Connect conntect the the newly created mongo backend.
+func (m *MetadataStore) Connect() error {
+	s, err := mgo.Dial(m.url)
+	if err != nil {
+		return err
+	}
+	m.session = s
+
+	// TODO: collection as env var, or param to New()? together with user/mongo
```
arschles: I think this is ok as a constant for now. we can make it an env var if we need to, later on

pkg/cdn/metadata/rdbms/rdbms.go
```diff
@@ -0,0 +1,31 @@
...
+type MetadataStore struct {
+	conn           *pop.Connection
+	connectionName string // settings name from database.yml
+}
+
+// NewRDBMSStorage  returns an unconnected RDBMS Metadata Storage
+// that satisfies the Getter and Setter interfaces. You must call
+// Connect() on the returned store before using it.
+// connectionName
+func NewRDBMSStorage(connectionName string) *MetadataStore {
```
arschles: This name also "stutters" because it will be called with `rdbms.NewRDBMSStorage` - can you change this name to `NewStorage` instead?

PR #152, opened 2018-06-06
[WIP] Adding skeleton for doing database merges

arschles:
This patch adds preliminary functionality for merging a diff (synched from another olympus cloud deployment) into the olympus database.

Requires #153 
Requires #154


cmd/olympus/actions/merge_db.go
```diff
@@ -0,0 +1,74 @@
...
+//		- Add operation adds to the CDN and saves to the module metadata DB
+//		- Delete operation adds tombstone to module metadata k/v store
+//
+// Both could be fixed by putting each 'for' loop into a (global) critical section
+func mergeDB(originURL string, diff dbDiff, getter cdn.Getter, saver cdn.Saver, deleter cdn.Deleter) error {
+	for _, added := range diff.Added {
+		if _, err := getter.GetForVersion(added.Module, added.Version); err != nil {
+			// the module/version already exists, is deprecated, or is
+			// tombstoned, so nothing to do
+			break
```
michalpristas: you will break here, but what about rest of the collection?

pkg/eventlog/event.go
```diff
@@ -1,20 +1,40 @@
...
+
 // Event is entry of event log specifying demand for a module.
+// It implements json.Marshaler
 type Event struct {
 	// ID is identifier, also used as a pointer reference target.
 	ID string `json:"_id" bson:"_id,omitempty"`
 	// Op is the operation on the event log. Valid values are "ADD"
 	// TODO: add support for "DEL"
-	Op string `json:"op" bson:"op"`
+	Op Op `json:"op" bson:"op"`
```
michalpristas: isn't this a bit confusing?

arschles: I think I can call this `EventOp`. I'll do that change in #153, which this PR requires

cmd/olympus/actions/merge_db.go
arschles: good catch - this should be `continue`

cmd/olympus/actions/merge_db.go
```diff
@@ -0,0 +1,84 @@
...
+	for _, deleted := range diff.Deleted {
+		// delete in the CDN
+		if err := deleter.Delete(deleted.Module, deleted.Version); err != nil {
+			log.Printf("error deleting deleted module %s/%s from CDN (%s)", deleted.Module, deleted.Version, err)
+			continue
+		}
+		// add tombstone to module metadata
+		if err := saver.SaveMetadata(deleted.Module, deleted.Version, eventlog.OpDel); err != nil {
+			log.Printf("error inserting tombstone for deleted module %s/%s (%s)", deleted.Module, deleted.Version, err)
+			return err
```
michalpristas: note: we need to be careful here with Deleter not to return an error when a module is already deleted, but to silently succeed.
this way we would not be able to process metadata change ever

pkg/cdn/getter.go
```diff
@@ -14,6 +18,16 @@ package cdn
...
+	// This method does not touch the CDN itself. Olympus should just
+	// redirect vgo to the CDN so that vgo downloads module data directly
+	// from there.
+	//
+	// If something went wrong getting the module metadata, returns a non-nil
+	// error.
 	Get(module string) (string, error)
+	// GetForVersion gets the module metadata about the given module/version.
+	// If something went wrong doing the get operation, returns a non-nil error.
+	GetForVersion(module, version string) (eventlog.Event, error)
```
michalpristas: not really sure if this should be here. in CDN you have stored raw module, no info about eventlog, operations... maybe we should extend eventlog to provide such a functionality instead of CDN

michalpristas: took offline

michalpristas:
@arschles GREEN :-) please check changes

cmd/olympus/actions/merge_db.go
```diff
@@ -0,0 +1,96 @@
...
+		// save module data to the CDN
+		zipBytes, err := ioutil.ReadAll(data.Zip)
+		if err != nil {
+			continue
+		}
+		if err := data.Zip.Close(); err != nil {
+			continue
+		}
+
+		if err := storage.Save(added.Module, added.Version, data.Mod, zipBytes, data.Info); err != nil {
```
marpio: Does the CDN Storage implement the storage.Backend interface? If that's the case do we need the cdn.Saver interface?

michalpristas: few changes we discussed with aaron here:
- there is no CDN storage, just regular storage and CDN attached to it. just like azure cdn to azure blob.
- we will have CDN Metadata K/V which holds redirect uris
- we will have eventlog which holds module,version,operation triplet, eventlog is append only

michalpristas: i will need to clean unneeded code as well. probably in separate PR

marpio: One more question - so I guess the storage.backend you're passing here is the aws S3/azure blob storage and so on, right?
Where will the CDN metadata (redirect URL) be created?

michalpristas: i guess on package processing based on configuration.
imagine you process a module, downloads sources uploads to s3 as `/mod/@v/v1.0` so CDN link stored will be env["CDN_DOMAIN"] + "/mod/@v/v1.0" = `aws-cdn/mod/@v/v1.0`
at this point you dont even need to store that as it can be computed.

now imagine your you will start using different storage, and keep old one as is. new packages goes to azure blob storage as `/mod/@v/v2.0` configuration is changed so CDN domain points to azure CDN and URI will be updated to env["CDN_DOMAIN"] + "/mod/@v/v2.0" = `azure-cdn/mod/@v/v1.0`

so now metadata for
- v1 says `aws-cdn/mod/@v/v1.0`
- v2  says `azure-cdn/mod/@v/v1.0`
both works.

you can easily delete aws if it is deprecated which will lead to cache miss and repopulating with azure backing storage.

does this make sense?

cmd/olympus/actions/diff_test.go
```diff
@@ -0,0 +1,6 @@
+package actions
+
+func (as *ActionSuite) TestBuildDiff() {
+	// r := as.Require()
```
arschles: TODO in a follow up :)

cmd/olympus/actions/merge_db.go
arschles: In https://github.com/gomods/athens/pull/170 I built the CDN driver for Azure and ended up removing the `cdn.Saver` interface ([here](https://github.com/gomods/athens/pull/170/files#diff-7b85c056fb0e17fbf182ac3616e0f2ca)). 

`pkg/cdn` can use `pkg/storage`'s saver. the other stuff (getter, lister, etc...) is only used by the proxy, because that's the only thing that actually serves source code (and `go.mod` and stuff). I implemented the Azure CDN getter (the thing that returns the redirect URL) and Azure storage Saver separately, and then composed the storage Saver into the CDN getter [here](https://github.com/gomods/athens/blob/353e1d1f619d611da20fa69a6735712b6713b678/pkg/cdn/metadata/azurecdn/getter.go)

PR #153, opened 2018-06-06
Adding a DB diff handler

arschles:
This adds an HTTP route to handle diffs that will be optimistically pushed from another Olympus deployment.


pkg/eventlog/event.go
```diff
@@ -1,20 +1,40 @@
...
+
 // Event is entry of event log specifying demand for a module.
+// It implements json.Marshaler
 type Event struct {
 	// ID is identifier, also used as a pointer reference target.
 	ID string `json:"_id" bson:"_id,omitempty"`
 	// Op is the operation on the event log. Valid values are "ADD"
 	// TODO: add support for "DEL"
-	Op string `json:"op" bson:"op"`
+	Op EventOp `json:"op" bson:"op"`
```
michalpristas: this seems to be fixed in other PR

arschles: @michalpristas I change this to `EventOp` in here, and that should address https://github.com/gomods/athens/pull/152/files#r193490061 as well

PR #161, opened 2018-06-07
implement file system based pointer registry

robjloranger:
includes tests <strike>and partial implementation

currently failing on second write for existing registered pointer</strike>


pkg/eventlog/fs/pointer_registry.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/spf13/afero"
+)
+
+// Registry is a pointer registry for olympus server event logs
+type Registry struct {
+	rootDir string
+	fs      afero.Fs
+}
+
+// registryData is a map[string]string used to gob encode the registry on disk
```
michalpristas: gob encode is obsolete in this comment

pkg/eventlog/fs/pointer_registry.go
```diff
@@ -0,0 +1,80 @@
...
+	f, err := r.fs.OpenFile(registryFilename, os.O_RDONLY|os.O_CREATE, 0440)
+	defer f.Close()
+	if err != nil {
+		return "", err
+	}
+
+	var data = make(registryData)
+
+	dec := json.NewDecoder(f)
+	err = dec.Decode(&data)
```
michalpristas: please make it more compact 

```go
if err := dec.Decode(&data); err != nil {
}
```

pkg/eventlog/fs/pointer_registry.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	if _, ok := data[deploymentID]; !ok {
```
michalpristas: here you can de-compact it and store it into value, you do map lookup twice

pkg/eventlog/fs/pointer_registry.go
```diff
@@ -0,0 +1,80 @@
...
+
+	dec := json.NewDecoder(f)
+	err = dec.Decode(&data)
+	if err != nil && err != io.EOF {
+		return err
+	}
+
+	data[deploymentID] = pointer
+
+	_, err = f.Seek(0, os.SEEK_SET)
```
michalpristas: single line

pkg/eventlog/fs/pointer_registry_mem_test.go
```diff
@@ -0,0 +1,52 @@
...
+	r.Equal("location1", gotten)
+
+	err = fs.registry.SetPointer("deployment1", "location2")
+	r.NoError(err)
+	gotten, err = fs.registry.LookupPointer("deployment1")
+	r.NoError(err)
+	r.Equal("location2", gotten)
+
+	gotten, err = fs.registry.LookupPointer("doesnt-exist")
+	r.Equal(err, eventlog.ErrDeploymentNotFound)
```
michalpristas: just a minor thing: first one should be expected then actual

pkg/eventlog/fs/pointer_registry.go
```diff
@@ -0,0 +1,78 @@
...
+		return "", eventlog.ErrDeploymentNotFound
+	}
+
+	return result, nil
+}
+
+// SetPointer both sets and updates the deployment's event log pointer
+func (r *Registry) SetPointer(deploymentID, pointer string) error {
+	f, err := r.fs.OpenFile(registryFilename, os.O_RDWR|os.O_CREATE, 0660)
+	defer f.Close()
```
michalpristas: what happens when error occurs? f is nil right. so you will call nil.Close()

pkg/eventlog/fs/pointer_registry.go
```diff
@@ -0,0 +1,78 @@
...
+
+// NewRegistry returns a file based implementation of a pointer registry
+func NewRegistry(rootDir string, filesystem afero.Fs) *Registry {
+	return &Registry{rootDir: rootDir, fs: filesystem}
+}
+
+// LookupPointer returns the pointer to the given deployment's event log
+func (r *Registry) LookupPointer(deploymentID string) (string, error) {
+	f, err := r.fs.OpenFile(registryFilename, os.O_RDONLY|os.O_CREATE, 0440)
+	defer f.Close()
```
michalpristas: same here

pkg/eventlog/fs/pointer_registry.go
```diff
@@ -0,0 +1,80 @@
...
+	if err := dec.Decode(&data); err != nil {
+		return "", err
+	}
+
+	result, ok := data[deploymentID]
+	if !ok {
+		return "", eventlog.ErrDeploymentNotFound
+	}
+
+	return result, f.Close()
```
marpio: Question: wouldn't the file stay open forever if we get an error doing decoding or ErrDeploymentNotFound? The same inside the SetPointer func

michalpristas: yes indeed good catch. @robjloranger can you put defer close right after error check for Open?

michalpristas:
thanks @robjloranger

PR #164, opened 2018-06-08
Add cdn metadata fs driver

marpio:
Adds the FS driver for #147 
I'm unsure what should happen when a module-baseURL entry already exists. Is there a scenario where we need to update an existing metadata entry?. The current impl returns an error but it means that updates are not possible.

The same actually also applies to the RDBMS driver. Right now if there are two or more entiries for the same module the Getter would just return the first one (so actually random one since there is no 'order by'). Saver saves everything because there is no unique index on the module column.


pkg/cdn/metadata/fs/fs_test.go
```diff
@@ -0,0 +1,52 @@
+package fs
+
+import (
+	"testing"
+
+	"github.com/spf13/afero"
+
```
michalpristas: remove newline pls

pkg/cdn/metadata/fs/saver.go
```diff
@@ -0,0 +1,40 @@
...
+	defer f.Close()
+	id, err := uuid.NewV4()
+	if err != nil {
+		return err
+	}
+	m := &metadata.CDNMetadataEntry{
+		ID:          id,
+		Module:      module,
+		RedirectURL: redirectURL,
+		CreatedAt:   time.Now().UTC(),
```
michalpristas: maybe create time upfront so we have exactly the same bits to start with

PR #166, opened 2018-06-11
Separate in-mem storage and test

### LINKED ISSUE #165: Tests for inmem storage
While working on #132 i found that inmem storage was not able to resolve saved package (no error on save).
No problem was found with mongo implementation, first request was a miss second one was served.

Implement tests for in memory storage and verify if Save->Get works.

marpio:
Fixes #165 
* adds separate mem pkg which uses the fs implementation under the hood
It's possible to call NewStorage() multiple time and get the same storage instance which makes the behavior consistent with other storage drivers


michalpristas:
cool @marpio this is how i worked around the issue in my PR. going to complete yours as you have it in separate package and incl. tests. gonna rebase my solution

PR #170, opened 2018-06-14
[WIP] Implementing a CDN driver for Azure

### LINKED ISSUE #29: Build a storage driver for Azure blob storage
It needs to be able to save module metadata & source (`go.mod`, add to the versions list, and the zip file), and also fetch it

After this is built, consider building drivers for other CDNs

arschles:
Fixes #29

- Implementing a `(github.com/gomods/athens/pkg/cdn/metadata).Getter` for Azure CDN
- Adding the GoGet middleware to Olympus
    - And using the Azure CDN metadata getter
- Implementing a `(github.com/gomods/athens/pkg/storage).Saver` for Azure CDN
    - Really, this is using Azure blob storage for now. Will add Azure CDN in the future

This PR adds a bunch of vendored dependencies, they are added in https://github.com/gomods/athens/commit/7fee6d96853368b079a47aa5f607ff5ec0c135fe only. Other commits have changes

cc/ @marstr @michalpristas @bketelsen


cmd/olympus/actions/app.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/gobuffalo/buffalo/middleware/csrf"
 	"github.com/gobuffalo/buffalo/middleware/i18n"
 	"github.com/gobuffalo/packr"
+
```
michalpristas: please remove newline

pkg/storage/azurecdn/storage.go
```diff
@@ -0,0 +1,87 @@
+package azurecdn
```
michalpristas: can we actualy rename this to azureblob or something similiar, now we will have azurecdn for this and for meta, and none of them is really a cdn implementation

pkg/storage/azurecdn/storage.go
```diff
@@ -0,0 +1,87 @@
...
+// For example:
+//
+//	<meta name="go-import" content="gomods.com/athens mod BaseURL()">
+func (s Storage) BaseURL() *url.URL {
+	return s.accountURL
+}
+
+// Save implements the (github.com/gomods/athens/pkg/storage).Saver interface.
+func (s *Storage) Save(module, version string, mod, zip, info []byte) error {
+	ctx := context.Background()
```
marstr: I see you've deleted the `Saver` interface, so I presume you're not yet bound to this method header. I'd update this to take a context as the first parameter, as so:

``` Go
func (s *Storage) Save(ctx context.Context, module, version string, mod, zip, info []byte) error {
```

I've found this pattern to be quite helpful, and adopting it later is a huge PITA.

pkg/cdn/metadata/azurecdn/getter.go
```diff
@@ -0,0 +1,18 @@
...
+// a (github.com/gomods/athens/pkg/storages/azurecdn).Storage implementation
+//
+// for Azure CDN
+type Metadata struct {
+	azurecdn.Storage
+}
+
+// Get implements the Getter interface
+func (m Metadata) Get(module string) (string, error) {
+	return m.BaseURL().String(), nil
```
michalpristas: what if CDN has a domain registered, this should be maybe configurable

michalpristas: and not tied to storage URL

pkg/storage/azurecdn/storage.go
```diff
@@ -0,0 +1,87 @@
...
+
+	pipe := azblob.NewPipeline(s.cred, azblob.PipelineOptions{})
+	serviceURL := azblob.NewServiceURL(*s.accountURL, pipe)
+	// according to the first example in:
+	// https://godoc.org/github.com/Azure/azure-storage-blob-go/2017-07-29/azblob
+	// ... container names need to be lower case
+	containerName := strings.ToLower(fmt.Sprintf("%s/@v", module))
```
marstr: There are few more rules about container names, you may want to take a look here:
https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names

pkg/storage/azurecdn/storage.go
```diff
@@ -0,0 +1,87 @@
...
+		return err
+	}
+	if _, err := modBlobURL.Upload(ctx, bytes.NewReader(info), httpHeaders("text/plain"), emptyMeta, emptyBlobAccessCond); err != nil {
+		// TODO: log
+		return err
+	}
+	if _, err := zipBlobURL.Upload(ctx, bytes.NewReader(zip), httpHeaders("application/octet-stream"), emptyMeta, emptyBlobAccessCond); err != nil {
+		// TODO: log
+		return err
+	}
```
marstr: Any reason you're not parallelizing these uploads?

pkg/cdn/metadata/azurecdn/getter.go
michalpristas: maybe fallback to this

arschles: I talked to @marstr about this and he had the same point. CDN by default uses account.azureedge.net so that problem is even if there's no domain registered. +1 on making it configurable. Mind if we do that in a follow-up?

pkg/storage/azurecdn/storage.go
michalpristas: saver is from CDN package

pkg/storage/azurecdn/storage.go
arschles: I'd like to keep it as CDN because I actually want to get Azure CDN working in here at some point. It'll have to save to azure blob storage like it does not and then it might have to register with azure CDN

pkg/storage/azurecdn/storage.go
arschles: +1 on taking a `context.Context`. we'd have to change [this Saver interface](https://github.com/gomods/athens/blob/master/pkg/storage/saver.go) though, so I think we should do that in a later PR

pkg/storage/azurecdn/storage.go
arschles: oh man, that kinda sucks. I'll use a static container name and make the blob URLs fancy

pkg/storage/azurecdn/storage.go
arschles: TODO 😄

arschles:
@michalpristas @marstr one more check over please?

pkg/storage/azurecdn/storage.go
```diff
@@ -0,0 +1,82 @@
...
+	serviceURL := azblob.NewServiceURL(*s.accountURL, pipe)
+	// rules on container names:
+	// https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names
+	//
+	// This container must exist
+	containerURL := serviceURL.NewContainerURL("gomodules")
+
+	infoBlobURL := containerURL.NewBlockBlobURL(fmt.Sprintf("%s/@v/%s.info", module, version))
+	modBlobURL := containerURL.NewBlockBlobURL(fmt.Sprintf("%s/@v/%s.mod", module, version))
+	zipBlobURL := containerURL.NewBlockBlobURL(fmt.Sprintf("%s/@v/%s.zip", module, version))
```
marstr: Between [the formerly referenced naming rules](https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#blob-names) and [RFC 3986 pg. 12](http://www.ietf.org/rfc/rfc3986.txt), I _believe_ you'll need to escape the '@' sign here.

pkg/storage/azurecdn/storage.go
```diff
@@ -0,0 +1,82 @@
...
+// Storage implements (github.com/gomods/athens/pkg/storage).Saver and
+// also provides a function to fetch the location of a module
+type Storage struct {
+	accountURL *url.URL
+	cred       azblob.Credential
+}
+
+// New creates a new azure CDN saver
+func New(accountName, accountKey string) (*Storage, error) {
+	u, err := url.Parse(fmt.Sprintf("https://%s.blob.core.windows.net", accountName))
```
michalpristas: is accountName URI safe? does it need sanitizing?

arschles: we'll decide it when we create the account, so we'll need to make it URI safe when we do

pkg/storage/azurecdn/storage.go
arschles: @marstr good point. I was curious about this so I tried it out and [it works](https://testvgo.blob.core.windows.net/gomodules/swtch.com/testmod/@v/list). Did I miss something?

pkg/storage/azurecdn/storage.go
arschles: @marstr ref: https://github.com/gomods/athens/issues/174

pkg/storage/azurecdn/storage.go
arschles: @marstr when I test this out, I'll report back if it really works or not

pkg/cdn/metadata/azurecdn/getter.go
arschles: Ref: https://github.com/gomods/athens/issues/178

pkg/storage/azurecdn/storage.go
marstr: If it works, it works! Something in the Blob library must be escaping it for you, or the documentation is just inaccurate.

PR #171, opened 2018-06-14
fix open tracing by setting operation name

bketelsen:
Set operation name explicitly in handlers to overcome issue with anonymous functions.

### DIFF
--- a/cmd/proxy/actions/cache_miss_handler.go
+++ b/cmd/proxy/actions/cache_miss_handler.go
@@ -13,7 +13,6 @@ import (
 func cacheMissHandler(next buffalo.Handler, w worker.Worker) buffalo.Handler {
 	return func(c buffalo.Context) error {
 		nextErr := next(c)
-
 		if isModuleNotFoundErr(nextErr) {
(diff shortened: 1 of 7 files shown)


cmd/proxy/actions/cache_miss_handler.go
```diff
@@ -48,7 +46,9 @@ func isModuleNotFoundErr(err error) bool {
 	if _, ok := err.(storage.ErrVersionNotFound); ok {
 		return ok
 	}
-
-	s := err.Error()
-	return strings.HasPrefix(s, "module ") && strings.HasSuffix(s, "not found")
+	if err != nil {
+		s := err.Error()
+		return strings.HasPrefix(s, "module ") && strings.HasSuffix(s, "not found")
```
arschles: any way to get away with checking error types instead of string matching?

michalpristas: i think we can, it was not working for me, because FS did not return type error and mongo did. so i fixed FS and one occurrence of string match, this slipped through

Issue #126, opened 2018-04-09
Endpoint for push synchronization


michalpristas:
With push synchronization, Olympus deployment notifies other deployments about the change.

When olympus deployment OA receives push notification from deployment OM, OA does not increase its pointer for OM.
Next time pull synchronization happens, it will be performed with latest pull sync pointer, changes received through push notifications will be skipped due to deduplication checks. 

Example:
```
T0 - OM has MxV1
T1 - OA joins and pulls from OM
	OA[MxV1], OM[MxV1]
T2 - OM receives MyV1 miss 
	OA[MxV1], OM[MxV1, MyV1]
T3 - OM pushes MyV1 change -- timeout
T4 - OM receives MzV1 miss
T5 - OM pushes MzV1 change
	OA[MxV1, MzV1], OM[MxV1, MyV1, MzV1]
T6 - OA pulls from pointer T1
		pulls OM[MyV1, MzV1] =dedup=> [MyV1] 
T7 - OA[MxV1, MzV1, MyV1], OM[MxV1, MyV1, MzV1]
```

Order of log entries does not matter.

arschles:
Note that https://docs.google.com/document/d/1R1iCuDUF8FPZpJalewVZMjSaqWOZn37MHjssayQRQs4/edit?usp=sharing is important background context for this issue.

marpio:
What should happen with the push notification? Should it be treated just like a cache miss from proxy, stored in a separate log or processed right away?

arschles:
It shouldn't work exactly like a cache miss from the proxy. Instead, it should be processed just as if it were a pull operation, except the log pointer shouldn't be updated (it will be reported as duplicate on the next pull)

michalpristas:
I would prefer having one/set of workers (with a set you need to sync about what is getting processed to avoid processing same item at the same time multiple times). background job per request might turn out bad and decrease perf/
we did not think this trough but, I imagine there could be a worker fed periodically with data pulled and decoupled from other Os. this very same worker would get fed by push notifications? 
@arschles

arschles:
Let me think this through some more and get back on Monday

marpio:
right, i didn't express myself clearly - by "creating a background job" i meant putting a job on a queue.

arschles:
Perfect. after thinking about this some more, I came to realize that's what you meant. In a production environment, we'll have workers running to consume from the job queue, and we can scale them up if the queue grows. I believe that a queue push per request is fine given that.

[...]

PR #181, opened 2018-06-15
[WIP] Push endpoint

marpio:
I'm trying to implement #126 
Could someone take a look and tell me if I'm on the right track?


arschles:
@marpio I'll take a look, thanks for opening this. I don't have much time this weekend, but I'll look monday at the latest.

cmd/olympus/actions/push.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/gomods/athens/pkg/payloads"
+)
+
+func pushNotificationHandler(w worker.Worker) func(c buffalo.Context) error {
+	return func(c buffalo.Context) error {
+		p := &payloads.PushNotification{}
+		if err := c.Bind(p); err != nil {
+			return err
+		}
+		pj, err := json.Marshal(p)
```
arschles: I think this is fine for now, but for my information: why did you decide to bind and then marshal back to bytes again?

cmd/olympus/actions/app.go
```diff
@@ -3,19 +3,29 @@ package actions
...
 	"github.com/gobuffalo/buffalo/middleware/csrf"
 	"github.com/gobuffalo/buffalo/middleware/i18n"
+	"github.com/gobuffalo/gocraft-work-adapter"
 	"github.com/gobuffalo/packr"
 	"github.com/gomods/athens/pkg/cdn/metadata/azurecdn"
 )
 
+const (
+	// PushWorkerName is the name of the worker processing pushes
+	PushWorkerName            = "pushProcessor"
```
michalpristas: change this to olympusWorker, you will have some conflicts as i already added one worker
if you're in doubt follow pattern in athens 1worker-2 handlers

cmd/olympus/actions/push.go
```diff
@@ -0,0 +1,59 @@
+package actions
+
+import (
+	"encoding/json"
+	"errors"
+
+	"github.com/gomods/athens/pkg/eventlog"
+	"github.com/gomods/athens/pkg/storage"
+
```
michalpristas: newline, pls remove and group them

cmd/olympus/actions/push.go
michalpristas: i think worker is not working well with more complicated types, json is more safe.

pkg/payloads/upload.go
```diff
@@ -12,3 +14,9 @@ type Module struct {
 	Name    string `json:"name"`
 	Version string `json:"version"`
 }
+
+// PushNotification is used to notify other Olympus instances about a new event
+type PushNotification struct {
+	Events    []eventlog.Event `json:"events"`
```
michalpristas: i'm in doubts here, push notification should be immediate action on a processed cache miss. it should send module, version and CDN uri where the package is.
and at this point you don't know subset of eventlog receiver is interested in as receiver holds the pointer of last acquired entry
so i guess, send single event. process as regular cachemiss but do not send push notification out.
db merge and subset of eventlog will be retrieved on regular pulls
@arschles

marpio: I assumed the events slice will contain single element but I thought using slice leaves the possibility to maybe send those in batches.

I'm not sure I understood it right but if we process the push event as a cache miss wouldn't we download the module using the generic/git fetcher? I thought I'll use mergedb because it already uses the CDN Uri.

cmd/olympus/actions/push.go
marpio: I'm not sure I follow @arschles . Do you mean why I don't just pass the request body to the worker?

pkg/payloads/upload.go
michalpristas: yes yes, of course, we need to use redirect uri from the request. by processing as regular cachemiss i meant, that you have a worker processing single events holding (module, version, uri). not well-picked comparison, my bad.
so to summarize, everything in this PR is right, the only thing differing from my state of mind is batch/single instance.

marpio: Ok, now I get it. My understanding was also that the push notification will be sent after processed cache miss. My consideration was that a slice will be more flexible but I agree that it can be confusing.

cmd/olympus/actions/app.go
marpio: Aren't there two workers in athens @michalpristas ?
https://github.com/gomods/athens/blob/c4ac27f58527e3cfc170cb0393f89f9a5a0df202/cmd/proxy/actions/app.go#L23-L26

michalpristas:
One worker, 2 handlers

ne 17. 6. 2018 o 23:17 marpio <notifications@github.com> napísal(a):

cmd/olympus/main.go
```diff
@@ -22,12 +22,13 @@ func main() {
 	}
 
 	w := app.Worker
-	if err := w.Register(actions.DownloadWorkerName, actions.GetPackageDownloaderJob(s, e, w)); err != nil {
+	if err := w.Register(actions.DownloadHandlerName, actions.GetPackageDownloaderJob(s, e, w)); err != nil {
```
michalpristas: 💯 cool

arschles:
@marpio I'll do a complete re-review on Monday so we can get this merged soon. Really sorry for the delay!

marpio:
No worries @arschles

cmd/olympus/actions/push.go
arschles: @marpio I was wondering why it calls `Bind` (which decodes bytes into `payloads.PushNotification{}`) and then immediately re-encodes the JSON back to bytes. I think I see now that the code is verifying that the JSON payload is correct. Is that right?

cmd/olympus/actions/push.go
```diff
@@ -0,0 +1,64 @@
...
+				workerPushNotificationKey: string(pj),
+			},
+		})
+	}
+}
+
+// GetProcessPushNotificationJob processes queue of push notifications
+func GetProcessPushNotificationJob(storage storage.Backend, eLog eventlog.Eventlog, w worker.Worker) worker.Handler {
+	return func(args worker.Args) (err error) {
+		// TODO: background for now
```
arschles: 👍

pkg/payloads/upload.go
arschles: That is the right understanding @michalpristas @marpio (see https://github.com/gomods/athens/wiki/The-Central-Package-Registry-(Olympus)#push-sync). I agree that a slice is easy to do and more flexible, so I agree with keeping it @marpio

arschles:
@marpio I'm ready to merge this whenever you're ready. I didn't do it yet because it's titled `[WIP]`, so let me know if/when you're ready and I'll hit the button

arschles:
@marpio and I talked on slack. he is ok with merging and says he's cool with writing tests in a follow-up

Issue #168, opened 2018-06-13
Build background job to download packages from GH


arschles:
There is code in https://github.com/gomods/athens/tree/master/pkg/repo/github that can pretty much do the download, we just have to build the background job (probably do it in a grift) that uses it to download stuff (it should hook up to Redis)

PR #182, opened 2018-06-15
Naive implementation of bg job processing downloads

michalpristas:
Basically just putting bits together in a background job

Fixes: #168


cmd/olympus/actions/actions_test.go
```diff
@@ -22,5 +22,4 @@ func (as *ActionSuite) Test_Cache_Miss_Route() {
 	mod.Version = "1.0.0"
 	res := as.JSON("/cachemiss").Post(mod)
 	as.Equal(200, res.Code)
-	as.Contains(res.Body.String(), "\"module\":\"moduleName\",\"version\":\"1.0.0\"")
```
arschles: was a test failing because of this?

michalpristas: yes, i'm not returning module info as method changes to just queue a job

Issue #174, opened 2018-06-14
Thread contexts down the stack


arschles:
Ref: https://github.com/gomods/athens/pull/170#discussion_r195553351

This also greatly helps us with tracing. cc/ @bketelsen 

Here's an example on how to add a tracing span to a function:

```go
func foo(ctx context.Context) {
    sp, ctx := buffet.ChildSpan("storage.save",context)
    defer sp.Close()
    // do stuff with ctx
```

arschles:
Definitely don't have to do this all in one PR. That would be a big refactor!

robjloranger:
I'm thinking if we start with just threading the context, not adding tracing as well, it should be pretty easy. Start in each app and only change one interface per PR.

bketelsen:
that's a nice logical way to do this 👍

manugupt1:
https://github.com/gomods/athens/pull/346

The above PR fixes it!

manugupt1:
Ref: https://github.com/gomods/athens/issues/392

arschles:
@manugupt1 since #346 is done, do you think we can close this issue?

Issue #156, opened 2018-06-06
Improve logging


arschles:
At least add line numbers to logs in server code.

The way to do this would be to create a new `log.Logger` with [`log.New()`](https://godoc.org/log#New). This would all be done in an `init.go` file in each package.

... and then this issue would also entail replacing `log.Abc()` calls with `logger.Abc()`

Here is an example `init.go`:

```go
...
var logger log.Logger
func init() {
    logger := log.New(os.Stdout, "", log.Lshortfile)
}
```

marwan-at-work:
Happy to pick this up and give it a shot. Feel free to assign it to me :)

michalpristas:
Feel free to start, I cannot assign it to anyone except few collaborators yet

robjloranger:
@marwan-at-work it's the usual fork > branch > PR process

PR #183, opened 2018-06-15
GCP storage saver

### LINKED ISSUE #158: Build storage driver for Google Cloud Storage
Ref #83

robjloranger:
closes #158
- [x] make environment vars for: bucket name, <strike>timeout in minutes</strike>
- [x] deal with errors properly during test suite setup and tear down
- [ ] <strike>find a way to abstract away the generation of the file paths in storage</strike>
- [x] log errors from the storage.Writer somewhere
- [ ] <strike>write some proper docs on how to use this</strike>


pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,102 @@
...
+
+// New returns a new Storage instance
+// authenticated using the provided ClientOptions
+func New(ctx buffalo.Context, projectID, bucket string, cred option.ClientOption) (*Storage, error) {
+	client, err := storage.NewClient(ctx, cred)
+	if err != nil {
+		return nil, fmt.Errorf("could not create new client: %s", err)
+	}
+	// TODO: client should be closed on any error or when done
+	// should this be done after save is complete?
```
arschles: @robjloranger you could provide a `Close` method on the `Storage` struct and then require that the caller call it after they're done with the saver. I think in most cases, they'll never want to close it (because in most cases there's gonna be a goroutine filling caches, etc...) but it'll be good to have

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,102 @@
...
+		if err := save(ctx, s.bucket, fmt.Sprintf("%s/@v/%s.info", module, version), info); err != nil {
+			errs <- err
+		} else {
+			errs <- nil
+		}
+	}(errs)
+
+	errsOut := make([]string, 3)
+	// wait for each routine above to send a value
+	for count := 0; count < 3; {
```
arschles: count doesn't look like it's used outside the loop, so just add a `count ++` at the end of the `for` statement. and you can remove the `count++` from below

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,102 @@
...
+			errs <- err
+		} else {
+			errs <- nil
+		}
+	}(errs)
+	go func(errs chan<- error) {
+		sp := buffet.ChildSpan("storage.save.info", ctx)
+		defer sp.Finish()
+
+		if err := save(ctx, s.bucket, fmt.Sprintf("%s/@v/%s.info", module, version), info); err != nil {
```
arschles: are there any restrictions on filenames in GCS?

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,102 @@
...
+func New(ctx buffalo.Context, projectID, bucket string, cred option.ClientOption) (*Storage, error) {
+	client, err := storage.NewClient(ctx, cred)
+	if err != nil {
+		return nil, fmt.Errorf("could not create new client: %s", err)
+	}
+	// TODO: client should be closed on any error or when done
+	// should this be done after save is complete?
+
+	// The bucket MUST already exist
+	bkt := client.Bucket(bucket)
```
arschles: any downside to just hard-coding bucket to `gomodules` or similar? maybe the caller could do it too, either way I can't think of a reason that two `Storage` instances should point to different buckets, you?

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// return concatenated error string if there is anything to report
+	if len(errsOut) < 0 {
+		return fmt.Errorf("one or more errors occured saving %s/@v/%s: %s", module, version, strings.Join(errsOut, ", "))
+	}
+	return nil
+}
+
+// save performs the actual write to a gcp storage bucket
+func save(ctx buffalo.Context, bkt *storage.BucketHandle, filename string, file []byte) error {
```
arschles: open a new span here?

pkg/storage/gcp/saver.go
robjloranger: I was just looking at the docs for the library and they mention the close does not need to be closed at program exit. So maybe we don't really need it.

pkg/storage/gcp/saver.go
robjloranger: just valid unicode up to 1024 bytes, no carriage return or line feed and not named .well-known/acme-challenge

pkg/storage/gcp/saver.go
robjloranger: not at all, I think I meant to change that.

robjloranger:
currently fails because the bucket doesn't exist. Also to create one we might need an actual live bucket for testing. Then have the test delete the bucket and contents on tear down

arschles:
Testify suite has setup and tear down functions that you could use for
setup and tear down. Should work regardless of whether mock or a live bucket

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,101 @@
...
+	defer sp.Finish()
+
+	errs := make(chan error, 3)
+	// dispatch go routine for each file to upload
+	// TODO: make this compact
+	go func(errs chan<- error) {
+		sp := buffet.ChildSpan("storage.save.module", ctx)
+		defer sp.Finish()
+
+		if err := save(ctx, s.bucket, fmt.Sprintf("%s/@v/%s.mod", module, version), mod); err != nil {
```
michalpristas:
```go
errs <- save(ctx, s.bucket, fmt.Sprintf("%s/@v/%s.mod", module, version), mod)
```

more readable

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,101 @@
...
+			errs <- err
+		} else {
+			errs <- nil
+		}
+	}(errs)
+
+	errsOut := make([]string, 3)
+	// wait for each routine above to send a value
+	for count := 0; count < 3; count++ {
+		err := <-errs
```
michalpristas: please use some preconfigured timeout when reading from errs. we dont want this to hang

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,101 @@
...
+		if err != nil {
+			errsOut = append(errsOut, err.Error())
+		}
+	}
+
+	// return concatenated error string if there is anything to report
+	if len(errsOut) < 0 {
+		return fmt.Errorf("one or more errors occured saving %s/@v/%s: %s", module, version, strings.Join(errsOut, ", "))
```
michalpristas: https://github.com/hashicorp/go-multierror
for consideration, not a must, this is perfectly fine

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,101 @@
...
+			errs <- err
+		} else {
+			errs <- nil
+		}
+	}(errs)
+	go func(errs chan<- error) {
+		sp := buffet.ChildSpan("storage.save.zip", ctx)
+		defer sp.Finish()
+
+		if err := save(ctx, s.bucket, fmt.Sprintf("%s/@v/%s.zip", module, version), zip); err != nil {
```
michalpristas: we need to abstract away the creation of blob name, you can use paths package aaron added recently.

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,101 @@
...
+	defer sp.Finish()
+
+	wc := bkt.Object(filename).NewWriter(ctx)
+	wc.ContentType = "application/octet-stream"
+	// TODO: set better access control?
+	wc.ACL = []storage.ACLRule{{storage.AllUsers, storage.RoleReader}}
+	if _, err := wc.Write(file); err != nil {
+		return err
+	}
+	if err := wc.Close(); err != nil {
```
michalpristas: return wc.Close()

robjloranger:
Cool, I'll check it out.

Sent from ProtonMail mobile

-------- Original Message --------

pkg/storage/gcp/saver.go
robjloranger: this one is done

pkg/storage/gcp/saver.go
robjloranger: like just a check on time elapsed? if so what do you think would be acceptable, 5min? this would be to upload whole zipped modules so I would expect due to variable network speeds it could take a while for larger libraries.

pkg/storage/gcp/saver.go
robjloranger: much more :)

pkg/storage/gcp/saver.go
robjloranger: I had a quick look at `pkg/paths` and am not sure how to use it for creating these blob names.

pkg/storage/gcp/saver.go
michalpristas: something configurable, we can change that any time. 5 minutes is ok to start with. once you hit the timeout log warning.
pattern is like this 
```go
switch{
case err := <-errs: 
    someting
case <- time.After(5 * time.Minutes):
   log.Println("WARNING: upload for moduls M version V, timed out")
}
```

michalpristas: or you can apply this pattern on save which makes more sense and gives same timout for each goroutine, benefit: go routine does not hang

```go 
switch {
   case errs <- save(ctx, s.bucket, fmt.Sprintf("%s/@v/%s.mod", module, version), mod):
   case <- time.After(5 * time.Minute): 
      log.Println("WARNING: upload for modul M version V, timed out")
}
```

robjloranger:
still need to do something about blob name generation and the CI will need the gcp sdk installed for tests to pass, it invokes dev_appserver.py behind the scenes.

robjloranger:
whoops, need to add the test dependency

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,75 @@
...
+	// wait for each routine above to send a value
+	for count := 0; count < 3; count++ {
+		err := <-errs
+		if err != nil {
+			errsOut = append(errsOut, err.Error())
+		}
+	}
+
+	// return concatenated error string if there is anything to report
+	if len(errsOut) < 0 {
```
michalpristas: when is len<0?
errsOut is initialized with length set to 3
use `errsOut := make([]string, 0, 3)`
also with make([]string, 3) and append you will end up with len==4 and value ["", "", "", "appended_value"]

robjloranger: Nice catch, I'm going over it all again before I write over this branch. Trying to keep a clean commit history.

michalpristas: it will be squashed into a single commit, so no need to clean the history just code :D

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,75 @@
...
+// writeToBucket performs the actual write to a gcp storage bucket
+func writeToBucket(ctx context.Context, bkt *storage.BucketHandle, filename string, file []byte) error {
+	wc := bkt.Object(filename).NewWriter(ctx)
+	wc.ContentType = "application/octet-stream"
+	// TODO: set better access control?
+	wc.ACL = []storage.ACLRule{{storage.AllUsers, storage.RoleReader}}
+	if _, err := wc.Write(file); err != nil {
+		return err
+	}
+	return wc.Close()
```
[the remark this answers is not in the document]
michalpristas: while defer close with ignoring error is widely accepted this is a nice pattern which might come of use:

```go
defer func(w *Writer) {		
		if err := w.Close(); err != nil {
			log.Fatal(err) // or handle otherwise
		}
	}(wc)
```

pkg/storage/gcp/saver.go
robjloranger: Haha oh 🤦‍♂️

pkg/storage/gcp/saver.go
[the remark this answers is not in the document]
marpio: Very cool!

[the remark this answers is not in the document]
robjloranger: Super cool, I was going to ask about that. I remembered a pattern using a go routine but not the pattern itself.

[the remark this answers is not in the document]
michalpristas: you can combine it with named output arguments and set err there.
but you might end up in a situation where err already has a value, in this case, I don't know what to return.

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	return nil
+}
+
+// save waits for writeToBucket to complete or times out after five minutes
+func save(ctx context.Context, errs chan<- error, bkt *storage.BucketHandle, module, version, ext string, content []byte) {
+	select {
+	case errs <- writeToBucket(ctx, bkt, fmt.Sprintf("%s/@v/%s.%s", module, version, ext), content):
+		return
+	case <-time.After(5 * time.Minute):
```
marpio: Shouldn't we define the timeout on the context and use it here? `case <-ctx.Done():`

michalpristas: yes indeed, but we need to make sure we'reusing ctx.WithTimeout and not Background one.
also it would be nice to have this value configurable. env should suffice

robjloranger:
So things that I need to do still:

- [ ] make environment vars for: bucket name, timeout in minutes
- [ ] deal with errors properly during test suite setup and tear down
- [ ] log errors from the storage.Writer somewhere
- [ ] find a way to abstract away the generation of the file paths in storage

pkg/storage/gcp/all_test.go
```diff
@@ -0,0 +1,88 @@
...
+	version string
+}
+
+func (g *GcpTests) SetupSuite() {
+	// this is the test project praxis-cab-207400.appspot.com
+	// administered by robbie <robjloranger@protonmail.com>
+	// the variable ATHENS_GCP_TEST_KEY should point to the test key json file
+	// and must be set for this test which otherwise will be skipped
+	creds := os.Getenv("ATHENS_GCP_TEST_KEY")
+	if len(creds) == 0 {
```
michalpristas: 👍

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,80 @@
...
+	go save(ctx, errs, s.bucket, module, version, "info", info)
+
+	errsOut := make([]string, 0, 3)
+	// wait for each routine above to send a value
+	for count := 0; count < 3; count++ {
+		err := <-errs
+		if err != nil {
+			errsOut = append(errsOut, err.Error())
+		}
+	}
```
michalpristas: please close the channel

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,80 @@
...
+	// return concatenated error string if there is anything to report
+	if len(errsOut) > 0 {
+		return fmt.Errorf("one or more errors occured saving %s/@v/%s: %s", module, version, strings.Join(errsOut, ", "))
+	}
+	return nil
+}
+
+// save waits for writeToBucket to complete or times out after five minutes
+func save(ctx context.Context, errs chan<- error, bkt *storage.BucketHandle, module, version, ext string, content []byte) {
+	ctxWT, cancel := context.WithTimeout(ctx, 5*time.Minute)
```
michalpristas: note: at the point this is configurable, please use seconds instead of minutes

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,80 @@
...
+		errs <- fmt.Errorf("WARNING: context deadline exceeded during write of %s version %s", module, version)
+	}
+}
+
+// writeToBucket performs the actual write to a gcp storage bucket
+func writeToBucket(ctx context.Context, bkt *storage.BucketHandle, filename string, file []byte) error {
+	wc := bkt.Object(filename).NewWriter(ctx)
+	defer func(w *storage.Writer) {
+		if err := w.Close(); err != nil {
+			// TODO: what kind of logging? the app's logger
```
michalpristas: log.Print - uses application logger and can be configured to stderr or sth else

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+	}(wc)
+	wc.ContentType = "application/octet-stream"
+	// TODO: set better access control?
+	wc.ACL = []storage.ACLRule{{Entity: storage.AllUsers, Role: storage.RoleReader}}
```
michalpristas: i don't know this api: is roleReader sufficient to perform writes?

pkg/storage/gcp/saver.go
robjloranger: ok perfect, I was wondering about that.

pkg/storage/gcp/saver.go
robjloranger: Yes, this sets the access for all users to the role of reader. Which would make the files publicly accessible, we can easily make it private instead by just not creating a custom rule. The application uses a service account which has configurable permissions, and is set to have full control over the storage of the GCP project.

pkg/storage/gcp/saver.go
robjloranger: Should I use log.Error or is Print sufficient?

michalpristas: error pls

robjloranger: I just use Print, the std log pkg does not have .Error like I was thinking. But all log to stderr, I think.

pkg/storage/gcp/all_test.go
```diff
@@ -33,8 +33,8 @@ func (g *GcpTests) SetupSuite() {
 	// administered by robbie <robjloranger@protonmail.com>
 	// the variable ATHENS_GCP_TEST_KEY should point to the test key json file
 	// and must be set for this test which otherwise will be skipped
-	creds := os.Getenv("ATHENS_GCP_TEST_KEY")
-	if len(creds) == 0 {
+	creds, ok := os.LookupEnv("ATHENS_GCP_TEST_KEY")
```
michalpristas: hint: you can use envy package we use widely in this project

robjloranger: :+1:

pkg/storage/gcp/saver.go
robjloranger: @arschles do you still think we need a close method here?

pkg/storage/gcp/saver.go
arschles: Not sure either. @michalpristas maybe we should add something? Could use it [here](https://github.com/gomods/athens/blob/2d25ccb3dea2bab4e208849cae58039ddf26e14e/pkg/storage/azurecdn/storage.go#L51) too

pkg/storage/gcp/saver.go
arschles: >Nice catch, I'm going over it all again before I write over this branch. Trying to keep a clean commit history.

I always think this too ...

pkg/storage/gcp/saver.go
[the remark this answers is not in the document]
arschles: > you can combine it with named output arguments and set err there.

I personally think that's really hard to reason about, since `defer`s are executed right before the function returned, and executed in stack order. I think logging is fine. 

One more request - can we avoid using `log.Fatal` in the `defer`? I'd like to avoid exiting the process in that case.

pkg/storage/gcp/saver.go
arschles: I think overall, `Save` should take a context itself and that can define the timeout

We can totally do that in a separate PR (see #174)

pkg/storage/gcp/saver.go
arschles: I'd be fine with leaving this as-is and fixing in a follow-up (see https://github.com/gomods/athens/issues/156)

arschles:
```ch-ch-changes🎶  …```

https://github.com/gomods/athens/pull/183/commits/bcbb01e3e09546d7940c851fe099b3eeca09ed00 - one of the best commit messages I've seen. TIL you can put emojis in commits 🎉

**Comment by arschles** [comment]
@robjloranger left a few more comments, but I'll do a complete re-review on Monday (tomorrow)

File: pkg/storage/gcp/saver.go (hunk 12 above)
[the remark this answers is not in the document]
Comment (robjloranger) [reply pkg/storage/gcp/saver.go]: I just used log.Printf

File: pkg/storage/gcp/saver.go (hunk 9 above)
Comment (michalpristas) [reply pkg/storage/gcp/saver.go]: @arschles sure thats' what i meant. we compose those filenames on too many places, let's ,make single func which handles this names, this will help with possible refactoring later

File: pkg/storage/gcp/saver.go (hunk 13 above)
Comment (michalpristas) [reply pkg/storage/gcp/saver.go]: i would go with shared timeout. goroutine should start right away but runtime don't give you the guarantee

File: pkg/storage/gcp/saver.go [hunk 20]
```diff
@@ -0,0 +1,92 @@
...
+		return fmt.Errorf("one or more errors occured saving %s/@v/%s: %s", module, version, strings.Join(errsOut, ", "))
+	}
+	return nil
+}
+
+// save waits for writeToBucket to complete or times out after five minutes
+func save(ctx context.Context, cancel func(), errs chan<- error, bkt *storage.BucketHandle, module, version, ext string, content []byte) {
+	select {
+	case errs <- writeToBucket(ctx, bkt, fmt.Sprintf("%s/@v/%s.%s", module, version, ext), content):
+		return
```
robjloranger: is this the correct way to do nothing and exit a select statement?

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	bkt := client.Bucket(bucketname)
+	return &Storage{bucket: bkt}, nil
+}
+
+// Save uploads the modules .mod, .zip and .info files for a given version
+func (s *Storage) Save(ctx context.Context, module, version string, mod, zip, info []byte) error {
+	errs := make(chan error, 3)
+	// create a context that will time out after 300 seconds / 5 minutes
+	ctxWT, _ := context.WithTimeout(ctx, 300*time.Second)
```
michalpristas: please use defer cancel(), it removes itself from parent context (clears resources)

robjloranger: awesome, thanks. I was wondering about that.

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// return concatenated error string if there is anything to report
+	if len(errsOut) > 0 {
+		return fmt.Errorf("one or more errors occured saving %s/@v/%s: %s", module, version, strings.Join(errsOut, ", "))
+	}
+	return nil
+}
+
+// save waits for writeToBucket to complete or times out after five minutes
+func save(ctx context.Context, errs chan<- error, bkt *storage.BucketHandle, module, version, ext string, file []byte) {
```
robjloranger: should this be named `upload` to be consistent with the azure implementation?

pkg/storage/gcp/all_test.go
```diff
@@ -0,0 +1,87 @@
...
+	ctx, done, err := aetest.NewContext()
+	defer done()
+	if err != nil {
+		g.T().Fatalf("could not create new aetest context: %s", err)
+	}
+	g.context = ctx
+	g.bucket = "staging.praxis-cab-207400.appspot.com"
+	// time stamped test module names will prevent concurrent test interference
+	g.module = "gcp-test" + time.Now().String()
+	g.version = "v1.2.3"
```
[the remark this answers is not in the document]
michalpristas: version is usually with a v as a prefix. 
if version is v1.2.3
then path needs to be module/@v/v1.2.3

pkg/storage/gcp/saver.go
michalpristas: even without return, there's no problem with that

pkg/storage/gcp/saver.go
michalpristas: yes, we can rename it, it will become more readable

pkg/storage/gcp/all_test.go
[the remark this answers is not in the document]
arschles: that's right, the version always needs to start with a `v`. so the path would always be `module/@v/v1.2.3`

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// New returns a new Storage instance authenticated using the provided
+// ClientOptions. The bucket name to be used will be loaded from the
+// environment variable ATHENS_GCP_BUCKET_NAME.
+//
+// The ClientOptions should provide permissions sufficient to read, write and
+// delete objects in google cloud storage for your project.
+func New(ctx context.Context, cred option.ClientOption) (*Storage, error) {
+	client, err := storage.NewClient(ctx, cred)
```
arschles: Since the client is created here, and it might not be returned below (in an error case, like if `envy.MustGet` returns an error), what do you think about returning a `func()` from this function that calls `client.Close()` if appropriate?

we may have talked about this before, but I don't remember exactly what we decided on, so if I'm repeating something that we already decided on, forgive me 😄

pkg/storage/gcp/saver.go
```diff
@@ -0,0 +1,93 @@
...
+	close(errs)
+
+	// return concatenated error string if there is anything to report
+	if len(errsOut) > 0 {
+		return fmt.Errorf("one or more errors occured saving %s/@v/%s: %s", module, version, strings.Join(errsOut, ", "))
+	}
+	return nil
+}
+
+// save waits for writeToBucket to complete or times out after five minutes
```
arschles: @robjloranger do you mind changing this comment to indicate that the function will time out after the context expires?

pkg/storage/gcp/saver.go
robjloranger: We can add this, and the Close method in the Storage struct. It is not required to be closed on program termination as per the gcp storage library docs, but if we do it would take a small amount off the GC.

arschles: ah yea, now I remember. if `gcp.New(...)` is only called once (it looks like it is?) then it seems ok to leave as-is

arschles:
@robjloranger this looks fine to merge, and we can do more in follow-ups. let me know if that's cool with you

robjloranger:
sounds good @arschles, just small improvements and better docs remain.

arschles:
@robjloranger I updated your branch & I'll merge when CI finishes. Also, can you create issues to do those follow-ups?

PR #186, opened 2018-06-20
fixing olympus dockerfile

arschles:


cmd/olympus/Dockerfile
```diff
@@ -6,14 +6,14 @@ RUN mkdir -p $GOPATH/src/github.com/gomods/athens/cmd/olympus
 WORKDIR $GOPATH/src/github.com/gomods/athens/cmd/olympus
 
 # this will cache the npm install step, unless package.json changes
-ADD cmd/olympus/package.json .
-ADD cmd/olympus/yarn.lock .
+ADD package.json .
```
michalpristas: i added this so it works from project root. if this works then i guess it's fine to get rid of them

PR #188, opened 2018-06-20
[WIP] Refactoring the Backend Connector

arschles:
And refactoring the mongo driver


pkg/storage/rdbms/rdbms.go
```diff
@@ -6,26 +6,22 @@ import (
...
 type ModuleStore struct {
-	conn           *pop.Connection
-	connectionName string // settings name from database.yml
+	conn *pop.Connection
+	// connectionName string // settings name from database.yml
 }
 
 // NewRDBMSStorage  returns an unconnected RDBMS Module Storage
 // that satisfies the Storage interface. You must call
 // Connect() on the returned store before using it.
```
michalpristas: update comment pls. you dont have Connect capability now

pkg/storage/mongo/mongo.go
```diff
@@ -4,32 +4,29 @@ import (
...
+	s     *mgo.Session
+	deets *ConnDetails
+	d     string // database
+	c     string // collection
+	url   string
 }
 
 // NewStorage returns an unconnected Mongo backed storage
 // that satisfies the Backend interface.  You must call
 // Connect() on the returned store before using it.
```
michalpristas: update comment pls. you dont have Connect capability now

pkg/storage/mongo/connect_util.go
```diff
@@ -0,0 +1,47 @@
...
+	Timeout  time.Duration
+	SSL      bool
+}
+
+func (c ConnDetails) String() string {
+	return fmt.Sprintf(`Host: %s, Port: %d, User: %s, Password: <redacted>, Timeout: %s`, c.Host, c.Port, c.User, c.Timeout)
+}
+
+// GetSession is a utility to create a session from a DB connection string
+// and credentials. The caller is responsible for calling Close on the returned
```
michalpristas: we caller has no ability to call Close at the moment, session is unexported and we do not provide Close

arschles:
These changes are gonna help with hosting Olympus as an MVP. I'll get tests & builds passing as soon as I can, and address your comments @michalpristas!

cmd/olympus/actions/storage.go
```diff
@@ -24,19 +26,19 @@ func GetStorage() (storage.BackendConnector, error) {
 			return nil, fmt.Errorf("missing disk storage root (%s)", err)
 		}
 		s := fs.NewStorage(rootLocation, afero.NewOsFs())
```
robjloranger: With the recent PR #199 this call now returns an error so will need to be updated

cmd/olympus/main.go
```diff
@@ -34,11 +34,7 @@ func main() {
 func getLocalStorage() (storage.Backend, error) {
```
robjloranger: this func could probably go as it essentially wraps `actions.GetStorage`

pkg/env/mongo.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	timeoutSec, err := strconv.Atoi(envy.Get("MONGO_CONN_TIMEOUT_SEC", "5"))
+	if err != nil {
+		return nil, fmt.Errorf("invalid MONGO_CONN_TIMEOUT_SEC (%s)", err)
+	}
+	sslBool, err := strconv.ParseBool(envy.Get("MONGO_SSL", "true"))
+	if err != nil {
+		return nil, fmt.Errorf("invalid MONGO_SSL (%s)", err)
+	}
+	details := &mongo.ConnDetails{
```
robjloranger: I think it's more readable to declare the variable then pass the address of it on ln 48, that's could just be preference though.

pkg/storage/mongo/connect_util.go
```diff
@@ -0,0 +1,47 @@
...
+// session.
+//
+// This function always establishes connections via TLS
```
robjloranger: This may no longer be true with the environment variable `MONGO_SSL` which could be false.

pkg/storage/mongo/mongo.go
```diff
@@ -38,6 +35,9 @@ func (m *ModuleStore) Connect() error {
 		Background: true,
 		Sparse:     true,
 	}
-	c := m.s.DB(m.d).C(m.c)
-	return c.EnsureIndex(index)
+	c := sess.DB(db).C(coll)
+	if err := c.EnsureIndex(index); err != nil {
+		return nil, err
+	}
+	return ms, nil
```
robjloranger: again I think it's better from a readability standpoint to explicitly pass the address of a value on return instead of declaring a variable of type pointer earlier on.

pkg/env/mongo.go
```diff
@@ -0,0 +1,49 @@
+package env
```
marpio: Michal added package "config" in his aws s3 storage PR. We should agree on a common name...

pkg/storage/mem/mem.go
```diff
@@ -30,6 +30,6 @@ func NewStorage() (storage.BackendConnector, error) {
 	}
 
 	s := fs.NewStorage(tmpDir, memFs)
-	memStorage = storage.NoOpBackendConnector(s)
+	memStorage = storage.NoOpBackend(s)
```
marpio: Do we need to wrap the FS storage in a NoOpBackend? FS storage should implement the Backend interface.

pkg/storage/noop_backend.go
```diff
@@ -0,0 +1,27 @@
...
+}
+func (n noOpBackend) List(module string) ([]string, error) {
+	return n.backend.List(module)
+}
+func (n noOpBackend) Save(module, version string, mod, zip, info []byte) error {
+	return n.backend.Save(module, version, mod, zip, info)
+}
+func (n noOpBackend) Delete(module, version string) error {
+	return n.backend.Delete(module, version)
+}
```
marpio: Do we need the NoOpBackend at all?

pkg/env/mongo.go
arschles: good call, thanks

pkg/storage/mongo/connect_util.go
arschles: not sure what you mean @michalpristas - there is a `Close` method defined on `mgo.Session`?

pkg/storage/mongo/connect_util.go
arschles: good catch, thanks @robjloranger

pkg/storage/mongo/mongo.go
arschles: will do, thanks @michalpristas

pkg/storage/mongo/mongo.go
arschles: agree - thanks!

pkg/storage/noop_backend.go
arschles: I don't think so... will remove in another PR

arschles: oops, sorry. I actually had to add this because I merged `BackendConnector` with `Backend`, so `NoopBackendConnector` had to become this

arschles:
This is a big & cross-cutting PR so I'm going to split it up. Into multiple smaller ones. Will close this when all of the smaller ones are in

arschles:
Here are the smaller pieces:

- Fix Mongo SSL
    - Centralize the mongo connection logic across event log and storage
- Make mongo collection names constants
- Env vars at the top of stack
- Remove BackendConnector
    - and just make `New` do the connect

PR #189, opened 2018-06-20
adding a "dev" target to the makefile

arschles:
This makes it easier to get started. You can type `make dev` on the cmdline and it executes the right docker-compose command to get all dependency services up

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,3 +16,12 @@ test:
 
 olympus-docker:
 	docker build -t gopackages/olympus -f cmd/olympus/Dockerfile .
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -16,3 +16,12 @@ test:
 
 olympus-docker:
 	docker build -t gopackages/olympus -f cmd/olympus/Dockerfile .
+
+dev:
+	# TODO: spin up all the services :)
```
arschles: this TODO is captured at https://github.com/gomods/athens/issues/191

Issue #191, opened 2018-06-21
Add all required services to the Makefile "dev" target


arschles:
see the comment that starts at https://github.com/gomods/athens/pull/189/files#diff-b67911656ef5d18c4ae36cb6741b7965R20

PR #192, opened 2018-06-21
spinning up required services in the makefile dev target

arschles:
fixes https://github.com/gomods/athens/issues/191


robjloranger:
Very nice

PR #200, opened 2018-06-26
Stream zips

### LINKED ISSUE #176: Try not to store zips in memory
Wherever possible, we should try to stream data up to CDNs and into proxy storage. Not sure if we can stream directly from GH to CDN/storage, but we should try wherever possible.

This is an optimization that would probably help for bit modules (i.e. Kubernetes)

marpio:
Fixes #176 and #210


pkg/storage/azurecdn/storage.go
```diff
@@ -67,10 +68,19 @@ func (s *Storage) Save(ctx context.Context, module, version string, mod, zip, in
 		_, err := url.Upload(ctx, content, httpHeaders(contentType), emptyMeta, emptyBlobAccessCond)
 		uploadErrs <- err
 	}
+	zipRs, ok := zip.(io.ReadSeeker)
```
marpio: Is there a better way?

cmd/olympus/actions/merge_db.go
```diff
@@ -24,38 +23,36 @@ import (
 // Both could be fixed by putting each 'for' loop into a (global) critical section
 func mergeDB(ctx context.Context, originURL string, diff dbDiff, eLog eventlog.Eventlog, storage storage.Backend) error {
 	for _, added := range diff.Added {
-		if _, err := eLog.ReadSingle(added.Module, added.Version); err != nil {
-			// the module/version already exists, is deprecated, or is
-			// tombstoned, so nothing to do
-			continue
-		}
+		err := func(ae eventlog.Event) error {
```
michalpristas: please separate func for readability

pkg/storage/azurecdn/storage.go
michalpristas: the point of this task should be to ease up on memory and not store the whole content of the zip file there. which is still happening with buffer somewhere up as well as with ioutil readall here. at least it looks like it - i just woke up

this one is not easy and i imagine you need to stream bytes as they come from storage or client.
so if we consume 20mb of RAM being idle, we should process 20mb of ram processing zips as well. there should be no difference.

marpio: Thanks for the comments @michalpristas ! 
You're right the module.MakeZip func uses Buffer but I guess I could use a io.Pipe there. What do you think?

About the azure cdn saver here - the problem is, unlike gcp where io.Reader is enough, azure needs an io.ReadSeeker. The func mergedb downloads the zip from another cdn so we could propably stream this directly from one cdn to another but I'm not sure how to do it because of this ReadSeeker requirement...

marpio: The only thing that comes to my mind is to save the zip to a temp file and upload it from there to the cdn or using something like this https://godoc.org/github.com/aws/aws-sdk-go/service/s3/s3manager

michalpristas: yeah file workaround, I thought about that too as a hackiest way how to do this. 
we need to be platform agnostic and provide a generic way which will work for any storage. 
this is not an easy task and would require a lot of playing around, maybe without any clean way how to achieve lower consumption. 

I will think about it as well. I'm just very limited in time these days.

marpio: Does it have to be generic for all types of storage? Couldn't we leave this to each storage implementation? i.e. azure provides api for uploading streams - `azblob.UploadStreamToBlockBlob` gcp and aws can do this as well.

michalpristas: yeah for sure. what i meant is that we need to design funcs in a way that func signatures are the same and all storages can sit behind one interface

pkg/module/module.go
```diff
@@ -23,40 +23,43 @@ type file struct {
 
 // MakeZip takes dir and module info and generates vgo valid zip
 // the dir must end with a "/"
-func MakeZip(fs afero.Fs, dir, module, version string) ([]byte, error) {
+func MakeZip(fs afero.Fs, dir, module, version string) io.Reader {
 	ignoreParser := getIgnoreParser(fs, dir)
-	buf := new(bytes.Buffer)
-	w := zip.NewWriter(buf)
+	pr, pw := io.Pipe()
```
michalpristas: 👍

marpio:
@arschles @michalpristas  At the moment the aws s3 and azure storage impl. uses `ioutil.ReadAll` and then `bytes.NewReader` internally because the upload funcs require `io.ReadSeeker`. 
I will create issues for those two.

michalpristas:
@marpio please link created issues here.

pkg/module/module.go
```diff
@@ -23,40 +23,43 @@ type file struct {
...
-	walkFunc := func(path string, info os.FileInfo, err error) error {
-		if err != nil || info == nil || info.IsDir() {
-			return err
-		}
+	go func() {
+		zw := zip.NewWriter(pw)
+		defer zw.Close()
 
-		fileName := getFileName(path, dir, module, version)
+		walkFunc := func(path string, info os.FileInfo, err error) error {
```
michalpristas: could you separate this for readability? it's pretty long

pkg/storage/mongo/saver.go
```diff
@@ -2,17 +2,23 @@ package mongo
...
+	"io"
+	"io/ioutil"
 
 	"github.com/gomods/athens/pkg/storage"
 )
 
 // Save stores a module in mongo storage.
-func (s *ModuleStore) Save(_ context.Context, module, version string, mod, zip, info []byte) error {
+func (s *ModuleStore) Save(_ context.Context, module, version string, mod []byte, zip io.Reader, info []byte) error {
+	z, err := ioutil.ReadAll(zip)
```
robjloranger: This is fine as `z` but could probably be a bit more readable as `zipBytes` as elsewhere in the PR

pkg/storage/rdbms/saver.go
```diff
@@ -2,17 +2,23 @@ package rdbms
...
+	"io"
+	"io/ioutil"
 
 	"github.com/gomods/athens/pkg/storage/rdbms/models"
 )
 
 // Save stores a module in rdbms storage.
-func (r *ModuleStore) Save(_ context.Context, module, version string, mod, zip, info []byte) error {
+func (r *ModuleStore) Save(_ context.Context, module, version string, mod []byte, zip io.Reader, info []byte) error {
+	z, err := ioutil.ReadAll(zip)
```
robjloranger: Again here, could be more readable as `zipBytes`

pkg/storage/azurecdn/storage.go
```diff
@@ -48,53 +48,88 @@ func (s *Storage) Save(ctx context.Context, module, version string, mod, zip, in
...
+	select {
+	case errChan <- save():
+	case <-ctx.Done():
+		errChan <- fmt.Errorf("uploading %s/%s.%s timed out", module, version, name)
+	}
+}
+
+func uploadStream(ctx context.Context, errChan chan<- error, containerURL azblob.ContainerURL, module, version, name string, stream io.Reader, opts uploadOptions) {
+	bufferSize := 1 * 1024 * 1024 // Size of the rotating buffers that are used when uploading
+	maxBuffers := 3               // Number of rotating buffers that are used when uploading
```
marpio: Get this from env/config or just sane defaults?

robjloranger: I would think some sane defaults should be fine.

pkg/storage/azurecdn/storage.go
```diff
@@ -49,49 +52,88 @@ func (s *Storage) Save(ctx context.Context, module, version string, mod []byte,
...
 
 	// TODO: take out lease on the /list file and add the version to it
 	//
 	// Do that only after module source+metadata is uploaded
 
 	return errors
 }
+
+func blockBlobURL(containerURL azblob.ContainerURL, module, version, name string) azblob.BlockBlobURL {
+	return containerURL.NewBlockBlobURL(fmt.Sprintf("%s/@v/%s.%s", module, version, name))
```
robjloranger: Here you can use `config.PackageVersionedName(module, version, name)` instead of `fmt.Sprintf`

arschles:
@marpio sorry this has taken so long. let's get it merged 😄 

when you're ready with a fix for https://github.com/gomods/athens/pull/200#discussion_r200758287, ping me in slack or on here and I'll merge right away

arschles:
oh it looks like you'll need to resolve conflicts too

PR #208, opened 2018-07-02
finish gcp implementation

robjloranger:
+ adds Close method for callers who wish to close the client
+ adds Storage.BaseURL to get the base public file url
+ <strike>abstracts full module file path into internal function 'key'</strike>
+ uses (./pkg/config/env).Timeout for uploads context w/ timeout
+ <strike>creates additonal New function to allow easier use when running on
appengine. The original New is now NewWithCredentials.</strike>
+ adds Getter implementation
+ adds Lister implementation
+ adds Checker and Deleter as well

will close #208


pkg/storage/gcp/saver.go
```diff
@@ -116,7 +116,7 @@ func (s *Storage) Save(ctx context.Context, module, version string, mod, zip, in
 // upload waits for either writeToBucket to complete or the context to expire
 func upload(ctx context.Context, errs chan<- error, bkt *storage.BucketHandle, module, version, ext string, file []byte) {
 	select {
-	case errs <- writeToBucket(ctx, bkt, key(module, version, ext), file):
+	case errs <- writeToBucket(ctx, bkt, fmt.Sprintf("%s/@v/%s.%s", module, version, ext), file):
```
michalpristas: sweet

michalpristas: i like your proactive conflict avoidance :-) in a good way

robjloranger: lol thanks

pkg/storage/gcp/storage.go
```diff
@@ -0,0 +1,78 @@
...
+	return s.baseURI
+}
+
+// Close calls the underlying storage and datastore client's close methods
+func (s *Storage) Close() error {
+	var errors error
+	if err := s.closeStorage(); err != nil {
+		errors = multierror.Append(errors, err)
+	}
+	if err := s.dsClient.Close(); err != nil {
```
michalpristas: this is not initialized can you nil check?

pkg/storage/gcp/gcp_test.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func exists(ctx context.Context, cred option.ClientOption, bucket, mod, ver string) error {
+	client, err := storage.NewClient(ctx, cred)
+	if err != nil {
+		return err
+	}
+	bkt := client.Bucket(bucket)
+
+	if _, err := bkt.Object(fmt.Sprintf("%s/@v/%s.%s", mod, ver, "mod")).Attrs(ctx); err != nil {
```
michalpristas: can you replace these fmt.Sprintf with `config.PackageVersionedName` please

pkg/storage/gcp/gcp.go
```diff
@@ -0,0 +1,78 @@
...
+//
+// The ClientOptions should provide permissions sufficient to read, write and
+// delete objects in google cloud storage for your project.
+// TODO: As well as datastore
+func NewWithCredentials(ctx context.Context, cred option.ClientOption) (*Storage, error) {
+	storage, err := storage.NewClient(ctx, cred)
+	if err != nil {
+		return nil, fmt.Errorf("could not create new storage client: %s", err)
+	}
+	bucketname, err := envy.MustGet("ATHENS_STORAGE_GCP_BUCKET")
```
michalpristas: env.GcpBucketName

pkg/storage/gcp/gcp.go
```diff
@@ -0,0 +1,78 @@
...
+	return s.baseURI
+}
+
+// Close calls the underlying storage and datastore client's close methods
+func (s *Storage) Close() error {
+	var errors error
+	if err := s.closeStorage(); err != nil {
+		errors = multierror.Append(errors, err)
+	}
+	if err := s.dsClient.Close(); s.dsClient != nil && err != nil {
```
marpio: Not sure but I think Close() will get executed before the nil check so it can still panic if client is nil.

michalpristas: yes it will. you need to do this ugly if s.dsClient != nil :-)

pkg/storage/gcp/gcp.go
arschles: @michalpristas I don't see that function anywhere. Did I miss it?

pkg/storage/gcp/gcp.go
```diff
@@ -0,0 +1,78 @@
...
+func NewWithCredentials(ctx context.Context, cred option.ClientOption) (*Storage, error) {
+	storage, err := storage.NewClient(ctx, cred)
+	if err != nil {
+		return nil, fmt.Errorf("could not create new storage client: %s", err)
+	}
+	bucketname, err := envy.MustGet("ATHENS_STORAGE_GCP_BUCKET")
+	if err != nil {
+		return nil, fmt.Errorf("could not load 'ATHENS_STORAGE_GCP_BUCKET': %s", err)
+	}
+	u, err := url.Parse(fmt.Sprintf("https://storage.googleapis.com/%s", bucketname))
```
arschles: can you pass in the url? we can't hard-code to GCP or any other specific object store

pkg/storage/gcp/gcp.go
arschles: I recommend we split into separate if statements. it is more lines of code, but easier to read:

```go
if s.dsClient == nil {
    return errors
}
if err := s.dsClient.Close(); err != nil {
    errors = multierror.Append(errors, err)
}
return errors
```

robjloranger: The check is no longer needed in my local changes. I'm proceeding with using datastore for now.

pkg/storage/gcp/gcp.go
robjloranger: It's been added as an optional override in pull #214

robjloranger:
because we can't use encrypted secrets in PRs from forks here's the tests as of this last commit.
2018-07-12-202155_1042x644_scrot

michalpristas:
@robjloranger i will try to find some time for review today. 
regarding test. we talked yesterday that from the longer perspective it would be better to have mocks of actual storage. so we can run tests even without services running and the troubles with sharing secrets would go away.
For now, I think it is ok to have it like we have but we should focus on this once we have registry and proxy up and running

marpio:
@robjloranger  I also think we should mock here. I understand your concerns - if we mock too much the tests can become meaningless but I think once the number of contributors increases it will be hard to manage and it would be great if everyone could run all tests locally.
Maybe once we're able to deploy proxy and registry we could real integration test on master?

pkg/storage/gcp/deleter.go
```diff
@@ -0,0 +1,31 @@
...
+	var errs error
+	if err := s.bucket.Object(config.PackageVersionedName(module, version, "mod")).Delete(ctx); err != nil {
+		errs = multierror.Append(errs, err)
+	}
+	if err := s.bucket.Object(config.PackageVersionedName(module, version, "info")).Delete(ctx); err != nil {
+		errs = multierror.Append(errs, err)
+	}
+	if err := s.bucket.Object(config.PackageVersionedName(module, version, "zip")).Delete(ctx); err != nil {
+		errs = multierror.Append(errs, err)
+	}
```
marpio: Just a note: we could parallelize this just like the uploads. Could be done in #255

pkg/storage/gcp/checker.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"context"
+	"fmt"
+)
+
+// Exists implements the (./pkg/storage).Checker interface
+// returning true if the module at version exists in storage
+func (s *Storage) Exists(module, version string) bool {
+	ctx := context.Background()
+	modName := fmt.Sprintf("%s/@v/%s.%s", module, version, "mod")
```
marpio: use config.PackageVersionedName

pkg/storage/gcp/gcp.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Storage implements the following interfaces:
+// (./pkg/storage).Saver
+// (./pkg/storage).Getter
+// (./pkg/storage).Lister
+// (./pkg/storage).Checker
+// (./pkg/storage).Deleter
+type Storage struct {
+	bucket       *storage.BucketHandle
+	dsClient     *datastore.Client
```
marpio: Is it used somewhere?. I coudn't find anything.

robjloranger:
I agree with you both. I'm already thinking of ways to mock it 😊 not sure if it should maybe be in another PR though. It might take me a while, another week depending on free time.

pkg/storage/gcp/gcp_test.go
```diff
@@ -0,0 +1,100 @@
...
+		r.NoError(err)
+		// check save was successful
+		err = exists(g.context, g.options, g.bucket, g.module, g.version)
+		r.NoError(err)
+	})
+
+	g.T().Run("Get from storage", func(t *testing.T) {
+		version, err := store.Get(g.module, g.version)
+		defer version.Zip.Close()
+		r.NoError(err)
```
marpio: defer after r.NoError(err) otherwise it can panic if version is nil

pkg/storage/gcp/lister.go
```diff
@@ -0,0 +1,45 @@
...
+
+		// kinda hacky looking at this time
+		if strings.HasSuffix(attrs.Name, ".info") {
+			segments := strings.Split(attrs.Name, "/")
+			// version should be last segment w/ .info suffix
+			last := segments[len(segments)-1]
+			version := strings.TrimSuffix(last, ".info")
+			versions = append(versions, version)
+		}
+	}
```
marpio: Sorry if I'm missing something completely obvious but shouldn't `List` be just a file which we update after Save and serve from CDN?

pkg/storage/gcp/getter.go
```diff
@@ -0,0 +1,48 @@
...
+	"io/ioutil"
+
+	"github.com/gomods/athens/pkg/config"
+	"github.com/gomods/athens/pkg/storage"
+)
+
+// Get retrieves a module at a specific version from storage as a (./pkg/storage).Version
+//
+// The caller is responsible for calling close on the Zip ReadCloser
+func (s *Storage) Get(module, version string) (*storage.Version, error) {
```
marpio: If we access the files from CDN do we even need that?

michalpristas:
i leave it up to you. if you cannot share secrets we're basically without tests anyway. so i would say take your time.
if you check my s3 PR i did a while back: https://github.com/gomods/athens/pull/194/files
i generated s3mock based on s3iface using `mockery`. so maybe this would be also possible for gcp?

pkg/storage/gcp/deleter.go
robjloranger: I think that's a great idea, I can do that this evening.

pkg/storage/gcp/gcp.go
robjloranger: I can't count how many times I've removed this :smiley: It keeps coming in from master

pkg/storage/gcp/lister.go
robjloranger: That feels like a more performant solution, so long as we make it concurrency safe. I mostly replicated functionality I saw the `fs` storage pkg providing.

pkg/storage/gcp/getter.go
robjloranger: As above, not sure. Just implementing the interfaces provided by the other storage packages. Will we be removing this?

pkg/storage/gcp/lister.go
marpio: Or is it meant for proxy?

pkg/storage/gcp/gcp.go
robjloranger: maybe that's a lie, not master. but I have been having some 'issues' merging master. I should probably just wait until I have actual free time to work on it instead of small bits.

pkg/storage/gcp/lister.go
robjloranger: It is part of the (./pkg/storage).Backend interface. Lister, Getter, Checker, Saver and Deleter

marpio: If we distribute the files via CDN then we can download them from url no?

marpio: But it could be still useful if the storage is used by proxy to store private modules.

robjloranger:
I think we can merge this is everyone is ok with where we are at right now.
#255 is going to include parallelization of the Deleter via #264
#263 is going to add mocks for local testing without live resources

pkg/storage/gcp/gcp.go
```diff
@@ -0,0 +1,63 @@
...
+func NewWithCredentials(ctx context.Context, cred option.ClientOption) (*Storage, error) {
+	storage, err := storage.NewClient(ctx, cred)
+	if err != nil {
+		return nil, fmt.Errorf("could not create new storage client: %s", err)
+	}
+	bucketname, err := env.GcpBucketName()
+	if err != nil {
+		return nil, err
+	}
+	u, err := url.Parse(fmt.Sprintf("https://storage.googleapis.com/%s", bucketname))
```
michalpristas: can you please use this on top of URI before setting baseURI
https://github.com/gomods/athens/blob/46699719a030675d6a0006d5bfcf176aef515fdf/pkg/config/env/cdn.go#L11

michalpristas: ok i see it in base uri, Ignore me

pkg/storage/gcp/getter.go
michalpristas: For CDN is not needed, if someone wants a proxy with this in place, it might be helpful

pkg/storage/gcp/getter_test.go
```diff
@@ -0,0 +1,3 @@
+package gcp
+
+// see gcp_test.go for a round-trip test that subsumes tests for the getter
```
michalpristas: is it possible to use package for storage test with GCP?

robjloranger: It will be, I want to get this in and then start on refactoring to be able to use a mock in tests. That issue, #263, might include moving to the shared package. I'm not sure yet. Maybe I will make two to keep them shorter. This one got big.

Either way I began to move it over after #215 went in but ran into issue cleaning up the resources after tests so decided to work on mocks first.

michalpristas: great 👍  for shorter PR, i would like to achieve a lot of smaller PRs which can be reviewed very fast

robjloranger: me too, I think we will be able to move a lot quicker over time this way.

PR #212, opened 2018-07-04
Support for exclude include

### LINKED ISSUE #96: Support for Exclude
Proxy by default fetches ALL modules reported from Olympus. Provide a way how to configure proxy with exclude configuration file
which changes this behavior

Example 1: Exclude modules 
```
- github.com/azure
- github.com/foo/bar
```
this configuration will exclude from processing all repos owned by azure and foo/bar repo.
Other repos by foo will be processed as designed

These prefixes should be supported:
```
vcs.com/owner/repo
vcs.com/owner
vcs.com
```
**NO** regexes

michalpristas:
I was thinking about two options here, first one with strong ordering and first rule matching wins.
The second one a bit complicated to implement but easier to use. No ordering reqs for user to take into account.

I went with second as it is more fun to write and the user is happier. 

Based on filter.conf file proxy decides which modules to process and which not.
If there is '-' rule for module or one of its parents it will not process the module and won't report cache miss to Olympus either. 

Example 
```
- a
+ a/b/c
```
this will process:
- d
- a/b/c
- a/b/c/d

wont process 

- a
- a/b

Second example
```
-
+ a/b/c
```
notice `-` with nothing after it - it means global exclude
This won't process everything except a/b/c and its children (a/b/c/d)

So the way it is decided for `a/b/c/d` is to check rule for `a/b/c/d` and then iterate upwards until non-default rule is found (rules: Default, Exclude, Include) 

Fixes #96

### DIFF
--- a/cmd/proxy/actions/cache_miss_fetcher.go
+++ b/cmd/proxy/actions/cache_miss_fetcher.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"io/ioutil"
 
+	"github.com/gomods/athens/pkg/modfilter"
+
 	"github.com/gobuffalo/buffalo/worker"
 	"github.com/gobuffalo/envy"
 	"github.com/gomods/athens/pkg/storage"
@@ -20,11 +22,15 @@ const (
 )
 
 // GetProcessCacheMissJob processes queue of cache misses and downloads sources from active Olympus
-func GetProcessCacheMissJob(s storage.Backend, w worker.Worker) worker.Handler {
+func GetProcessCacheMissJob(s storage.Backend, w worker.Worker, mf *modfilter.ModFilter) worker.Handler {
 	return func(args worker.Args) (err error) {
 		module, version, err := parseArgs(args)
 		if err != nil {
 			return err
+		}
+
+		if !mf.ShouldProcess(module) {
+			return fmt.Errorf("Module %s is excluded", module)
 		}
 
 		if s.Exists(module, version) {
--- a/cmd/proxy/actions/cache_miss_fetcher.go
+++ b/cmd/proxy/actions/cache_miss_fetcher.go
@@ -7,7 +7,7 @@ import (
 
 	"github.com/gobuffalo/buffalo/worker"
 	"github.com/gomods/athens/pkg/config/env"
-	"github.com/gomods/athens/pkg/modfilter"
+	"github.com/gomods/athens/pkg/module"
 	"github.com/gomods/athens/pkg/storage"
 	olympusStore "github.com/gomods/athens/pkg/storage/olympus"
 )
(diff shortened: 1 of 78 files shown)


pkg/modfilter/modfilter.go
```diff
@@ -0,0 +1,191 @@
...
+func newRule(r Rule) ruleNode {
+	rn := ruleNode{}
+	rn.next = make(map[string]ruleNode)
+	rn.rule = r
+
+	return rn
+}
+
+func getConfigLines() ([]string, error) {
+	configName := configurationFileName
```
michalpristas: this i will update with new config package once it's in

pkg/modfilter/modfilter.go
```diff
@@ -0,0 +1,192 @@
...
+
+	"github.com/gomods/athens/pkg/config/env"
+)
+
+var (
+	pathSeparator = "/"
+)
+
+// ModFilter is a filter of modules
+type ModFilter struct {
```
arschles: can you name this `Filter`? that way, the type will read `mod.Filter` when it's used - instead of the redundant `mod.ModFilter`

pkg/modfilter/modfilter.go
```diff
@@ -0,0 +1,192 @@
...
+// Configuration consists of two operations + for include and - for exclude
+// e.g.
+//    - github.com/a
+//    + github.com/a/b
+// will communicate all modules except github.com/a and its children, but github.com/a/b will be communicated
+// example 2:
+//   -
+//   + github.com/a
+// will exclude all items from communication except github.com/a
+func NewModFilter() *ModFilter {
```
arschles: can you name this `NewFilter()`? then the call will be `mod.NewFilter()` which will be less redundant

cmd/proxy/actions/cache_miss_fetcher.go
```diff
@@ -20,13 +21,17 @@ const (
...
-func GetProcessCacheMissJob(s storage.Backend, w worker.Worker) worker.Handler {
+func GetProcessCacheMissJob(s storage.Backend, w worker.Worker, mf *modfilter.ModFilter) worker.Handler {
 	return func(args worker.Args) (err error) {
 		module, version, err := parseArgs(args)
 		if err != nil {
 			return err
 		}
 
+		if !mf.ShouldProcess(module) {
+			return fmt.Errorf("Module %s is excluded", module)
```
arschles: can you make this an error type so we can check for it in other places?

cmd/proxy/actions/cache_miss_handler.go
```diff
@@ -6,11 +6,12 @@ import (
...
 	"github.com/gobuffalo/buffalo/worker"
+	"github.com/gomods/athens/pkg/modfilter"
 	"github.com/gomods/athens/pkg/paths"
 	"github.com/gomods/athens/pkg/storage"
 )
 
-func cacheMissHandler(next buffalo.Handler, w worker.Worker) buffalo.Handler {
+func cacheMissHandler(next buffalo.Handler, w worker.Worker, mf *modfilter.ModFilter) buffalo.Handler {
 	return func(c buffalo.Context) error {
 		nextErr := next(c)
```
arschles: wouldn't we want to check if the module should not be processed before we go to the next step?

pkg/modfilter/modfilter.go
arschles: also, do you mind adding a note here that this struct is not concurrency safe? not sure if that matters right now, but it'll be nice to have that later on at least as a reminder 😄

pkg/config/env/filter.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "github.com/gobuffalo/envy"
+
+const defaultConfigurationFileName = "filter.conf"
+
+// IncludeExcludeFileName specifies file name for include exclude filter
+// If no filename is specified it fallbacks to 'filter.conf'
+func IncludeExcludeFileName() string {
+	return envy.Get("ATHENS_FILTER_FILENAME", defaultConfigurationFileName)
+}
```
arschles: I ❤️ this env package!

Issue #193, opened 2018-06-21
Add support for Azure CDN


arschles:
This is a follow-on to https://github.com/gomods/athens/issues/29. Originally, that issue was going to support Azure CDN, but I stopped after I supported Azure Blog Storage. This would be to add support for Azure CDN.

Here's what I think most of the work would be (I'm not 100% sure how Azure CDN works):

- Add support for custom URLs per-module in the metadata driver (right now, it's the same base URL for all modules)
- Add some configuration to Olympus to make CDN work. I'm not really sure exactly what config is needed

PR #214, opened 2018-07-06
Added configurable CDN endpoint

### LINKED ISSUE #178: Handle CDN custom domains properly
Ref https://github.com/gomods/athens/pull/170#discussion_r195549007

### LINKED ISSUE #195: Add support for AWS CDN
This is a follow-on to #159. Originally, that issue was going to support AWS CDN, but I stopped after I supported S3 Storage. This would be to add support for AWS CDN.

Here's what I think most of the work would be (I'm not 100% sure how Azure CDN works):

Prereq is #193 then this task is just about applying patterns set there to S3 storage

michalpristas:
This should be enough for serving CDN links. 
Connecting CDN to backing storage is always a manual task on storage setup. 

Fixes: #193 
Fixes #195 
Fixes #178

### DIFF
--- a/pkg/config/env/cdn.go
+++ b/pkg/config/env/cdn.go
@@ -1,0 +1,23 @@
+package env
+
+import (
+	"net/url"
+
+	"github.com/gobuffalo/envy"
+)
+
+// CdnEndpointWithDefault returns CDN endpoint if set
+// if not it should default to clouds default blob storage endpoint e.g
+func CdnEndpointWithDefault(value *url.URL) *url.URL {
+	rawURI, err := envy.MustGet("CDN_ENDPOINT")
+	if err != nil {
+		return value
+	}
+
+	uri, err := url.Parse(rawURI)
+	if err != nil {
+		return value
+	}
+
+	return uri
+}
--- a/pkg/config/env/cdn.go
+++ b/pkg/config/env/cdn.go
@@ -6,9 +6,9 @@ import (
 	"github.com/gobuffalo/envy"
 )
 
-// CdnEndpointWithDefault returns CDN endpoint if set
+// CDNEndpointWithDefault returns CDN endpoint if set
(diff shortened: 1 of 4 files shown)


pkg/config/env/cdn.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"net/url"
+
+	"github.com/gobuffalo/envy"
+)
+
+// CdnEndpointWithDefault returns CDN endpoint if set
+// if not it should default to clouds default blob storage endpoint e.g
+func CdnEndpointWithDefault(value *url.URL) *url.URL {
```
arschles: can you use `CDN` instead of `Cdn` here? that's a go convention to upper-case all letters in an acronym

pkg/storage/gcp/saver.go
```diff
@@ -35,7 +37,23 @@ func New(ctx context.Context, cred option.ClientOption) (*Storage, error) {
 		return nil, err
 	}
 	bkt := client.Bucket(bucketname)
-	return &Storage{bucket: bkt}, nil
+
+	u, err := url.Parse(fmt.Sprintf("https://storage.googleapis.com/%s", bucketname))
```
arschles: can you pass in the URL string here? we can't hard-code to GCS or any specific system

robjloranger: The BaseURL method will check for an override env variable and return that if set, this hard coded one is the default for google cloud storage urls.

michalpristas: yep this one is fallback uri to default blob store public URI. in case CDN is not configured this needs to be as it is. i was thinking if it should be moved to config or here. and came to understanding with my inner self that it is ok here, as it is bound to a specific storage provider and should not change

robjloranger:
I think just the conflicts in master and the little one, https://github.com/gomods/athens/pull/214#discussion_r200766115, and we can merge this.

PR #215, opened 2018-07-06
Common testing ground for storage

### LINKED ISSUE #173: investigate checking error types instead of using string matching
see:
https://github.com/gomods/athens/pull/171#discussion-diff-195561658R51

michalpristas:
Common testing ground for testing storages.

+ Not found test as example

I have no issue for this one, i just wanted to do it but it should Fix #173 as a side effect


pkg/storage/storage_tests/storage_test.go
```diff
@@ -0,0 +1,71 @@
...
+type StorageTests struct {
+	*suite.Model
+	storages []storage.Backend
+}
+
+func (d *StorageTests) SetupTest() {
+	ra := d.Require()
+
+	// fs
+	memFs := afero.NewOsFs()
```
robjloranger: Did you intend to use `NewMemMapFs()`?

michalpristas: nono this is filesystem storage. 
most of these setups are taken from its actual tests

robjloranger: Oh ok, it's just the name that threw me off.

robjloranger: It's funny because I named that in the test :-1:

pkg/storage/storage_tests/storage_test.go
```diff
@@ -0,0 +1,71 @@
...
+	ra.NoError(mongoStore.Connect())
+
+	d.storages = append(d.storages, mongoStore)
+
+	// // rdbms
+	conn := d.DB
+	rdbmsStore := rdbms.NewRDBMSStorageWithConn(conn)
+	d.Model.SetupTest()
+
+	d.storages = append(d.storages, rdbmsStore)
```
arschles: @michalpristas how do you feel about inverting this test setup, so that the individual storage driver indicates how to create itself, and then the common tests exercise common test functionality. Here's a skeleton:

```go
type StorageTests struct {
    *suite.Model
    CreateStorage func() (storage.Storage, func(), error) // the individual storage driver sets this
    storage storage.Storage
    storageCloser func()
}

func (s StorageTests) SetupTest() {
    ra := d.Require()
    stg, close, err := s.CreateStorage()
    ra.NoError(err)
    s.storage = stg
    s.storageCloser = close
}

func (s StorageTests) TearDownTest() {
    s.storageCloser()
}

func (s StorageTests) TestGetSaveListRoundTrip() {
    // .....
}
```

michalpristas: make sense, on the note of why not to move it. circular dependency when specific storage is using structs defined in storage root. and then test in storage root using specific storage. so i went with this

robjloranger: I think there should be no tests in storage root that test functionality from sub packages. Those tests should be in each package.

For example in my WIP for gcp I have a new test for both modules and versions not being found.

michalpristas: i think that in some cases it is good to have test like this. 
for example, this not found err. we rely on this error in our logic. so it is ok to have that in common place. 
specific things needs to be tested in each of them but not found or basic save-get roundtrip can be extracted

robjloranger: Ok cool, and we can eventually move all the general implementation tests out into a suite of sorts?

michalpristas: i forgot about one little detail, structs inside one test package cannot be used in another one. which results in as much complicated code as it is now so i will keep this as is

arschles: @michalpristas RE the separate test package: that totally makes sense what you did.

regarding the idea of "inverting" the tests, I didn't understand what you meant in https://github.com/gomods/athens/pull/215#discussion_r200892945 - do you think it's a bad idea?

arschles:
@michalpristas can you resolve conflicts on this branch?

michalpristas:
My plan for this PR was to keep it slim and easy to review by just providing backbone with one example (not found test)
But today it went wrong, i got greedy and migrated other tests as well. I will let other test for other PRs I promise.

pkg/storage/storage_tests/storage_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	d.module = "testmodule"
+	d.version = "v1.0.0"
+	d.mod = []byte("123")
+	d.zip = []byte("456")
+	d.info = []byte("789")
+
+}
+
+func TestDiskStorage(t *testing.T) {
```
robjloranger: Is this testing only disk based storage or all storages?

pkg/storage/storage_tests/storage_test.go
```diff
@@ -0,0 +1,124 @@
...
+	retVersion := listedVersions[0]
+	r.Equal(d.version, retVersion, hrn)
+	gotten, err := ts.Storage().Get(d.module, d.version)
+	r.NoError(err, hrn)
+	defer gotten.Zip.Close()
+	// TODO: test the time
+	r.Equal(gotten.Mod, d.mod)
+	zipContent, err := ioutil.ReadAll(gotten.Zip)
+	r.NoError(err, hrn)
+	r.Equal(zipContent, d.zip, hrn)
```
robjloranger: Small but makes test errors confusing, `r.Equal` takes the expected value first, followed by the actual value.

pkg/storage/storage_tests/storage_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// mongo
+	mongoStore, err := mongo.NewTestSuite(d.Model)
+	ra.NoError(err)
+	d.storages = append(d.storages, mongoStore)
+
+	// rdbms
+	rdbmsStore, err := rdbms.NewTestSuite(d.Model)
+	d.Model.SetupTest()
+	d.storages = append(d.storages, rdbmsStore)
```
arschles: can we put each of these tests runs into its own package?

for example, in the `disk` package:

```go
func TestDiskStorages(t *testing.T) {
    diskSuite, err := NewTestSuite(d.Model)
    if err != nil {
        t.Fatal(err)
    }
    tst := storagetests.Test(diskSuite)
    tst.SetupTest()
    tst.Run()
}
```

pkg/storage/storage_tests/storage_test.go
michalpristas: all storages, just a copy-paste issue

pkg/storage/storage_tests/storage_test.go
michalpristas: another copy-paste, funny because last time I had the same comment in some PR

robjloranger: oh hey, that was me :) which is why I remembered

pkg/storage/storage_tests/storage_test.go
michalpristas: i create module_storage package
another one would be cdn_storage, up to separate PR, i will create issue for this once this is in

PR #216, opened 2018-07-06
[WIP] Adding support for Mongo SSL Connections

arschles:
This replaces part of https://github.com/gomods/athens/pull/188


pkg/storage/mongo/mongo.go
```diff
@@ -2,35 +2,38 @@ package mongo
...
+		d:           "athens",
+		c:           "modules",
+	}
 }
 
-// Connect conntect the the newly created mongo backend.
+// Connect connects the the newly created mongo backend.
 func (m *ModuleStore) Connect() error {
-	s, err := mgo.Dial(m.url)
+	s, err := mongoutil.GetSession(m.connDetails, "athens")
```
michalpristas: is "athens" `m.d`?

pkg/storage/mongoutil/connection.go
```diff
@@ -0,0 +1,39 @@
+package mongoutil
```
michalpristas: can we do mongo/storage/conn instead of util?

pkg/storage/mongo/mongo.go
arschles: yup. now that you pointed that out, I should pull that out into a constant

pkg/storage/mongoutil/connection.go
arschles: yup, absolutely

pkg/config/env/mongo.go
```diff
@@ -67,3 +74,52 @@ func MongoConnectionTimeoutWithDefault(value string) string {
...
+	if err != nil {
+		return nil, err
+	}
+
+	password, err := MongoPassword()
+	if err != nil {
+		return nil, err
+	}
+
+	timeoutSecStr := MongoConnectionTimeoutWithDefault(defaultMongoTimeoutSec)
```
michalpristas: can we reuse https://github.com/gomods/athens/blob/514ac4965992b165cd6a63fde71a2ba29c309d24/pkg/config/env/timeout.go#L11
or does it need to be different one

pkg/cdn/metadata/mongo/mongo.go
```diff
@@ -1,29 +1,32 @@
...
+	session     *mgo.Session
+	connDetails *mongoutil.ConnDetails
+	db          string
+	col         string
 }
 
 // NewStorage returns an unconnected Mongo backed storage
 // that satisfies the Storage interface.  You must call
 // Connect() on the returned store before using it.
-func NewStorage(url, dbName string) *MetadataStore {
```
michalpristas: still mongoutil. `go vet ./...` should scream about it

pkg/config/env/mongo.go
arschles: I think it should be one specific to mongo. going forward we should probably have other specialize timeouts with defaults like this one. the `Timeout()` function seems a bit generic

pkg/cdn/metadata/mongo/mongo.go
```diff
@@ -1,29 +1,32 @@
...
-	return &MetadataStore{url: url, db: dbName}
+func NewStorage(connDetails *mongoutil.ConnDetails, dbName string) *MetadataStore {
+	return &MetadataStore{connDetails: connDetails, db: dbName}
 }
 
 // Connect conntect the the newly created mongo backend.
 func (m *MetadataStore) Connect() error {
-	s, err := mgo.Dial(m.url)
+	conn, err := mongoutil.GetSession(m.connDetails, m.db)
```
michalpristas: sess?

pkg/config/env/mongo.go
```diff
@@ -67,3 +74,52 @@ func MongoConnectionTimeoutWithDefault(value string) string {
...
+	timeoutSec, err := strconv.Atoi(timeoutSecStr)
+	if err != nil {
+		return nil, err
+	}
+
+	sslStr := MongoSSLWithDefault("true")
```
michalpristas: can this instead of returning string, return a parsed bool and in case it is not parsable fallback to default provided one?

pkg/user/mongo/user.go
```diff
@@ -12,31 +13,33 @@ import (
...
 // UserStore represents a UserStore implementation backed by mongo.
 type UserStore struct {
-	s   *mgo.Session
-	d   string // database
-	c   string // collection
-	url string
+	s           *mgo.Session
+	d           string // database
+	c           string // collection
+	connDetails *mongoutil.ConnDetails
```
michalpristas: mongoutil

arschles:
This PR is so old, rebase will be really hard. It's small enough that I'll reopen a PR against the current tip of master

arschles:
I'm closing this in favor of #331

PR #217, opened 2018-07-07
Fix upload nil pointer dereference

marwan-at-work:
When running the CLI's upload command for the first time, I got a panic if the `.gitignore` file did not exist. This is because of the difference between a *nil interface* and a *nil pointer*. This PR fixes it, but I also noticed that we ignore the error for reading a .gignore file, we should potentially check if it's `os.IsNotExist` err and maybe panic otherwise? Happy to update the pr based on the behavior you'd like.

### DIFF
--- a/pkg/module/module.go
+++ b/pkg/module/module.go
@@ -67,19 +67,26 @@ func walkFunc(fs afero.Fs, zw *zip.Writer, dir, module, version string, ignorePa
 
 func getIgnoreParser(fs afero.Fs, dir string) ignore.IgnoreParser {
 	gitFilePath := filepath.Join(dir, gitIgnoreFilename)
-	gitParser, _ := compileIgnoreFileAndLines(fs, gitFilePath, gitIgnoreFilename)
+	gitParser := compileIgnoreFileAndLines(fs, gitFilePath, gitIgnoreFilename)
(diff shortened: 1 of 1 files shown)


michalpristas:
Maybe initializing it with accept all struct on err and returning this. Getting rid of error in return completely if it is ignored. Just a quick idea

From: Marwan Sulaiman
Sent: Sunday, July 8, 2018 12:53 AM
To: gomods/athens
Cc: Subscribed
Subject: [gomods/athens] Fix upload nil pointer dereference (#217)

When running the CLI for the first time, I got a panic if the .gitignore file did not exist. This is because of the difference between a nil interface and a nil pointer. This PR fixes it, but I also noticed that we ignore the error for reading a .gignore file, we should potentially check if it's os.IsNotExist err and maybe panic otherwise? Happy to update the pr based on the behavior you'd like.

You can view, comment on, or merge this pull request online at:
  https://github.com/gomods/athens/pull/217
Commit Summary
• Fix upload nil pointer dereference
File Changes
• M pkg/module/module.go (11) 
Patch Links:
• https://github.com/gomods/athens/pull/217.patch
• https://github.com/gomods/athens/pull/217.diff

marwan-at-work:
@michalpristas updated :)

Issue #218, opened 2018-07-08
Zeus: Upload Module Endpoint Does Not Cover All URL Characters


marwan-at-work:
The upload endpoint in the proxy does not account for dashes and other characters valid in a url, my github url is `github.com/marwan-at-work/somemodule` and so the upload returns a `404`. 

I suggest that the endpoint should just be `POST /admin/upload` and move module path and the version to the request body.

marwan-at-work:
fixed by #219

PR #219, opened 2018-07-08
Zeus: move upload endpoint params to request body

marwan-at-work:
Fixes https://github.com/gomods/athens/issues/218

### DIFF
--- a/cmd/proxy/actions/app_proxy.go
+++ b/cmd/proxy/actions/app_proxy.go
@@ -2,6 +2,7 @@ package actions
 
 import (
 	"github.com/gobuffalo/buffalo"
+	"github.com/gomods/athens/pkg/config/env"
 	"github.com/gomods/athens/pkg/storage"
 )
 
@@ -11,7 +12,12 @@ func addProxyRoutes(app *buffalo.App, storage storage.Backend) error {
 	app.GET("/{module:.+}/@v/{version}.info", cacheMissHandler(versionInfoHandler(storage), app.Worker))
 	app.GET("/{module:.+}/@v/{version}.mod", cacheMissHandler(versionModuleHandler(storage), app.Worker))
 	app.GET("/{module:.+}/@v/{version}.zip", cacheMissHandler(versionZipHandler(storage), app.Worker))
-	app.POST("/admin/upload/{module:.+}/{version}", uploadHandler(storage))
+
+	// Only allow proxy module uploads during development.
+	if env.GoEnvironmentWithDefault("") == "development" {
+		app.POST("/admin/upload/{module:.+}/{version}", uploadHandler(storage))
+	}
+
 	app.POST("/admin/fetch/{module:.+}/{owner}/{repo}/{ref}/{version}", fetchHandler(storage))
 	return nil
 }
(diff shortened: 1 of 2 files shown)


cmd/proxy/actions/app_proxy.go
```diff
@@ -11,7 +11,7 @@ func addProxyRoutes(app *buffalo.App, storage storage.Backend) error {
 	app.GET("/{module:.+}/@v/{version}.info", cacheMissHandler(versionInfoHandler(storage), app.Worker))
 	app.GET("/{module:.+}/@v/{version}.mod", cacheMissHandler(versionModuleHandler(storage), app.Worker))
 	app.GET("/{module:.+}/@v/{version}.zip", cacheMissHandler(versionZipHandler(storage), app.Worker))
-	app.POST("/admin/upload/{module:[a-zA-Z./]+}/{version}", uploadHandler(storage))
```
michalpristas: maybe removing this restriction should be enough. i like having module and version identifiers in a path, it identifies the resource. so I would prefer finding a way how to make this work and use req body as last resort.
does this make sense?

michalpristas: why not letting marawan handle this as this is wonderful first PR

robjloranger: Sure no problem 😊

marwan-at-work: Thanks @michalpristas and @robjloranger 😄  -- For now, I made the regex just like the other ones. @robjloranger can you point me to a regex or a doc as to how Go module import paths should be matched? thanks!

michalpristas: i think this should be fine. can you also add `@v/`?

marwan-at-work: @michalpristas happy to. But would that potentially confuse people to think that this path is part of Vgo's Download Protocol?

michalpristas: yeah probably. and as i think about it we will not want this in prod. so maybe we should add FCB to use this enpoint only in dev

marwan-at-work: Would you like me to add that check as part of this PR? Is there a global variable to indicate whether runtime is dev or prod? 

Another random question: According to the wiki, Zeus can "host private modules". So I assumed this is the endpoint to which you'd do that. So wouldn't you want it in production as well? 

Thanks!

michalpristas: regarding env check: https://github.com/gomods/athens/blob/58d42a1b571d60c40313c91da3f4fd468d024dd0/pkg/config/env/go.go#L11

about private modules. in v2 we will allow uploading of custom packages, right now it should be all discoverable. so private means it sits in private VCS (private gitlab instance).
so you host your code in gitlab instance deployed within companys VPN.
your private proxy  is deployed in same VPN
so it is all visible.
another way how to understand `private` is as a private repo in github visible to organization. in this case we need to provide a way how to make this package available through proxy. we don't really have this scenario thought through. 
@arschles correct me if i'm wrong

marwan-at-work: ah, then maybe once the CLI is removed, this endpoint can be removed as well? For now, I put a check on top of the route. Let me know if that's what you were looking for. Thanks!

As for `private` github repos, the proxy can just use a GitHub Token to get access to private repos. So when a company deploys the proxy, they can just provide that token and the proxy knows how to use it.

arschles:
@marwan-at-work since we're going to remove the CLI and the upload endpoint (see https://github.com/gomods/athens/issues/220), I am going to close this

PR #222, opened 2018-07-08
Olympus: remove potential PROXY_URL when running vgo get

marwan-at-work:
If we set the PROXY_URL to point to Zeus, and zues misses a cache, it will delegate the module fetch to olympus which in turn uses `vgo get`. We have to make sure that vgo get doesn't hit the proxy again causing an infinite loop.

### DIFF
--- a/pkg/repo/goget.go
+++ b/pkg/repo/goget.go
@@ -123,8 +123,7 @@ func getSources(fs afero.Fs, gopath, repoRoot, repoURI, version string) (string,
 	disableCgo := "CGO_ENABLED=0"
 
 	cmd := exec.Command("vgo", "get", fullURI)
-	cmd.Env = os.Environ()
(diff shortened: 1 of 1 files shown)


arschles:
@marwan-at-work - I updated this branch to tip of master. also, there was a CI failure before I did that: https://travis-ci.org/gomods/athens/builds/401554263?utm_source=github_status&utm_medium=notification

pkg/repo/goget.go
```diff
@@ -123,8 +123,9 @@ func getSources(fs afero.Fs, gopath, repoRoot, repoURI, version string) (string,
 	disableCgo := "CGO_ENABLED=0"
 
 	cmd := exec.Command("vgo", "get", fullURI)
-	cmd.Env = os.Environ()
```
michalpristas: this environ was there to provide git/ mercurial...

PR #225, opened 2018-07-09
Olympus: default port to 3001

marwan-at-work:
So that Zeus can happily run at 3000.

### DIFF
--- a/cmd/proxy/actions/cache_miss_reporter.go
+++ b/cmd/proxy/actions/cache_miss_reporter.go
@@ -43,6 +43,7 @@ func reportCacheMiss(module, version string) error {
 	if err != nil {
 		return err
 	}
+	req.Header.Set("Content-Type", "application/json")
 
 	client := http.Client{
 		Timeout: 30 * time.Second,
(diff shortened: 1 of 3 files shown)


pkg/config/env/olympus.go
```diff
@@ -6,3 +6,9 @@ import "github.com/gobuffalo/envy"
 func OlympusGlobalEndpointWithDefault(value string) string {
 	return envy.Get("OLYMPUS_GLOBAL_ENDPOINT", value)
 }
+
+// OlympusHTTPPort returns the port that the olympus server is running on;
+// should default to 30001.
```
arschles: prob a typo: `30001` => `3001`?

marwan-at-work: fixed

cmd/proxy/actions/cache_miss_reporter.go
```diff
@@ -43,6 +43,7 @@ func reportCacheMiss(module, version string) error {
 	if err != nil {
 		return err
 	}
+	req.Header.Set("Content-Type", "application/json")
```
michalpristas: 👍

PR #226, opened 2018-07-09
Olympus: default to PORT env var

marwan-at-work:
So that there won't be an confusion between Buffalo's PORT convention and ours.

### DIFF
--- a/pkg/config/env/olympus.go
+++ b/pkg/config/env/olympus.go
@@ -10,5 +10,5 @@ func OlympusGlobalEndpointWithDefault(value string) string {
 // OlympusHTTPPort returns the port that the olympus server is running on;
 // should default to 3001.
 func OlympusHTTPPort(value string) string {
-	return envy.Get("OLYMPUS_HTTP_PORT", value)
+	return envy.Get("PORT", value)
 }


kteb:
Few points I would change but it could stay at it is:

I would have renamed the function OlympusHTTPPort to Port (it could be used by Zeus or anything else later).

To stay as much as possible close to the buffalo default behavior, I would have had also an ADDR variable. (in dev mode Buffalo by default listen only on 127.0.0.1, overwriting it in this case for me make sense (but the same should be done on Zeus).

At some point I was worried about the case where ADDR was a unix socket, but looking at the buffalo source it should be fine: https://github.com/gobuffalo/buffalo/blob/master/options.go#L92

@marwan-at-work a quick read of those docs (it's really quick) could help to get some part of buffalo you're missing at the moment: 
* https://gobuffalo.io/en/docs/proxy
* https://gobuffalo.io/en/docs/systemd

marwan-at-work:
@kteb thank you! Updated the PR to just be Port

kteb:
In my opinion if you set the env file I'll revert all the changes, they bring nothing and add a small risk to have collision with Buffalo.

Issue #221, opened 2018-07-08
update docs


arschles:
The docs are way out of date. They include how to set up the CLI to upload code (see https://github.com/gomods/athens/issues/220 for related), etc...

The docs should be updated to show:

- How to run the registry if you want (done)
- How to run the proxy (done)
- How to set up vgo to talk to either

arschles:
We've done [a lot of work on docs](https://github.com/gomods/athens/issues?utf8=%E2%9C%93&q=is%3Aclosed+label%3Adocs+) and everything in the OP is fixed. Closing

PR #227, opened 2018-07-09
WIP: Removing the CLI

### LINKED ISSUE #220: remove the CLI
since the source of code is VCSs now, and the registry/proxy pulls that code in the background, we don't need the CLI anymore, since its purpose was to upload code to the registry/proxy. Right now, it's a burden for us to keep it up to date with endpoint changes, etc...

This issue is for removing the CLI code and removing the "upload" endpoints as well

arschles:
Still TODO:

- Removing CLI references from docs
- Remove the CLI label after merging this

Fixes #220
Ref #221 because it's related to updating docs


arschles:
thanks @marwan-at-work - I am going to wait for CI and then merge

PR #229, opened 2018-07-09
Removing the registry renderer

arschles:
The proxy should only render its own templates


michalpristas:
Good catch @arschles 
Totaly unrelated: I would like to scan codebase for unused packages and code to clean it up a bit (not everything unused is not needed)

arschles:
>Totaly unrelated: I would like to scan codebase for unused packages and code to clean it up a bit (not everything unused is not needed)

That would be great. Feel like starting that?

PR #230, opened 2018-07-09
WIP: always downloading a fixed buffalo CLI version

arschles:
As of this writing, latest is v0.12.3

### DIFF
--- a/scripts/get_buffalo.sh
+++ b/scripts/get_buffalo.sh
@@ -2,7 +2,9 @@
 
 BUFFALO_URL="https://github.com/gobuffalo/buffalo/releases/download/v0.12.3/buffalo_0.12.3_linux_amd64.tar.gz"
 TAR_GZ="buffalo_0.12.3_linux_amd64.tar.gz"
+BUFFALO_TARGET_BIN="./bin/buffalo"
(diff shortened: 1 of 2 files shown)


scripts/get_buffalo.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+BUFFALO_URL="https://github.com/gobuffalo/buffalo/releases/download/v0.12.3/buffalo_0.12.3_linux_amd64.tar.gz"
+TAR_GZ="buffalo_0.12.3_linux_amd64.tar.gz"
+
+curl -L -o ${TAR_GZ} ${BUFFALO_URL}
+tar -xvf ${TAR_GZ}
+mv buffalo-no-sqlite /usr/local/bin/buffalo
```
michalpristas: dont you need `+x`

arschles: I think it should already be executable in the tarball that gets downloaded, but no harm in doing so!

arschles:
good idea @robjloranger. change incoming

arschles:
@robjloranger just updated. mind taking another look?

PR #235, opened 2018-07-09
Moving design docs from the wiki

arschles:
Todo after merging this:

- Redirect the [wiki](https://github.com/gomods/athens/wiki/Proxies-(Zeus)) [pages](https://github.com/gomods/athens/wiki/The-Central-Package-Registry-(Olympus)) to the new proxy.md and registry.md pages

Partially addresses #221

### DIFF
--- a/PROXY.md
+++ b/PROXY.md
@@ -1,61 +1,68 @@
 # The Athens Proxy
 
-Athens comes with a module proxy that you can run inside your own organization.
-This proxy will do two things for you:
+The Athens project has two components, the [central registry](./REGISTRY.md) and edge proxies.
+This document details the latter.
 
-* Cache modules locally, so that developers inside your organization do not
-  have to fetch modules from the public internet
-* Redirect module fetch operations to a specific location on the internet
+# The Role of the Proxy
 
-It's intended that clients set the `GOPROXY` environment variable on their
-machine to the Athens proxy address before they start developing. That
-enables the following:
+We intend proxies to be deployed primarily inside of enterprises to:
 
-```console
-go get github.com/a/b
+- Host private modules
+- Exclude access to public modules
+- Cache public modules
+
+Importantly, a proxy is not intended to be a complete _mirror_ of an upstream registry. For public modules, its role is to cache and provide access control.
+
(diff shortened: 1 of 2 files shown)


PROXY.md
```diff
@@ -1,61 +1,68 @@
...
+First and foremost, a proxy exposes the same vgo download protocol as the registry. Since it doesn't have the multi-cloud requirements as the registry does, it supports simpler backend data storage mechanisms. We plan to release a proxy with several backends including:
 
-# Proxy Caches
+- In-memory
+- Disk
+- RDBMS
+- Cloud blob storage
 
-The proxy can be configured to store cached modules in several places:
+Users who want to target a proxy configure their `vgo` CLI to point to the proxy, and then execute commands as normal.
```
marwan-at-work: Should we mention that the configuration is through `GOPROXY` env var? Just in case people wanna test it out

PROXY.md
```diff
@@ -1,61 +1,68 @@
...
-* Disk: this is appropriate for single-node deployments. We don't recommend
-  running an athens proxy on a single node, so we don't recomment using disk
-  storage in a large scale deployment. _This backend is not yet complete_
-* RDBMS: this is appropriate for multi-node deployments. We recommend this
-  storage backend for larger scale deployments. _This backend is not yet complete_
+# Cache Misses
 
-The proxy storage drivers are responsible for storing _all_ information
-concerning a module:
+When a user requests a module `MxV1` from a proxy and the proxy doesn't have `MxV1` in its cache, it first determines whether `MxV1` is private or not private.
```
marwan-at-work: how does it determine that a module is indeed private? just for my own curiosity lol

robjloranger: Based on the module filters set up by the org running their proxy. It's up to them to define what paths host private modules.

PROXY.md
arschles: I was thinking of creating a separate section / document for how to test the thing (probably in `DEVELOPMENT.md`) and linking to it from here

robjloranger: I think that's a good idea

PR #237, opened 2018-07-09
Adding an "on the shoulders of giants" section and document

arschles:


arschles:
I stole it from the [Buffalo project](https://github.com/gobuffalo/buffalo/blob/master/SHOULDERS.md)

PR #238, opened 2018-07-10
Olympus: implement download protocol

marwan-at-work:
- Implement the download protocol in Olympus
- Move the handler logic into a separate package so they can be shared


cmd/proxy/actions/app_proxy.go
```diff
@@ -2,16 +2,20 @@ package actions
...
 
 func addProxyRoutes(app *buffalo.App, storage storage.Backend, mf *module.Filter) error {
 	app.GET("/", proxyHomeHandler)
-	app.GET("/{module:.+}/@v/list", listHandler(storage))
-	app.GET("/{module:.+}/@v/{version}.info", cacheMissHandler(versionInfoHandler(storage), app.Worker, mf))
-	app.GET("/{module:.+}/@v/{version}.mod", cacheMissHandler(versionModuleHandler(storage), app.Worker, mf))
-	app.GET("/{module:.+}/@v/{version}.zip", cacheMissHandler(versionZipHandler(storage), app.Worker, mf))
+
+	// Download Protocol
+	app.GET(download.ListPath, download.ListHandler(storage, proxy))
```
michalpristas: not complaining, just an idea can we start names of these constant with Path?
just like we do with errors: **Err**MyNewError
so when I code I type Err and get all the errors in the package, here i would type Path and got the list of the paths

michalpristas: it feels weird but ok at the same time

pkg/download/doc.go
```diff
@@ -0,0 +1,6 @@
+// Package download provides buffalo handlers
+// that implement vgo's Download Protocol.
+// This is so that both Zeus and Olympus
+// can share the same download protocol
+// implementation.
```
arschles: wooo package docs 👍

arschles:
CI passed after updating from master. Merging!

PR #251, opened 2018-07-12
proxy: fixes missing proxy template check

### LINKED ISSUE #250: proxy: authenticity_token missing within context
#248 defaults to disabled CSRF but the application template uses the variable `authenticity_token`. It should be evaluated if it's set and then be used within the template.

I've prepared a fix for this and will open a PR.

tobiaskohlbau:
Fixes #250


robjloranger:
Nice work @tobiaskohlbau

PR #252, opened 2018-07-12
proxy: remove dangling registry templates

tobiaskohlbau:
Removes templates for registry within proxy cmd subdirectory.


michalpristas:
👍 your eyes are merciless

PR #254, opened 2018-07-12
Stream zip up to AWS S3 storage

### LINKED ISSUE #211: Stream zip up to AWS S3 storage
PR #200 changes the `storage.Saver` interface so that it accepts zip as `io.Reader` but the AWS S3 storage impl. uses `ioutil.ReadAll` to get a byte slice which gets then uploaded.

This could probably help:
https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3manager/#Uploader

marpio:
Fixes #211


pkg/repo/github/commitinfo.go
```diff
@@ -25,7 +27,7 @@ func getCommitInfo(owner, repo, tag string) (CommitInfo, error) {
 
 	uri := fmt.Sprintf(fetchCommitURI, owner, repo, tag)
 
-	client := http.Client{Timeout: timeout * time.Second}
+	client := http.Client{Timeout: env.Timeout()}
```
arschles: Nothing to do in this PR, but I created an issue to rename the Timeout function

arschles: link to issue: https://github.com/gomods/athens/issues/261

PR #256, opened 2018-07-12
SkipDead for workers

michalpristas:
setting skip dead for workers so they don't end up accumulated in a dead queue


cmd/proxy/main.go
```diff
@@ -20,14 +23,11 @@ func main() {
...
-		log.Fatal(err)
-	}
-
-	if err := w.Register(actions.ReporterWorkerName, actions.GetCacheMissReporterJob(w, mf)); err != nil {
-		log.Fatal(err)
+	if w, ok := app.Worker.(*gwa.Adapter); ok {
+		registerWithOptions(w, s)
+	} else {
+		w := app.Worker
+		registerNoOptions(w, s)
```
marpio: Couldn't we create the worker here and pass it to the app? We wouldn't need the type assertion. Just like in Olympus/main.go

michalpristas: yes, but we have auth.go, I did not want to create it there and pass it down

cmd/olympus/main.go
```diff
@@ -76,6 +81,6 @@ func getWorker() worker.Worker {
 			},
 		},
 		Name:           actions.OlympusWorkerName,
-		MaxConcurrency: 25,
+		MaxConcurrency: 5,
```
marwan-at-work: Is this maximum Goroutines or Redis Connections? If Goroutines, then it's kinda small no? 
Also the other `getWorker` function for the proxy still uses `25` :)

cmd/proxy/main.go
marwan-at-work: I don't think we need `registerNoOptions` at all no? We're always using `*gwa.Adapter` and since `actions.App()` doesn't accept any arguments, it's always gonna initialize an app with a redis worker. I'm gonna add a general comment in the review a little more

cmd/olympus/main.go
michalpristas: yes i want to use number of goroutines, but wnated to share this before i go offline, i will continue on this tomorrow

pkg/config/env/olympus.go
```diff
@@ -1,8 +1,30 @@
...
 
 // OlympusGlobalEndpointWithDefault returns Olympus global endpoint defined by OLYMPUS_GLOBAL_ENDPOINT.
 func OlympusGlobalEndpointWithDefault(value string) string {
 	return envy.Get("OLYMPUS_GLOBAL_ENDPOINT", value)
 }
+
+// AthensMaxConcurrency retrieves maximal level of concurrency based on ATHENS_MAX_CONCURRENCY.
+// Defaults to number of cores if env is not set.
+func AthensMaxConcurrency() int {
+	defaultMaxConcurrency := runtime.NumCPU()
```
marwan-at-work: if maxConcurrency is `goroutines` can't be like 100 times the number of cpus? Just curious :)

michalpristas:
@marpio check new tests in Olympus action_test against real redis. I moved creation of worker into the app, so it is united through the whole codebase. so i have to get rid of fakeworker and think about other way how to test it

pkg/config/env/olympus.go
michalpristas: it can, but from my experience more is not better, and this is most of the time the sweet spot. 
but we can override in any case

marpio:
@michalpristas I like it because it tests more of the real thing.
My only concern would be that if we ever use different worker implementations that rely on some cloud service (Azure Bus was mentioned) then it will be difficult to test  because of authentication etc. On the other hand we can always use Redis in test so we should be fine.

marwan-at-work:
@marpio i agree. This is similar to storage tests. We should have a concise way of testing all of our interface-implementations. 

Maybe write the tests once against a particular interface, and then we can pass it different implementations? We'd still be testing the "real thing" and not some mock interface, but we won't have to write different tests per implementation.

What do you think @michalpristas ?

This can be another pr of course.

pkg/config/env/olympus.go
```diff
@@ -1,8 +1,30 @@
...
+	maxConcurrencyEnv, err := envy.MustGet("ATHENS_MAX_CONCURRENCY")
```
robjloranger: here you could use `envy.Get("ATHENS_MAX_CONCURRENCY", defaultMaxConcurrency)` and skip the err check

pkg/config/env/olympus.go
```diff
@@ -1,8 +1,47 @@
...
+	}
+
+	return mc
+}
+
+// WorkerMaxFails retrieves maximal level of concurrency based on ATHENS_WORKER_MAX_FAILS.
+// Defaults to 5.
+func WorkerMaxFails() uint {
+	defaultMaxFails := uint(5)
+	maxFailsEnv, err := envy.MustGet("ATHENS_WORKER_MAX_FAILS")
```
robjloranger: same as https://github.com/gomods/athens/pull/256/files#r202444723

pkg/config/env/olympus.go
michalpristas: it accepts string, i wanted to avoid these funny conversions

robjloranger: ohh, yes my mistake

arschles:
totally agree on centralizing the worker creation logic. we have a good place to centralize configs in `./pkg/config`, so maybe we could create a `./pkg/worker` to put all the worker registration and logic in there?

also RE testing - totally agree on mock implementations vs. real integration tests. as an example, the Kubernetes project (which has lots of dependencies on cloud services) runs its full integration suite on master builds only, and in a test queue. some of them nightly, but I can't remember which or if they still do that. we probably don't need to get that fancy.

anyway, I'm going to create an issue to talk about our tests in more detail (I heard this was a topic of discussion in yesterday's meeting too!)

arschles:
Here is the issue: https://github.com/gomods/athens/issues/262

PR #264, opened 2018-07-13
CDN parallel uploads

### LINKED ISSUE #255: Unify code for parallel uploades in S3/azure/gcp storages
At the moment each Saver uploades .zip,.mod and .info in parallel but does this a little bit different.
It would be good to use common code for that.

marpio:
Fixes #255 
also refactoring of the gcp bucket iface


pkg/storage/blobstoreutil/blob_store_util.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/gomods/athens/pkg/config"
+	multierror "github.com/hashicorp/go-multierror"
+)
+
+// Uploader takes a stream and saves it to the block storage under a given path
```
robjloranger: Was it meant to say `blob storage` here?

pkg/storage/blobstoreutil/blob_store_util.go
```diff
@@ -0,0 +1,44 @@
+package blobstoreutil
```
michalpristas: please something like storage/module/upload.go
so we dont end up with `util`packages

pkg/storage/blobstoreutil/blob_store_util.go
```diff
@@ -0,0 +1,44 @@
...
+	multierror "github.com/hashicorp/go-multierror"
+)
+
+// Uploader takes a stream and saves it to the block storage under a given path
+type Uploader func(ctx context.Context, path, contentType string, stream io.Reader) error
+
+// UploadModule saves .info, .mod and .zip files to the blob storage in parallel.
+// Returns multierror containing errors from all uploads and timeouts
+func UploadModule(ctx context.Context, module, version string, info, mod, zip io.Reader, uploader Uploader) error {
+	const numUpload = 3
```
michalpristas: now you can put this into a package level consts

michalpristas:
direction looks good to me

pkg/storage/module/delete.go
```diff
@@ -0,0 +1,42 @@
...
+// Returns multierror containing errors from all deletes and timeouts
+func Delete(ctx context.Context, module, version string, delete Deleter) error {
+	errChan := make(chan error, numFiles)
+
+	del := func(ext string) {
+		p := config.PackageVersionedName(module, version, ext)
+		select {
+		case errChan <- delete(ctx, p):
+		case <-ctx.Done():
+			errChan <- fmt.Errorf("deleting %s timed out", p)
```
marwan-at-work: Is this context Done always a timeout and not cancelled? why not just send `ctx.Error()` to `errChan` so that we pass along exactly why the context was done?

pkg/storage/module/upload.go
```diff
@@ -0,0 +1,45 @@
...
+// Returns multierror containing errors from all uploads and timeouts
+func Upload(ctx context.Context, module, version string, info, mod, zip io.Reader, uploader Uploader) error {
+	errChan := make(chan error, numFiles)
+
+	save := func(ext, contentType string, stream io.Reader) {
+		p := config.PackageVersionedName(module, version, ext)
+		select {
+		case errChan <- uploader(ctx, p, contentType, stream):
+		case <-ctx.Done():
+			errChan <- fmt.Errorf("uploading %s timed out", p)
```
marwan-at-work: same as the delete one

pkg/storage/module/delete.go
robjloranger: I like that, seems cleaner.

pkg/storage/azurecdn/storage.go
```diff
@@ -8,26 +8,26 @@ import (
 	"net/url"
 
 	"github.com/Azure/azure-storage-blob-go/2017-07-29/azblob"
-	"github.com/gomods/athens/pkg/config"
 	"github.com/gomods/athens/pkg/config/env"
-	multierror "github.com/hashicorp/go-multierror"
+	m "github.com/gomods/athens/pkg/storage/module"
```
michalpristas: something more descriptive please

michalpristas: module would cause conflicts, but for example modupl. or uploader as we use this only to m.Upload.
moduploader.Upload?
just few ideas, none of those seems 100% fine

pkg/storage/gcp/saver.go
```diff
@@ -20,47 +18,16 @@ import (
...
-
-// upload waits for either writeToBucket to complete or the context expires
-func upload(ctx context.Context, errs chan<- error, bkt Bucket, module, version, ext, contentType string, file io.Reader) {
-	select {
-	case errs <- writeToBucket(ctx, bkt, module, version, ext, contentType, file):
-		return
-	case <-ctx.Done():
-		errs <- fmt.Errorf("WARNING: context deadline exceeded during write of %s version %s", module, version)
-	}
+	err := modupl.Upload(ctx, module, version, bytes.NewReader(info), bytes.NewReader(mod), zip, s.upload)
```
robjloranger: maybe you can just return the function call here

pkg/storage/module/delete_test.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func TestDelete(t *testing.T) {
+	suite.Run(t, new(DeleteTests))
+}
+
+func (m *DeleteTests) TestDeleteTimeout() {
+	r := m.Require()
+	envy.Set("ATHENS_TIMEOUT", "1")
+	defer envy.Set("ATHENS_TIMEOUT", "300")
```
robjloranger: I'm wondering about the use of these defer for the environment variable. What was the reasoning? It seems like each test uses a timeout of 1 second and sets it back to 300. If you are trying to keep the original value for other tests, you can instead use `(m *DeleteTests) BeforeTest()` and `(m *DeleteTests) AfterTest()` .

pkg/storage/module/delete_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+type DeleteTests struct {
+	suite.Suite
+}
+
+func TestDelete(t *testing.T) {
+	suite.Run(t, new(DeleteTests))
+}
+
+func (m *DeleteTests) TestDeleteTimeout() {
```
robjloranger: I think it may be more idiomatic to use the first letter of the type. So here `(d *DeleteTests)` would be better.

pkg/storage/azurecdn/storage.go
```diff
@@ -7,9 +7,8 @@ import (
 	"io"
 	"net/url"
 
-	"github.com/gomods/athens/pkg/config"
 	"github.com/gomods/athens/pkg/config/env"
-	multierror "github.com/hashicorp/go-multierror"
+	modupl "github.com/gomods/athens/pkg/storage/module"
```
arschles: do you mind keeping this without the package alias? code in this file can then call `module.Thing` that way and it'll be a tiny bit clearer

same with the below `modupl` package alias

pkg/storage/module/delete_test.go
```diff
@@ -0,0 +1,55 @@
...
+	r := m.Require()
+	envy.Set("ATHENS_TIMEOUT", "1")
+	defer envy.Set("ATHENS_TIMEOUT", "300")
+	err := Delete(context.Background(), "mx", "1.1.1", delWithTimeout)
+
+	me := err.(*multierror.Error)
```
robjloranger: I'm not sure if you need to cast to a `multierror.Error` here. I think the method `.Error` will return the full error string of all errors collected. Maybe I'm confused, but the error returned from delete is a multierror already? Or does it lose those that type after being returned.

pkg/storage/gcp/lister.go
```diff
@@ -2,11 +2,33 @@ package gcp
...
 func (s *Storage) List(module string) ([]string, error) {
 	ctx := context.Background()
-	return s.bucket.ListVersions(ctx, module)
+	unfiltered, err := s.bucket.List(ctx, module)
+	if err != nil {
+		return nil, err
+	}
+	versions := make([]string, 0, 10)
+	for _, n := range unfiltered {
+		// kinda hacky looking at this time
```
arschles: if you have any ideas for making this less hacky, can you create an issue for it @marpio ?

pkg/storage/module/delete_test.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func delWithTimeout(ctx context.Context, path string) error {
+	time.Sleep(2 * time.Second)
+	return nil
+}
+
+func delWithErr(ctx context.Context, path string) error {
+	return errors.New("some err")
+}
```
arschles: 💯 for these tests, and the ones below! thanks @marpio !!

pkg/storage/s3/storage.go
```diff
@@ -76,43 +75,17 @@ func (s Storage) BaseURL() *url.URL {
...
-		err := <-errChan
-		if err != nil {
-			errors = multierror.Append(errors, err)
-		}
-	}
-	close(errChan)
-
-	return errors
+	err := modupl.Upload(ctx, module, version, bytes.NewReader(info), bytes.NewReader(mod), zip, s.upload)
+	return err
```
robjloranger: Here you can just return the function call.

pkg/storage/module/upload_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+type UploadTests struct {
+	suite.Suite
+}
+
+func TestUpload(t *testing.T) {
+	suite.Run(t, new(UploadTests))
+}
+
+func (m *UploadTests) TestUploadTimeout() {
```
robjloranger: Same as before, `(u *UploadTests)` might be more idiomatic.

pkg/storage/gcp/lister.go
robjloranger: that was my comment from another PR :) but I haven't thought about it yet. It might in fact just appear hacky.

pkg/storage/module/delete_test.go
robjloranger: This one is more a question.

pkg/storage/azurecdn/storage.go
marpio: @arschles this would cause conflicts with the variable name. I'm not entirely happy with the pkg name so maybe you have some better idea?

pkg/storage/module/delete_test.go
marpio: You're right - the cast isn't needed.

pkg/storage/s3/storage.go
marpio: I wanted to leave it here because at some point we  have to check the error and update the list file. I'll write a comment here.

pkg/storage/module/delete_test.go
marpio: Delete returns an `error` and I want to check the no. of errors so I call `WrappedErrors()`

pkg/storage/gcp/lister.go
marpio: I actually think this is fine. I'll extract a func.

pkg/storage/gcp/saver.go
marpio: We need to update the list file before we return so I've put a comment here.

pkg/storage/azurecdn/storage.go
robjloranger: Maybe just `mod`, I think it's the `upl` part that makes it less clear. In the past I've done things like `athensModule` or `athModule` etc. Thoughts @arschles ?

marpio: mod would also cause conflicts

PR #272, opened 2018-07-14
Zeus: remove duplicate '/' handler assignment

robjloranger:
The home '/' route handler for the proxy app is assigned to `proxyHomeHandler` in (./cmd/proxy/app_proxy.go) line 11.
However on line 112 of (./cmd/proxy/app.go) it is reassigned to the `homeHandler`. Since we can't have two handlers I removed the one from `app.go` for using `homeHandler`.

I am wondering which handler is preferred however, as the 'homeHandler' also includes a new tracing child span and a sleep for one millisecond. If whichever of the two is not needed I will remove that as well.

### DIFF
--- a/cmd/proxy/actions/app.go
+++ b/cmd/proxy/actions/app.go
@@ -109,8 +109,6 @@ func App() (*buffalo.App, error) {
 		}
 		app.Use(T.Middleware())
 
-		app.GET("/", homeHandler)
-
 		if err := addProxyRoutes(app, store, mf); err != nil {
 			err = fmt.Errorf("error adding proxy routes (%s)", err)
 			return nil, err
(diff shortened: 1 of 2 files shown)


cmd/proxy/actions/app.go
```diff
@@ -109,8 +109,6 @@ func App() (*buffalo.App, error) {
 		}
 		app.Use(T.Middleware())
 
-		app.GET("/", homeHandler)
```
michalpristas: homeHandler  func will need to be removed as well

robjloranger: Cool, I was waiting for confirmation :smile: when we add tracing we can add a span to the `proxyHomeHandler` if desired.

PR #274, opened 2018-07-15
Athens: introduce log package

marwan-at-work:
This PR does two things: 

1. Creates a formatter for buffalo's logs, just so that we can follow the flow of handlers more clearly when developing/debugging only.

2. Creates our own `log` package which will be the broad logging mechanism for the whole Athens system. This PR just defines the behavior of logging so that it can be discussed/tweaked if necessary before changing all the logs in the code base. 

Once this PR is resolved, further prs should come in to spit logs through this package instead of the standard library

Fixes https://github.com/gomods/athens/issues/156

### DIFF
--- a/cmd/proxy/actions/app_proxy.go
+++ b/cmd/proxy/actions/app_proxy.go
@@ -3,18 +3,24 @@ package actions
 import (
 	"github.com/gobuffalo/buffalo"
 	"github.com/gomods/athens/pkg/download"
+	"github.com/gomods/athens/pkg/log"
 	"github.com/gomods/athens/pkg/module"
 	"github.com/gomods/athens/pkg/storage"
 )
 
-func addProxyRoutes(app *buffalo.App, storage storage.Backend, mf *module.Filter) error {
+func addProxyRoutes(
+	app *buffalo.App,
+	storage storage.Backend,
+	mf *module.Filter,
+	lggr *log.Logger,
+) error {
 	app.GET("/", proxyHomeHandler)
 
 	// Download Protocol
 	app.GET(download.PathList, download.ListHandler(storage, proxy))
-	app.GET(download.PathVersionInfo, cacheMissHandler(download.VersionInfoHandler(storage, proxy), app.Worker, mf))
-	app.GET(download.PathVersionModule, cacheMissHandler(download.VersionModuleHandler(storage), app.Worker, mf))
-	app.GET(download.PathVersionZip, cacheMissHandler(download.VersionZipHandler(storage), app.Worker, mf))
+	app.GET(download.PathVersionInfo, cacheMissHandler(download.VersionInfoHandler(storage, proxy), app.Worker, mf, lggr))
+	app.GET(download.PathVersionModule, cacheMissHandler(download.VersionModuleHandler(storage), app.Worker, mf, lggr))
+	app.GET(download.PathVersionZip, cacheMissHandler(download.VersionZipHandler(storage), app.Worker, mf, lggr))
 
 	app.POST("/admin/fetch/{module:[a-zA-Z./]+}/{owner}/{repo}/{ref}/{version}", fetchHandler(storage))
 	return nil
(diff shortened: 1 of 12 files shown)


cmd/proxy/actions/app_proxy.go
```diff
@@ -3,18 +3,24 @@ package actions
...
 	"github.com/gomods/athens/pkg/module"
 	"github.com/gomods/athens/pkg/storage"
 )
 
-func addProxyRoutes(app *buffalo.App, storage storage.Backend, mf *module.Filter) error {
+func addProxyRoutes(
+	app *buffalo.App,
+	storage storage.Backend,
+	mf *module.Filter,
+	lggr *log.Logger,
```
robjloranger: I like this, it's a lot more readable.

pkg/log/buffalo.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return &buffaloLogger{e}
+}
+
+func (bf *buffaloLogger) WithFields(fields map[string]interface{}) buffalo.Logger {
+	e := bf.FieldLogger.WithFields(fields)
+	return &buffaloLogger{e}
+}
+
+func fmtBuffaloErr(msg string) ([]byte, error) {
```
robjloranger: does this need to return a nil error? If it's only used at the package level it should probably just return the `[]byte` and the caller can return a nil in place of an error when necessary.

pkg/log/log.go
```diff
@@ -0,0 +1,68 @@
...
+
+	return &Logger{Logger: l}
+}
+
+// Entry is an abstraction to the
+// Logger and the logrus.Entry
+// so that *Logger always creates
+// an Entry copy which ensures no
+// Fields are being overwritten.
+type Entry interface {
```
michalpristas: can we separate this into entry.go

pkg/log/format.go
```diff
@@ -0,0 +1,23 @@
+package log
+
+import (
+	"github.com/sirupsen/logrus"
+)
+
+func getGCPFormatter() logrus.Formatter {
+	return &logrus.JSONFormatter{
+		FieldMap: logrus.FieldMap{
+			logrus.FieldKeyLevel: "severity",
```
michalpristas: 👍

pkg/log/buffalo.go
```diff
@@ -0,0 +1,70 @@
...
+		color.CyanString("handler:"),
+		entry.Data["method"],
+		entry.Data["path"],
+		status,
+	)
+
+	return []byte(str), nil
+}
+
+type buffaloLogger struct{ logrus.FieldLogger }
```
michalpristas: and separate this into buffalo_formatter.go logger.go

pkg/config/env/cloud_runtime.go
```diff
@@ -0,0 +1,11 @@
+package env
+
+import (
+	"github.com/gobuffalo/envy"
+)
+
+// CloudRuntime returns the Cloud Provider
+// underneath which the Proxy/Registry is running.
+func CloudRuntime() string {
```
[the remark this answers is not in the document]
marwan-at-work: @michalpristas It's a bit different because the Go Env can be "development", "staging", or "production". Meaning, you can potentially be on GCP but still running as "staging". CloudRuntime is just to say what cloud are you running on. "development" is overlapping from both because you can be running on your local machine. But that name can be changed to "local" or "none" if you think is better.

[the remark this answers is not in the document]
marwan-at-work: @robjloranger updated to "none"

robjloranger:
I :green_heart: that there is color output

michalpristas:
@marwan-at-work  was there a passing test?

michalpristas:
no i noticed failed test so wondered if CI passed, if they're ok then i have no problem with ignoring Netlify

marwan-at-work:
@michalpristas i believe the CI passed. But I still haven't figured out running the tests locally so I should start doing that 😅

robjloranger:
Ya CI passed but Netlify deploy failed. Although we should probably write some test for the log package. see #279

PR #278, opened 2018-07-15
Athens: remove c.Error

marwan-at-work:
This PR does two things: 

1. Replaces c.Error with c.Render because of https://github.com/gobuffalo/buffalo/issues/1171
2. Replace c.String with c.JSON so that we put quotes around the string and be application/json compliant 

When integration with vgo and clients becomes more important than now, all of our responses should conform to a specific way but for now it's not a biggie.


pkg/download/version_module.go
```diff
@@ -25,7 +26,8 @@ func VersionModuleHandler(getter storage.Getter) func(c buffalo.Context) error {
 
 		version, err := getter.Get(params.Module, params.Version)
 		if storage.IsNotFoundError(err) {
-			return c.Error(http.StatusNotFound, fmt.Errorf("%s@%s not found", params.Module, params.Version))
+			msg := fmt.Errorf("%s@%s not found", params.Module, params.Version)
```
robjloranger: this should be fmt.Sprintf

marwan-at-work: nice catch

cmd/olympus/actions/goget.go
```diff
@@ -30,7 +30,8 @@ func goGetMeta(c buffalo.Context, getter cdnmetadata.Getter) error {
 	}
 	loc, err := getter.Get(params.Module)
 	if err != nil {
-		return c.Error(http.StatusNotFound, fmt.Errorf("module %s does not exist", params.Module))
+		msg := fmt.Sprintf("module %s does not exist", params.Module)
```
michalpristas: just thinking about reusing ErrModuleVersionNotFound from storage package and returning that.

PR #280, opened 2018-07-15
abstract gcp bucket into interface

robjloranger:
This will set us up to create the mock for #263

### DIFF
--- a/pkg/storage/gcp/bucket.go
+++ b/pkg/storage/gcp/bucket.go
@@ -1,0 +1,21 @@
+package gcp
+
+import (
+	"context"
+	"io"
+)
+
+// Bucket defines the storage bucket TODO
+type Bucket interface {
+	// DeleteModule removes a module @ version from the bucket
+	DeleteModule(ctx context.Context, module, version string) error
+	// GetReader returns a reader for the module filename and an error if any
+	GetReader(ctx context.Context, filename string) (io.ReadCloser, error)
+	// GetWriter returns a new writer for the module filename
+	// This will overwrite any exists file stored under the same module, version
+	GetWriter(ctx context.Context, filename string) io.WriteCloser
+	// ListVersions returns a slice of version strings for a module
+	ListVersions(ctx context.Context, module string) ([]string, error)
+	// ObjectExists returns true if the module @ version exists in the bucket
+	ObjectExists(ctx context.Context, module, version string) bool
+}
--- a/pkg/storage/gcp/bucket.go
+++ b/pkg/storage/gcp/bucket.go
@@ -5,17 +5,17 @@ import (
 	"io"
 )
 
-// Bucket defines the storage bucket TODO
+// Bucket provides file operations for a Google Cloud Storage resource.
 type Bucket interface {
-	// DeleteModule removes a module @ version from the bucket
-	DeleteModule(ctx context.Context, module, version string) error
-	// GetReader returns a reader for the module filename and an error if any
-	GetReader(ctx context.Context, filename string) (io.ReadCloser, error)
-	// GetWriter returns a new writer for the module filename
-	// This will overwrite any exists file stored under the same module, version
-	GetWriter(ctx context.Context, filename string) io.WriteCloser
-	// ListVersions returns a slice of version strings for a module
+	// Delete removes the file module/@v/version.extension
+	Delete(ctx context.Context, module, version string) error
+	// Open returns a reader for module/@v/version.extension and any error
+	Open(ctx context.Context, module, version, extension string) (io.ReadCloser, error)
+	// Write returns a new writer for module/@v/version.extension
+	// This writer will overwrite any existing file stored at the same path
+	Write(ctx context.Context, module, version, extension string) io.WriteCloser
+	// ListVersions returns a slice of versions for a module and any error
 	ListVersions(ctx context.Context, module string) ([]string, error)
-	// ObjectExists returns true if the module @ version exists in the bucket
-	ObjectExists(ctx context.Context, module, version string) bool
+	// Exists returns true if the module @ version exists
+	Exists(ctx context.Context, module, version string) bool
 }
--- a/pkg/storage/gcp/bucket_cloud.go
+++ b/pkg/storage/gcp/bucket_cloud.go
@@ -1,0 +1,78 @@
+package gcp
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"strings"
+
+	"cloud.google.com/go/storage"
+	"github.com/gomods/athens/pkg/config"
+	s "github.com/gomods/athens/pkg/storage"
+	multierror "github.com/hashicorp/go-multierror"
+	"google.golang.org/api/iterator"
+)
+
+type gcpBucket struct {
+	*storage.BucketHandle
+}
+
+func (b *gcpBucket) DeleteModule(ctx context.Context, module, version string) error {
+	if !b.ObjectExists(ctx, module, version) {
+		return s.ErrVersionNotFound{Module: module, Version: version}
+	}
+	var errs error
+	if err := b.Object(config.PackageVersionedName(module, version, "mod")).Delete(ctx); err != nil {
+		errs = multierror.Append(errs, err)
+	}
+	if err := b.Object(config.PackageVersionedName(module, version, "info")).Delete(ctx); err != nil {
+		errs = multierror.Append(errs, err)
+	}
+	if err := b.Object(config.PackageVersionedName(module, version, "zip")).Delete(ctx); err != nil {
+		errs = multierror.Append(errs, err)
+	}
+	return errs
+}
+
+func (b *gcpBucket) GetReader(ctx context.Context, filename string) (io.ReadCloser, error) {
+	return b.Object(filename).NewReader(ctx)
+}
+
+func (b *gcpBucket) GetWriter(ctx context.Context, filename string) io.WriteCloser {
+	return b.Object(filename).NewWriter(ctx)
+}
+
+func (b *gcpBucket) ListVersions(ctx context.Context, module string) ([]string, error) {
+	it := b.Objects(ctx, &storage.Query{Prefix: module})
+
+	versions := make([]string, 0, 10)
+	for {
+		attrs, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("could not iterate over query: %s", err)
+		}
+
+		// kinda hacky looking at this time
+		if strings.HasSuffix(attrs.Name, ".info") {
+			segments := strings.Split(attrs.Name, "/")
+			// version should be last segment w/ .info suffix
+			last := segments[len(segments)-1]
+			version := strings.TrimSuffix(last, ".info")
+			versions = append(versions, version)
+		}
+	}
+
+	if len(versions) == 0 {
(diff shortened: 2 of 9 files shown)


robjloranger:
It's not a lot of use to anyone but the tests do pass still. Soon you won't have to take my word for it :laughing:

pkg/storage/gcp/gcp.go
```diff
@@ -10,9 +10,16 @@ import (
 	"google.golang.org/api/option"
 )
 
+// Bucket defines the subset of functionality provided by the
+// cloud.google.com/go/storage.BucketHandle required by this package
+type Bucket interface {
+	Object(name string) *storage.ObjectHandle
+	Objects(ctx context.Context, q *storage.Query) *storage.ObjectIterator
+}
+
```
marpio: Wouldn't it be easier to do something like this?
```
type Bucket interface {
	GetWriter(ctx context.Context, fileName string) *storage.Writer
	GetItems(ctx context.Context, module string) []string
}

```
and then something similar to:
```
func (b *gcpClient) GetWriter(ctx context.Context, fileName string) *storage.Writer {
  return  b.bkt.Object(fileName).NewWriter(ctx)
}
```
or some fake.
Otherwise you have to create all the intermediary objects like the ObjectHandle etc.

robjloranger: Much easier :grin: I like it, I was dreading the next step with my impl.

pkg/storage/gcp/bucket.go
```diff
@@ -0,0 +1,21 @@
+package gcp
+
+import (
+	"context"
+	"io"
+)
+
+// Bucket defines the storage bucket TODO
```
marwan-at-work: maybe a little explanation about what a bucket means in GCP terms would be great for documentation :)

pkg/storage/gcp/bucket.go
```diff
@@ -0,0 +1,21 @@
...
+type Bucket interface {
+	// DeleteModule removes a module @ version from the bucket
+	DeleteModule(ctx context.Context, module, version string) error
+	// GetReader returns a reader for the module filename and an error if any
+	GetReader(ctx context.Context, filename string) (io.ReadCloser, error)
+	// GetWriter returns a new writer for the module filename
+	// This will overwrite any exists file stored under the same module, version
```
marwan-at-work: existing*

pkg/storage/gcp/bucket.go
```diff
@@ -0,0 +1,21 @@
...
+	"context"
+	"io"
+)
+
+// Bucket defines the storage bucket TODO
+type Bucket interface {
+	// DeleteModule removes a module @ version from the bucket
+	DeleteModule(ctx context.Context, module, version string) error
+	// GetReader returns a reader for the module filename and an error if any
+	GetReader(ctx context.Context, filename string) (io.ReadCloser, error)
```
marwan-at-work: `GetReader` is a bit ambiguous, what about `GetObject` or `Open` etc?

pkg/storage/gcp/bucket.go
```diff
@@ -0,0 +1,21 @@
+package gcp
+
+import (
+	"context"
+	"io"
+)
+
+// Bucket defines the storage bucket TODO
+type Bucket interface {
```
marwan-at-work: I'm a tiny bit confused about this interface. Some methods seem to work directly with a whole module (DeleteModule, ListVersions) but some are about files/objects within modules? If that's true that maybe a little documentation about why that is would be great. Otherwise, maybe having this interface not know anything about modules and just work with GCP terminology, or the other way around.

pkg/storage/gcp/bucket_cloud.go
```diff
@@ -0,0 +1,78 @@
...
+	s "github.com/gomods/athens/pkg/storage"
+	multierror "github.com/hashicorp/go-multierror"
+	"google.golang.org/api/iterator"
+)
+
+type gcpBucket struct {
+	*storage.BucketHandle
+}
+
+func (b *gcpBucket) DeleteModule(ctx context.Context, module, version string) error {
```
marwan-at-work: Does GCP Storage have nesting capabilities? If so, an enhancement would be to nest the three "objects" below within one path and so you don't have to do 3 roundtrips just to delete a module, you can just delete the parent path. Doesn't have to be in this pr of course.

pkg/storage/gcp/deleter.go
```diff
@@ -3,29 +3,17 @@ package gcp
...
-	multierror "github.com/hashicorp/go-multierror"
 )
 
 // Delete implements the (./pkg/storage).Deleter interface and
 // removes a version of a module from storage. Returning ErrNotFound
 // if the version does not exist.
 func (s *Storage) Delete(module, version string) error {
 	ctx := context.Background()
-	if exists := s.Exists(module, version); !exists {
+	if exists := s.bucket.ObjectExists(ctx, module, version); !exists {
```
marwan-at-work: we can remove this whole check since `s.bucket` already does it.

pkg/storage/gcp/bucket.go
robjloranger: I'll fix it. It was because I didn't want the extra code to convert module, version into the file path. It will just require the signature to include the file extension.

robjloranger: So to clarify, it does not operate on a file within the module. but the module at the path filename.

pkg/storage/gcp/bucket.go
robjloranger: I'me trying to decide on better names for both. They both only return the reader/writer, how about `Open` and `Write`.

pkg/storage/gcp/checker.go
```diff
@@ -2,20 +2,11 @@ package gcp
...
 import (
 	"context"
-
-	"github.com/gomods/athens/pkg/config"
 )
 
 // Exists implements the (./pkg/storage).Checker interface
 // returning true if the module at version exists in storage
 func (s *Storage) Exists(module, version string) bool {
 	ctx := context.Background()
```
marwan-at-work: should the `storage.Backend` interface take a `context.Context`  in all of its methods? They are all remote calls by default so imo they should. Otherwise, we should at least do a `context.WithTimeout` instead of Background. Doesn't have to be in this pr.

robjloranger: That's a great call. We should open an issue regarding the timeout. Not so backends are network. Most but not file or memory. However we do intend to thread context down the whole stack so it can't hurt to start somewhere that makes sense. Then just throw a timeout on the context before passing it to one in the backend calls. Keep in mind that the file uploads already share a default timeout of five minutes.

robjloranger: For sure another PR though

pkg/storage/gcp/bucket_cloud.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"cloud.google.com/go/storage"
+	"github.com/gomods/athens/pkg/config"
+	s "github.com/gomods/athens/pkg/storage"
+	multierror "github.com/hashicorp/go-multierror"
+	"google.golang.org/api/iterator"
+)
+
+// gcpBucket implements the (./pkg/storage/gcp).Bucket interface
+type gcpBucket struct {
```
michalpristas: @robjloranger  just an after merge note: in this case it is ok to have iface and impl together

pkg/storage/gcp/bucket.go
```diff
@@ -0,0 +1,21 @@
...
+	// Open returns a reader for module/@v/version.extension and any error
+	Open(ctx context.Context, module, version, extension string) (io.ReadCloser, error)
+	// Write returns a new writer for module/@v/version.extension
+	// This writer will overwrite any existing file stored at the same path
+	Write(ctx context.Context, module, version, extension string) io.WriteCloser
+	// ListVersions returns a slice of versions for a module and any error
+	ListVersions(ctx context.Context, module string) ([]string, error)
+	// Exists returns true if the module @ version exists
+	Exists(ctx context.Context, module, version string) bool
+}
```
marpio: I think it would be more elegant if we could decouple the `Bucket` from our domain. i.e. the Bucket doesn't have to know anything about `module, version, extension` it knows only `path`. i.e.
`Open(ctx context.Context, path string) (io.ReadCloser, error)`

This would make the parallel uploads much easier to implement because I could handle uploading of the different parts (info, mod, zip) inside of the Upload func.

pkg/storage/gcp/bucket_cloud.go
```diff
@@ -0,0 +1,79 @@
...
+	if err := b.Object(config.PackageVersionedName(module, version, "info")).Delete(ctx); err != nil {
+		errs = multierror.Append(errs, err)
+	}
+	if err := b.Object(config.PackageVersionedName(module, version, "zip")).Delete(ctx); err != nil {
+		errs = multierror.Append(errs, err)
+	}
+	return errs
+}
+
+func (b *gcpBucket) Open(ctx context.Context, module, version, extension string) (io.ReadCloser, error) {
```
marpio: This could be `Open(ctx context.Context, path string) (io.ReadCloser, error)`

pkg/storage/gcp/bucket_cloud.go
```diff
@@ -0,0 +1,79 @@
...
+	return b.Object(config.PackageVersionedName(module, version, extension)).NewReader(ctx)
+}
+
+func (b *gcpBucket) Write(ctx context.Context, module, version, extension string) io.WriteCloser {
```
marpio: `Write(ctx context.Context, path string) io.WriteCloser`

pkg/storage/gcp/bucket_cloud.go
```diff
@@ -0,0 +1,79 @@
...
+
+		// kinda hacky looking at this time
+		if strings.HasSuffix(attrs.Name, ".info") {
+			segments := strings.Split(attrs.Name, "/")
+			// version should be last segment w/ .info suffix
+			last := segments[len(segments)-1]
+			version := strings.TrimSuffix(last, ".info")
+			versions = append(versions, version)
+		}
+	}
```
marpio: I wouldn't put the logic here because it will be never tested. Either create a separate func doing all the spliting etc. and use it here or return just a collection of `attrs.Name` and do the spliting inside lister.go

pkg/storage/gcp/bucket_cloud.go
robjloranger: I only split them because there will soon be the mock implementation as well.

pkg/storage/gcp/bucket_cloud.go
```diff
@@ -0,0 +1,79 @@
...
+		return s.ErrVersionNotFound{Module: module, Version: version}
+	}
+	var errs error
+	if err := b.Object(config.PackageVersionedName(module, version, "mod")).Delete(ctx); err != nil {
+		errs = multierror.Append(errs, err)
+	}
+	if err := b.Object(config.PackageVersionedName(module, version, "info")).Delete(ctx); err != nil {
+		errs = multierror.Append(errs, err)
+	}
+	if err := b.Object(config.PackageVersionedName(module, version, "zip")).Delete(ctx); err != nil {
```
marpio: I think `Delete` should delete only one file just like Write writes only one file. Also because it will not be tested.

robjloranger: I agree, and the check for not found should be moved

PR #283, opened 2018-07-17
Download module version from CDN

### LINKED ISSUE #240: Implement cdn.Download
At the moment the impl. looks  like this:
```
func Download(url, module, version string) (*storage.Version, error) {
	// TODO
	return nil, nil
}
```
It should download module from the cdn.

marpio:
Fixes #240 
Enables module download from every source providing info, mod and zip endpoints defined by download protocol.

### DIFF
--- a/cmd/olympus/actions/merge_db.go
+++ b/cmd/olympus/actions/merge_db.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/gomods/athens/pkg/config/env"
-
 	"github.com/gomods/athens/pkg/eventlog"
 	"github.com/gomods/athens/pkg/module"
 	"github.com/gomods/athens/pkg/storage"
--- a/cmd/olympus/actions/merge_db.go
+++ b/cmd/olympus/actions/merge_db.go
@@ -22,10 +22,10 @@ import (
 //		- Delete operation adds tombstone to module metadata k/v store
 //
 // Both could be fixed by putting each 'for' loop into a (global) critical section
-func mergeDB(ctx context.Context, originURL string, diff dbDiff, eLog eventlog.Eventlog, storage storage.Backend) error {
+func mergeDB(ctx context.Context, originURL string, diff dbDiff, eLog eventlog.Eventlog, storage storage.Backend, downloader cdn.ModuleDownloader) error {
 	var errors error
 	for _, added := range diff.Added {
-		if err := add(ctx, added, originURL, eLog, storage); err != nil {
+		if err := add(ctx, added, originURL, eLog, storage, downloader); err != nil {
 			errors = multierror.Append(errors, err)
 		}
 	}
@@ -42,7 +42,7 @@ func mergeDB(ctx context.Context, originURL string, diff dbDiff, eLog eventlog.E
 	return errors
 }
 
-func add(ctx context.Context, event eventlog.Event, originURL string, eLog eventlog.Eventlog, storage storage.Backend) error {
+func add(ctx context.Context, event eventlog.Event, originURL string, eLog eventlog.Eventlog, storage storage.Backend, downloader cdn.ModuleDownloader) error {
 	if _, err := eLog.ReadSingle(event.Module, event.Version); err != nil {
 		// the module/version already exists, is deprecated, or is
 		// tombstoned, so nothing to do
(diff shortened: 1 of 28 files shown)


pkg/cdn/download.go
```diff
@@ -1,13 +1,93 @@
...
-	// TODO
-	return nil, nil
+func Download(ctx context.Context, baseURL, module, version string) (*storage.Version, error) {
+	tctx, cancel := context.WithTimeout(ctx, env.Timeout())
+	defer cancel()
+	getReq := func(ext string) (*http.Request, error) {
+		return getRequest(tctx, baseURL, module, version, ext)
+	}
+
+	infoReq, err := getReq(".info")
```
michalpristas: can you do downloads in parallel?

pkg/cdn/download.go
```diff
@@ -1,13 +1,93 @@
...
 // Download downloads the module/version from url. Returns a storage.Version
 // representing the downloaded module/version or a non-nil error if something
 // went wrong
-func Download(url, module, version string) (*storage.Version, error) {
-	// TODO
-	return nil, nil
+func Download(ctx context.Context, baseURL, module, version string) (*storage.Version, error) {
+	tctx, cancel := context.WithTimeout(ctx, env.Timeout())
+	defer cancel()
+	getReq := func(ext string) (*http.Request, error) {
```
michalpristas: i have a mixed feeling about this

pkg/cdn/download.go
```diff
@@ -1,13 +1,93 @@
...
+	ver := storage.Version{
+		Info: info,
+		Mod:  mod,
+		Zip:  zipRes.Body,
+	}
+	return &ver, nil
+}
+
+func getResBytes(req *http.Request) ([]byte, error) {
+	res, err := http.DefaultClient.Do(req)
```
michalpristas: careful about defaultClient it does not have timeout set up and can hang. 
if we stick with ctx.WithTimeout this should not be a problem, but we can set Timeout also here just to make sure we work if somebody removes WithTimeout

cmd/olympus/actions/push.go
```diff
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 
+	"github.com/gomods/athens/pkg/cdn"
+
```
marwan-at-work: remove space

pkg/cdn/download.go
```diff
@@ -1,13 +1,93 @@
...
+}
+
+func join(baseURL string, module, version, ext string) (string, error) {
+	u, err := url.Parse(baseURL)
+	if err != nil {
+		return "", err
+	}
+	packageVersionedName := config.PackageVersionedName(module, version, ext)
+	u.Path = path.Join(u.Path, packageVersionedName)
+	return fmt.Sprint(u), nil
```
marwan-at-work: just call `u.String()`

pkg/module/download.go
```diff
@@ -0,0 +1,131 @@
...
+		zipReq, err := getRequest(tctx, baseURL, module, version, ".zip")
+		if err != nil {
+			zip, zipErr = nil, err
+			return
+		}
+		zip, zipErr = getResBody(zipReq, timeout)
+	}()
+	wg.Wait()
+
+	if infoErr != nil {
```
robjloranger: You could probably use multierror here instead of having one error/check for each file.

pkg/module/download.go
```diff
@@ -0,0 +1,137 @@
...
+		if err != nil {
+			zip, zipErr = nil, err
+			return
+		}
+		zip, zipErr = getResBody(zipReq, timeout)
+	}()
+	wg.Wait()
+
+	var errors error
+	if infoErr != nil {
```
robjloranger: I think you can append the errors during the check inside the anonymous funcs for each file, then just check once here if `errors` is not nil and return it if needed. It would save the extra checks for each value

marpio: @robjloranger is `multierror.Append` thread-safe?

robjloranger: Good call, I don't think so.

cmd/olympus/actions/merge_db.go
```diff
@@ -5,8 +5,10 @@ import (
 	"log"
 	"time"
 
-	"github.com/gomods/athens/pkg/cdn"
+	"github.com/gomods/athens/pkg/config/env"
+
```
robjloranger: another space, just a tiny thing.

robjloranger:
It looks like all concerns have been addressed in this one. @arschles and @marwan-at-work can you confirm?

Issue #285, opened 2018-07-17
enable in-memory workers in the proxy


arschles:
instead of always using redis (see [here](https://github.com/gomods/athens/blob/eb0768f3fe476dd68f4635a7ad60c3e059d0ff44/cmd/proxy/actions/app.go#L145)) we should allow in-memory proxies for local development

kengbailey:
I'll go ahead and take this as my first PR. I'm able to build the run the proxy using `buffalo dev` so I should be all setup.

Seeing as this has the "good first issue" tag, I don't want to over think this. The solutions is probably much simpler than I'm thinking right now. I see that the proxy is currently using `buffalo/worker.Worker` with a redis backend. I have two questions:

1. By in memory you mean that the worker queue is stored in memory( as opposed to redis)?
2. Looking at the [buffalo workers doc](https://www.gobuffalo.io/en/docs/workers/), would worker.Simple be the best solution for this?

arschles:
Hey Ken - awesome that you’re joining us - welcome!!!

You asked and answered your two questions 100% correctly 😊

One note to keep in mind as you implement - can you make sure that folks
can configure whether they want in-mem or redis using an environment
variable?

robjloranger:
:+1: see https://github.com/gomods/athens/blob/master/pkg/config/env/ for examples regarding our env package, https://github.com/gomods/athens/blob/master/pkg/config/env/redis.go for the current redis ones.

kengbailey:
@arschles @robjloranger Thanks 😊👍

arschles:
Totally cool Ken. If you get some time you come back and want to
contribute, we’ll be here.

Meanwhile, have an amazing holiday! 🎉

kengbailey:
I appreciate that!

kdvy:
I'd like to take this as my first contribution to Athens if that's okay.

Thanks to the investigation done by @kengbailey, I think I have a good idea of what to do. 

I noticed that Olympus also has the same redis backing for processing push notifications, would it be useful to add a different env variable to control whether we use a worker.Simple or Redis there as well ?

kdvy:
Had a quiet night in so I decided to write the code for both Olympus and Athens. Happy to make any changes/ remove as necessary.

PR #286, opened 2018-07-17
WIP: enable in-memory workers

arschles:
So far, only hard-coding in-memory workers in the proxy. Plan is to enable them via a config flag, mostly for local dev.

Fixes https://github.com/gomods/athens/issues/285


arschles:
This might be obsolete after #290 is done. Leaving open because I'm not sure if/why we'd still need workers at that point

arschles:
Thanks to the awesome @kdvy, I can close this in favor of #400

Issue #279, opened 2018-07-15
Tests for Log pkg


robjloranger:
Recently #274 added a custom log package.

We need some basic unit tests.

PR #289, opened 2018-07-17
Athens: add log tests to cloud formatters

marwan-at-work:
Fixes https://github.com/gomods/athens/issues/279

### DIFF
--- a/pkg/log/log_test.go
+++ b/pkg/log/log_test.go
@@ -1,0 +1,92 @@
+package log
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+type input struct {
+	name          string
+	cloudProvider string
+	level         string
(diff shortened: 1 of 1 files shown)


pkg/log/log_test.go
```diff
@@ -0,0 +1,92 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			lggr := New(tc.cloudProvider, tc.level)
+			var buf bytes.Buffer
+			lggr.Out = &buf
+			e := lggr.WithFields(tc.fields)
+			tc.logFunc(e)
+			out := buf.String()
+			expected := tc.output
+			if strings.Contains(expected, "%v") {
+				expected = fmt.Sprintf(tc.output, time.Now().Format(time.RFC3339))
```
Kunde21: `s/tc.output/expected/`

arschles: This would be a good place to use https://godoc.org/github.com/stretchr/testify/require. would you mind doing that?

Issue #290, opened 2018-07-17
Big Kahuna: Do cache fills synchronously


arschles:
We've had some discussion about whether to do cache fills synchronously or asynchronously. We've discussed how we can make a `vgo get` call still succeed if we do cache fills asynchronously (and return a 404 to the user), and created https://github.com/golang/go/issues/26334 to discuss a possibility.

I've mostly argued for asynchronously filling caches until now, but I'm changing my mind because I think doing so makes the CLI (`vgo`, soon to be Go modules in 1.11), the architecture, and the conceptual model of both the registry and the proxy easier to grasp.

Here's exactly what I'm suggesting on the servers:

- For a cache miss on a `GET /list` request, we return only the tags from the upstream. No fetching `go.mod` or other assets
    - In the registry, "the upstream" means the VCS
    - In the proxy, "the upstream" means the VCS or the registry, or just 404-ing to the client, depending on exclude rules
- For a cache miss on a `/vX.Y.Z.info`, `vX.Y.Z.info` or `vX.Y.Z.zip` request, we fetch and store all three assets and return the appropriate one

With this machinery in place, the CLI won't need to consider whether to "back off" to the VCS. If it's not in the proxy/registry, it doesn't exist, and that invariant makes https://github.com/golang/go/issues/26334 no longer necessary.

marwan-at-work:
@arschles if you haven't started working on it already, I've already written some of the code for how you described `/latest`. I can push that as a separate PR.

arschles:
@marwan-at-work I'd love that!

I'm gonna work in increments starting with moving the fetcher logic in https://github.com/gomods/athens/pull/291 and then to putting synchronous fetches into each endpoint, so I'll skip the `/list` endpoint if you've pushed it up already

arschles:
We're breaking this up into a few pieces:

- Cleanup: refactor the fetcher logic (https://github.com/gomods/athens/pull/291)
- Cleanup: remove the proxy fetch handler (https://github.com/gomods/athens/pull/314)
- Install synchronous fetcher logic in the various endpoints as necessary
- Implement the `/list` endpoint (https://github.com/gomods/athens/pull/300)

arschles:
Ok right now I have a ton of PRs up that attack this problem, and I've opened some against other branches. Here's how it all looks:

- The root of everything is https://github.com/gomods/athens/pull/291. That's the thing that needs reviewing first 😄 
- #319, #316 and #314 are opened against #291 and they need #291 merged before I can continue with them

Issue #288, opened 2018-07-17
make the proxy download metadata and code from the registry using the standard vgo endpoints


arschles:

michalpristas:
closing as we are parking olympus, we can re-open later if needed

PR #291, opened 2018-07-18
Moving the fetch logic into 'pkg/module'

arschles:
This copies the stuff in `./pkg/repo` into `./pkg/module`, but leaves `./pkg/repo` (see #314 and #316 for removing `./pkg/repo`).

It also refactors the interface for how we download code from an upstream (VCS), access it, and prepare it for upload to storage/CDN/wherever. During the copy from `./pkg/repo`, I only retained the `go get` implementation in the new package, and not the Github API implementation (see #288 for why)

I am doing this refactor to prepare for #290, where I'll use the new functionality to synchronously fetch packages. I'll do that in a separate PR 😄 

TODO:

- [x] Make the `(./pkg/module).diskRef` look in the global cache for the `.info`, `.mod`, and `.zip` files (right now I believe it looks in the wrong place)
- [x] ~Remove `./pkg/repo`~ (see #316)

This is part of #290 
There are a bunch of PRs opened against this branch. Check out https://github.com/gomods/athens/issues/290#issuecomment-406735602 for what they are


robjloranger:
This one is meaty, I'll take a proper look tomorrow morning sometime. There is a failing test though.

.gitignore
```diff
@@ -24,3 +24,6 @@ cmd/proxy/proxy
 cmd/olympus/olympus
 test-keys.json
 tmp
+cmd/olympus/bin/olympus
+cmd/proxy/bin/proxy
+.vs
```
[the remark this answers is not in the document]
arschles: @robjloranger good catch on these lines. I am going to back these out and just wait for #303 to go in, so we can discuss there

arschles:
@robjloranger thanks! I'll fix that test today and hopefully finish this up as well.

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,169 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/afero"
+)
+
+const (
+	tmpRepoDir = "%s-%s" // owner-repo-ref
```
robjloranger: I think this is probably fine as is, but at first I wasn't sure it the comment was meant to reference three parts of the string. If you agree it's a bit confusing, maybe something like `owner-ref`, `owner-version` etc

pkg/module/ref.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/gomods/athens/pkg/storage"
+)
+
+// Ref points to a module somewhere
+type Ref interface {
+	// Clear frees the storage & resources that the module uses. Calls to Read after you call
+	// this function may fail, regardless of whether this function returns nil or not
+	Clear() error
+	// Read reads the module into memory and returns it. The caller should call
+	// the returned function.
```
robjloranger: Which function? Something in the `storage.Version`?

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,169 @@
...
+
+	cmd := exec.Command("vgo", "get", fullURI)
+	// PATH is needed for vgo to recognize vcs binaries
+	// this breaks windows.
+	cmd.Env = []string{"PATH=" + os.Getenv("PATH"), gopathEnv, cacheEnv, disableCgo}
+	cmd.Dir = repoRoot
+
+	packagePath := filepath.Join(gopath, "src", "mod", "cache", "download", repoURI, "@v")
+
+	o, err := cmd.CombinedOutput()
```
robjloranger: I'm wondering if instead of the error check and eventual return, we can just have the switch.

Since the default case is the same as the return, and if the cmd output contains the limit error we return our custom error.
The only other check is on the files existing, in which case the only time the default case will be executed is when there is some actual error from the vgo get command.

marpio: I think you don't want to do these checks (isLimit and checkFiles) unnecessarily.
The alternative would be to do something like 
`case err != nil && isLimit`
`case err != nil && checkFiles`
but I'm not sure it's that much better.

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		switch {
+		case isLimitHit(o):
+			// github quota exceeded
+			return packagePath, ErrLimitExceeded
+		case checkFiles(fs, packagePath, version) == nil:
```
marpio: Shouldn't this be `case checkFiles(fs, packagePath, version) != nil:`

pkg/module/disk_ref.go
```diff
@@ -0,0 +1,73 @@
...
+//
+// You should always call this function after you fetch a module into a DiskRef
+func (d *diskRef) Clear() error {
+	d.closer()
+	return d.fs.RemoveAll(d.root)
+}
+
+// read is the Ref interface implementation.
+func (d *diskRef) Read() (*storage.Version, error) {
+	ver := &storage.Version{}
```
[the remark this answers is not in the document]
arschles: @robjloranger so would you prefer that I do `var ver *storage.Version` at the beginning of the function?

pkg/module/go_get_fetcher.go
arschles: @robjloranger I'll abstract this into a function so that it's much easier to read and more type-safe

pkg/module/go_get_fetcher.go
arschles: good catch, change incoming

pkg/module/go_get_fetcher.go
arschles: I've simplified this to not do the switch inside the if. hopefully it's a bit more clear what's going on, even if it's more nested

pkg/module/disk_ref.go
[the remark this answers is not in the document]
robjloranger: No because that would still be a pointer type, I find it's more clear to read the return of a variable address in place of returning a variable that is not clearly a pointer. i.e. `var ver storage.Version` then on return `return &ver`

marwan-at-work:
@arschles Besides the above comments, I have a bit of an important question: how are going to be able to `/list` modules through `go get`? Is there a command that lets you list all the upstream tags for all different VCS systems?

marwan-at-work:
@arschles my pr list will only cover the github part, i've also implemented bitbucket locally. But we might want to borrow from Go's code if we want to cover *all* of git. I've looked into the code and it's not and easy cut/paste. But should be doable. I have a bunch of thoughts as I spent sometime looking at all of this this weekend.

pkg/module/disk_ref.go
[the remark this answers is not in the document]
arschles: ah, ok got it. I'll make that change

pkg/module/disk_ref.go
```diff
@@ -0,0 +1,83 @@
...
+
+	infoFile, err := d.fs.Open(filepath.Join(d.root, fmt.Sprintf("%s.info", d.version)))
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	d.filesToClose = append(d.filesToClose, infoFile)
```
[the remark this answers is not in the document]
arschles: good point. `diskRef.Read()` should work regardless of how many times you call it, but in this case it'll open up 3 file handles on each call and not close them. bummer! fix incoming

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// given a filesystem, gopath, repository root, module and version, runs 'vgo get'
+// on module@version from the repoRoot with GOPATH=gopath, and returns the location
+// of the module cache. returns a non-nil error if anything went wrong. always returns
+// the location of the module cache so you can delete it if necessary
+func getSources(fs afero.Fs, gopath, repoRoot, module, version string) (string, error) {
+	version = strings.TrimPrefix(version, "@")
+	if !strings.HasPrefix(version, "v") {
+		version = "v" + version
```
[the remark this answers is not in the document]
arschles: yea, doesn't this do that?

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	uri := strings.TrimSuffix(module, "/")
+
+	fullURI := fmt.Sprintf("%s@%s", uri, version)
+
+	gopathEnv := fmt.Sprintf("GOPATH=%s", gopath)
+	cacheEnv := fmt.Sprintf("GOCACHE=%s", filepath.Join(gopath, "cache"))
+	disableCgo := "CGO_ENABLED=0"
+
+	cmd := exec.Command("vgo", "get", fullURI)
```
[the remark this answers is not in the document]
arschles: :+1: - great idea. change incoming

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,156 @@
...
+	if !strings.HasPrefix(version, "v") {
+		version = "v" + version
+	}
+	uri := strings.TrimSuffix(module, "/")
+
+	fullURI := fmt.Sprintf("%s@%s", uri, version)
+
+	gopathEnv := fmt.Sprintf("GOPATH=%s", gopath)
+	cacheEnv := fmt.Sprintf("GOCACHE=%s", filepath.Join(gopath, "cache"))
+	disableCgo := "CGO_ENABLED=0"
```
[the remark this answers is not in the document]
arschles: I copied this code from `./pkg/repo` so I'm not entirely familiar, but my guess would be so that we can run this code inside of non-libc runtimes, like an `alpine` docker container

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,156 @@
...
+	cmd.Env = []string{"PATH=" + os.Getenv("PATH"), gopathEnv, cacheEnv, disableCgo}
+	cmd.Dir = repoRoot
+
+	packagePath := filepath.Join(gopath, "src", "mod", "cache", "download", module, "@v")
+
+	o, err := cmd.CombinedOutput()
+	if err != nil {
+		// github quota exceeded
+		if isLimitHit(o) {
+			return packagePath, errors.New("github API limit hit")
```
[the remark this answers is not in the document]
arschles: :+1: - change incoming

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		// one or more of the expected files doesn't exist
+		if err := checkFiles(fs, packagePath, version); err != nil {
```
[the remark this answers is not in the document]
arschles: Yes I believe so. Change incoming

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		// another error in the output
+		return packagePath, err
+	}
+
+	return packagePath, nil
+}
+
+func checkFiles(fs afero.Fs, path, version string) error {
+	if _, err := fs.Stat(filepath.Join(path, version+".mod")); err != nil {
```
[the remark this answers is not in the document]
arschles: Will do, I'll embed it.

pkg/module/fetcher.go
```diff
@@ -0,0 +1,10 @@
+package module
+
+// Fetcher fetches module from an upstream source
+type Fetcher interface {
```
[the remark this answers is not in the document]
arschles: not sure I understand correctly. would the `DownloadProtocol` interface replace both the `Fetcher` and `Ref`? If I misunderstood that, then everything below is totally worthless 😆. Otherwise, read on!

I'm asking because I like the separation between fetching and reading. The separation is obviously useful for disk based fetch/read scenarios (i.e. the thing gets put on disk once and read N times), but also could be useful if we wanted to somehow read directly into a CDN or somewhere else (if we had functionality to do that in the future)

From a super subjective readability perspective, fetching, then getting a reference to a place that the stuff was fetched, and then being able to read from that reference makes sense to me.

arschles:
@marwan-at-work I think the best we can do at the moment is what you do in your PR. Do you happen to know if there's a PR in golang/go for adding the ability to list versions for a module (and if not would you be cool creating one)?

arschles:
Ok I think I addressed everyone's comments but CI is failing now 😆 

Once I get on the plane (🤞 for wifi!) I'll look into that

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,159 @@
...
+
+// Hacky thing makes vgo not to complain
+func prepareStructure(fs afero.Fs, repoRoot string) error {
+	// vgo expects go.mod file present with module statement or .go file with import comment
+	gomodPath := filepath.Join(repoRoot, "go.mod")
+	gomodContent := []byte("module mod")
+	if err := afero.WriteFile(fs, gomodPath, gomodContent, 0666); err != nil {
+		return err
+	}
+
```
robjloranger: per the comment on line 80 above, do we need both files to make vgo happy? or just one of the two?

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,159 @@
...
+	// setup the module with barebones stuff
+	if err := prepareStructure(g.fs, modPath); err != nil {
+		// TODO: return a ref for cleaning up the goPathRoot
+		return nil, err
+	}
+
+	cachePath, err := getSources(g.goBinaryName, g.fs, goPathRoot, modPath, mod, ver)
+	if err != nil {
```
robjloranger: you can simplify all of this to just
```go
return newDiskRef(g.fs, cachePath, ver), err
```
the error will be nil or not and you want to return the ref either way.

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,159 @@
...
+		// TODO: return a ref that cleans up the goPathRoot
+		return newDiskRef(g.fs, cachePath, ver), err
+	}
+	// TODO: make sure this ref also cleans up the goPathRoot
```
robjloranger: Another issue of this PR?

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		// github quota exceeded
+		if isLimitHit(o) {
+			return packagePath, pkgerrors.WithMessage(err, "github API limit hit")
+		}
+		// another error in the output
+		return packagePath, err
+	}
+	// make sure the expected files exist
+	if err := checkFiles(fs, packagePath, version); err != nil {
```
robjloranger: This could be simplified to just
```go
return packagePath, checkFiles(fs, packagePath, version)
```
as you always return `packagePath` and the error will be nil or not executing `checkFiles`.

pkg/module/go_get_fetcher.go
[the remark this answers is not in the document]
marwan-at-work: @arschles i meant you shouldn't need this whole block in the first place. `version` is always going to include `v` and therefore you can remove the whole`if !strings.HasPrefix(version, "v")` block. You might want to call `semver.IsValid` here if you just want to double check the version is sane.

pkg/module/go_get_fetcher_test.go
```diff
@@ -0,0 +1,43 @@
+package module
+
+import (
+	"github.com/spf13/afero"
+	"io/ioutil"
+)
+
+func (s *ModuleSuite) TestNewGoGetFetcher() {
+	r := s.Require()
+	fs := s.fs
```
robjloranger: I know it's just a test but no need for the extra variable, you could just call
```go
fetcher, err := NewGoGetFetcher(s.goBinaryName, s.fs, repoURI, version)
```
on the next line.

pkg/module/go_get_fetcher_test.go
```diff
@@ -0,0 +1,43 @@
...
+	r.True(ok)
+	r.Equal(repoURI, goGetFetcher.repoURI)
+	r.Equal(version, goGetFetcher.version)
+}
+
+func (s *ModuleSuite) TestGoGetFetcherFetch() {
+	r := s.Require()
+	// we need to use an OS filesystem because fetch executes vgo on the command line, which
+	// always writes to the filesystem
+	fs := afero.NewOsFs()
```
robjloranger: Here as well, only a test but could instead just pass the call below as
```go
fetcher, err := NewGoGetFetcher(s.goBinaryName, afero.NewOsFs(), repoURI, version)
```

pkg/module/go_get_fetcher_test.go
```diff
@@ -0,0 +1,43 @@
...
+	ref, err := fetcher.Fetch(repoURI, version)
+	r.NoError(err)
+	ver, err := ref.Read()
+	r.NoError(err)
+
+	r.True(len(ver.Info) > 0)
+
+	r.True(len(ver.Mod) > 0)
+
+	zipBytes, err := ioutil.ReadAll(ver.Zip)
```
robjloranger: does `ver.Zip` still need to be closed?

marwan-at-work:
@arschles I haven't seen anything about go being able to list versions from remote. But I know where that code lives so I can potentially borrow it lol. Happy to chat more about this outside this pr

pkg/module/fetcher.go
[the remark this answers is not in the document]
arschles: oh, I see! I love the idea about wrapping. would it be ok if we see how it goes with using the `Fetcher` and `Ref` in #319 before we implement the wrapper? I'm asking because I bet it'll help us build the abstraction right

pkg/module/disk_ref_test.go
```diff
@@ -0,0 +1,63 @@
...
+//
+// returns a non-nil error if anything went wrong. the file will be closed
+// regardless of what this function returns
+func createAndWriteFile(fs afero.Fs, filename, data string) error {
+	fileHandle, err := fs.Create(filename)
+	if err != nil {
+		return err
+	}
+	defer fileHandle.Close()
+	if _, err := fileHandle.Write([]byte(data)); err != nil {
```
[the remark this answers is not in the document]
arschles: I'm cool with that

pkg/module/fetch.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+// Fetch downloads the module@version using the fetcher and stores it in storage
+func Fetch(ctx context.Context, s storage.Backend, fetcher Fetcher, mod, version string, mf *Filter) error {
+	if !mf.ShouldProcess(mod) {
+		return NewErrModuleExcluded(mod)
+	}
+
+	if s.Exists(ctx, mod, version) {
+		return nil
```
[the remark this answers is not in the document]
arschles: good idea, I'll add a custom error

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,159 @@
...
+		repoURI:      repoURI,
+		version:      version,
+		goBinaryName: goBinaryName,
+	}, nil
+}
+
+// Fetch downloads the sources and returns path where it can be found. Fetch will always
+// return a non-nil Ref.
+//
+// TODO: If an error is returned, the Fetch method will likely fail, but you should always call
```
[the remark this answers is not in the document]
arschles: This is actually outdated documentation so I'll remove it, but there are TODOs below that I will create issues for - thanks for the reminder!

pkg/module/disk_ref_test.go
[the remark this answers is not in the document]
robjloranger: actually, on that note just
```go
return fileHandle.Write([]byte(data))
```

pkg/module/errors.go
```diff
@@ -16,3 +16,20 @@ func (e *ErrModuleExcluded) Error() string {
...
+// ErrModuleAlreadyFetched is an error returned when you try to fetch the same module@version
+// more than once
+type ErrModuleAlreadyFetched struct {
+	module  string
+	version string
+}
+
+// Error is the error interface implementation
+func (e *ErrModuleAlreadyFetched) Error() string {
+	return fmt.Sprintf("%s@%s was already fetched", e.module, e.version)
```
robjloranger: There is a new package for this pattern, `%s@%s` `config.FmtModVer(mod, ver string) string`

pkg/module/go_get_fetcher.go
robjloranger: Here I was thinking the above can replace all of lines 131-134

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// Fetch downloads the sources and returns path where it can be found. This function will
+// always return a ref, even if it returns a non-nil error.
+//
+// If an error was returned, the returned ref's Read method will return an error, but you
+// should always call ref.Clear() on the returned ref
+func (g *goGetFetcher) Fetch(mod, ver string) (Ref, error) {
+	var ref Ref
+	ref = noopRef{}
```
robjloranger: Since this is immediately assigned to an empty value it could just be `ref := noopRef{}`

pkg/module/go_get_fetcher.go
robjloranger: line 83 now

pkg/module/fetch.go
```diff
@@ -0,0 +1,32 @@
...
+		return err
+	}
+	defer moduleRef.Clear()
+
+	// pretend like moduleLoc has $version.info, $version.mod and $version.zip in it :)
+	module, err := moduleRef.Read()
+	if err != nil {
+		return err
+	}
+
```
marpio: Shouldn't we do defer module.Zip.Close() here? Save doesn't call Close AFAIK.

marpio: What will happen if Clear() is called before zip gets closed? What is the defer calls order?

pkg/module/go_get_fetcher.go
arschles: yea, for now we do. https://github.com/golang/go/issues/26577 will alleviate this problem though

pkg/module/go_get_fetcher.go
arschles: yep - #329

pkg/module/go_get_fetcher.go
arschles: I replaced the `if err := checkFiles` and did what you recommended in https://github.com/gomods/athens/pull/291#discussion_r204963367. I felt like simplifying the `if err != nil` block above would hurt readability though. Cool with you?

pkg/module/go_get_fetcher_test.go
arschles: sounds good - will do!

pkg/module/go_get_fetcher_test.go
arschles: yes, it does. for a follow-up PR, do you have any ideas for making this simpler? right now, `fetcher.Fetch()` closes the info and mod files, but leaves the zip file open and forces you to close it for you. that feels a little weird and maybe like a leaky abstraction...

I don't want to block this PR on that if that's ok with everyone here, but I'm wondering if we can do better in a follow-up

arschles:
>UPDATE: go list -m -versions seems to do the trick. However, you have to be inside a go module for it to work, we can hack around that easily. But sort of weird.

@marwan-at-work we could file an issue similar to https://github.com/golang/go/issues/26577 for listing versions. What do you think? Also, we have a hack in here to make a directory into a go module (i.e. add an empty `func main() {}` and an empty `go.mod`) so we could in theory use that hack to make `go list` think it's inside a go module?

pkg/module/errors.go
arschles: 👍

pkg/module/go_get_fetcher.go
arschles: unfortunately not because then `ref` is a `noopRef` type instead of a `Ref` type, and we need it to be so that `ref = newDiskRef()` works below

pkg/module/fetch.go
arschles: Good catch - yes, we need to call it. `defer`s execute in stack order, so the zip would be closed before the clear. I wrote an example to verify: https://play.golang.org/p/Q0x563MPJoi

pkg/module/go_get_fetcher_test.go
robjloranger: For sure. I'm sure there will be lots of iterations on the tests moving forward.

pkg/module/go_get_fetcher.go
robjloranger: :+1:

pkg/module/disk_ref.go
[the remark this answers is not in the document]
robjloranger: Will this still need a fix?

robjloranger:
Just want to double check with you two @marpio and @marwan-at-work that your concerns were addressed.

marpio:
Lgtm @robjloranger

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,163 @@
...
+		goBinaryName: goBinaryName,
+	}, nil
+}
+
+// Fetch downloads the sources and returns path where it can be found. This function will
+// always return a ref, even if it returns a non-nil error.
+//
+// If an error was returned, the returned ref's Read method will return an error, but you
+// should always call ref.Clear() on the returned ref
+func (g *goGetFetcher) Fetch(mod, ver string) (Ref, error) {
```
marwan-at-work: so why do we always wanna return a Ref? why can't we just return `nil`? If it's a matter of calling `ref.Clear()` we can always make sure we clear things inside this function when an error occurs, no?

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,163 @@
...
+	return afero.WriteFile(fs, sourcePath, sourceContent, 0666)
+}
+
+// given a filesystem, gopath, repository root, module and version, runs 'vgo get'
+// on module@version from the repoRoot with GOPATH=gopath, and returns the location
+// of the module cache. returns a non-nil error if anything went wrong. always returns
+// the location of the module cache so you can delete it if necessary
+func getSources(goBinaryName string, fs afero.Fs, gopath, repoRoot, module, version string) (string, error) {
+	version = strings.TrimPrefix(version, "@")
+	if !strings.HasPrefix(version, "v") {
```
marwan-at-work: from my previous review: 

you shouldn't need this whole block in the first place. version is *always* going to include v and therefore you can remove the whole `if !strings.HasPrefix(version, "v")` block. If you really don't trust the caller, then calling `semver.IsValid` is probably more wholesome. But imo, just going out and saying `go get module@version` is good enough because Go will validate the version anyway

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,163 @@
...
+	"path/filepath"
+	"strings"
+
+	pkgerrors "github.com/pkg/errors"
+	"github.com/spf13/afero"
+)
+
+var (
+	// ErrLimitExceeded signals that github.com refused to serve the request due to exceeded quota
+	ErrLimitExceeded = errors.New("github limit exceeded")
```
marwan-at-work: This seems to be not used anywhere, furthermore it seems that `goget.go` and goget_test.go` are still around. Can we delete them in this pr?

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,163 @@
...
+	cmd.Env = []string{"PATH=" + os.Getenv("PATH"), gopathEnv, cacheEnv, disableCgo}
+	cmd.Dir = repoRoot
+
+	packagePath := filepath.Join(gopath, "src", "mod", "cache", "download", module, "@v")
+
+	o, err := cmd.CombinedOutput()
+	if err != nil {
+		// github quota exceeded
+		if isLimitHit(o) {
+			return packagePath, pkgerrors.WithMessage(err, "github API limit hit")
```
marwan-at-work: let's make sure we propagate the rate limit status all the way back to vgo so it knows to show the right message. You can do `return packagePath, errors.E(op, err, status.StatusTooManyRequests)` -- we should also add that as a known `Kind` in the errors package. So in the future we can do something like 
`if errors.Kind(err) == errors.KindRateLimit` or wrap that in a function like `if errors.ErrRateLimit(err)`

pkg/module/errors.go
```diff
@@ -16,3 +20,20 @@ func (e *ErrModuleExcluded) Error() string {
 func NewErrModuleExcluded(module string) error {
 	return &ErrModuleExcluded{module: module}
 }
+
+// ErrModuleAlreadyFetched is an error returned when you try to fetch the same module@version
+// more than once
+type ErrModuleAlreadyFetched struct {
```
marwan-at-work: you can remove this struct and its methods :)

pkg/module/fetcher.go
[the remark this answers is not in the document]
marwan-at-work: sounds good to me 👍

.travis.yml
```diff
@@ -15,6 +15,7 @@ env:
     - MINIO_SECRET_KEY=minio123
     - ATHENS_MONGO_STORAGE_URL=mongodb://127.0.0.1:27017
     - CODE_COV=1
+    - GO_BINARY_NAME=vgo
```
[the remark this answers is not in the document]
robjloranger: That would probably be more clear

pkg/module/go_get_fetcher.go
robjloranger: It was used on ~129 after the `isLimitHit` check

.travis.yml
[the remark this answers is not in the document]
arschles: 👍

pkg/module/go_get_fetcher.go
arschles: >also, why does Fetch takes a mod/ver when NewGoGetFetcher also takes a mod and ver?

good catch, only `Fetch` should take the mod/ver

pkg/module/go_get_fetcher.go
arschles: I'd rather leave it as two - it feels a bit more readable

pkg/module/go_get_fetcher.go
```diff
@@ -0,0 +1,163 @@
...
+func prepareStructure(fs afero.Fs, repoRoot string) error {
+	// vgo expects go.mod file present with module statement or .go file with import comment
+	gomodPath := filepath.Join(repoRoot, "go.mod")
+	gomodContent := []byte("module mod")
+	if err := afero.WriteFile(fs, gomodPath, gomodContent, 0666); err != nil {
+		return err
+	}
+
+	sourcePath := filepath.Join(repoRoot, "mod.go")
+	sourceContent := []byte(`package mod // import "mod"`)
```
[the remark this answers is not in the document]
arschles: good catch

pkg/module/go_get_fetcher.go
arschles: (instead of relying on `go get` to tell us that the version is invalid when we call)

arschles:
@marwan-at-work after I fix CI and remove that struct, will do - thanks!

PR #302, opened 2018-07-19
removing the home handler on olympus

arschles:
Related to #314, which removes the home handler on the proxy

### DIFF
--- a/pkg/storage/azurecdn/storage.go
+++ b/pkg/storage/azurecdn/storage.go
@@ -7,9 +7,8 @@ import (
 	"io"
 	"net/url"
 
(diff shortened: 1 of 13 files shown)


arschles:
Yep - just 404

marpio:
@arschles can we also delte `./cmd/olympus/templates/olympus/index.html` now?

arschles:
@marpio @robjloranger are you ok if I leave the template dir but just delete the `index.html`? I'm asking because the `application.html` and `_flash.html` are reusable and we might want to use them later on...

bketelsen:
OK on deleting the handler, but this will be an actual HTML end point at some future point.

PR #303, opened 2018-07-19
adding ignores

arschles:
for vscode config dir and binaries


.gitignore
```diff
@@ -24,3 +24,6 @@ cmd/proxy/proxy
 cmd/olympus/olympus
 test-keys.json
 tmp
+.vs
```
robjloranger: has it changed? it used to be `.vscode`

robjloranger: oh weird, I thought it was the workspace settings folder

arschles:
@robjloranger hrm, maybe it's coming from somewhere else? [this](https://stackoverflow.com/questions/31878901/should-i-add-the-visual-studio-2015-vs-folder-to-source-control) and [this](https://social.msdn.microsoft.com/Forums/en-US/1ea61876-d755-40e5-996f-0159fea46b10/vs-folder) indicate that it's a VS 2015 config dir.

i have no idea

anyway, are you open to adding while I figure out wtf is going on with my vscode install?

robjloranger:
Sure

Issue #262, opened 2018-07-13
Split up our tests (unit, integration, end to end)


arschles:
Following https://github.com/gomods/athens/pull/256#issuecomment-404931246 and I believe there was discussion on this in yesterday's meeting (7/12/2018).

Here's my opinion on how we should split up our tests:

1. Unit tests - only operate on in-memory things
2. Integration tests - operate on services that a developer can run on their machine via docker compose
3. Full end to end - operate on everything including live cloud services

(1) is easy for anyone to run and we have pretty good docs (IMO, not sure if peeps agree?) for how to set up your environment for (2). (3) is just starting to become a thing in @robjloranger's https://github.com/gomods/athens/pull/208. I think he added a switch in the code to turn the live GCP tests on or off, so that might be a good thing to build on top of.

michalpristas:
I totally agree on that, maybe we can talk about 3. I would like to see how it works in k8s. 
in my current org we have real services, isolated storages... which ofc cost money.

marwan-at-work:
For my dev set up I'd like two things: 
1. Have one command to bootstrap only things I need to run in dev mode. At this point: it's just redis, mongo, jaegar and the registry/proxy servers. And another command to tear down everything and remove persistent data (docker volume prune)

2. Have a clean slate backdoor to not have to take down everything and reboot everything once again just so that we have a clean cache/storage.

arschles:
@marwan-at-work I feel like our test code and test/dev dependencies are related enough to keep in the same issue for now. Would you be up for PR-ing the command to start up things only needed in dev mode? That'd be a good start down this path.

marpio:
I was thinking, maybe we could use minio for almost everything. We already have storage backed by minio. It is s3 compatible so in theory we could use it a CDN storage for the registry (not sure if aws-sdk can be used to talk to minio though). The only thing that's missing is the event log impl. on top of minio (at the moment mongo only). Then we would only need minio, redis and jaeger for proxy and registry development. What do you think?

marpio:
I was actually only thinking about the development setup @arschles sorry for not being clear on that.
It would be nice if we would only need minio, jeager and redis for the proxy and registry development env.
We still need the others dependencies for tests.

arschles:
@marpio that makes sense. What do you think about making unit tests completely free of outside dependencies? For example, #343 is removing all dependencies on external GCP, we could do the same thing for RDBMS, Jaeger, and so on. I'm not sure if this is wise but I want to hear what you think

[...]

PR #313, opened 2018-07-20
Adding registry build to the makefile

arschles:
We were only building the proxy

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ dev:
 	docker-compose -p athensdev up -d mongo
 	docker-compose -p athensdev up -d redis
 	docker-compose -p athensdev up -d minio
+	docker-compose -p athensdev up -d jaeger
 	echo "sleeping for a bit to wait for the DB to come up"
 	sleep 5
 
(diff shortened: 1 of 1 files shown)


arschles:
@marwan-at-work the `dev` and `dev-teardown` endpoints already existed, so I added jaeger to the `dev` target and `docker volume prune` to the `dev-teardown` target

marwan-at-work:
@arschles the problem is that i keep commenting out `mysql`, `postgres` and `minio`, because they are not used in development and they make building/tearing down the dev environment really slow. Maybe im just neurotic 😄

arschles:
Oh I see, good point. I remember there was some discussion in a dev meeting about coming up with a lighter-weight dev environment (as opposed to a test environment). I created https://github.com/gomods/athens/issues/262 to try and capture that discussion. Would you mind adding a comment in there about turning off those services for development? I would totally love for us to come up with a good way to run a sane, non-crazy-memory-intensive dev environment

I'm going to merge this for now since it makes some progress on the build

Issue #312, opened 2018-07-20
Remove the fetch route, handler & gomod parser


arschles:
The fetch route is [here](https://github.com/gomods/athens/blob/b1f178072f482cf69b7111a6366db383ad62c0c2/cmd/proxy/actions/app_proxy.go#L25) and the handler is [here](https://github.com/gomods/athens/blob/b1f178072f482cf69b7111a6366db383ad62c0c2/cmd/proxy/actions/fetch.go#L20)

This handler was used by the CLI (which is obsolete and was removed), and we don't need it anymore. This issue is for removing the fetch handler, the fetch route, and the gomod parser which was used by the fetch handler

ryancox:
is there a gomod parser library elsewhere that could be used / improved?

arschles:
@ryancox yes, it's in a package inside `golang/go`, I think an `internal` package. We're going to delegate as much of this as possible to calling the `go` CLI so we don't have to depend on code inside of the go toolchain

cc/ @rsc @marwan-at-work, both of whom have been part of discussions on this topic

arschles:
This is a subset of #335, closing

PR #314, opened 2018-07-20
Removing the fetch handler & gomod parser

arschles:
Fixes https://github.com/gomods/athens/issues/312
Requires #316 and #291. Check out more details on why this is opened against #291 at https://github.com/gomods/athens/issues/290#issuecomment-406735602

Also removes:

- The `/` route on the proxy (similar PR for olympus: https://github.com/gomods/athens/pull/302)
- The `./pkg/repo` package (this fetcher logic is moved and refactored in https://github.com/gomods/athens/pull/291)

This is part of #290


arschles:
This requires https://github.com/gomods/athens/pull/291 and then some changes in `GetPackageDownloaderJob`

arschles:
Closing this now because I'm going to do this and #316 all in one pass, after #290 is all done

That's all tracked by #335

PR #319, opened 2018-07-20
Doing synchronous cache filling for the .info endpoint

arschles:
This patch shows how we could synchronously fill in storage when the proxy or registry receive a request to '.../version.info' and the package isn't in storage.

Requires #291. Check out more details on why this is opened against #291 at https://github.com/gomods/athens/issues/290#issuecomment-406735602

Part of #290

The only commit that makes changes on top of #291 is https://github.com/gomods/athens/commit/f8cffc6a6bb6da80dfda6fa4c67a9af348dd4d6a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 .PHONY: build
 build:
 	cd cmd/proxy && buffalo build
+	cd cmd/olympus && buffalo build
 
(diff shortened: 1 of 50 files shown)


arschles:
I am gonna hold off on this one until #291 is in

Issue #275, opened 2018-07-15
Introduce errors package


marwan-at-work:
Athens is quite a big project with a lot of moving parts and dependencies. We should have a consistent way to construct and communicate errors through the codebase. 

We should have our own errors package with an `Error` struct holds relevant Athens information about a system error. 

An Error struct can hold information about what led to the creation of that error: module path, module operation, storage, user etc. 

Constructing an error should be easy as `errors.E(op, "error message", moduleInfo, severityLevel)`

Logging an error should be as easy as `log.SystemErr(err)` which will internally cast the error interface into the Error struct to probe the severityLevel and the operation to do structured logging. 

inspiration: https://commandcenter.blogspot.com/2017/12/error-handling-in-upspin.html

marpio:
Also from the same author: https://middlemost.com/failure-is-your-domain/

michalpristas:
while reading the article I realized I've read it before, but it was a nice recap.
It would be great to have something like this with Athens but first, we need to find a common delimiter for our errors. Just like with the mentioned Upspin they developed the pkg with the problem in mind to solve.

We don't play with errors just yet so maybe our understanding of how we should approach this might be a bit off. But I would start with following Upspin error for implementation/API part and composing our errors of:
- Module
- Version
- Op (operation like cache-miss, source-pull, ...)
- Kind (just like with upspin)
- Err (cause of error)

So to sum it up, yeah let's 🔨 it. Nice and searchable errors rocks

marwan-at-work:
I agree, we should at least have one canonical way to construct errors and then as we progress we'll see what kind of information/behavior it should have. But we should start conforming our errors before it gets outta hand :P 

I'll take a stab at it today or this week

arschles:
Sounds like a plan! I see some packages where errors are pretty
standardized so I can go error-ify those and we can move in increments

arschles:
I wonder if https://github.com/pkg/errors would be useful here?

PR #320, opened 2018-07-22
Athens: introduce pkg/errors

marwan-at-work:
Fixes https://github.com/gomods/athens/issues/275

Todo: write tests for the errors package and the logger's SystemErr method. 

I'd like to gather some feedback first before writing the tests. 

Things to look out for: 

1. I updated our logger to be "Athens' Errors Friendly" using the SystemErr method.
2. I updated download.VersionZipHandler to use pkg/errors, this is so you can check out what it's like to use the package. 
3. ~I'm using a copy of cmd/go's semver validation to determine whether a string goes into the Version field. We can use Mastermind's package instead (as suggested by @arschles) but we need to confirm that they both agree to the same semver rules (so far I know cmd/go enforces a "v" prefix while Mastermind does not). We can also avoid semver validation by just creating an errors.V type just like errors.Op. Let me know :)~

### DIFF
--- a/cmd/olympus/actions/app.go
+++ b/cmd/olympus/actions/app.go
@@ -62,6 +62,7 @@ func App(config *AppConfig) (*buffalo.App, error) {
 			return nil, err
 		}
 
+		lggr := log.New(env.CloudRuntime(), env.LogLevel())
 		app = buffalo.New(buffalo.Options{
 			Addr: port,
 			Env:  ENV,
@@ -117,7 +118,7 @@ func App(config *AppConfig) (*buffalo.App, error) {
 		app.GET(download.PathList, download.ListHandler(config.Storage, renderEng))
 		app.GET(download.PathVersionInfo, download.VersionInfoHandler(config.Storage, renderEng))
 		app.GET(download.PathVersionModule, download.VersionModuleHandler(config.Storage, renderEng))
-		app.GET(download.PathVersionZip, download.VersionZipHandler(config.Storage))
+		app.GET(download.PathVersionZip, download.VersionZipHandler(config.Storage, renderEng, lggr))
 
 		app.ServeFiles("/", assetsBox) // serve files from the public directory
 	}
--- a/pkg/download/version_zip.go
+++ b/pkg/download/version_zip.go
@@ -8,6 +8,7 @@ import (
 	"github.com/bketelsen/buffet"
 	"github.com/gobuffalo/buffalo"
 	"github.com/gobuffalo/buffalo/render"
+	"github.com/gomods/athens/pkg/config"
 	"github.com/gomods/athens/pkg/errors"
 	"github.com/gomods/athens/pkg/log"
 	"github.com/gomods/athens/pkg/paths"
@@ -28,12 +29,11 @@ func VersionZipHandler(getter storage.Getter, eng *render.Engine, lggr *log.Logg
 		params, err := paths.GetAllParams(c)
 		if err != nil {
 			lggr.SystemErr(errors.E(op, err))
-			c.Response().WriteHeader(http.StatusInternalServerError) // 500 because handler should not be called in the first place.
+			c.Render(http.StatusInternalServerError, nil) // 500 because handler should not be called in the first place.
 			return nil
 		}
 
-		mod, ver := params.Module, params.Version
-		version, err := getter.Get(mod, ver)
+		version, err := getter.Get(params.Module, params.Version)
 		if err != nil {
 			lvl := logrus.ErrorLevel
 			status := http.StatusInternalServerError
@@ -41,23 +41,24 @@ func VersionZipHandler(getter storage.Getter, eng *render.Engine, lggr *log.Logg
 			// TODO: move this function to pkg/errors
 			if storage.IsNotFoundError(err) {
 				lvl = logrus.DebugLevel
-				msg = fmt.Sprintf("%s@%s not found", mod, ver)
+				msg = fmt.Sprintf("%v not found", config.FmtModVer(params.Module, params.Version))
 				status = http.StatusNotFound
 			}
 
-			lggr.SystemErr(errors.E(op, errors.M(mod), errors.V(ver), lvl, err))
+			lggr.SystemErr(errors.E(op, errors.M(params.Module), errors.V(params.Version), lvl, err))
 			return c.Render(status, eng.JSON(msg))
 		}
 		defer version.Zip.Close()
 
-		// should we write the status file *after* we ensure io.Copy is
-		// successful or not?
-		c.Response().WriteHeader(http.StatusOK)
+		status := http.StatusOK
 
 		_, err = io.Copy(c.Response(), version.Zip)
 		if err != nil {
-			lggr.SystemErr(errors.E(op, errors.M(mod), errors.V(ver), err))
+			status = http.StatusInternalServerError
+			lggr.SystemErr(errors.E(op, errors.M(params.Module), errors.V(params.Version), err))
 		}
+
+		c.Response().WriteHeader(status)
 
 		return nil
 	}
--- a/pkg/download/version_zip.go
+++ b/pkg/download/version_zip.go
@@ -7,34 +7,58 @@ import (
 
 	"github.com/bketelsen/buffet"
 	"github.com/gobuffalo/buffalo"
+	"github.com/gobuffalo/buffalo/render"
+	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/log"
 	"github.com/gomods/athens/pkg/paths"
 	"github.com/gomods/athens/pkg/storage"
+	"github.com/sirupsen/logrus"
 )
 
 // PathVersionZip URL.
 const PathVersionZip = "/{module:.+}/@v/{version}.zip"
 
 // VersionZipHandler implements GET baseURL/module/@v/version.zip
-func VersionZipHandler(getter storage.Getter) func(c buffalo.Context) error {
+func VersionZipHandler(getter storage.Getter, eng *render.Engine, lggr *log.Logger) func(c buffalo.Context) error {
+	const op errors.Op = "download.VersionZipHandler"
+
 	return func(c buffalo.Context) error {
 		sp := buffet.SpanFromContext(c)
 		sp.SetOperationName("versionZipHandler")
 		params, err := paths.GetAllParams(c)
 		if err != nil {
-			return err
-		}
-		version, err := getter.Get(params.Module, params.Version)
-		if storage.IsNotFoundError(err) {
-			return c.Error(http.StatusNotFound, fmt.Errorf("%s@%s not found", params.Module, params.Version))
-		} else if err != nil {
-			return err
+			lggr.SystemErr(errors.E(op, err))
+			c.Response().WriteHeader(http.StatusInternalServerError) // 500 because handler should not be called in the first place.
+			return nil
 		}
 
+		mod, ver := params.Module, params.Version
+		version, err := getter.Get(mod, ver)
+		if err != nil {
+			lvl := logrus.ErrorLevel
+			status := http.StatusInternalServerError
+			msg := http.StatusText(status)
+			// TODO: move this function to pkg/errors
+			if storage.IsNotFoundError(err) {
+				lvl = logrus.DebugLevel
+				msg = fmt.Sprintf("%s@%s not found", mod, ver)
+				status = http.StatusNotFound
+			}
+
+			lggr.SystemErr(errors.E(op, mod, ver, lvl, err))
+			return c.Render(status, eng.JSON(msg))
+		}
 		defer version.Zip.Close()
 
+		// should we write the status file *after* we ensure io.Copy is
+		// successful or not?
 		c.Response().WriteHeader(http.StatusOK)
 
 		_, err = io.Copy(c.Response(), version.Zip)
-		return err
+		if err != nil {
+			lggr.SystemErr(errors.E(op, mod, ver, err))
+		}
+
+		return nil
 	}
 }
--- a/pkg/download/version_zip.go
+++ b/pkg/download/version_zip.go
@@ -45,7 +45,7 @@ func VersionZipHandler(getter storage.Getter, eng *render.Engine, lggr *log.Logg
 				status = http.StatusNotFound
 			}
 
-			lggr.SystemErr(errors.E(op, mod, ver, lvl, err))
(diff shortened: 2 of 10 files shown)


pkg/errors/errors.go
```diff
@@ -0,0 +1,158 @@
...
+	e := Error{Op: op}
+	for _, a := range args {
+		switch a := a.(type) {
+		case error:
+			e.Err = a
+		case string:
+			switch {
+			case isModule(a):
+				e.Module = a
+			case isVersion(a):
```
marpio: Did I understand it right - if we would have to wrap every version into errors.V to get rid of this?

pkg/errors/errors.go
```diff
@@ -0,0 +1,158 @@
...
+// as to what caused this error so that
+// callers can implement logic around it.
+type Error struct {
+	// Kind categories Athens errors into a smaller
+	// subset of errors. This way we can generalize
+	// what an error really is: such as "not found",
+	// "bad request", etc. The official categories
+	// are HTTP status code but the ones we use are
+	// imported into this package.
+	Kind     int
```
marpio: Could this be an enum?

marwan-at-work: Do you mean have it as `type Kind int` and create an `iota` list? Upspin does it the way you're suggesting, however, I actually think we might hit two birds with one stone if we made the Kind be simple HTTP Statuses which could help us deduce external API errors more easily.

marpio: Do we have to use iota? I was thinking about something like this (option without iota) https://blog.learngoprogramming.com/golang-const-type-enums-iota-bc4befd096d3

marwan-at-work: So in this case we already have everything as an enum. See Line 12 for the enums declaration, and KindText for stringifying the enum. The only difference is that `Kind` is not its own special type. Should I do `type Kind int`?

marwan-at-work: Bearing in mind, the user will now have to do `errors.E(op, Kind(httpStatusCode)` instead of `errors.E(op, httpStatusCode)`, or we can do the casting inside the function `case int: err.Kind = Kind(a)`

marpio: Oh, so I can use http status and also define a custom kind? I think this is a great idea.

marwan-at-work: Yep! I think it can play out nicely. We'll see 😄

pkg/errors/errors.go
marpio: I think it will be easy to forget to add errors.V/M so I would leave it as it is but maybe others have different opinion. I could live with both I guess.

marwan-at-work: Unfortunately I had to remove `isModule` because there are third party modules with absolutely no `/` -- for example `go.opencensus.io` is a valid package on its own. So I had to replace it with `errors.M` -- And now that I've replaced it with `errors.M` we might as well do `errors.V`. You can look at how that looks in two of the errors. I think it's not too terrible. And we'll be able to remember because now `errors.E()` only takes a string for an error.

pkg/errors/errors.go
```diff
@@ -0,0 +1,153 @@
...
+
+// E is a helper function to construct an Error type
+// Operation always comes first, module path and version
+// come second, they are optional. Args must have at least
+// an error or a string to describe what exactly went wrong.
+// You can optionally pass a Logrus severity to indicate
+// the log level of an error based on the context it was constructed in.
+func E(op Op, args ...interface{}) error {
+	if len(args) == 0 {
+		panic("errors.E called with 0 args")
```
robjloranger: I'm not sure of a better option but I prefer to avoid calling panic anywhere other than main.

robjloranger: Maybe similar to line 85 below, we should return an error with a message reporting that no arguments were passed.

pkg/errors/errors.go
```diff
@@ -0,0 +1,153 @@
...
+
+	if e.Err == nil {
+		e.Err = errors.New("no error message provided")
+	}
+
+	return e
+}
+
+// Severity returns the log level of an error
+// if non exists, then it's an Error level because
```
robjloranger: s/if non exists, then it's an Error level because/if none exists, then the level is Error because/

pkg/paths/path.go
```diff
@@ -1,17 +1,19 @@
...
 )
 
 // GetModule gets the module from the path of a ?go-get=1 request
 func GetModule(c buffalo.Context) (string, error) {
+	const op errors.Op = "paths.GetModule"
+
 	module := c.Param("module")
 	if module == "" {
-		return "", fmt.Errorf("module missing")
+		return "", errors.E(op, "module missing")
```
robjloranger: I know it was there before, but maybe the string could be more descriptive. Something like "missing module parameter"?

pkg/paths/path.go
```diff
@@ -24,13 +26,15 @@ type AllPathParams struct {
...
+	const op errors.Op = "paths.GetAllParams"
 	mod, err := GetModule(c)
 	if err != nil {
-		return nil, err
+		return nil, errors.E(op, err)
 	}
 	version := c.Param("version")
 	if version == "" {
-		return nil, fmt.Errorf("version not found")
+		return nil, errors.E(op, "version not found")
```
robjloranger: Same here, present from another commit but the string passed could better describe that the version parameter is missing and not the version itself.

pkg/errors/errors.go
```diff
@@ -0,0 +1,153 @@
...
+	e, ok := err.(Error)
+	if !ok {
+		return KindUnexpected
+	}
+
+	if e.Kind != 0 {
+		return e.Kind
+	}
+
+	return Kind(e.Err)
```
robjloranger: Just a question, is there a difference between `e.Kind` and `Kind(e.Err)`? Does the latter get the Kind of the child or parent?

robjloranger:
Quick note, I added the [WIP] as this is not yet ready for final review. Once you have time to address concerns and add your tests please remove the [WIP] from the title. Also at that time feel free to request a few reviewers here.

pkg/errors/errors.go
marpio: @robjloranger It's a recursion. It goes through all children until something is found.

robjloranger: Cool, thanks. I was thinking something like that. Just not at home to properly dig through the code base.

cmd/olympus/actions/app.go
```diff
@@ -117,7 +118,7 @@ func App(config *AppConfig) (*buffalo.App, error) {
 		app.GET(download.PathList, download.ListHandler(config.Storage, renderEng))
```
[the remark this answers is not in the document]
marwan-at-work: @robjloranger All routes for sure. I will open an issue about adding a logging middleware to add some sane defaults to every request (like request path, method etc), and then handlers could use that resulting `Entry`.

pkg/download/version_zip.go
```diff
@@ -7,34 +7,58 @@ import (
...
+			return nil
 		}
 
+		mod, ver := params.Module, params.Version
+		version, err := getter.Get(mod, ver)
+		if err != nil {
+			lvl := logrus.ErrorLevel
+			status := http.StatusInternalServerError
+			msg := http.StatusText(status)
+			// TODO: move this function to pkg/errors
```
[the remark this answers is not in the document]
marwan-at-work: @robjloranger probably not :) would like to do a separate pr for moving our defined errors in this package

pkg/download/version_zip.go
```diff
@@ -7,34 +7,58 @@ import (
...
+			if storage.IsNotFoundError(err) {
+				lvl = logrus.DebugLevel
+				msg = fmt.Sprintf("%s@%s not found", mod, ver)
```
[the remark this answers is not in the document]
marwan-at-work: @robjloranger will do it in this pr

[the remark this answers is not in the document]
marwan-at-work: @robjloranger there's already a function with that name that takes an extension. I will create another name, like `FmtModVer`?

pkg/download/version_zip.go
```diff
@@ -7,34 +7,58 @@ import (
...
 		if err != nil {
-			return err
-		}
-		version, err := getter.Get(params.Module, params.Version)
-		if storage.IsNotFoundError(err) {
-			return c.Error(http.StatusNotFound, fmt.Errorf("%s@%s not found", params.Module, params.Version))
-		} else if err != nil {
-			return err
+			lggr.SystemErr(errors.E(op, err))
+			c.Response().WriteHeader(http.StatusInternalServerError) // 500 because handler should not be called in the first place.
```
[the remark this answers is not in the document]
marwan-at-work: @robjloranger is it okay to pass a `nil` Render in the second argument of `c.Render`? I just want to return a status code without body strings or json etc.

pkg/errors/errors.go
marwan-at-work: will do

pkg/paths/path.go
marwan-at-work: @robjloranger in my opinion we should even panic here. Because that's the Buffalo Handler misbehaving or we did not write the path regex correctly 😄 but will change the string text for now.

pkg/download/version_zip.go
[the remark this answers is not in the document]
robjloranger: Ok, can you create an issue for it so we don't forget? :)

cmd/olympus/actions/app.go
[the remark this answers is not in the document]
robjloranger: :+1:

pkg/download/version_zip.go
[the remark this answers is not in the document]
robjloranger: Oh sorry, I should have been more clear. I meant similar to the existing one. `FmtModVer` should be good.

pkg/download/version_zip.go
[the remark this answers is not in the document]
robjloranger: This is probably good as is then, This one was more a question as I wasn't sure.

robjloranger:
I'll give this a better look over later this morning.

pkg/download/version_zip.go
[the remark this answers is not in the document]
marwan-at-work: @robjloranger https://github.com/gomods/athens/issues/327

cmd/olympus/actions/app.go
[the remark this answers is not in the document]
marwan-at-work: @robjloranger https://github.com/gomods/athens/issues/328

pkg/download/version_zip.go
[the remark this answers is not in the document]
robjloranger: lol oh :grin:

pkg/errors/errors.go
```diff
@@ -0,0 +1,160 @@
...
+func E(op Op, args ...interface{}) error {
+	if len(args) == 0 {
+		msg := "errors.E called with 0 args"
+		_, file, line, ok := runtime.Caller(1)
+		if ok {
+			msg = fmt.Sprintf("%v - %v:%v", msg, file, line)
+		}
+		return errors.New(msg)
+	}
+	e := Error{Op: op}
```
robjloranger: Would we gain a bit of clarity if this was moved to the top and we used it for the return on 72 above? Similar to 93 below, then we would have the operation chain intact.

pkg/errors/errors_test.go
```diff
@@ -0,0 +1,65 @@
...
+	msg := "test error"
+
+	err := E(op, msg)
+	require.Equal(t, logrus.ErrorLevel, Severity(err))
+
+	err = E(op, msg, logrus.WarnLevel)
+	require.Equal(t, logrus.WarnLevel, Severity(err))
+
+	err = E(op, err)
+	require.Equal(t, logrus.WarnLevel, Severity(err))
```
robjloranger: do we need to check the warning level is set twice? or was this meant to check another logging level?

marwan-at-work: @robjloranger yes this is meant to check only the bottom level, this is because when the parent doesn't have a Severity, we take the Severity of the child.

marwan-at-work:
@robjloranger updated

pkg/errors/doc.go
```diff
@@ -0,0 +1,22 @@
...
+// it knows how to deal with errors constructed from this package in a debuggable way.
+// To construct an Athens error, call the errors.E function. The E function takes
+// an Op and a variadic interface{}, but the values of the Error struct are what you can
+// pass to it. Values such as the error Kind, Module, Version, Error Message,
+// and Seveirty (seriousness of an error) are all optional. The only truly required value is
+// the errors.Op so you can construct a traceable stack that leads to where
+// the error happened. However, adding more information can help catch an issue
+// quicker and would help Cloud Log Monitoring services be more efficient to maintainers
+// as you can run queries on Athens Errors such as "Give me all errors of KindUnexpected"
+// or "Give me all errors where caused by a particular Module"
```
arschles: this godoc is 🔥

pkg/errors/errors.go
```diff
@@ -0,0 +1,158 @@
...
+			e.Err = a
+		case string:
+			e.Err = errors.New(a)
+		case M:
+			e.Module = a
+		case V:
+			e.Version = a
+		case logrus.Level:
+			e.Severity = a
+		case int:
```
arschles: should we have a `default` here?

pkg/errors/errors.go
```diff
@@ -0,0 +1,158 @@
...
+// stack trace.
+func Ops(err Error) []Op {
+	ops := []Op{err.Op}
+	for {
+		embeddedErr, ok := err.Err.(Error)
+		if !ok {
+			break
+		}
+
+		ops = append(ops, embeddedErr.Op)
```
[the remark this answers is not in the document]
arschles: @marpio which package?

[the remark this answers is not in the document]
marwan-at-work: @arschles this was from a previous implementation. I used semver to figure out if a string was semver. Same for Modules which proven to be potentially very buggy, so i removed them and replaced them with `M` and `V` to be explicit and not buggy. More writing tho, but im ok with it

pkg/errors/errors.go
marwan-at-work: @arschles I don't think so, but I'm happy either way. I'm not sure what the default would do here, the only thing it can do is either Panic or log "unknown type" which means we'd have to pass a logger to every error or just use fmt.Println which i think we should avoid. wdyt?

arschles: I'm with you! Maybe we'll figure out what to put in a default here after this thing gets some more usage, but for now I'm good leaving it out

PR #322, opened 2018-07-23
storage: add context to storage.Exists and storage.Delete

tbpg:
This PR touches lots of files, but each one only has a few mechanical changes.

Note: I did `Exists` and `Delete` at the same time since every `Delete` function calls `Exists`. This way, we can avoid having this PR add `context.TODO()` everywhere and needing another PR to update those lines again.

Part of #174.

### DIFF
--- a/pkg/eventlog/multireader.go
+++ b/pkg/eventlog/multireader.go
@@ -1,6 +1,7 @@
 package eventlog
 
 import (
+	"context"
 	"errors"
 
 	"github.com/gomods/athens/pkg/storage"
@@ -119,5 +120,5 @@ func exists(event Event, log []Event, checker storage.Checker) bool {
 		}
 	}
 
-	return checker.Exists(event.Module, event.Version)
+	return checker.Exists(context.Background(), event.Module, event.Version)
 }
--- a/pkg/eventlog/multireader.go
+++ b/pkg/eventlog/multireader.go
@@ -120,5 +120,5 @@ func exists(event Event, log []Event, checker storage.Checker) bool {
 		}
 	}
 
-	return checker.Exists(context.Background(), event.Module, event.Version)
+	return checker.Exists(context.TODO(), event.Module, event.Version)
 }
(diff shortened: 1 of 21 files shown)


pkg/eventlog/multireader.go
```diff
@@ -119,5 +120,5 @@ func exists(event Event, log []Event, checker storage.Checker) bool {
 		}
 	}
 
-	return checker.Exists(event.Module, event.Version)
+	return checker.Exists(context.Background(), event.Module, event.Version)
```
Kunde21: Use `context.TODO()` here, as a placeholder until context is threaded through ReadSingle.

robjloranger: I'm interested in what others think about this, they seem to return the same thing. Other than the name is there a reason to use one over the other?

tbpg: Done. No, `context.Background()` is more or less assumed to be correct, whereas `context.TODO()` is a placeholder to update later. Also, "TODO is recognized by static analysis tools that determine whether Contexts are propagated correctly in a program." https://golang.org/pkg/context/#TODO

pkg/storage/gcp/getter.go
```diff
@@ -14,7 +14,7 @@ import (
 // The caller is responsible for calling close on the Zip ReadCloser
 func (s *Storage) Get(module, version string) (*storage.Version, error) {
 	ctx := context.Background()
-	if exists := s.Exists(module, version); !exists {
+	if exists := s.Exists(context.TODO(), module, version); !exists {
```
marwan-at-work: why not pass the paxx the `ctx` value from one line above? You can change it from `Background` to `TODO` if you'd like.

tbpg: Done. Good catch!

robjloranger:
LGTM, merge if your settled @marwan-at-work

arschles:
@marwan-at-work since your comment was addressed in https://github.com/gomods/athens/pull/322#discussion_r204785563, I'm going to merge this 😄

PR #323, opened 2018-07-23
Specify mysql should be installed as part of development bootstrap

es:
After following the setup guide in `DEVELOPMENT.md`, I still ran into an issue running tests:
```
ERRO[0000] Error: exec: "mysql": executable file not found in $PATH
```

Turns out `mysql` must be installed locally for `buffalo` to work. This PR adds that dependency to the bootstrap guide.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -3,7 +3,7 @@
 Both the registry and the proxy are written using the [Buffalo](https://gobuffalo.io/) framework. We chose
 this framework to make it as straightforward as possible to get your development environment up and running.
 
-You'll need Buffalo [v0.11.0](https://github.com/gobuffalo/buffalo/releases/tag/v0.11.0) or later to get started on Athens,
+You'll need Buffalo [v0.12.4](https://github.com/gobuffalo/buffalo/releases/tag/v0.12.4) or later to get started on Athens,
 so be sure to download the CLI and put it into your `PATH`.
 
 # Services that Athens Needs


robjloranger:
Can you open an issue to discuss your problem? I just removed mysql and ran `make dev` with success. We use docker for database services and so should not require mysql to be installed.

robjloranger:
double checked and I can run both Olympus and Zeus with no installation of mysql via `buffalo dev`

arschles:
I can’t confirm that bc I’m pretty sure it is because buffalo test uses it
under the hood. We could do some work though to remove that as a
requirement if we wanted. That test command used the mysql cli to run
migrations but we could do that in code. What do you think?

robjloranger:
@arschles in that case either or is fine by me. Merging this or getting rid of the requirement. How can I invoke the test to verify?

arschles:
Sorry I didn’t catch - which part are you trying to verify?

robjloranger:
Just that mysql is in fact required. I have run our make dev script and also run both development app servers without it so just wondering what might be looking for it. What script/test etc is trying to use mysql. Is it just the tests for our packages? @marpio mentioned it was an old requirement so wanted to be certain before we add it as a hard installation requirement.

marpio:
The gobuffalo/pop PR which removes the mysql client dependency for schema loading: https://github.com/gobuffalo/pop/pull/125

es:
It looks like @marpio was right. I was running `v0.11.0` - once I updated to `v0.12.4`, the tests ran fine without MySQL client installed locally. 

I've modified the PR to specify buffalo v0.12.4 or higher is required.

arschles:
Thank you! 😊

PR #336, opened 2018-07-26
download: add list from go cli

marwan-at-work:


pkg/download/goget/goget.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+	v, err := ref.Read()
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+
+	return v.Zip, nil
+}
```
arschles: the code in this file looks very similar to the `./pkg/module/go_get_fetcher.go` and `disk_ref.go` - any way you could build the goget protocol on top of those two primitives?

marwan-at-work: @arschles I am. Line 162 above in this function

arschles: I take this back completely - I see you are using the Fetcher and Ref!

arschles:
Ref #290 - this doesn't synchronously do cache fills but it's related because it's listing tags synchronously

cmd/proxy/actions/app_proxy.go
```diff
@@ -16,8 +17,9 @@ func addProxyRoutes(
 ) error {
 	app.GET("/", proxyHomeHandler)
 
+	dp := goget.New()
```
robjloranger: A nonsense nit but maybe `gg`, I guess `dp` was download protocol?

marwan-at-work: @robjloranger yeah dp for download protocol because the handlers don't care what the implementation is (go get, github, storage, etc) -- I think it would be really cool if in the very near future we made one download.Protocol interface that will do most of the magic for us: go get, storage, cache filling, etc.

marwan-at-work: And by magic, I don't really mean magic in the meta programming sense lol

pkg/download/download.go
```diff
@@ -0,0 +1,17 @@
+package download
+
+import (
+	"context"
+	"io"
+
+	"github.com/gomods/athens/pkg/storage"
+)
+
+// Protocol is the download protocol defined by cmd/go
```
arschles: > defined by cmd/go

do you mean that this is taken from `cmd/go`?

pkg/download/goget/goget.go
```diff
@@ -0,0 +1,179 @@
...
+	if err := afero.WriteFile(fs, gomodPath, gomodContent, 0666); err != nil {
+		return errors.E(op, err)
+	}
+	sourcePath := filepath.Join(repoRoot, "mod.go")
+	sourceContent := []byte(`package mod`)
+	if err := afero.WriteFile(fs, sourcePath, sourceContent, 0666); err != nil {
+		return errors.E(op, err)
+	}
+	return nil
+}
```
arschles: this code would be super useful in the `Fetcher.Fetch` method. can we move it down to there and then make list use it?

pkg/download/goget/goget_test.go
```diff
@@ -0,0 +1,37 @@
...
+	name    string
+	mod     string
+	version string
+}
+
+// TODO(marwan): we should create Test Repos under github.com/gomods
+// so we can get reproducible results from live VCS repos.
+// For now, I cannot test that github.com/pkg/errors returns v0.8.0
+// from goget.Latest, because they could very well introduce a new tag
+// in the near future.
```
arschles: sounds like a great idea! mind creating an issue for it?

pkg/download/download.go
marwan-at-work: Sorry the wording is definitely misleading. I meant they mirror the HTTP requests that cmd/go expects to use. I'll update the comment.

pkg/download/goget/goget.go
marwan-at-work: @arschles this is already there, I just copied it over. Should I export the function from that package?

pkg/download/goget/goget_test.go
marwan-at-work: yep

pkg/download/goget/goget.go
```diff
@@ -0,0 +1,161 @@
...
+	ref, err := fetcher.Fetch(mod, ver)
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+	defer ref.Clear()
+	v, err := ref.Read()
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+	defer v.Zip.Close()
```
robjloranger: It's kind of the same thing but you could probably just call Close as the zip is never used.

pkg/download/goget/goget.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	pseudoInfo := strings.Split(lr.Version, "-")
+	if len(pseudoInfo) < 3 {
+		return nil, errors.E(op, fmt.Errorf("malformed pseudoInfo %v", lr.Version))
+	}
+	return &storage.RevInfo{
+		Name:    pseudoInfo[2],
```
robjloranger: Should the name and the short name be the same? I'm not familiar.

marwan-at-work: @robjloranger the name is meant to be the full sha. But it should work without it. I'm currently testing that.

pkg/download/goget/goget.go
```diff
@@ -0,0 +1,161 @@
...
+	ref, err := fetcher.Fetch(mod, ver)
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+	defer ref.Clear()
+	v, err := ref.Read()
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+	defer v.Zip.Close()
```
robjloranger: Same as with Info, could just close zip now as it's not used.

pkg/download/goget/goget_test.go
robjloranger: Can we update the comment to ref: https://github.com/gomods/athens/issues/340

pkg/download/goget/goget_test.go
```diff
@@ -0,0 +1,37 @@
...
+	{"list vanity", "golang.org/x/tools", "latest"},
+}
+
+func TestList(t *testing.T) {
+	dp := New()
+	ctx := context.Background()
+
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			_, err := dp.List(ctx, tc.mod) // TODO ensure list is correct per TODO above.
```
robjloranger: Will this be in this PR?

robjloranger:
No worries then @arschles, just wanted to make sure it didn't go missed.

pkg/download/goget/goget_test.go
marwan-at-work: Probably after

arschles:
@carolynvs, @marwan-at-work and @Kunde21 discussed this PR on a screenshare. we are merging now, and doing follow-up refactors in #337 

🎉

PR #337, opened 2018-07-26
add middlewares to fill the cache

arschles:
also adding a middleware to fill in the module and version names

cc/ @marwan-at-work - this could be an alternative or complement (not sure?) to the download protocol  you introduced in #336 

this makes progress toward #290 because the cache fill middleware can wrap the zip, mod and info endpoints to give them sync. cache filling "for free"


pkg/download/cache_fill_middleware.go
```diff
@@ -0,0 +1,62 @@
...
+		versionInfo, err := getterSaver.Get(module, version)
+		if storage.IsNotFoundError(err) {
+			ref, err := fetcher.Fetch(module, version)
+			if err != nil {
+				err := errors.E(op, err)
+				lggr.SystemErr(err)
+				return err
+			}
+			defer ref.Clear()
+			version, err := ref.Read()
```
robjloranger: Is this meant to escape this block? Currently it is never used besides the deferred zip close below.

pkg/download/cache_fill_middleware.go
```diff
@@ -0,0 +1,62 @@
...
+				lggr.SystemErr(err)
+				return err
+			}
+			defer version.Zip.Close()
+		} else if err != nil {
+			err := errors.E(op, err)
+			lggr.SystemErr(err)
+			return err
+		}
+		c.Set("versionInfo", version)
```
robjloranger: Is this meant to set the `versionInfo` param to the string `version` from 23?

pkg/download/params_middleware.go
```diff
@@ -0,0 +1,20 @@
+package download
+
+import (
+	"github.com/gobuffalo/buffalo"
+	"github.com/gomods/athens/pkg/paths"
+)
+
+// ParamsMiddleware fetches module and version params from c and calls next with "value"
+// and "version" strings in the context. Access them with c.Value("module") and c.Value("version")
+func ParamsMiddleware(next buffalo.Handler) buffalo.Handler {
```
marwan-at-work: Why do we need middleware for this if we can just call path.GetAllParams from inside a handler?

marwan-at-work: PS. if it's the error handling part, paths.GetAllParams shouldn't really return an error. It can safely panic because that means we got something seriously screwed up in our http paths.

pkg/download/cache_fill_middleware.go
arschles: good call - this is supposed to be `versionInfo, err = ref.Read()`. change incoming

pkg/download/cache_fill_middleware.go
arschles: this should set to `versionInfo`, thanks for catching this!

pkg/download/cache_fill_middleware.go
```diff
@@ -0,0 +1,62 @@
...
+
+		if err := getterSaver.Save(
```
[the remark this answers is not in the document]
arschles: yup, that's a big bug on my part, thanks for catching

arschles:
@marwan-at-work RE your comment on overloading context - I was following the [buffalo middleware docs](https://gobuffalo.io/en/docs/middleware/) here. I def understand the type-safety argument and normally I'd be for it, but I think it's a tough tradeoff in this case because we exchange a bit of "magic" (setting `interface{}` context values) for less code in the handlers.

Would you be ok with submitting just the download protocol bits from https://github.com/gomods/athens/pull/336 in a new PR so we can get that in separately from the `/list` stuff in #336, and see how it looks in here?

arschles:
Closing in favor of #342

PR #338, opened 2018-07-26
Finish spans in the download handlers

arschles:
I don't know a ton about buffet and the opentracing library, but I saw that we're calling `defer sp.Finish()` in other parts of the code, so I figured we should do it here too.

cc/ @manugupt1 @robjloranger

### DIFF
--- a/pkg/download/list.go
+++ b/pkg/download/list.go
@@ -19,8 +19,7 @@ const PathList = "/{module:.+}/@v/list"
 func ListHandler(dp Protocol, lggr *log.Logger, eng *render.Engine) func(c buffalo.Context) error {
 	const op errors.Op = "download.ListHandler"
 	return func(c buffalo.Context) error {
-		sp := buffet.SpanFromContext(c)
-		sp.SetOperationName("listHandler")
+		sp := buffet.SpanFromContext(c).SetOperationName("listHandler")
 		defer sp.Finish()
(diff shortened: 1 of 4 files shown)


pkg/download/list.go
```diff
@@ -18,8 +18,8 @@ const PathList = "/{module:.+}/@v/list"
 // ListHandler implements GET baseURL/module/@v/list
 func ListHandler(lister storage.Lister, eng *render.Engine) func(c buffalo.Context) error {
 	return func(c buffalo.Context) error {
-		sp := buffet.SpanFromContext(c)
-		sp.SetOperationName("listHandler")
+		sp := buffet.SpanFromContext(c).SetOperationName("listHandler")
```
manugupt1: I do not know much either, but I am going through the docs as we speak.
From the docs, it looks like this wont really work as this does not start the span. You need to start the span somehow. 

One way is to directly use the function `sp.StartSpanFromContext(ctx, op_name)`
If you want to use this then you can probably. I like this better just because it can automatically wrap around a new span if the earlier context had a span. 

```
		sp := buffet.SpanFromContext(c).SetOperationName("listHandler")
		sp.Tracer().StartSpan()
                defer sp.Finish()
```

TBH, I have not seen in production. Once this is done, I would like to run olympus try it out and see what the results look like.

manugupt1:
Nice catch! thanks for doing this.

PR #342, opened 2018-07-27
Backfill the cache on a miss

carolynvs:
I am also stealing the glory and including Marwan's /latest endpoint.


arschles:
Ref #290

PR #343, opened 2018-07-27
GCP storage tests with mock

fedepaol:
Replacing the real GCP implementation with a mock bucket in order to perform local tests.

Fixes https://github.com/gomods/athens/issues/263


pkg/storage/gcp/gcp.go
```diff
@@ -45,6 +45,14 @@ func NewWithCredentials(ctx context.Context, cred option.ClientOption) (*Storage
 	}, nil
 }
 
+func newWithBucket(bkt Bucket) *Storage {
+	return &Storage{
+		bucket:       bkt,
+		baseURI:      nil,
```
marpio: We will have to test the BaseURL function so this shouldn't be nil. Maybe an argument to the newWithBucket function.

pkg/storage/gcp/gcp.go
```diff
@@ -45,6 +45,14 @@ func NewWithCredentials(ctx context.Context, cred option.ClientOption) (*Storage
 	}, nil
 }
 
+func newWithBucket(bkt Bucket) *Storage {
```
marpio: We will have to export this if we want to use a common test suite.

pkg/storage/gcp/bucket_mock.go
```diff
@@ -0,0 +1,95 @@
...
+	r.bucketMock.lock.RUnlock()
+	return nil
+}
+
+func (r *bucketWriter) Close() error {
+	r.bucketMock.lock.Unlock()
+	return nil
+}
+
+func (r *bucketWriter) Write(p []byte) (int, error) {
```
arschles: should this do a 

```go
r.lock.Lock()
defer r.lock.Unlock()
```

?

pkg/storage/gcp/bucket_mock.go
```diff
@@ -0,0 +1,95 @@
...
+	path string
+}
+
+func (r *bucketReader) Close() error {
+	r.bucketMock.lock.RUnlock()
+	return nil
+}
+
+func (r *bucketWriter) Close() error {
+	r.bucketMock.lock.Unlock()
```
arschles: can you help me understand why you need the unlock here and above?

pkg/storage/gcp/bucket_mock.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (m *bucketMock) Delete(ctx context.Context, path string) error {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+	delete(m.db, path)
+	return nil
+}
+
+func (m *bucketMock) Open(ctx context.Context, path string) (io.ReadCloser, error) {
```
arschles: does this need to use a read lock before you do the lookup?

pkg/storage/gcp/gcp.go
fedepaol: agreed

pkg/storage/gcp/gcp.go
fedepaol: That's correct. That could be changed after exporting it into the common test suite though

pkg/storage/gcp/bucket_mock.go
fedepaol: It might sound an overkill since the test are all called sequentially (and there are probably better ways to achieve the same result), however what I wanted to achieve was to protect the db until the whole read / write was done, treating it as a resource. I took leverage of the fact that Read and Write return a io.ReadCloser / io.WriteCloser and that the user of the api is expected to release the resource by closing the reader / writer.
So the mutex is taken while returning the reader / writer and is released while closing the reader / writer.

Mutexing the single Read / Write call would not be enough if the payload is consumed in more than one read / write call. Consider the case of a read that reads one byte at the time while a writer changes the record from a different goroutine for example.

pkg/storage/gcp/bucket_mock.go
fedepaol: Nope, as per the comment above. The lock is taken while returning the Writer, and released in Close

pkg/storage/gcp/bucket_mock.go
fedepaol: You are right, it does. Will change it

fedepaol:
@arschles commented inline, if the locking / unlocking machinery makes sense I'll push the nipticks, otherwise I can change that.

pkg/storage/gcp/bucket_mock.go
marpio: While we're at it - we could record the call to Close() (setting a bool is probably enough) and check it later in a test so that we know that it was called. Just a thought. What do you think @fedepaol ?

pkg/storage/gcp/bucket_mock.go
```diff
@@ -0,0 +1,95 @@
...
+	return nil
+}
+
+func (r *bucketWriter) Write(p []byte) (int, error) {
+	_, ok := r.bucketMock.db[r.path]
+	if !ok {
+		r.bucketMock.db[r.path] = make([]byte, 0)
+	}
+
+	r.bucketMock.db[r.path] = append(r.bucketMock.db[r.path], p...)
```
robjloranger: Does this mean that if you call write on a file that exists it will append the contents?

fedepaol: Yes. That was intended for handling writes that took more than one pass but has the side effect that different writers returned by different `Write` calls will append the content. 
Will change the `Write()` implementation to reset the map entry before returning a new `Writer`.

robjloranger: :+1: Thanks, it would just keep more consistent as the GCP library. Using `Object.Write` overwrites any existing file with the same name.

pkg/storage/gcp/bucket_mock.go
arschles: @fedepaol thank you for that amazing response and explanation. that lock-unlock flow you described sounds great to me! Also, +1 on the counter that @marpio mentioned 👍

pkg/storage/gcp/all_test.go
```diff
@@ -21,59 +18,20 @@ var (
...
-		return err
-	}
-	if err := bkt.Object(config.PackageVersionedName(mod, ver, "info")).Delete(ctx); err != nil {
-		return err
-	}
-	if err := bkt.Object(config.PackageVersionedName(mod, ver, "zip")).Delete(ctx); err != nil {
-		return err
-	}
-	return nil
+	g.url, _ = url.Parse(fmt.Sprintf("https://storage.googleapis.com/testbucket"))
```
robjloranger: no need for the format string printer.

fedepaol: That's correct, fixed!

robjloranger:
It looks like @arschles concerns were addressed so I'm going to merge this once the CI passes from updating the branch.

arschles:
🎉🎉🎉🎉🎉🎉🎉🎉🎉

Issue #329, opened 2018-07-25
Return a diskRef that can delete the entire gopath root


arschles:
After #291, the disk ref (in `./pkg/module/disk_ref.go`) will return a `diskRef` that will clean up the zip, info and mod files. It should delete the _entire_ GOPATH root directory that it owns

arschles:
Also, when this is done, the `TestGoGetFetcherFetch()` test in `./pkg/module/go_get_fetcher_test.go` should be updated to test that the entire GOPATH is deleted for that ref

robjloranger:
is this a duplicate of #318?

Issue #318, opened 2018-07-20
Make disk refs clean up their entire gopath


arschles:
After https://github.com/gomods/athens/pull/291, there will be a `diskRef` type that implements `(./pkg/module).Ref`. These `diskRef` types are intended to read a mod, zip and info from disk into memory and return them to the caller. They're also used to clean those files up but they don't do so completely. This issue is for making their `Clean` method able to clean up the entire path that they own

rchakra3:
I'd like to work on this!

marwan-at-work:
I believe this is fixed.

PR #354, opened 2018-07-29
Ensure diskRef cleans up everything in it's own GOPATH

rchakra3:
This PR addresses https://github.com/gomods/athens/issues/329

Adding info about the root dir and module name to the diskRef and adding the `getPackagePath` function to construct the path to the cached module seemed the cleanest way to achieve what was desired - though I'm open to suggestions for a cleaner implementation :)

Edit - #329 was closed in favor of #318

### DIFF
--- a/pkg/module/disk_ref.go
+++ b/pkg/module/disk_ref.go
@@ -2,7 +2,6 @@ package module
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 
@@ -56,24 +55,16 @@ func (d *diskRef) Read() (*storage.Version, error) {
 	var ver storage.Version
 
 	packagePath := getPackagePath(d.root, d.module)
-	infoFile, err := d.fs.Open(filepath.Join(packagePath, fmt.Sprintf("%s.info", d.version)))
-	if err != nil {
-		return nil, errors.WithStack(err)
-	}
-	defer infoFile.Close()
 
-	info, err := ioutil.ReadAll(infoFile)
+	infoFile := filepath.Join(packagePath, fmt.Sprintf("%s.info", d.version))
+	info, err := afero.ReadFile(d.fs, infoFile)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
 	ver.Info = info
 
-	modFile, err := d.fs.Open(filepath.Join(packagePath, fmt.Sprintf("%s.mod", d.version)))
-	if err != nil {
-		return nil, errors.WithStack(err)
-	}
-	defer modFile.Close()
-	mod, err := ioutil.ReadAll(modFile)
+	modFile := filepath.Join(packagePath, fmt.Sprintf("%s.mod", d.version))
+	mod, err := afero.ReadFile(d.fs, modFile)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
--- a/pkg/module/disk_ref.go
+++ b/pkg/module/disk_ref.go
@@ -3,6 +3,7 @@ package module
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 
 	"github.com/gomods/athens/pkg/storage"
@@ -15,14 +16,16 @@ import (
 // Do not create this struct directly. use newDiskRef
 type diskRef struct {
 	root    string
+	module  string
 	fs      afero.Fs
 	version string
 }
 
-func newDiskRef(fs afero.Fs, root, version string) *diskRef {
+func newDiskRef(fs afero.Fs, root, module, version string) *diskRef {
 	return &diskRef{
 		fs:      fs,
 		root:    root,
+		module:  module,
 		version: version,
 	}
 }
@@ -31,6 +34,20 @@ func newDiskRef(fs afero.Fs, root, version string) *diskRef {
 //
 // You should always call this function after you fetch a module into a DiskRef
 func (d *diskRef) Clear() error {
+
+	// This is required because vgo ensures dependencies are read-only
+	// See https://github.com/golang/go/issues/24111 and
+	// https://go-review.googlesource.com/c/vgo/+/96978
+	walkFn := func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		return d.fs.Chmod(path, 0770)
+	}
+	err := afero.Walk(d.fs, d.root, walkFn)
+	if err != nil {
+		return err
+	}
 	return d.fs.RemoveAll(d.root)
 }
 
@@ -38,7 +55,8 @@ func (d *diskRef) Clear() error {
 func (d *diskRef) Read() (*storage.Version, error) {
 	var ver storage.Version
 
-	infoFile, err := d.fs.Open(filepath.Join(d.root, fmt.Sprintf("%s.info", d.version)))
+	packagePath := getPackagePath(d.root, d.module)
+	infoFile, err := d.fs.Open(filepath.Join(packagePath, fmt.Sprintf("%s.info", d.version)))
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
@@ -50,7 +68,7 @@ func (d *diskRef) Read() (*storage.Version, error) {
 	}
 	ver.Info = info
 
-	modFile, err := d.fs.Open(filepath.Join(d.root, fmt.Sprintf("%s.mod", d.version)))
+	modFile, err := d.fs.Open(filepath.Join(packagePath, fmt.Sprintf("%s.mod", d.version)))
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
@@ -61,7 +79,7 @@ func (d *diskRef) Read() (*storage.Version, error) {
 	}
 	ver.Mod = mod
 
-	sourceFile, err := d.fs.Open(filepath.Join(d.root, fmt.Sprintf("%s.zip", d.version)))
+	sourceFile, err := d.fs.Open(filepath.Join(packagePath, fmt.Sprintf("%s.zip", d.version)))
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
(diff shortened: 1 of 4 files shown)


pkg/module/go_get_fetcher.go
```diff
@@ -34,38 +34,30 @@ func (g *goGetFetcher) Fetch(mod, ver string) (Ref, error) {
 	// setup the GOPATH
 	goPathRoot, err := afero.TempDir(g.fs, "", "athens")
 	if err != nil {
-		// TODO: return a ref for cleaning up the goPathRoot
-		// https://github.com/gomods/athens/issues/329
 		ref.Clear()
-		return ref, err
+		return newDiskRef(g.fs, goPathRoot, "", ""), err
```
marwan-at-work: Can we safely return nil here, and remove the ref.Clear() one line above?

pkg/module/go_get_fetcher.go
```diff
@@ -34,38 +34,30 @@ func (g *goGetFetcher) Fetch(mod, ver string) (Ref, error) {
...
 	}
 
 	// setup the module with barebones stuff
 	if err := Dummy(g.fs, modPath); err != nil {
-		// TODO: return a ref for cleaning up the goPathRoot
-		// https://github.com/gomods/athens/issues/329
 		ref.Clear()
-		return ref, err
+		return newDiskRef(g.fs, goPathRoot, "", ""), err
```
marwan-at-work: same as above

pkg/module/go_get_fetcher.go
```diff
@@ -34,38 +34,30 @@ func (g *goGetFetcher) Fetch(mod, ver string) (Ref, error) {
...
 	}
 
-	cachePath, err := getSources(g.goBinaryName, g.fs, goPathRoot, modPath, mod, ver)
+	err = getSources(g.goBinaryName, g.fs, goPathRoot, modPath, mod, ver)
 	if err != nil {
-		// TODO: return a ref that cleans up the goPathRoot
-		// https://github.com/gomods/athens/issues/329
 		ref.Clear()
-		return nil, err
+		return newDiskRef(g.fs, goPathRoot, "", ""), err
```
marwan-at-work: same as above

pkg/module/go_get_fetcher.go
```diff
@@ -34,38 +34,30 @@ func (g *goGetFetcher) Fetch(mod, ver string) (Ref, error) {
...
 	}
 	// TODO: make sure this ref also cleans up the goPathRoot
 	// https://github.com/gomods/athens/issues/329
-	return newDiskRef(g.fs, cachePath, ver), err
+	return newDiskRef(g.fs, goPathRoot, mod, ver), err
```
marwan-at-work: s/err/nil

pkg/module/disk_ref.go
```diff
@@ -61,7 +79,7 @@ func (d *diskRef) Read() (*storage.Version, error) {
 	}
 	ver.Mod = mod
 
-	sourceFile, err := d.fs.Open(filepath.Join(d.root, fmt.Sprintf("%s.zip", d.version)))
+	sourceFile, err := d.fs.Open(filepath.Join(packagePath, fmt.Sprintf("%s.zip", d.version)))
```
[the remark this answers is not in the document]
marwan-at-work: In fact, could you make `ver.Zip` below take `&refCloser{zip: sourceFile, ref: d}` and then you can in this file do the following: 

```golang
type refCloser struct {
	zip  io.ReadCloser
	ref Ref
}

func (rc *refCloser) Close() error {
	rc.zip.Close()
	return rc.Ref.Clear()
}

func (rc *refCloser) Read(p []byte) (n int, err error) {
	return rc.zip.Read(p)
}
```

This way we ensure a path is cleaned up once its Zip is fully closed. 

This can be another pr since it addresses https://github.com/gomods/athens/issues/355 but it's also very closely related to what this pr does. so it's up to you :)

pkg/module/disk_ref.go
```diff
@@ -31,14 +34,29 @@ func newDiskRef(fs afero.Fs, root, version string) *diskRef {
 //
 // You should always call this function after you fetch a module into a DiskRef
 func (d *diskRef) Clear() error {
+
+	// This is required because vgo ensures dependencies are read-only
+	// See https://github.com/golang/go/issues/24111 and
+	// https://go-review.googlesource.com/c/vgo/+/96978
+	walkFn := func(path string, info os.FileInfo, err error) error {
```
marwan-at-work: in unix you can pass an `-R` flag and it will chmod all the sub files and directories. Is there an equivalent in Go? just wondering so we don't have to walk the whole dir with potentially many files.

pkg/module/disk_ref.go
```diff
@@ -31,14 +34,29 @@ func newDiskRef(fs afero.Fs, root, version string) *diskRef {
...
 	return d.fs.RemoveAll(d.root)
 }
 
 // read is the Ref interface implementation.
 func (d *diskRef) Read() (*storage.Version, error) {
 	var ver storage.Version
 
-	infoFile, err := d.fs.Open(filepath.Join(d.root, fmt.Sprintf("%s.info", d.version)))
+	packagePath := getPackagePath(d.root, d.module)
+	infoFile, err := d.fs.Open(filepath.Join(packagePath, fmt.Sprintf("%s.info", d.version)))
```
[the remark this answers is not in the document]
rchakra3: that's a good point!

pkg/module/disk_ref.go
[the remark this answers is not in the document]
rchakra3: absolutely! I'll make that change

pkg/module/go_get_fetcher.go
rchakra3: If we return a nil here the caller will not be able to call .Clear() to delete the temp dir created on disk for the GOPATH (that's done in line 35). This applies to all the subsequent comments as well. 

Yes I believe we could remove the ref.Clear() above since this it's using a NoOp ref, but thought it made sense to call Clear once we're done with the ref in case there is actually some cleanup involved in the future. For now, yes we can definitely get rid of it!

pkg/module/go_get_fetcher.go
rchakra3: Oops good catch!

pkg/module/disk_ref.go
rchakra3: I tried looking but was unsuccessful - do you know of something?

rchakra3:
@marwan-at-work thanks for the review! I'll address the changes soon

pkg/module/disk_ref.go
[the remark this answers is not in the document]
rchakra3: Actually I've commented on the issue for some clarification when you get a chance.

rchakra3:
@marwan-at-work I'm going to submit a separate PR for #355 based on our discussion. Does everything else look good?

pkg/module/go_get_fetcher.go
```diff
@@ -29,43 +29,29 @@ func NewGoGetFetcher(goBinaryName string, fs afero.Fs) Fetcher {
...
-	ref := noopRef{}
 
 	// setup the GOPATH
 	goPathRoot, err := afero.TempDir(g.fs, "", "athens")
 	if err != nil {
-		// TODO: return a ref for cleaning up the goPathRoot
-		// https://github.com/gomods/athens/issues/329
-		ref.Clear()
-		return ref, err
+		return newDiskRef(g.fs, goPathRoot, "", ""), err
```
marwan-at-work: return nil, err

pkg/module/go_get_fetcher.go
```diff
@@ -29,43 +29,29 @@ func NewGoGetFetcher(goBinaryName string, fs afero.Fs) Fetcher {
...
 	}
 	sourcePath := filepath.Join(goPathRoot, "src")
 	modPath := filepath.Join(sourcePath, getRepoDirName(mod, ver))
 	if err := g.fs.MkdirAll(modPath, os.ModeDir|os.ModePerm); err != nil {
-		// TODO: return a ref for cleaning up the goPathRoot
-		// https://github.com/gomods/athens/issues/329
-		ref.Clear()
-		return ref, err
+		return newDiskRef(g.fs, goPathRoot, "", ""), err
```
marwan-at-work: return nil, err

pkg/module/go_get_fetcher.go
```diff
@@ -29,43 +29,29 @@ func NewGoGetFetcher(goBinaryName string, fs afero.Fs) Fetcher {
...
 	}
 
 	// setup the module with barebones stuff
 	if err := Dummy(g.fs, modPath); err != nil {
-		// TODO: return a ref for cleaning up the goPathRoot
-		// https://github.com/gomods/athens/issues/329
-		ref.Clear()
-		return ref, err
+		return newDiskRef(g.fs, goPathRoot, "", ""), err
```
marwan-at-work: return nil, err

pkg/module/go_get_fetcher.go
```diff
@@ -29,43 +29,29 @@ func NewGoGetFetcher(goBinaryName string, fs afero.Fs) Fetcher {
...
 	}
 
-	cachePath, err := getSources(g.goBinaryName, g.fs, goPathRoot, modPath, mod, ver)
+	err = getSources(g.goBinaryName, g.fs, goPathRoot, modPath, mod, ver)
 	if err != nil {
-		// TODO: return a ref that cleans up the goPathRoot
-		// https://github.com/gomods/athens/issues/329
-		ref.Clear()
-		return nil, err
+		return newDiskRef(g.fs, goPathRoot, "", ""), err
```
marwan-at-work: return nil, err

rchakra3:
@marwan-at-work I've handled cleaning up the temp directory created by the diskref before returning errors. It made sense since `Clear` will not be a part of the interface anymore.

Thanks for the review and comments :)

rchakra3:
@robjloranger thanks for taking a look!

Let me know when it's good to merge, I'll rebase it one last time. My branch keeps getting out of date :)

marwan-at-work:
@rchakra3 this is good to merge. Please rebase before I hit the merge button :)

marwan-at-work:
Thank you @rchakra3 - let me know if you need any help with issue 335

Issue #358, opened 2018-07-29
Stream Zip Modules to User and to Storage simultaneously 


marwan-at-work:
When we do a cache fill, we risk completely reading the zip file out (in order to store it to backend), so we end up giving the user an empty reader. 

As a first implementation, we should just do storage.Get once storage.Save succeeds and that's the easiest way to do it. In the future, we should be able to stream from the upstream directly to storage and to the client that requested the zip.

Issue #357, opened 2018-07-29
Mongo: stream zip modules


marwan-at-work:
Our MongoDB implementation of storage.Backend stores entire zip modules in json objects which makes it super slow and inefficient. We can use MongoDB's GridFS feature to store/stream zip modules instead.

arschles:
Waaaaaat. That’s awesome. Let’s measure latencies?

Issue #356, opened 2018-07-29
Encode/Decode path according to Go Modules


marwan-at-work:
Modules are encoded in a weird way for upper case packages and currently they blow up on athens. We probably need to copy/paste from go's repository or implement our own encode/decoder

This is from `go help goproxy`

>To avoid problems when serving from case-sensitive file systems,
the < module > and < version > elements are case-encoded, replacing every
uppercase letter with an exclamation mark followed by the correponding
lower-case letter: github.com/Azure encodes as github.com/!azure.

Encoder https://github.com/golang/go/blob/master/src/cmd/go/internal/module/module.go#L431
Decoer is somewhere there as well. 

To test: run Athens against `github.com/NYTimes/gizmo`

PR #359, opened 2018-07-29
Athens: encode mods + fix handler bugs + stream mongo

marwan-at-work:
Fixes https://github.com/gomods/athens/issues/358, https://github.com/gomods/athens/issues/357, and half of https://github.com/gomods/athens/issues/356

Also some handler clean ups where we were calling WriteHeader twice.

### DIFF
--- a/pkg/download/version_module.go
+++ b/pkg/download/version_module.go
@@ -1,8 +1,6 @@
 package download
 
 import (
-	"net/http"
-
 	"github.com/bketelsen/buffet"
 	"github.com/gobuffalo/buffalo"
 	"github.com/gobuffalo/buffalo/render"
@@ -23,18 +21,12 @@ func VersionModuleHandler(dp Protocol, lggr *log.Logger, eng *render.Engine) buf
 		if err != nil {
 			err = errors.E(op, errors.M(mod), errors.V(ver), err)
 			lggr.SystemErr(err)
-			c.Render(errors.Kind(err), nil)
+			return c.Render(errors.Kind(err), nil)
 		}
 		verInfo.Zip.Close()
-		status := http.StatusOK
-		_, err = c.Response().Write(verInfo.Mod)
-		if err != nil {
-			err = errors.E(op, errors.M(mod), errors.V(ver), err)
-			status = http.StatusInternalServerError
-			lggr.SystemErr(err)
-		}
 
-		c.Response().WriteHeader(status)
-		return nil
+		// Calling c.Response().Write will write the header directly
+		// and we would get a 0 status in the buffalo logs.
+		return c.Render(200, eng.String(string(verInfo.Mod)))
 	}
 }
--- a/pkg/download/version_zip.go
+++ b/pkg/download/version_zip.go
@@ -2,7 +2,6 @@ package download
 
 import (
 	"io"
-	"net/http"
 
 	"github.com/bketelsen/buffet"
 	"github.com/gobuffalo/buffalo"
@@ -25,7 +24,7 @@ func VersionZipHandler(dp Protocol, lggr *log.Logger, eng *render.Engine) buffal
 		if err != nil {
 			err := errors.E(op, errors.M(mod), errors.V(ver), err)
 			lggr.SystemErr(err)
-			return c.Render(http.StatusInternalServerError, nil)
+			return c.Render(errors.Kind(err), nil)
 		}
 		defer verInfo.Zip.Close()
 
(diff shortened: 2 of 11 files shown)


pkg/paths/decode.go
```diff
@@ -0,0 +1,301 @@
...
+	"strings"
+	"unicode"
+	"unicode/utf8"
+)
+
+// Ripped from cmd/go
+
+// DecodePath returns the module path of the given safe encoding.
+// It fails if the encoding is invalid or encodes an invalid path.
+func DecodePath(encoding string) (path string, err error) {
```
michalpristas: please include table test with good/bad inputs, also serves as docs/examples

pkg/download/version_module.go
```diff
@@ -23,18 +21,12 @@ func VersionModuleHandler(dp Protocol, lggr *log.Logger, eng *render.Engine) buf
...
-			err = errors.E(op, errors.M(mod), errors.V(ver), err)
-			status = http.StatusInternalServerError
-			lggr.SystemErr(err)
-		}
 
-		c.Response().WriteHeader(status)
-		return nil
+		// Calling c.Response().Write will write the header directly
+		// and we would get a 0 status in the buffalo logs.
+		return c.Render(200, eng.String(string(verInfo.Mod)))
```
arschles: 👍

pkg/download/version_zip.go
```diff
@@ -25,17 +24,18 @@ func VersionZipHandler(dp Protocol, lggr *log.Logger, eng *render.Engine) buffal
...
+		// Calling c.Response().Write will write the header directly
+		// and we would get a 0 status in the buffalo logs.
+		c.Render(200, nil)
 		_, err = io.Copy(c.Response(), verInfo.Zip)
 		if err != nil {
-			status = http.StatusInternalServerError
 			lggr.SystemErr(errors.E(op, errors.M(mod), errors.V(ver), err))
 		}
 
-		c.Response().WriteHeader(status)
```
arschles: is it cool with you to not send an error back down to the client in this case? and are we just unable to do it b/c of the buffalo bug?

pkg/paths/path.go
```diff
@@ -31,9 +31,9 @@ func GetAllParams(c buffalo.Context) (*AllPathParams, error) {
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
-	version := c.Param("version")
-	if version == "" {
-		return nil, errors.E(op, "missing version paramater")
+	version, err := DecodeVersion(c.Param("version"))
```
arschles: can you remind me why we need `DecodeVersion`? and `DecodePath`? it seems like a lot of duplicated code that would be _amazing_ if we could avoid, so I'm wondering if there's any way around it

pkg/storage/mongo/getter.go
```diff
@@ -1,18 +1,18 @@
...
 
 	"github.com/globalsign/mgo/bson"
+	"github.com/gomods/athens/pkg/errors"
 	"github.com/gomods/athens/pkg/storage"
 	opentracing "github.com/opentracing/opentracing-go"
 )
 
 // Get a specific version of a module
 func (s *ModuleStore) Get(module, vsn string) (*storage.Version, error) {
+	const op errors.Op = "mongo.Get"
```
arschles: @marwan-at-work I'm really liking this standardized error handling. just thought you should know - not really related to this PR

🎉

pkg/download/version_zip.go
marwan-at-work: @arschles Yes. The only possible error is that if the connection dropped or we went some falsy data, which in both cases the client will figure out. Plus, per the comment I put above, it's really our only option :)

pkg/paths/decode.go
marwan-at-work: will do

pkg/paths/path.go
marwan-at-work: @arschles find the reason here :) https://github.com/gomods/athens/issues/356

pkg/storage/mongo/getter.go
marwan-at-work: @arschles I appreciate you 🕺

marwan-at-work:
@arschles yes, I can split the PR. The difference between GridFS and loading the whole zip into memory doesn't need benchmarking, it's visible to the naked eye lol

arschles:
Fair enough! Thanks for splitting up

PR #361, opened 2018-07-30
Add support for Go Modules

komuw:
# what
- I added support for Go modules

# why
- Go 1.11 is a few weeks away(if not days) 

# how
- I ran `go mod -init` which used the existing `Gopkg.lock` file to create a `go.mod` file
- go mod -init didn't however add `gobuffalo/plus` which I had to add manually via; `go get github.com/gobuffalo/plush@v3.7.7`
- I also removed the `pkg/config/env/go.go`'s `GoPath` function[1]; this is because;
  - that function does not appear to be used to do anything useful, it is only used in init func of proxy/actions/app.go[2] and even there it doesn't do much except panic if it does not exist
  - I do not think we ought to be relying on the existence of $GOPATH, especially post go version 1.11
  - I'm using a Go version compiled from tip and working outside $GOPATH

- I also removed `https://github.com/gomods/athens/tree/master/testmodule` since it is not in use and appears to have been an experiment cc @arschles 

# ref;
1. https://github.com/gomods/athens/blob/53df0f7c6ae93cdef88833b6514c98497110cc25/pkg/config/env/go.go#L15-L23

2. https://github.com/gomods/athens/blob/53df0f7c6ae93cdef88833b6514c98497110cc25/cmd/proxy/actions/app.go#L42-L49

### DIFF
--- a/pkg/module/disk_ref.go
+++ b/pkg/module/disk_ref.go
@@ -5,8 +5,9 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/paths"
 	"github.com/gomods/athens/pkg/storage"
-	"github.com/pkg/errors"
 	"github.com/spf13/afero"
 )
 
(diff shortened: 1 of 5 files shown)


arschles:
Really really really really cool! Would you be ok if we hold off on this
until Go 1.11 is officially released?

Also, what do you think about hosting a proxy that our project can use? We
can do that as a separate PR if you like the idea

komuw:
@arschles 
`Would you be ok if we hold off on this until Go 1.11?`
I'm okay with that

`what do you think about hosting a proxy that our project can use?`
I'm not sure I understand fully, do you mean: just like the way there's going to be a `microsoft.registry.golang.org` etc  that in the same vein there should also be a `microsoft.proxy.golang.org` etc?

komuw:
superseded by:

1. https://github.com/gomods/athens/pull/369
2 https://github.com/gomods/athens/pull/370
3. https://github.com/gomods/athens/pull/371

PR #362, opened 2018-07-30
Make it easier to run all the same commands that Travis does 

carolynvs:
* Extract the commands from `.travis.yml` into `./scripts/`
* Add targets to our make file to run groups of scripts
* Changed the buffalo install so that it doesn't do all it's dirty work in the repo repo, because it modifies existing files like our license and readme
* Add a contributing guide that explains our new make targets

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,8 @@
+Hurray! We are glad that you want to contribute to our project! 👍
+
+## Verify your work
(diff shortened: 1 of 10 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,8 @@
+Hurray! We are glad that you want to contribute to our project! 👍
+
+## Verify your work
```
michalpristas: this is joy

PR #363, opened 2018-07-30
Run E2E tests during CI

### LINKED ISSUE #353: Add end-to-end test for the proxy to CI
Now that #290 is done, a `GOPROXY=proxyurl go get stuff` works against a running proxy. We should totally do that in our CI tests!

cc/ @carolynvs since you were interested in this

carolynvs:
* Add `./scripts/test_e2e.sh` which builds a go program that relies on go modules using GOPROXY.
* Add `make test-e2e` target and document it. It's important to run `make setup-dev-env` once, before running the tests.
* Use the same setup for travis and local development for the storage services

Closes #353

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 services:
-  - mongodb
-  - postgresql
-  - redis-server
+  - docker
 language: go
 install: false
 go:
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ before_script:
   - buffalo db create
   - buffalo db migrate up
 script:
-  - make verify test-unit
+  - make verify test-unit test-e2e
 after_success:
   - if [ "${CODE_COV}" == "1" ]; then
       curl -s https://codecov.io/bash -o codecov && bash codecov -X fix;
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,9 @@ env:
     - MINIO_SECRET_KEY=minio123
     - ATHENS_MONGO_STORAGE_URL=mongodb://127.0.0.1:27017
     - CODE_COV=1
-    - GO_BINARY_PATH=vgo
     - TMPDIR=/tmp/
+    - GO_SOURCE=${TMPDIR}go
+    - GO_BINARY_PATH=${TMPDIR}go/bin/go
 before_script:
   - make setup-dev-env
   - wget "https://dl.minio.io/server/minio/release/linux-amd64/minio"
--- a/scripts/test_e2e.sh
+++ b/scripts/test_e2e.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+# test_e2e.sh
+# Execute end-to-end (e2e) tests to verify that everything is working right
+# from the end user perpsective
+set -xeuo pipefail
+
+REPO_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )/.."
+
+# Use a version of Go that supports Go Modules
+export GO111MODULES=on
+GOMOD_CACHE=$(go env GOPATH)/src/mod
+GO_SOURCE=${GO_SOURCE:=$(go env GOPATH)/src/golang.org/x/go}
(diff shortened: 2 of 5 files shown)


carolynvs:
Oops, I think I dorked something up on this PR. Please ignore and I'll fix later.

arschles:
Thanks @carolynvs ! Let me know if I can help on this in any way. These tests would be awesome to have in here 👍 🎉

carolynvs:
@arschles Ok! I've gotten our travis and local dev setup more consistent and a very simple end-to-end test passing.

Someone may prefer to rewrite it in Go, vs Bash, but this at least gets us some coverage for now! 😀

carolynvs:
At some point we may be able to use gimme go1.11beta2 in our travis config instead of building go from source, I didn't want to make that jump at this time though. I'm pretty sure we may need to make more changes once start building Athens using go 1.11

.travis.yml
```diff
@@ -1,7 +1,5 @@
 services:
-  - mongodb
-  - postgresql
-  - redis-server
+  - docker
```
michalpristas: this escalated quickly 👍

scripts/test_e2e.sh
```diff
@@ -0,0 +1,53 @@
...
+# test_e2e.sh
+# Execute end-to-end (e2e) tests to verify that everything is working right
+# from the end user perpsective
+set -xeuo pipefail
+
+REPO_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )/.."
+
+# Use a version of Go that supports Go Modules
+export GO111MODULES=on
+GOMOD_CACHE=$(go env GOPATH)/src/mod
```
komuw: @carolynvs 
In Go1.11beta the location of the module cache is `$GOPATH/src/mod`
However on tip the location has changed to `$GOPATH/pkg/mod`[1]

Maybe you could just declare;
```bash
clearGoModCache () {
  # The sudo is a necessary workaround until go is fixed
  sudo rm -fr $GOPATH/*
}
```

**ref:**
1. https://go-review.googlesource.com/c/go/+/126755

scripts/test_e2e.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+# test_e2e.sh
+# Execute end-to-end (e2e) tests to verify that everything is working right
+# from the end user perpsective
+set -xeuo pipefail
```
komuw: @carolynvs 
I do not know what version of bash travisCI uses,
but Bash version 4.3 and older doesn't handle `set -u` very well in some circumstances(arrays)
But I don't think you are using arrays so, this may be fine

https://github.com/anordal/shellharden/blob/master/how_to_do_things_safely_in_bash.md#how-to-begin-a-bash-script

scripts/test_e2e.sh
```diff
@@ -0,0 +1,53 @@
...
+  # Cleanup after our tests
+  pkill buffalo || true
+  popd 2> /dev/null || true
+}
+trap teardown EXIT
+
+# Start the proxy in the background and wait for it to be ready
+export GO_BINARY_PATH=${GO_SOURCE}/bin/go
+cd $REPO_DIR/cmd/proxy
+pkill buffalo || true # cleanup old buffalos
```
komuw: `cleanup old buffalos` :smile:

scripts/test_e2e.sh
robjloranger: I think this should be ignored until the change is in an official Go release as we do not work with go from tip.

carolynvs:
I've updated the PR with the recent changes from master.

michalpristas:
@carolynvs i think this is good to go, do you plan to do any more changes?

carolynvs:
This is ready to merge! I can’t , not a committee on this repo, so anyone has ca feel free to :shipit:

michalpristas:
I'm just bringing this even with master and will merge

robjloranger:
Thanks, was just about to as well 😁 sorry @carolynvs I got distracted at work yesterday.

PR #365, opened 2018-07-30
backend/mongo: use GridFS to stream large zips

marwan-at-work:
fixes https://github.com/gomods/athens/issues/357


pkg/storage/mongo/getter.go
```diff
@@ -24,9 +24,17 @@ func (s *ModuleStore) Get(ctx context.Context, module, vsn string) (*storage.Ver
 		}
 		return nil, err
 	}
+
+	zipName := s.gridFileName(module, vsn)
+	fs := s.s.DB(s.d).GridFS("fs")
+	f, err := fs.Open(zipName)
```
michalpristas: dont you need to close the file? i see gridfs has exported method for that

michalpristas: thanks @marwan-at-work missed that.

michalpristas: i'm thinking whether or not to rename ZIp -> ZipReadCloser 
so it's obvious you have to close it

marwan-at-work: Zip *is* a ReadCloser. As long as your code editor shows you the type, then it should be clear. But if you think being more verbose is better, happy to change it :)

michalpristas: no i guess there's no need. it's not a common practice so lets avoid it.

marwan-at-work: @michalpristas sounds good. Let me know if this PR is good on your side : )

PR #366, opened 2018-07-30
athens: encode/decode modules per cmd/go rules

marwan-at-work:
fixes https://github.com/gomods/athens/issues/356


arschles:
Yesssss!

marwan-at-work:
appreciate it 🕺

Issue #355, opened 2018-07-29
Make the zip's io.ReadCloser delete the entire goPath


marwan-at-work:
Currently the buffalo handlers don't know anything about disks and refs as they are low level implementations of getting modules. However, we need a way to ensure things are cleaned up once a module is fully used. 

Therefore when `diskRef` returns a *storage.Version, we should have the io.ReadCloser of its zip be able to delete the entire temp gopath since there's practically no need for that path once the client downloads the zip, all other fields are in memory so we're good.

rchakra3:
Does this mean the expected behavior would be:
```
ref := fetcher.Fetch(..)
ref.zip.Read()
ref.zip.Close()
```
I thought the expectation was that the caller calls ref.Clear() ? Something like:
```
ref := fetcher.Fetch(..)
ref.zip.Read()
ref.zip.Close()
ref.Clear()
```

Which would actually delete everything on disk

rchakra3:
Gotcha, that makes sense. 
I just went through the various calls and I personally think it would be cleaner to remove Clear from the Ref interface and let zip.Close() handle cleanup. 

The alternative would be that we read the zip file into memory and pass that around (which is mentioned in the comments for diskref) - that way the flow would be to call `ref.Clear()` right after `ref.Read()`. That would be nice to have but could possibly end up using large amounts of memory

marwan-at-work:
@rchakra3 yes with the alternative we would run out of memory very fast. Feel free to remove the Clear method from the interface definition. Thank you!

arschles:
All good with me. Can you add really good docs (with code samples hopefully
😊) that explain how to clean up? Thank you thank you!

PR #367, opened 2018-07-31
zip.Close() also clears the GOPATH of the underlying diskRef

rchakra3:
This PR address #355

### DIFF
--- a/pkg/module/disk_ref.go
+++ b/pkg/module/disk_ref.go
@@ -38,13 +38,6 @@ func (rc *zipReadCloser) Read(p []byte) (n int, err error) {
 	return rc.zip.Read(p)
 }
 
-func newZipReadCloser(zip io.ReadCloser, ref *diskRef) io.ReadCloser {
-	return &zipReadCloser{
-		zip: zip,
-		ref: ref,
-	}
-}
-
 func newDiskRef(fs afero.Fs, root, module, version string) *diskRef {
 	return &diskRef{
 		fs:      fs,
@@ -111,7 +104,7 @@ func (d *diskRef) Read() (*storage.Version, error) {
 	//
 	// if we close, then the caller will panic, and the alternative to make this work is
 	// that we read into memory and return an io.ReadCloser that reads out of memory
-	ver.Zip = newZipReadCloser(sourceFile, d)
+	ver.Zip = &zipReadCloser{sourceFile, d}
 
 	return &ver, nil
 }
--- a/pkg/module/disk_ref.go
+++ b/pkg/module/disk_ref.go
@@ -2,6 +2,7 @@ package module
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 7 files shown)


pkg/module/go_get_fetcher_test.go
```diff
@@ -32,8 +34,34 @@ func (s *ModuleSuite) TestGoGetFetcherFetch() {
...
 
-	r.NoError(ref.Clear())
+	// close the version's zip file (which also cleans up the underlying diskref's GOPATH) and expect it to fail again
+	r.NoError(ver.Zip.Close())
 	ver, err = ref.Read()
 	r.NotNil(err)
 	r.Nil(ver)
 }
+
+func ExampleFetch() {
```
rchakra3: This is a little verbose and doesn't use helper functions since it's mean't to be an example that appears in the godocs

pkg/module/disk_ref.go
```diff
@@ -78,7 +100,7 @@ func (d *diskRef) Read() (*storage.Version, error) {
 	//
 	// if we close, then the caller will panic, and the alternative to make this work is
 	// that we read into memory and return an io.ReadCloser that reads out of memory
-	ver.Zip = sourceFile
+	ver.Zip = newZipReadCloser(sourceFile, d)
```
marwan-at-work: no need for a constructor, just do `ver.Zip = &zipReadCloser{sourceFile, d)`

pkg/module/disk_ref.go
```diff
@@ -20,6 +21,29 @@ type diskRef struct {
...
+func (rc *zipReadCloser) Close() error {
+	rc.zip.Close()
+	return clearFiles(rc.ref.fs, rc.ref.root)
+}
+
+func (rc *zipReadCloser) Read(p []byte) (n int, err error) {
+	return rc.zip.Read(p)
+}
+
+func newZipReadCloser(zip io.ReadCloser, ref *diskRef) io.ReadCloser {
```
marwan-at-work: you can remove this function and just inline below.

rchakra3: yeah that makes sense, that was unnecessarily verbose :)

Thanks for the review!

pkg/module/go_get_fetcher_test.go
```diff
@@ -32,8 +34,34 @@ func (s *ModuleSuite) TestGoGetFetcherFetch() {
...
+	repoURI := "github.com/arschles/assert"
+	version := "v1.0.0"
+	goBinaryName := os.Getenv("GO_BINARY_PATH")
```
robjloranger: Here you can use (./pkg/config/environment).GoBinPath

pkg/module/go_get_fetcher_test.go
```diff
@@ -32,8 +34,34 @@ func (s *ModuleSuite) TestGoGetFetcherFetch() {
...
+	if goBinaryName == "" {
+		goBinaryName = "go"
```
robjloranger: After above you can omit this check/set as the pkg defaults to go.

pkg/module/go_get_fetcher_test.go
rchakra3: I did see that. Since this is meant to be an example in the godoc, I wanted to keep it really simple so the reader could understand how to use Fetch without looking anything else up.

Definitely open to switching to `GoBinPath()` if you believe that's better here.

marwan-at-work: "GO_BINRAY_PATH" is an Athens convention anyway so it's probably ok to do `env.GoBinPath()`

rchakra3:
Ok, makes sense! I'll push an update soon

rchakra3:
There were some merge conflicts so I've just rebased. I think this is good to merge now

pkg/module/disk_ref_test.go
```diff
@@ -58,6 +59,42 @@ func (m *ModuleSuite) TestDiskRefReadAndClear() {
...
+	_, err := m.fs.Stat(filePath)
+	r.NoError(err)
+
+	// Now clear the files
+	err = clearFiles(m.fs, root)
+	r.NoError(err)
+
+	// Validate the file has been deleted
+	_, err = m.fs.Stat(filePath)
+	expErr := fmt.Sprintf("open %s: file does not exist", filePath)
```
marwan-at-work: @rchakra3 use `r.True(os.IsNotExist(err))`

marwan-at-work:
@rchakra3 I'll merge this for now since it's a minor comment and feel free to push another pr for it :)

Issue #327, opened 2018-07-25
move defined errors into pkg/errors


marwan-at-work:
now that we have pkg/errors, we should move all user defined errors such as storage/NotExist into pkg/errors

PR #373, opened 2018-07-31
Athens: move storage.NotFound to errors.NotFound

marwan-at-work:
Fixes https://github.com/gomods/athens/issues/327

Removed ErrNotFound and ErrVersionNotFound into one NotFound error in the errors package. 

All the plumbing.


pkg/download/download.go
```diff
@@ -49,7 +49,7 @@ func (p *protocol) List(ctx context.Context, mod string) ([]string, error) {
 func (p *protocol) Info(ctx context.Context, mod, ver string) ([]byte, error) {
 	const op errors.Op = "protocol.Info"
 	v, err := p.s.Get(ctx, mod, ver)
-	if storage.IsNotFoundError(err) {
+	if errors.ErrNotFound(err) {
```
michalpristas: can we, as this is check, stick with IsErrNotFound?

marwan-at-work: @michalpristas how about `IsNotFoundErr` ? to stay consist with std lib's `os.IsNotExist` -- grammatically speaking.

michalpristas: 👍 Great!

PR #376, opened 2018-08-01
goget cache: its a moving target

marwan-at-work:
Go caches its modules inside of `GOPATH/pkg` instead of `GOPATH/src` similar to dep. 

Commit: https://github.com/golang/go/commit/b8f42d74e87aeec189f53e9fdb2a7e6026c099b1#diff-18bbe9fc05441fe3191725c83637e767

### DIFF
--- a/pkg/module/go_get_fetcher.go
+++ b/pkg/module/go_get_fetcher.go
@@ -146,5 +146,5 @@ func getRepoDirName(repoURI, version string) string {
 
 // getPackagePath returns the path to the module cache given the gopath and module name
 func getPackagePath(gopath, module string) string {
-	return filepath.Join(gopath, "src", "mod", "cache", "download", module, "@v")
+	return filepath.Join(gopath, "pkg", "mod", "cache", "download", module, "@v")
 }


pkg/module/go_get_fetcher.go
```diff
@@ -146,5 +146,5 @@ func getRepoDirName(repoURI, version string) string {
 
 // getPackagePath returns the path to the module cache given the gopath and module name
 func getPackagePath(gopath, module string) string {
-	return filepath.Join(gopath, "src", "mod", "cache", "download", module, "@v")
+	return filepath.Join(gopath, "pkg", "mod", "cache", "download", module, "@v")
```
michalpristas: this as well

https://github.com/gomods/athens/blob/441e21d28113ad3de1e1fd757300ee4d9cd7804b/pkg/module/go_get_fetcher.go#L38

michalpristas: sorry, ignore,

michalpristas:
this will need to go in when this commit makes it into release. as we're not targeting latest vgo it gets better with adoption

PR #377, opened 2018-08-01
Goget Fetcher should error out if gobinpath is not valid

### LINKED ISSUE #372: Check "go" exists in your path before running server
We should have gogetfetcher.New() stat the goBinPath and returns an error if go cannot be found.

devdinu:
* Using `which gobinary` to verify whether the given binary exists, and if its invalid application exits. 
* Fixes instantiation of fetcher in each `(gg *goget) Version()` call.

Closes #372.

### DIFF
--- a/pkg/download/goget/goget.go
+++ b/pkg/download/goget/goget.go
@@ -22,15 +22,23 @@ import (
 // New returns a download protocol by using
 // go get. You must have a modules supported
 // go binary for this to work.
-func New() download.Protocol {
-	return &goget{
-		goBinPath: env.GoBinPath(),
-		fs:        afero.NewOsFs(),
+func New() (download.Protocol, error) {
+	goBin := env.GoBinPath()
+	fs := afero.NewOsFs()
+	mf, err := module.NewGoGetFetcher(goBin, fs)
+	if err != nil {
+		return nil, fmt.Errorf("go get fetcher can't be initialised with gobinpath: %s", goBin)
 	}
+	return &goget{
+		goBinPath: goBin,
+		fetcher:   mf,
+		fs:        fs,
+	}, nil
 }
 
 type goget struct {
 	goBinPath string
+	fetcher   module.Fetcher
 	fs        afero.Fs
 }
 
@@ -139,11 +147,7 @@ func (gg *goget) Zip(ctx context.Context, mod, ver string) (io.ReadCloser, error
 
 func (gg *goget) Version(ctx context.Context, mod, ver string) (*storage.Version, error) {
 	const op errors.Op = "goget.Version"
-	fetcher, err := module.NewGoGetFetcher(gg.goBinPath, gg.fs)
-	if err != nil {
-		return nil, errors.E(op, err)
-	}
-	ref, err := fetcher.Fetch(mod, ver)
+	ref, err := gg.fetcher.Fetch(mod, ver)
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
--- a/pkg/download/goget/goget.go
+++ b/pkg/download/goget/goget.go
@@ -25,19 +25,22 @@ import (
 func New() (download.Protocol, error) {
 	const op errors.Op = "goget.New"
 	goBin := env.GoBinPath()
-	mf, err := module.NewGoGetFetcher(goBin)
+	fs := afero.NewOsFs()
+	mf, err := module.NewGoGetFetcher(goBin, fs)
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
 	return &goget{
 		goBinPath: goBin,
 		fetcher:   mf,
+		fs:        fs,
 	}, nil
 }
 
 type goget struct {
 	goBinPath string
 	fetcher   module.Fetcher
+	fs        afero.Fs
 }
 
 func (gg *goget) List(ctx context.Context, mod string) ([]string, error) {
@@ -88,13 +91,12 @@ func (gg *goget) Latest(ctx context.Context, mod string) (*storage.RevInfo, erro
 }
 
 func (gg *goget) list(op errors.Op, mod string) (*listResp, error) {
-	fs := afero.NewOsFs()
-	hackyPath, err := afero.TempDir(fs, "", "hackymod")
+	hackyPath, err := afero.TempDir(gg.fs, "", "hackymod")
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
-	defer fs.RemoveAll(hackyPath)
-	err = module.Dummy(fs, hackyPath)
+	defer gg.fs.RemoveAll(hackyPath)
+	err = module.Dummy(gg.fs, hackyPath)
 	cmd := exec.Command(
 		gg.goBinPath,
 		"list", "-m", "-versions", "-json",
--- a/pkg/download/goget/goget.go
+++ b/pkg/download/goget/goget.go
@@ -23,11 +23,12 @@ import (
 // go get. You must have a modules supported
 // go binary for this to work.
 func New() (download.Protocol, error) {
+	const op errors.Op = "goget.New"
 	goBin := env.GoBinPath()
 	fs := afero.NewOsFs()
 	mf, err := module.NewGoGetFetcher(goBin, fs)
 	if err != nil {
-		return nil, fmt.Errorf("go get fetcher can't be initialised with gobinpath: %s", goBin)
+		return nil, errors.E(op, err)
 	}
 	return &goget{
 		goBinPath: goBin,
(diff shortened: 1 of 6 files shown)


pkg/download/goget/goget.go
```diff
@@ -22,15 +22,23 @@ import (
...
-func New() download.Protocol {
-	return &goget{
-		goBinPath: env.GoBinPath(),
-		fs:        afero.NewOsFs(),
+func New() (download.Protocol, error) {
+	goBin := env.GoBinPath()
+	fs := afero.NewOsFs()
+	mf, err := module.NewGoGetFetcher(goBin, fs)
+	if err != nil {
+		return nil, fmt.Errorf("go get fetcher can't be initialised with gobinpath: %s", goBin)
```
michalpristas: swallowed root error

pkg/module/go_get_fetcher.go
```diff
@@ -19,11 +19,16 @@ type goGetFetcher struct {
 }
 
 // NewGoGetFetcher creates fetcher which uses go get tool to fetch modules
-func NewGoGetFetcher(goBinaryName string, fs afero.Fs) Fetcher {
+func NewGoGetFetcher(goBinaryName string, fs afero.Fs) (Fetcher, error) {
+	cmd := exec.Command("which", goBinaryName)
```
michalpristas: same thing with PATH

pkg/download/goget/goget.go
devdinu: since its custom built error, `Invalid go binary: exit status` thought this error would be more apt. will add the cause to this error.

michalpristas: yes i understand.
but newgoget can return another error in a future so it would be good to know root cause of this.

```go
fmt.Errorf("go get fetcher can't be initialised with gobinpath '%s': %v", goBin, err)
```

pkg/module/go_get_fetcher.go
devdinu: Earlier i had the `stat` implementation, the problem was, it would require the path of `vgo`, and we can't give simply `vgo`, even though its in PATH but `which vgo` will succeed. 

The tradeoff here is, If `gobin` is not in path, we've to give the absolute path.

marwan-at-work: @devdinu  @michalpristas why not just do this: 
```golang
func NewGoGetFetcher(goBinaryName string, fs afero.Fs) (Fetcher, error) {
  err := exec.Command(goBinaryName).Run()
  if _, ok := err.(*exec.ExitError); !ok {
    return nil, err
  }
  ... // go binary exists
}

Comment (marwan-at-work) [reply pkg/module/go_get_fetcher.go]: and side note, we should add an errors.Op and wrap it :)

Comment (michalpristas) [reply pkg/module/go_get_fetcher.go]: ok i understood that this PR is about early failure considering that goBinaryName is correct either `vgo` or absolute path and verifying that it is existent. 
maybe we can do something like this
- which vgo - should succ for `vgo` fail for vgo not within PATH
- on ^failure: stat goBinPath, if exists, ok, if not err
?

Comment (michalpristas) [reply pkg/module/go_get_fetcher.go]: @marwan-at-work looks good. it will fail even if vgo exists but does not have +x which would pass with stat and which

Comment (marwan-at-work) [reply pkg/module/go_get_fetcher.go]: @michalpristas that's a good thing right?

File: pkg/download/goget/goget.go [hunk 3]
```diff
@@ -22,15 +22,24 @@ import (
...
+func New() (download.Protocol, error) {
+	const op errors.Op = "goget.New"
+	goBin := env.GoBinPath()
+	fs := afero.NewOsFs()
+	mf, err := module.NewGoGetFetcher(goBin, fs)
+	if err != nil {
+		return nil, errors.E(op, fmt.Errorf("Protocol can't be initialised with gobinpath: %s, error: %v", goBin, err))
 	}
+	return &goget{
+		goBinPath: goBin,
```
marwan-at-work: you can remove goBinPath and fs from this struct now.

pkg/module/go_get_fetcher.go
```diff
@@ -19,11 +19,15 @@ type goGetFetcher struct {
 }
 
 // NewGoGetFetcher creates fetcher which uses go get tool to fetch modules
-func NewGoGetFetcher(goBinaryName string, fs afero.Fs) Fetcher {
+func NewGoGetFetcher(goBinaryName string, fs afero.Fs) (Fetcher, error) {
+	err := validGoBinary(goBinaryName)
```
marwan-at-work: `const op errors.Op = "module.NewGoGetFetcher"`

pkg/module/go_get_fetcher.go
```diff
@@ -19,11 +19,15 @@ type goGetFetcher struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("Invalid go binary: %v", err)
```
marwan-at-work: `return nil, errors.E(op, err)`

pkg/module/go_get_fetcher.go
```diff
@@ -148,3 +152,12 @@ func getRepoDirName(repoURI, version string) string {
 func getPackagePath(gopath, module string) string {
 	return filepath.Join(gopath, "src", "mod", "cache", "download", module, "@v")
 }
+
+func validGoBinary(name string) error {
+	err := exec.Command(name).Run()
```
marwan-at-work: `const op errors.Op = "module.validGoBinary"`

pkg/module/go_get_fetcher.go
```diff
@@ -20,8 +20,7 @@ type goGetFetcher struct {
 
 // NewGoGetFetcher creates fetcher which uses go get tool to fetch modules
 func NewGoGetFetcher(goBinaryName string, fs afero.Fs) (Fetcher, error) {
-	cmd := exec.Command("which", goBinaryName)
-	_, err := cmd.CombinedOutput()
+	err := validGoBinary(goBinaryName)
```
michalpristas: just a small thing

```
if err := validGoBinary(goBinaryName); err != nil {
		return nil, fmt.Errorf("Invalid go binary: %v", err)			return nil, fmt.Errorf("Invalid go binary: %v", err)
	}		}
```

pkg/module/go_get_fetcher.go
```diff
@@ -148,3 +152,12 @@ func getRepoDirName(repoURI, version string) string {
 func getPackagePath(gopath, module string) string {
 	return filepath.Join(gopath, "src", "mod", "cache", "download", module, "@v")
 }
+
+func validGoBinary(name string) error {
+	err := exec.Command(name).Run()
+	_, ok := err.(*exec.ExitError)
+	if err != nil && !ok {
+		return err
```
marwan-at-work: `return errors.E(op, err)`

pkg/download/goget/goget_test.go
```diff
@@ -35,3 +43,19 @@ func TestList(t *testing.T) {
 		})
 	}
 }
+
+func TestInvalidGoBinError(t *testing.T) {
```
marwan-at-work: we don't need this test, because we are unit testing `TestGoGetFetcherError` which takes care of this scenario.

devdinu: This test would ensure we're propagating the error from the call.

marwan-at-work: @devdinu it's not necessary to test that. If we're not propagating en error, that means we're not checking it. If we're not checking, that means we're doing something seriously wrong and it's ok to panic.

pkg/download/goget/goget.go
```diff
@@ -22,15 +22,24 @@ import (
...
+		fetcher:   mf,
+		fs:        fs,
+	}, nil
 }
 
 type goget struct {
 	goBinPath string
+	fetcher   module.Fetcher
```
marwan-at-work: can you remove `goBinPath` and `fs` from the struct definition? They were being used to construct a module.Fetcher but now that it's being constructed just once, we can remove them.

pkg/module/go_get_fetcher_test.go
```diff
@@ -46,7 +56,7 @@ func ExampleFetch() {
 	repoURI := "github.com/arschles/assert"
 	version := "v1.0.0"
 	goBinaryName := env.GoBinPath()
-	fetcher := NewGoGetFetcher(goBinaryName, afero.NewOsFs())
+	fetcher, err := NewGoGetFetcher(goBinaryName, afero.NewOsFs())
```
marwan-at-work: make sure you handle the error on this line, or just use `_` to show that it's intentional we don't want to handle the error :)

pkg/download/goget/goget.go
marwan-at-work: This comment ^ is not true, turns out we use both things in goget.list() -- my b

PR #371, opened 2018-07-31
Add support for Go Modules

komuw:
# what
- I added support for Go modules

# why
- Go 1.11 is now available https://golang.org/dl/

# how
- I ran `go mod -init` which used the existing `Gopkg.lock` file to create a `go.mod` file
- go mod -init didn't however add `gobuffalo/plus` which I had to add manually via; `go get github.com/gobuffalo/plush@v3.7.7`

Fixes #462

PR #378, opened 2018-08-01
Remove dep files

komuw:
# what
- Removed dep files(`Gopkg.lock`, `Gopkg.toml`)
- Removed vendor/ dirctory

# why
- So as to use Go modules[1] for dependency management.

# notes:
-  This is dependent on:
1. Pull request: https://github.com/gomods/athens/pull/371
2. Pull request: https://github.com/gomods/athens/pull/563

# ref
1. https://tip.golang.org/cmd/go/#hdr-Modules__module_versions__and_more


komuw:
This should be blocked until Go version 1.11 comes out. 
I expect travis-ci to fail since it uses `dep` for package management. 
I have created a github issue[1] to update travis-ci to use Go version 1.11 and Go modules for dependency management.
I'll work on issue; https://github.com/gomods/athens/issues/379 

I'm working on the assumption that everyone is okay if we drop dep in favour of Go modules for dependency management once Go version 1.11 comes out

**ref:**
1. https://github.com/gomods/athens/issues/379

marwan-at-work:
I don't think we should remove the Vendor folder just yet, we can definitely remove Dep and switch to Go Modules. Keeping vendor gives us the option to use both at the same time (through `go build -mod=vendor` until $GOPATH/pkg proves stable. 

Plus, our builds will be much slower without vendor until we actually have a Proxy running for our CI

robjloranger:
Ref: https://github.com/gomods/athens/issues/518#issuecomment-415995566

robjloranger:
I think you need to remove the vendor directory again @komuw

komuw:
hi @robjloranger 
I had removed vendor, but re-added it (using `go mod vendor`) 
because of @marwan-at-work ' comment; https://github.com/gomods/athens/pull/378#issuecomment-415977922

what do you guys think?
personally, I would like to remove vendor.

robjloranger:
I think the build is failing because the vendor directory is present. At least that's what the build logs seem to be saying.

marwan-at-work:
Our go.mod file seems inconsistent from what Gopkg.lock requires 
For example: github.com/gobuffalo/tags is in v2 in Gopkg.lock while it's v1.9.6 in go.mod 
Furthermore, if you delete go.mod and rebuild it from Gopkg.lock you get even weirder results.

marwan-at-work:
I just pushed a pr to hopefully fix our go.mod file, if that's the issue then just hitting the "update branch" button after my pr is merged could solve stuff https://github.com/gomods/athens/pull/589

komuw:
@robjloranger 
I know why the build is failing;
1. for `GO111MODULE=auto` it is because I'm trying to run tests as:      
 `go test -mod=vendor -race -coverprofile cover.out -covermode atomic ./...`   
whereas `-mod=vendor` is only available in `GO111MODULE=on`

2. for `GO111MODULE=on` it is because i'm trying:
  `go run -mod=vendor .`    
but run doesn't seem to support that flag.  I'll try and use:
  `export 'GOFLAGS=-mod=vendor' && go run .`   

@marwan-at-work 
`github.com/gobuffalo/tags` is an indirect dependency and not a direct one. 
I think we should only tell Go modules what exact versions of `direct` dependencies that we want to depend on, and Go modules will choose whichever versions of indirect depencies it thinks are okay according to MVS. 
I think, that if we dictate to Go modules the versions of indirect deps to use; we may end up messing MVS. 
But if it turns out that there are direct deps from Gopkg.lock that Go modules was unable to pick right versions for, then we should definitely fix those.

I just haven't had time to fix the issues above. I'll try and dedicate time to fixing them.

komuw:
@marwan-at-work @robjloranger 

I've fixed the outstanding issues. re-review

komuw:
@marwan-at-work 
It is not that `-mod=vendor` doesn't work; it is that we are running ci in a matrix:
https://github.com/gomods/athens/blob/ea1503f0718789714645b4c8405b3b12d9550e63/.travis.yml#L16

once with `GO111MODULE=on` and again with `GO111MODULE=auto` 
Since travisCI checks out code inside $GOPATH it means `GO111MODULE=auto`  is equivalent to `GO111MODULE=off`. 
So in effect we are running CI;
a) once with `GO111MODULE=on`
b) again with `GO111MODULE=off`

`-mod=vendor` works when `GO111MODULE=on`   
however when `GO111MODULE=off` then `-mod=vendor` does not work and this is by design. 

So if we wanted to use `-mod=vendor` in our travisCI we would have to special case our scripts. 
So for example the script; https://github.com/gomods/athens/blob/ea1503f0718789714645b4c8405b3b12d9550e63/scripts/test_unit.sh#L11 
would have to change to;
```
if [ GO111MODULE=on ]
then
    go test -mod=vendor -race -coverprofile cover.out -covermode atomic ./...
else
    go test -race -coverprofile cover.out -covermode atomic ./...
fi
```
I'm not really sure that we want to do those special casing all over the place.     

An alternative would be; we decide we will run our CI only in `GO111MODULE=on` mode. 
That way we can run all our go commands using `-mod=vendor` .


Personally; I'm of the opinion that we should just remove the whole `vendor/` directory and just get our dependencies from VCS or from the Athens proxy(once it is ready).
On my machine with bandwidth of 970Kbps(downlink) and the Go modules cache wiped out, the command;
```sh
rm -rf ~/go/pkg/mod/cache/ && \
/usr/bin/time -p make test
```
takes
```
real       313.56
user        26.97
sys         20.81
```
It's bound to be orders of magnitude faster on travisCI with their gigabit links.

In summary, I see a few alternatives:
1. we remove `vendor/` completely and get dependencies from vcs/athens
2. we keep `vendor/` and special case our ci scripts to use `-mod=vendor`  when `GO111MODULE=on` and when `GO111MODULE=off` they just download from vcs/athens
3. ??

I would vote for option 1.

komuw:
the script;
```bash
mkdir /tmp/hey && cd /tmp/hey

cat <<EOD > hey.go
package main
import (
  "fmt"
	"github.com/pkg/errors"
)
func main() {
	fmt.Println("\n", errors.New("hey hey."))
}
EOD

export GO111MODULE=on
go mod init hey
go mod vendor
go run -mod=vendor .
```
works.  
but the script;
```bash
mkdir /tmp/hey && cd /tmp/hey

cat <<EOD > hey.go
package main
import (
  "fmt"
	"github.com/pkg/errors"
)
func main() {
	fmt.Println("\n", errors.New("hey hey."))
}
EOD

export GO111MODULE=off
go mod init hey
go mod vendor
go run -mod=vendor .
```
does not.

marwan-at-work:
@komuw we should **absolutely** not download from VCS on every CI build. Just look at this issue I opened yesterday: https://github.com/gobuffalo/uuid/issues/4 (which you have resolved!) 
What makes this even less secure is that we actually build a container image on every build. So I am strongly against this. However, I'd love to hear other opinions 😄 (so ccing @michalpristas @robjloranger @arschles and anyone else who'd like to voice in) 

I think the easiest solution is that we should always have GO111MODULE=on. If we point to a Proxy (the microsoft one) that's a nicer solution of course, but Aaron said it was gonna go away at some point so it's not a stable decision to make.

marwan-at-work:
As a reference: https://youtu.be/sln-gJaURzk?t=38m37s

komuw:
> I think the easiest solution is that we should always have GO111MODULE=on

I'm agreeable to that;
```bash
GO111MODULE=on && \
go test -mod=vendor blah blah
```

komuw:
the downside is, everytime someone adds a new dependency to `go.mod` they have to remember to run 
`go mod vendor` and then git commit the changes in vendor.   
with dep, there was nothing to remember since `dep ensure` would also update your vendor directory.  
maybe we should add a check in ci that if `go.mod` has changed, there should be a corresponding change in `vendor` and if not fail ci.

marwan-at-work:
@komuw the build should fail if the user did not remember to run `go mod vendor`

komuw:
@marwan-at-work 

> the build should fail if the user did not remember to run go mod vendor
:+1: good point, the build will fail automatically; we do not need to add a ci check for that.

I'll implement this tomorrow pending input from other people.

komuw:
@marwan-at-work hi. 

I've effected the required changes.

.drone.yml
```diff
@@ -10,7 +10,7 @@ pipeline:
   build:
     image: golang:1.10
```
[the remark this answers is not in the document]
robjloranger: :+1:

robjloranger:
Just the change Piotr mentioned then this looks good.

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/beorn7/perks v0.0.0-20180321164747-3a771d992973 // indirect
 	github.com/bketelsen/buffet v0.1.5
 	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
+	github.com/codegangsta/negroni v0.3.0 // indirect
```
marwan-at-work: I did a global search on this import path (and others) and they seem to be not imported at all by any of the vendor directory. You probably ran `go mod tidy` and ended up with something like this issue: https://github.com/golang/go/issues/27428

komuw: I do not think it was added incorrectly, it is indeed an indirect dep. 

`go mod why -m github.com/codegangsta/negroni`
```bash
# github.com/codegangsta/negroni
github.com/gomods/athens/cmd/olympus/actions
github.com/unrolled/secure
github.com/unrolled/secure.test
github.com/codegangsta/negroni
```

marwan-at-work: @komuw tests of dependencies are not included in vendor, nor are they tested when you say `go test ./...` -- but in any case. This is not biggie, just something to be aware of :) thank you for fixing the PR!

PR #380, opened 2018-08-01
GCP: get credentials implicitly

marwan-at-work:
Update constructor and its doc to reflect that Cloud Storage credentials can be obtained automatically by either being logged in to gcloud, being inside GCP, or having the GOOGLE_APPLICATION_CREDENTIALS set.

### DIFF
--- a/pkg/config/env/gcp.go
+++ b/pkg/config/env/gcp.go
@@ -6,12 +6,18 @@ import (
 	"github.com/gobuffalo/envy"
 )
 
-// GcpBucketName returns Google Cloud Storage bucket name defined by ATHENS_STORAGE_GCP_BUCKET
-func GcpBucketName() (string, error) {
+// GCPBucketName returns Google Cloud Storage bucket name defined by ATHENS_STORAGE_GCP_BUCKET
+func GCPBucketName() (string, error) {
 	env, err := envy.MustGet("ATHENS_STORAGE_GCP_BUCKET")
 	if err != nil {
 		return "", fmt.Errorf("missing Google Cloud storage bucket name: %s", err)
 	}
 
 	return env, nil
+}
+
+// GCPPRojectID returns the project id on which the project
+// is running or the cloud storage is using.
+func GCPPRojectID() string {
+	return envy.Get("GOOGLE_CLOUD_PROJECT", "")
 }
--- a/pkg/config/env/gcp.go
+++ b/pkg/config/env/gcp.go
@@ -16,8 +16,8 @@ func GCPBucketName() (string, error) {
 	return env, nil
 }
 
-// GCPPRojectID returns the project id on which the project
+// GCPProjectID returns the project id on which the project
 // is running or the cloud storage is using.
-func GCPPRojectID() string {
+func GCPProjectID() string {
 	return envy.Get("GOOGLE_CLOUD_PROJECT", "")
 }
--- a/pkg/storage/gcp/gcp.go
+++ b/pkg/storage/gcp/gcp.go
@@ -3,11 +3,13 @@ package gcp
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"net/url"
 
 	"cloud.google.com/go/storage"
 	"github.com/gomods/athens/pkg/config/env"
-	"google.golang.org/api/option"
+	"github.com/gomods/athens/pkg/errors"
+	"google.golang.org/api/googleapi"
 )
 
 // Storage implements the (./pkg/storage).Backend interface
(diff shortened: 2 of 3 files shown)


pkg/storage/gcp/gcp.go
```diff
@@ -3,40 +3,48 @@ package gcp
...
-		return nil, err
+		return nil, errors.E(op, err)
 	}
 	u, err := url.Parse(fmt.Sprintf("https://storage.googleapis.com/%s", bucketname))
 	if err != nil {
-		return nil, err
+		return nil, errors.E(op, err)
 	}
 	bkt := gcpBucket{storage.Bucket(bucketname)}
+	err = bkt.Create(ctx, env.GCPPRojectID(), nil)
```
robjloranger: What happens if the project ID is not set?

marwan-at-work: @robjloranger it returns `googleapi: Error 400: Unknown project id: , invalid` which is clear enough for us to fatal out of the boot up with that message.

pkg/config/env/gcp.go
```diff
@@ -6,12 +6,18 @@ import (
...
 		return "", fmt.Errorf("missing Google Cloud storage bucket name: %s", err)
 	}
 
 	return env, nil
 }
+
+// GCPPRojectID returns the project id on which the project
+// is running or the cloud storage is using.
+func GCPPRojectID() string {
+	return envy.Get("GCP_PROJECT_ID", "")
```
tbpg: This variable is usually named `GOOGLE_CLOUD_PROJECT`. Some libraries will look for that variable automatically.

pkg/storage/gcp/gcp.go
```diff
@@ -3,40 +3,48 @@ package gcp
...
+// New returns a new Storage instance authenticated using the provided
 // ClientOptions. The bucket name to be used will be loaded from the
 // environment variable ATHENS_STORAGE_GCP_BUCKET.
 //
-// The ClientOptions should provide permissions sufficient to read, write and
-// delete objects in google cloud storage for your project.
-func NewWithCredentials(ctx context.Context, cred option.ClientOption) (*Storage, error) {
-	storage, err := storage.NewClient(ctx, cred)
+// If you're on your dev machine, make sure you're logged in to gcloud.
+// If you're outside GCP, make sure you have GOOGLE_APPLICATION_CREDENTIALS set.
```
tbpg: On a dev machine, setting `GOOGLE_APPLICATION_CREDENTIALS` should be enough. Can link to https://cloud.google.com/docs/authentication/getting-started for more detail.

Suggested:
```
// If you're not running on GCP, set the GOOGLE_APPLICATION_CREDENTIALS environment variable
// to the path of your service account file. If you're running on GCP (e.g. AppEngine),
// credentials will be automatically provided.
// See https://cloud.google.com/docs/authentication/getting-started.
```

pkg/storage/gcp/gcp.go
```diff
@@ -3,40 +3,48 @@ package gcp
...
 type Storage struct {
 	bucket       Bucket
 	baseURI      *url.URL
 	closeStorage func() error
+	projectID    string
 }
 
-// NewWithCredentials returns a new Storage instance authenticated using the provided
+// New returns a new Storage instance authenticated using the provided
 // ClientOptions. The bucket name to be used will be loaded from the
```
tbpg:
```
// New returns a new Storage instance backed by a Google Cloud Storage bucket.
```

(No more ClientOptions argument.)

pkg/config/env/gcp.go
tbpg: Should this be documented anywhere (OK if the answer is yes, but we don't know where, yet.)

robjloranger: It should, Maybe in the above with https://github.com/gomods/athens/pull/380#discussion_r207253615

marwan-at-work: you mean the GCP set up in general or this environment variable? If the first then, yes I was thinking of having a README in every storage implementation that shows you its set up and why we have that implementation in the first place.

tbpg: A bit of both. Sounds good to me.

robjloranger: :+1:

cmd/proxy/.env
```diff
@@ -1 +1,4 @@
 ATHENS_MONGO_STORAGE_URL=mongodb://127.0.0.1:27017
+ATHENS_STORAGE_GCP_BUCKET=marwan-athens-test
```
tbpg: Revert?

pkg/config/env/go.go
```diff
@@ -13,5 +13,5 @@ func GoEnvironmentWithDefault(value string) string {
 // GoBinPath returns the path to the go binary to use, defined by
 // GO_BINARY_PATH. This value can be a name on the PATH, or the full path
 func GoBinPath() string {
-	return envy.Get("GO_BINARY_PATH", "go")
+	return envy.Get("GO_BINARY_PATH", "/Users/208581/Desktop/gosource/go/bin/go")
```
tbpg: Revert?

pkg/config/env/gcp.go
michalpristas: m: capital R

PR #399, opened 2018-08-03
Download: test against controlled repositories

### LINKED ISSUE #340: Create Test Repos in gomods org
We should create test repos with certain tags so that we can test against them in various scenarios such as download protocol, git fetching etc.

TODO: Adjust goget_test.go

marwan-at-work:
This PR introduces GitHub integration tests to our pkg/download/goget Protocol. This tests that goget is getting the right version data from live repos especially /list and /latest.

I've created a new organization: github.com/athens-artifacts to store dummy repositories. Please ping me if you'd like write access to it.

Fixes #340

### DIFF
--- a/pkg/download/goget/goget_test.go
+++ b/pkg/download/goget/goget_test.go
@@ -1,27 +1,35 @@
 package goget
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"io/ioutil"
+	"path/filepath"
+	"strings"
 	"testing"
+	"time"
 
+	"github.com/gomods/athens/pkg/storage"
 	"github.com/stretchr/testify/require"
 )
 
-type testCase struct {
-	name    string
-	mod     string
-	version string
+type listTest struct {
+	name string
+	path string
+	tags []string
 }
 
-// TODO(marwan): we should create Test Repos under github.com/gomods
-// so we can get reproducible results from live VCS repos.
-// For now, I cannot test that github.com/pkg/errors returns v0.8.0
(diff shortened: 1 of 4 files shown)


pkg/download/goget/goget_test.go
```diff
@@ -1,40 +1,200 @@
...
 	dp, err := New()
 	require.NoError(t, err, "failed to create protocol")
 	ctx := context.Background()
 
-	for _, tc := range tt {
+	for _, tc := range listTests {
+		t.Run(tc.name, func(t *testing.T) {
+			versions, err := dp.List(ctx, tc.path)
+			require.NoError(t, err)
+			require.EqualValues(t, tc.tags, versions)
```
michalpristas: are we going to mimick this?

https://github.com/golang/go/blob/b7d3f4c0b2f421c6b7bf6ce3533f1b34b470b39d/src/cmd/go/internal/modload/help.go#L238-L239

michalpristas: ^ this is obviously for @latest endpoint

marwan-at-work: @michalpristas yeah I actually got confused about that. From what I can tell, `@latest` is never called for an actual tag. cmd/go always calls `/@v/list` first then if the list is empty it will call `@latest`. But maybe we should just follow what the spec says? I can do that in a separate issue/PR

michalpristas: yeah i was thinking about exactly the same thing. not very clear to me. 
but if we take specs as a contract we probably should stick to it.

PR #400, opened 2018-08-04
In memory workers config

### LINKED ISSUE #285: enable in-memory workers in the proxy

kdvy:
Corrected typo where OLYMPUS was referenced as ATHENS in comment.

Added environment variable config for in-memory mock of redis for both Athens and Olympus. (Separate commits just in case). 

Default behaviour is to use redis workers (for backwards compatibility), with the in-memory queue used if *__REDIS_MOCK_IN_MEM is set to true.

Fixed #285


cmd/proxy/actions/app.go
```diff
@@ -130,6 +130,24 @@ func App() (*buffalo.App, error) {
...
+	}
+	return registerRedis(ctx, s, mf)
+}
+
+func registerInMem(ctx context.Context, s storage.Backend, mf *module.Filter) (worker.Worker, error) {
+	w := worker.NewSimple()
+	if err := w.Register(FetcherWorkerName, GetProcessCacheMissJob(ctx, s, w, mf)); err != nil {
+		return nil, err
+	}
+	if err := w.Register(ReporterWorkerName, GetCacheMissReporterJob(w, mf)); err != nil {
```
michalpristas: no need for this one

michalpristas: basically proxy is sync so it does not need workers now.
please keep one in Olympus
@marwan-at-work correct me if i'm wrong

marpio: I think we can still make use of the worker in proxy. If the proxy doesn't have the requested module we could redirect go to Olympus and at the same time queue a job to fill the cache. Otherwise go would have to wait for Olympus to download the module, store it in cache, then for proxy to download it, store it in it's own cache and then send to go.

arschles: @kdvy @michalpristas @marpio @marwan-at-work are you all ok with removing this for now? It's essentially dead code until we implement the redirect logic that @marpio mentioned in ☝️

pkg/config/env/redis.go
```diff
@@ -1,15 +1,39 @@
...
 // RedisQueuePortWithDefault returns Redis queue port used by workers defined by ATHENS_REDIS_QUEUE_PORT.
 // Standard port is 6379
 func RedisQueuePortWithDefault(value string) string {
 	return envy.Get("ATHENS_REDIS_QUEUE_PORT", value)
 }
 
-// OlympusRedisQueuePortWithDefault returns Redis queue port used by workers defined by ATHENS_REDIS_QUEUE_PORT.
+// RedisMockInMem determines whether an in-memory worker is used to mock Redis for Proxy (Athens)
+func RedisMockInMem() bool {
+	boolStr := envy.Get("ATHENS_REDIS_MOCK_IN_MEM", "false")
```
arschles: @kdvy can you set up an environment variable that lets the user specify what kind of background workers to use? For example, what do you think of `ATHENS_BACKGROUND_WORKER_TYPE`, which could be `memory` or `redis`?

Same with the below one for olympus

arschles:
Closes #286

cmd/proxy/actions/app.go
michalpristas: i'm ok with removing, more dead code == less readable code

cmd/olympus/actions/app.go
```diff
@@ -127,6 +129,26 @@ func App(config *AppConfig) (*buffalo.App, error) {
...
 
 func getWorker(store storage.Backend, eLog eventlog.Eventlog) (worker.Worker, error) {
+	workerType := env.OlympusBackgroundWorkerType()
+	switch workerType {
+	case "redis":
+		return registerRedis(store, eLog)
+	case "memory":
+		return registerInMem(store, eLog)
+	default:
+		return nil, fmt.Errorf("Provided worker type %s not one of redis|memory", workerType)
```
michalpristas: let's put some sane default here, no need to be this restricted. Log warning and default to inmem as it does not require deps to be setup.

kdvy: Made the change to in-memory worker for default, now prints a message using the standard log package. Thanks for the review.

PR #403, opened 2018-08-04
test: try using beta3 from travis, no build

robjloranger:


arschles:
@robjloranger does this overlap with #404 at all?

robjloranger:
Found solution

Issue #396, opened 2018-08-03
Split backend's storage.Getter interface


marwan-at-work:
I'd like to get up/down votes to get a sense of what's preferred: 

Should we split the backend's storage.Getter interface into three independent methods? 

Currently this is the interface: 

```golang
type Getter interface {
	Get(ctx context.Context, module, vsn string) (*Version, error)
}
```

Should we make it like this: 

```golang
type Getter interface {
	Info(ctx context.Context, module, vsn string) ([]byte, error)
	GoMod(ctx context.Context, module, vsn string) ([]byte, error)
	Zip(ctx context.Context, module, vsn string) (io.ReadCloser, error)
}
```

### Motivation: 
1. `cmd/go` never asks for all three files at the same time, so why load them up? 
2. We have to be very careful when to close and when *not* to close a zip reader. This can easily introduce memory/disk leaks.

arschles:
Big ups from me for this. Your point about having to be careful to close the zip really resonates with me.

PR #406, opened 2018-08-04
Storage: split storage.Getter interface

marwan-at-work:
fixes https://github.com/gomods/athens/issues/396

This is my first time going around the storage packages and I started to see some room for improvement and clean up. Like getting rid of the Connector (we have an issue for that I believe) and not exposing implementations for no reason (*Storage) vs (*storage). As well as removing mocks that will never be used outside of testing (GCP). 

However, this pr only takes care of issue 396. I'll leave cleaning up for other issues and prs but I'd love to start doing it :)


cmd/proxy/actions/cache_miss_fetcher.go
```diff
@@ -61,7 +61,21 @@ func parseArgs(args worker.Args) (string, string, error) {
 
 func getModuleInfo(ctx context.Context, module, version string) (*storage.Version, error) {
 	os := olympusStore.NewStorage(GetOlympusEndpoint())
-	return os.Get(ctx, module, version)
+	var v storage.Version
+	var err error
+	v.Info, err = os.Info(ctx, module, version)
+	if err != nil {
```
michalpristas: can we do this async, and with multierror?

michalpristas: also we need to start thinking about concurrent cache fills,

pkg/storage/olympus/getter.go
```diff
@@ -1,80 +1,82 @@
...
+		msg := fmt.Errorf("GET %v returned unexpected status: %v", zipURI, resp.StatusCode)
+		return nil, errors.E(op, msg)
+	}
 
-	return &storage.Version{
-		Info: info,
-		Mod:  mod,
-		Zip:  ioutil.NopCloser(bytes.NewReader(zip)),
-	}, nil
+	return resp.Body, nil
```
michalpristas: i'm on the fence here. while this is readcloser and it is correct to return it, it will keep socket opened until zip is processed which might easily lead to `out of sockets`

marwan-at-work: @michalpristas part of my clean up is that I'd like to suggest removing this whole package. issues 349 and 309 will make it so that we always redirect to Olympus so we never have to "stream from it". And even if we do, it will be in the shape of a download.Protocol interface, which goes to our Proxy storage first then back to the user. Let me know if that does or doesn't make sense :)

michalpristas: Yes it does @marwan-at-work Thanks.

cmd/proxy/actions/cache_miss_fetcher.go
marwan-at-work: @michalpristas another clean up question: do we still want the cache_miss_fetcher? or should it be deleted?

michalpristas: i think we will need it in a modified form.
From the flow described in #346 we talked that proxy on a miss redirects to Olympus and schedules a job to populate itself from O after some (configured) time. 
This might be a place where cache_miss_fetcher might be of use

pkg/download/download.go
```diff
@@ -86,39 +90,55 @@ func (p *protocol) Latest(ctx context.Context, mod string) (*storage.RevInfo, er
 
 func (p *protocol) GoMod(ctx context.Context, mod, ver string) ([]byte, error) {
```
robjloranger: I'd like to recommend a different name, but can't think of a better one. At first I thought `Mod` but that may be confused for the _entire_ module. Then maybe `ModInfo` but that would seem to represent the module's `RevInfo`. So good as is I guess :grin:

pkg/storage/gcp/bucket_cloud.go
```diff
@@ -58,5 +58,5 @@ func (b *gcpBucket) List(ctx context.Context, prefix string) ([]string, error) {
 
 func (b *gcpBucket) Exists(ctx context.Context, path string) bool {
 	_, err := b.Object(path).Attrs(ctx)
-	return err == nil
+	return err == storage.ErrObjectNotExist
```
robjloranger: Should this be `!=`?

If the error was nil, the result will be false. This will return true only when the object does not exist.

pkg/storage/minio/getter.go
```diff
@@ -3,53 +3,62 @@ package minio
...
 	"github.com/gomods/athens/pkg/errors"
-	"github.com/gomods/athens/pkg/storage"
 	minio "github.com/minio/minio-go"
 	opentracing "github.com/opentracing/opentracing-go"
 )
 
-func (v *storageImpl) Get(ctx context.Context, module, version string) (*storage.Version, error) {
+func (v *storageImpl) Info(ctx context.Context, module, vsn string) ([]byte, error) {
+	const op errors.Op = "minio.Info"
 	sp, ctx := opentracing.StartSpanFromContext(ctx, "storage.minio.Get")
```
robjloranger: `"storage.minio.Info"`

pkg/storage/minio/getter.go
```diff
@@ -3,53 +3,62 @@ package minio
...
 	}
-	zipPath := fmt.Sprintf("%s/source.zip", versionedPath)
-	zipReader, err := v.minioClient.GetObject(v.bucketName, zipPath, minio.GetObjectOptions{})
+
+	return info, nil
+}
+
+func (v *storageImpl) GoMod(ctx context.Context, module, vsn string) ([]byte, error) {
+	const op errors.Op = "minio.GetMod"
+	sp, ctx := opentracing.StartSpanFromContext(ctx, "storage.minio.Get")
```
robjloranger: `"storage.minio.GoMod"`

pkg/storage/minio/getter.go
```diff
@@ -3,53 +3,62 @@ package minio
...
-		return nil, err
+		return nil, transformNotFoundErr(op, module, vsn, err)
 	}
-	info, err := ioutil.ReadAll(infoReader)
+
+	return mod, nil
+}
+func (v *storageImpl) Zip(ctx context.Context, module, vsn string) (io.ReadCloser, error) {
+	const op errors.Op = "minio.Zip"
+	sp, ctx := opentracing.StartSpanFromContext(ctx, "storage.minio.Get")
```
robjloranger: `"storage.minio.Zip"`

pkg/download/get_module_params.go
```diff
@@ -0,0 +1,16 @@
+package download
+
+import (
+	"github.com/gobuffalo/buffalo"
+	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/paths"
+)
+
+func getModuleParams(op errors.Op, c buffalo.Context) (string, string, error) {
```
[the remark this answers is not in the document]
michalpristas: named returns but only with a combination with non-blank returns. it lowers readability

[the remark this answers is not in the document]
marwan-at-work: 💯

cmd/proxy/actions/cache_miss_fetcher.go
marwan-at-work: @michalpristas is it ok to do that in a separate PR? This pr is already big and bulky. I'll push one right after :)

pkg/storage/gcp/bucket_cloud.go
marwan-at-work: @robjloranger if err nil, the result *should* be false. And if err is netiher nil nor notExist, then we'll leave it up to the getter to find out what the error is when they try to get that object. Is that ok? Otherwise, we have to make the `Exists` in `Checker` interface return an `error`. What do you think?

cmd/proxy/actions/cache_miss_fetcher.go
michalpristas: yes, i will always 👍 for smaller, more readable PRs

pkg/download/download.go
```diff
@@ -86,39 +90,55 @@ func (p *protocol) Latest(ctx context.Context, mod string) (*storage.RevInfo, er
...
+	return goMod, nil
 }
 
 func (p *protocol) Zip(ctx context.Context, mod, ver string) (io.ReadCloser, error) {
 	const op errors.Op = "protocol.Zip"
-	v, err := p.s.Get(ctx, mod, ver)
+	zip, err := p.s.Zip(ctx, mod, ver)
 	if errors.IsNotFoundErr(err) {
-		v, err = p.fillCache(ctx, mod, ver)
+		err = p.fillCache(ctx, mod, ver)
```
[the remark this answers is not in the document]
marwan-at-work: done

pkg/storage/gcp/bucket_cloud.go
marwan-at-work: Sounds good. @robjloranger let me know if this pr looks ok then will merge.

marwan-at-work: @robjloranger you're right i should be doing a more thorough check: 
```golang
	_, err := b.Object(path).Attrs(ctx)
	if err == nil {
          return true
        } else if err == storage.ErrObjectNotExist {
           return false
        } 
        
        TODO: this means that something errored out and we should return false, err
        return false
```

marwan-at-work: what do you think?

robjloranger: Awesome 😀

marwan-at-work: @robjloranger ok i've updated the PR :) let me know if it's good to merge! 
PS. we should start having a GCP set up in our CI, what do you think ?

robjloranger:
Thanks @marwan-at-work

Issue #397, opened 2018-08-03
some scripts are not portable


fhs:
Few issues with `get_buffalo.sh`:
* It uses `$TMPDIR`, which is not always set in Linux
* It downloads Linux binary only
* It doesn't clean up temporary directory after it's done
* Maybe it should be updated to use 0.12.4 instead of 0.12.3? It seem like extra maintenance keeping the version up to date when one can just `get get` the latest.

arschles:
@fhs would you be up for cleaning the script up?

fhs:
> @fhs would you be up for cleaning the script up?

Looks like robjloranger has started the work (PR #410).

It seems that these scripts were expected to be run in travis. The main issue is that `CONTRIBUTING.md` and `DEVELOPMENT.md` asks the reader to run `make setup-dev-env`. So, either the docs need to be updated or the scripts need to become portable.

robjloranger:
I'm pretty sure this is fixed, between the buffalo script changes, temp directory fix and power shell scripts.

PR #410, opened 2018-08-04
Fix up e2e script

### LINKED ISSUE #405: Travis: use gimme instead of building go
Currently the build scripts clone and build go 1.11beta3 from source. This is available in Travis so we should use that instead of downloading and compiling during a build.

I'll take this one.

robjloranger:
+ use correct variable name for module support, `GO111MODULE`
+ don't set `GO111MODULE` until after running `buffalo dev`, see #412
+ fixed old path for test repo in e2e, now `happy-path`
+ set go version in travis
+ remove building from source and artifacts

Bonus: builds should now average ~2min faster

Closes #405


scripts/get_dev_tools.sh
```diff
@@ -7,15 +7,4 @@ set -xeuo pipefail
 go get github.com/golang/lint/golint
 go get github.com/golang/dep/cmd/dep
 
-GO_VERSION="go1.11beta3"
```
michalpristas: what i liked on this is that this allowed me to use the stable version of Go on my dev env. while new way requires me to stay up to date with latest release

marwan-at-work: @michalpristas you can still have stable/unstable versions in your dev env. This is just a set up script, so I'm not sure how you're using it. In any case, we can potentially separate the CI set up script from the dev set up script if it makes sense

robjloranger:
I should actually also include an update to the docs about installing go for development. Or #418 would do this for me.

robjloranger:
Agreed, #415 can make use of that after this is in

scripts/get_dev_tools.sh
robjloranger: I think splitting up the scripts for dev vs CI makes sense. Maybe during #397

arschles: Ref this thread: https://github.com/gomods/athens/pull/418#discussion_r208064390

PR #426, opened 2018-08-08
Benchmark storages

### LINKED ISSUE #382: Benchmark Storage Performance
Let's get an idea of which of the backend implementations work best for our use cases (scalability, streaming, etc.) 

Simple benchmark tests should be a good start. But in the future, we should have the ability to run benchmarks against staged deployments to get more accurate results.

devdinu:
- list, save, delete, exits operations with the storages are benchmarked. closes #382.
- changed storage/testsuite cleanup to return error
- changed `MongoConnectionTimeoutSecWithDefault` to return time.Duration instead of string

Cleanup/Post:
- could run benchmarks as part of ci
- create required storage without TestSuite,Model.


pkg/config/env/mongo.go
```diff
@@ -56,14 +58,20 @@ func MongoPassword() (string, error) {
...
+	if err != nil {
+		return time.Duration(defTimeout) * time.Second
+	}
+	return time.Duration(timeout) * time.Second
 }
 
 // MongoSSLWithDefault returns Athens Mongo Storage SSL flag defined by MONGO_SSL.
 // Defines whether or not SSL should be used.
 func MongoSSLWithDefault(value string) string {
+	//TODO: could return time.Duration
```
michalpristas: ?? this is a flag

pkg/storage/minio/test_suite.go
```diff
@@ -42,12 +42,16 @@ func (ts *TestSuite) StorageHumanReadableName() string {
...
 // Cleanup tears down test
-func (ts *TestSuite) Cleanup() {
+func (ts *TestSuite) Cleanup() error {
 	minioClient, _ := minio.New(ts.endpoint, ts.accessKeyID, ts.secretAccessKey, false)
 	doneCh := make(chan struct{})
 	defer close(doneCh)
 	objectCh := minioClient.ListObjectsV2(ts.bucketName, "", true, doneCh)
 	for object := range objectCh {
-		ts.Require().NoError(minioClient.RemoveObject(ts.bucketName, object.Key))
+		//TODO: could return multi error and clean other objects
```
michalpristas: could you include TODO in this PR?

pkg/storage/mongo/mongo.go
```diff
@@ -23,9 +25,10 @@ func NewStorage(url string) *ModuleStore {
 
 // Connect conntect the the newly created mongo backend.
 func (m *ModuleStore) Connect() error {
-	s, err := mgo.Dial(m.url)
+	timeout := env.MongoConnectionTimeoutSecWithDefault(1)
+	s, err := mgo.DialWithTimeout(m.url, timeout)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to connect with %s, error: %v", m.url, err)
```
michalpristas: we have an issue for more granular timeouts i think this is one of the cases when it can be used. 
Timeout for a request should not be the same as a timeout for dialing a connection.

pkg/storage/storage_tests/module_storage/storage_benchmark_test.go
```diff
@@ -0,0 +1,131 @@
...
+	for _, store := range getStores(b) {
+
+		backend := store.Storage()
+
+		for i := 0; i < b.N; i++ {
+			err := backend.Save(context.Background(), fmt.Sprintf("%s-%d", module, i), version, mod, bytes.NewReader(zip), info)
+			require.NoError(b, err, "Save for storage %s failed", backend)
+		}
+
+		b.Run(fmt.Sprintf("save module backend %s", store.StorageHumanReadableName()), func(b *testing.B) {
```
michalpristas: /s/save/delete

pkg/storage/storage_tests/module_storage/storage_benchmark_test.go
```diff
@@ -0,0 +1,131 @@
...
+			for i := 0; i < b.N; i++ {
+				err := backend.Save(context.Background(), fmt.Sprintf("%s-%d", module, i), version, mod, bytes.NewReader(zip), info)
```
michalpristas: again this should test delete

pkg/storage/storage_tests/module_storage/storage_benchmark_test.go
```diff
@@ -0,0 +1,131 @@
...
+	for _, store := range getStores(b) {
+
+		backend := store.Storage()
+
+		for i := 0; i < b.N; i++ {
+			err := backend.Save(context.Background(), fmt.Sprintf("%s-%d", module, i), version, mod, bytes.NewReader(zip), info)
+			require.NoError(b, err, "Save for storage %s failed", backend)
+		}
+
+		b.Run(fmt.Sprintf("save module backend %s", store.StorageHumanReadableName()), func(b *testing.B) {
```
michalpristas: /s/save module/exists module/

pkg/storage/storage_tests/module_storage/storage_benchmark_test.go
```diff
@@ -0,0 +1,131 @@
...
+			for i := 0; i < b.N; i++ {
```
michalpristas: can you test positive/negative exists? 
positive = will find something
negative = will look for non existent module

pkg/storage/mongo/mongo.go
devdinu: Do you want this to be solved in this PR, or we could raise a separate PR which handles the `storage/mongo.go` database name, this granular dial connection env vars, and other required vars.

michalpristas: nono, just pointing out so we don't forget

pkg/storage/storage_tests/module_storage/storage_benchmark_test.go
```diff
@@ -0,0 +1,136 @@
...
+		backend := store.Storage()
+
+		for i := 0; i < b.N; i++ {
+			err := backend.Save(context.Background(), fmt.Sprintf("%s-%d", module, i), version, mod, bytes.NewReader(zip), info)
+			require.NoError(b, err, "delete for storage %s failed", backend)
+		}
+
+		b.Run(fmt.Sprintf("delete module backend %s", store.StorageHumanReadableName()), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				err := backend.Save(context.Background(), fmt.Sprintf("%s-%d", module, i), version, mod, bytes.NewReader(zip), info)
```
michalpristas: `err := backend.Save`, this should be delete

devdinu: My bad, should never copy any text :P

pkg/storage/storage_tests/module_storage/storage_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+func BenchmarkStorageList(b *testing.B) {
+	module, version := "athens_module", "1.0.1"
+	zip, info, mod := []byte("zip_data"), []byte("info"), []byte("mod_info")
+
+	for _, store := range getStores(b) {
+
+		backend := store.Storage()
+
+		require.NoError(b, backend.Save(context.Background(), module, version, mod, bytes.NewReader(zip), info), "Save for storage %s failed", backend)
+
```
marwan-at-work: make sure you reset the timer here and in other benchmarks to clarify that this is an initial set up.

pkg/storage/storage_tests/module_storage/storage_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+func BenchmarkStorageExists(b *testing.B) {
+	module, version := "athens_module", "1.0.1"
+	zip, info, mod := []byte("zip_data"), []byte("info"), []byte("mod_info")
+
+	for _, store := range getStores(b) {
+
+		backend := store.Storage()
+
+		for i := 0; i < b.N; i++ {
+			err := backend.Save(context.Background(), fmt.Sprintf("%s-%d", module, i), version, mod, bytes.NewReader(zip), info)
```
marwan-at-work: this `N` can be different than the `N` inside the `Run` function, furthermore `N` actually changes and its not constant depending on how long the operations take over each iteration. For this test, you don't need to create multiple modules. Just create 1 module and test `backend.Exists` N times over that one module.

pkg/storage/storage_tests/module_storage/storage_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+			}
+		})
+		require.NoError(b, store.Cleanup())
+	}
+}
+
+func getStores(b *testing.B) []storage.TestSuite {
+	var stores []storage.TestSuite
+
+	//TODO: create the instance without model or TestSuite
```
marwan-at-work: would be much more readable. But also having the name attached to a storage is important when a test fails or when naming the benchmark. We should also have a flag or an indicator to say we only want to benchmark `mongo, minio, gcp` and so that we can easily compare 2 or 3 storages against each other without having to spin up ALL implementations.

pkg/storage/storage_tests/module_storage/storage_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+func BenchmarkStorageDelete(b *testing.B) {
+	module, version := "athens_module", "1.0.1"
+	zip, info, mod := []byte("zip_data"), []byte("info"), []byte("mod_info")
+
+	for _, store := range getStores(b) {
+
+		backend := store.Storage()
+
+		for i := 0; i < b.N; i++ {
+			err := backend.Save(context.Background(), fmt.Sprintf("del-%s-%d", module, i), version, mod, bytes.NewReader(zip), info)
```
marwan-at-work: the outer `N` can be different from the inner `N`, and therefore some of those `backend.Delete` functions are probably being terminated early on the `Exists` check giving inaccurate benchmark results. Your only option is to save and delete at the same time inside `b.Run` and that should be accurate enough to compare between storages.

pkg/storage/storage_tests/module_storage/storage_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+			require.NoError(b, err, "save for storage %s module: %s failed", backend, i)
+		}
+
+		b.Run(fmt.Sprintf("delete module backend %s", store.StorageHumanReadableName()), func(b *testing.B) {
```
marwan-at-work: just put the storage name because the benchmark will already attach this test's parent name as such: 
`BenchmarkStorageDelete/mongo` which is easier to read than `BenchmarkStorageDelete/delete_module_backend_mongo`  (same for all other tests)

pkg/storage/storage_tests/module_storage/storage_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+
+		b.Run(fmt.Sprintf("exists module backend %s", store.StorageHumanReadableName()), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				backend.Exists(context.Background(), fmt.Sprintf("%s-%d", module, i), version)
+			}
+		})
+
+		b.Run(fmt.Sprintf("non existent module backend %s", store.StorageHumanReadableName()), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
```
marwan-at-work: this can be a separate benchmark with no setting up or deleting anything. Would be more readable.

pkg/storage/storage_tests/module_storage/storage_benchmark_test.go
devdinu: created two `b.Run` one for existing and other non-existing modules, would it be apt to pull it out as a separate benchmark?

for existing tests, we gotta save the modules, so i could pull out a `x := b.N` constant for saving x modules, and while doing `exists` could do  `n %  x`. would that work?

marwan-at-work: @devdinu for existing to test you only need to save `1` module, not `N` module :) just check that this one module Exists N times.

devdinu: i was thinking for single module `same query` it would get `cached` and that wouldn't give us the right number.

marwan-at-work: @devdinu yes load testing is separate, though very important for scalability. I'll create an issue for it. As for `cached`, that's Okay. We don't care how the backend implementation does things. We just wanna compare them against each other. For example, if Mongo caches things but GCP does not, then we wanna see that Mongo is actually faster on average.

pkg/storage/minio/test_suite.go
michalpristas: will you do multierror in this PR?

michalpristas:
other than mutlierror part, which can be in a separate PR, it looks good

scripts/benchmark.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+go test -v -bench=. $(find . -iname '*storage*test.go' -not -path '/vendor/') -run=^$
```
marwan-at-work: why not just `go test ./... -bench=. -run=^$` -- vendor gets excluded automatically, and we test all benchmarks since this is a global Makefile command.

devdinu: That ran benchmark against all packages thought it didn't have any benchmark files, and this output was much better :)

pkg/storage/mongo/mongo.go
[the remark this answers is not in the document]
devdinu: Changed it. This still misses the `url`, could be helpful at times when env is not loaded properly, and how do we make the logging print the operation?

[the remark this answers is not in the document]
marwan-at-work: you can still embed the url if you'd like. `errors.E(op, fmt.Errorf(...)`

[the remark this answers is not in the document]
marwan-at-work: @devdinu no need to log the operation. If things failed right here, then the caller should fail the boot of the server

PR #436, opened 2018-08-09
Update pkg/storage/backend.go

robjloranger:
Backend was defined before the `Reader` interface was designed. This updates the `Backend` to embed the `Reader` instead of having `Lister`, `Getter` and `Checker`. Also updates the comment.


robjloranger:
oops, not an interface. No working from the web UI apparently

Issue #425, opened 2018-08-07
pkg/module: switch from go get to go mod download


marwan-at-work:
`go get` attempts to traverse all dependencies and downloads them and so 
1. it blows up in certain cases where a dependency has a build error. 
2. it's slow because you don't need to download all deps from the Proxy perspective, nor do you need to compile anything. 

Once CL 128355 goes in (or a new beta is released), we should switch `go mod download`. We should also look into the error field in the struct it returns to see if we can better use error handling in our fetcher.

michalpristas:
good catch @marwan-at-work

PR #448, opened 2018-08-10
Fetcher: switch to mod download

marwan-at-work:
Can't merge until our CI builds a Go version with the new command. 

Fixes https://github.com/gomods/athens/issues/425

### DIFF
--- a/pkg/module/go_get_fetcher.go
+++ b/pkg/module/go_get_fetcher.go
@@ -91,7 +91,7 @@ func getSources(goBinaryName string, fs afero.Fs, gopath, repoRoot, module, vers
 	disableCgo := "CGO_ENABLED=0"
 	enableGoModules := "GO111MODULE=on"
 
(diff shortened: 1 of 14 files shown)


pkg/module/go_get_fetcher.go
```diff
@@ -91,7 +91,7 @@ func getSources(goBinaryName string, fs afero.Fs, gopath, repoRoot, module, vers
 	disableCgo := "CGO_ENABLED=0"
 	enableGoModules := "GO111MODULE=on"
 
-	cmd := exec.Command(goBinaryName, "get", fullURI)
+	cmd := exec.Command(goBinaryName, "mod", "download", fullURI)
```
arschles: @marwan-at-work we won't have to do all the gymnastics to create the `main.go` and the `go.mod` files anymore, right?

marwan-at-work: @arschles the answer is unfortunately still no :)

Issue #284, opened 2018-07-17
switch to envconfig for configuration 


arschles:
the `./pkg/config/env` package has lots and lots of `envy.Get` (and friends) calls, with defaults, type conversions, etc...

let's remove all that duplicated code and use [envconfig](https://github.com/kelseyhightower/envconfig)

marwan-at-work:
@arschles Should we have a config struct with all of our expected environment variables?

robjloranger:
I think so @marwan-at-work. What do you think about using an init to first load the config, then the subsequent calls can return needed variables?

michalpristas:
👍 for a configuration file. keep dev defaults there. we can inject/replace the file with test/prod env easily on deploy

marwan-at-work:
What about having a json or toml representation of that config struct and so instead of populating env variables and keeping track of them, we just have one file where all this stuff lives. 
Something like this 

```golang
type Config struct {
  GoBinPath string
  OlympusGlobalURL string
  DebugLevel string
  ...
}
```

And then we can have several json representations for different environments: 

config.local.json
```json
{
  "GoBinPath": "go",
  "OlympusGlobalURL: "http://localhost:3001",
  "DebugLevel": "debug"
}
```

and then config.prod.json

```json
{
  "GoBinPath": "go",
  "OlympusGlobalURL": "https://www.olympus.com",
  "DebugLevel": "warning"
}
```

This way we don't need envconfig, or we can use it for overrides in the config if we really want to. Which is cool w/ me.

rchakra3:
I'll work on this!

Issue #261, opened 2018-07-13
Make 'env.Timeout' more specific


arschles:
I think we should have separate timeouts for different services. For example, mongo connection timeouts (see https://github.com/gomods/athens/pull/216), HTTP timeouts, etc...

Continuation of https://github.com/gomods/athens/pull/254#discussion_r202407219

robjloranger:
Agreed, I thought that as well but forgot to create the follow up.

Web requests and DB connections should have more sane defaults, maybe 60s?

Issue #190, opened 2018-06-20
Consolidate all environment variable fetching at the top of the stack


arschles:
I was figuring out how to deploy Olympus and it was pretty hard to find all the env vars it uses, and which of those it requires. It'll be easier when we have docs but even then it's nicer to have things in one place.

Everything uses `envy.Get` or `envy.MustGet`, so we need to:

1. Move all those commands to the top of the stack (i.e. near the `main`)
2. ... possibly create a utility function to get all the env vars (consider using [envconfig](https://github.com/kelseyhightower/envconfig) instead of envy for consolidating)
3. Pass env vars down the stack

marpio:
I would like to start this one

marpio:
Does #209 close this?

michalpristas:
I believe it does, to some extent. Next step is documenting all the things necessary to setup preferably with a script setting it for you.
Then I would consider this issue closed.

PR #453, opened 2018-08-10
add a configuration file

rchakra3:
Just want to preface this by saying this is the first time I've worked with handling such a large number of configs in go, and this might not be the right approach at all! Please let me know if that's the case! :)

I've used a combination of [viper](https://github.com/spf13/viper) and [validator](https://github.com/go-playground/validator) to achieve the following behavior:
- Read in a toml file (There's an example file with descriptions)
- Set properties from the following sources (in decreasing order of priority):
  - Environment variables
  - properties defined in the config file
  - defaults
- The resulting configuration struct is then validated. 
  - If any properties are provided for a storage backend, then all the required properties should be specified. For example, if a minio endpoint is provided, but secrets are not - a validation error will occur 

I wanted to make sure I was headed down the right path before starting to integrate these changes with the proxy and olympus.

*If* this is the right approach, in it's final state I would expect this to address https://github.com/gomods/athens/issues/190, https://github.com/gomods/athens/issues/284 and to some extent https://github.com/gomods/athens/issues/261


config.full.toml
```diff
@@ -0,0 +1,122 @@
...
+#         # Env override: MONGO_PASSWORD
+#         Password = ""
+#         # Defaults to Global Timeout
+#         # Env override: MONGO_CONN_TIMEOUT_SEC
+#         TimeoutSeconds = 30
+#     [Storage.RDBMS]
+#         # Env override: ATHENS_RDBMS_STORAGE_NAME
+#         Name = ""
+#         # Defaults to Global Timeout
+#         TimeoutSeconds = 30
```
arschles: @rchakra3 is this an example file (if so, that's awesome by the way!)? if so, can you put a comment indicating that at the top?

rchakra3:
@arschles yes I was hoping to make those changes in smaller increments. 
I did start down that path but realized there would be a TON of small changes all over. 

IMO it might be easier to test/review if we make smaller subsequent PRs for various parts of the codebase and slowly move over to using this instead of env.*. What do you think?

pkg/config/parse.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	// validate all required fields have been populated
+	validate := validator.New()
+	if err := validate.Struct(config); err != nil {
+		return nil, err
+	}
+	return &config, nil
+}
+
+func setDefaultsAndEnvBindings() {
```
marwan-at-work: This function might be hard to maintain, why don't we use `envconfig` to make our struct more declarative. Something like this: 

```golang
type Config struct {
  EnableCSRFProtection bool
  GoEnv                string         `validate:"required",envconfig:"GO_ENV",default:"development"`
  GoBinary string `validate:"required" envconfig: "GO_BINARY_PATH" default: "go"
}
```
If viper provides the same functionality, then we don't need envconfig. 

And in the initializer, you basically just do 3 steps in a row: 
1. marshal TOML into struct
2. parse envconfig (override TOML if env var exists)
3. validate struct

robjloranger: I agree, it doesn't look like viper provides something like this so we should look into `envconfig`. Viper just uses [pelletier/go-toml](github.com/pelletier/go-toml) to load a map of values.

rchakra3:
Sorry, I've been out for most of the weekend!

envconfig's a great suggestion for maintainability, I'll switch to using that instead of the method.

@robjloranger thanks for the detailed review! :) I'm going to get started on addressing your comments as well.

rchakra3:
I think I've addressed all the initial comments. Going to start adding some tests now!

scripts/check_gofmt.sh
```diff
@@ -5,4 +5,4 @@
 set -euo pipefail
 
 GO_FILES=$(find . -iname '*.go' -type f | grep -v /vendor/)   # All the .go files, excluding vendor/
-test -z $(gofmt -s -l $GO_FILES | tee /dev/stderr)
+#test -z $(gofmt -s -l $GO_FILES | tee /dev/stderr)
```
robjloranger: Just forgot to remove commenting from previous commit :slightly_smiling_face:

rchakra3: That's one last thing I need to figure out :(

gofmt fails on the travis build server for some reason and I haven't been able to figure out why. After spending an hour on it I decided to first validate everything else was working fine. Now, back to figuring that out! :)

rchakra3: Well. Looks like something I did fixed it. That was odd!

robjloranger: very odd, I wonder what it was

pkg/config/parse.go
```diff
@@ -0,0 +1,99 @@
...
+	EnableCSRFProtection bool           `envconfig:"ATHENS_ENABLE_CSRF_PROTECTION" default:"false"`
+	Proxy                *ProxyConfig   `validate:""`
+	Olympus              *OlympusConfig `validate:""`
+	Storage              *StorageConfig `validate:""`
+}
+
+// ParseConfigFile parses the given file into an athens config struct
+func ParseConfigFile(configFile string) (*Config, error) {
+
+	// attempt to read the given config file
```
devdinu: `envconfig` supports reading from environment or `yml` file. Any reason to have a `toml` file?

pkg/config/parse.go
```diff
@@ -0,0 +1,99 @@
...
+
+// envOverride uses Environment variables and specified defaults to override unspecified properties
+func envOverride(config *Config) error {
+	if err := envconfig.Process("athens", config); err != nil {
+		return err
+	}
+	return nil
+}
+
+func validateConfig(c Config) error {
```
devdinu: envconfig has `tag` which does this job `required: true`

pkg/config/rdbms.go
```diff
@@ -0,0 +1,7 @@
+package config
+
+// RDBMSConfig specifies the properties required to use an RDBMS as the storage backend
+type RDBMSConfig struct {
+	Name    string `validate:"required" envconfig:"ATHENS_RDBMS_STORAGE_NAME"`
+	Timeout int
```
devdinu: could rename it to `TimeoutInMs` with the duration unit so we know the usage.

vendor/github.com/go-playground/locales/currency/currency.go
```diff
@@ -0,0 +1,308 @@
+package currency
+
+// Type is the currency type associated with the locales currency enum
+type Type int
```
devdinu: we could remove the vendor directory and add it to `.gitignore` since we've the dependencies with dep. This could go as a separate PR/issue. and others could add more if there's any reason for those checkedin dependencies.

devdinu:
Awesome, @rchakra3 infact we could remove all the `env.Vars()` if we're gonna depend on this config package, by exposing these via functions. We could create a issue and track it, if we wanna remove `env`  package.

pkg/config/parse.go
rchakra3: I didn't think [envconfig](https://github.com/kelseyhightower/envconfig) has support for yaml files. Am I missing something?

No specific reason for toml over yaml, I just personally prefer it for configs.

The idea with this change was to move towards using config files over only environment variables so we could have a few distinct configs (dev/test/prod) that would be easy to switch between and reason about

pkg/config/storage.go
```diff
@@ -0,0 +1,74 @@
+package config
+
+// StorageConfig provides configs for various storage backends
+type StorageConfig struct {
+	CDN   *CDNConfig   `validate:""`
```
marwan-at-work: do we need to put `validate:""`?? can we not just omit it? same as below.

pkg/config/storage.go
```diff
@@ -0,0 +1,74 @@
...
+	Disk  *DiskConfig  `validate:""`
+	GCP   *GCPConfig   `validate:""`
+	Minio *MinioConfig `validate:""`
+	Mongo *MongoConfig `validate:""`
+	RDBMS *RDBMSConfig `validate:""`
+}
+
+func setDefaultTimeouts(s *StorageConfig, defaultTimeout int) {
```
marwan-at-work: do we need this function if we already have `default` struct tags in the Config?

config.example.toml
```diff
@@ -0,0 +1,144 @@
...
+
+[Proxy]
+    # StorageType sets the type of storage backend the proxy will use.
+    # Possible values are memory, disk, mongo, postgres, sqlite, cockroach, mysql
+    # Defaults to mongo
+    # Env override: ATHENS_STORAGE_TYPE
+    StorageType = "mongo"
+    # Port sets the port the proxy listens on
+    # Env override: PORT
+	Port = ":3000"
```
michalpristas: why theose spaces? is it intentional?

pkg/config/cdn.go
```diff
@@ -0,0 +1,24 @@
...
+
+// CDNConfig specifies the properties required to use a CDN as the storage backend
+type CDNConfig struct {
+	Endpoint string `envconfig:"CDN_ENDPOINT"`
+	Timeout  int    `validate:"required"`
+}
+
+// CDNEndpointWithDefault returns CDN endpoint if set
+// if not it should default to clouds default blob storage endpoint e.g
+func (c *CDNConfig) CDNEndpointWithDefault(value *url.URL) *url.URL {
```
marwan-at-work: Do we need this function? Can we not just work with the URL as a string and not a *ur.URL?

pkg/config/proxy.go
```diff
@@ -0,0 +1,9 @@
+package config
+
+// ProxyConfig specifies the properties required to run the proxy
+type ProxyConfig struct {
+	StorageType           string `validate:"required" envconfig:"ATHENS_STORAGE_TYPE" default:"mongo"`
+	OlympusGlobalEndpoint string `validate:"required" envconfig:"OLYMPUS_GLOBAL_ENDPOINT" default:"olympus.gomods.io"`
```
michalpristas: i think this changed in some recent PR to localhost

pkg/config/parse.go
```diff
@@ -0,0 +1,99 @@
...
+	"runtime"
+
+	"github.com/BurntSushi/toml"
+	"github.com/kelseyhightower/envconfig"
+	validator "gopkg.in/go-playground/validator.v9"
+)
+
+var validate *validator.Validate
+
+func getValidator() *validator.Validate {
```
marwan-at-work: this is a little bit magical. If the *global* variable is not nil then return it, otherwise return a new one. Not sure why we have to set up things this way. We can just initialize `validate` globally on line 12 like this: `var validate = validator.New()` and that's it. Or, we can remove this function as well as the global variable and just initialize a new validator when we wanna use it in other funcs.

pkg/config/rdbms.go
robjloranger: It's in seconds and derived from the global timeout currently.

pkg/config/parse.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return nil, err
+	}
+	confStr := string(confBytes)
+	return parseConfig(confStr)
+}
+
+func parseConfig(confStr string) (*Config, error) {
+	var config Config
+	if _, err := toml.Decode(confStr, &config); err != nil {
```
marwan-at-work: you can get rid of all the stuff in lines 41-46 and just use `toml.DecodeFile` and pass the file path :)

pkg/config/parse.go
```diff
@@ -0,0 +1,99 @@
...
+	// set default values that are dependent on the runtime
+	setRuntimeDefaults(&config)
+
+	// set default timeouts for storage backends if not already set
+	setDefaultTimeouts(config.Storage, config.Timeout)
+
+	// delete invalid storage backend configs
+	// envconfig initializes *all* struct pointers, even if there are no corresponding defaults or env variables
+	// this method prunes all such invalid configurations
+	deleteInvalidStorageConfigs(config.Storage)
```
marwan-at-work: do we need to delete them? no harm in having them around as opposed to accidentally deleting something we need.

vendor/github.com/go-playground/locales/currency/currency.go
robjloranger: We are currently committing the vendor directory, this prevents upstream VCS changes from breaking builds.

pkg/config/rdbms.go
devdinu: so renaming to`TimeoutSec` would help.

pkg/config/parse.go
devdinu: Oops, my bad. yaml support is in viper. If its plain file with `export`, running `go test` in different directory would be simpler. toml works too.

pkg/config/rdbms.go
robjloranger: That would be another PR, we do intend to have more than the one timeout. If there is not an issue open yet, can you please open one for that? This PR however should just move us from where we are to using a config file.

config.example.toml
```diff
@@ -0,0 +1,144 @@
+# This is an example configuration with all supported properties explicitly set
+# Most properties can be overriden with environment variables specified in this file
+# Most properties also have defaults (mentioned in this file) if they are not set in either the config file or the corresponding environment variable
+
+# GoBinary returns the path to the go binary to use. This value can be a name of a binary in your PATH, or the full path
+# Defaults to "go"
+# Env override: GO_BINARY_PATH
+GoBinary = "go"
```
manugupt1: Should it default to go1.11rc?

robjloranger: It defaults to `go` because we assume the installation environment has an appropriate version installed.

When not running in a Docker or similar environment where we have an appropriate version of go installed, the operator is expected to set `GO_BINARY_PATH`. Which would currently be `go1.11rc1`, or the full path if the binary is not in the user's `$PATH`.

config.example.toml
rchakra3: Wow great catch! Not intentional :)

pkg/config/rdbms.go
devdinu: relevant to this https://github.com/gomods/athens/issues/261

pkg/config/storage.go
rchakra3: It ends up skipping properties that don't have that tag, and since we want to run the validation for CDNConfig, MongoConfig, etc, we need it here!

pkg/config/storage.go
rchakra3: this sets timeouts for storage backends to the global timeout value if they haven't been specified. Since it's not a static value (it's whatever the global timeout is set to), I couldn't think of a better way to do this!

config.example.toml
manugupt1: Makes sense!

pkg/config/cdn.go
rchakra3: I've left this in for the `WithDefault` behavior.

Currently the default passed in is the base URI of the storage implementation being used. We could add logic here to default to the Endpoint/URI of the storage config defined in `StorageType`. That would require this config package to be aware of a little more business logic, and if we're doing that we could also just move all the logic in the `GetStorage()` method in `actions/storage.go` to this package as well.

pkg/config/parse.go
rchakra3: I broke it apart to make it a little easier to test, but I'm happy to use `DecodeFile` too if you think that's better :)

pkg/config/parse.go
rchakra3: The problem it solves is that invalid configs (created by envconfig by default) would fail overall validation of the Config struct later. 

I was planning on passing in our logger when I start moving code to use this package so we could log the fact that we're ignoring certain storage configs.

pkg/config/parse.go
rchakra3: good point :)

pkg/config/proxy.go
rchakra3: didn't realize, fixed!

rchakra3:
@marwan-at-work done!

rchakra3:
Ok I just found a bug in this. Please don't merge yet!

rchakra3:
While working on using these configs in the codebase I found an issue - There is no way to know if the value returned by `envconfig` is a default or an Environment variable. 

By extension, when overriding the values provided by the config file with the struct returned by `envconfig` the default value as per envconfig will always override the values provided by the config file. I didn't catch this before because the example config file I was using had the same values as the defaults.

I've added tests to test this behavior with my last commit as well as setting defaults in code. I don't think there's a way to work around this.

config.example.toml
```diff
@@ -0,0 +1,154 @@
...
+        Bucket = "gomods"
+    [Storage.Mongo]
+        # Full URL for mongo storage
+        # Env override: ATHENS_MONGO_STORAGE_URL
+        URL = "mongo.example.com"
+        # Timeout for networks calls made to Mongo in seconds
+        # Defaults to Global Timeout
+        # Env override: MONGO_CONN_TIMEOUT_SEC
+        Timeout = 300
+    [Storage.RDBMS]
```
marwan-at-work: remove this.

pkg/config/parse.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/BurntSushi/toml"
+	"github.com/kelseyhightower/envconfig"
+	validator "gopkg.in/go-playground/validator.v9"
+)
+
+// Config provides configuration values for all components
+type Config struct {
+	GoEnv                string         `validate:"required" envconfig:"GO_ENV"`
+	GoBinary             string         `validate:"required" envconfig:"GO_BINARY_PATH"`
+	LogLevel             string         `validate:"required" split_words:"true"`
```
marwan-at-work: why do we need to split words on log level?

pkg/config/olympus.go
```diff
@@ -0,0 +1,20 @@
...
+	StorageType       string `validate:"required" envconfig:"ATHENS_STORAGE_TYPE"`
+	WorkerType        string `validate:"required" envconfig:"OLYMPUS_BACKGROUND_WORKER_TYPE"`
+	RedisQueueAddress string `validate:"required" envconfig:"OLYMPUS_REDIS_QUEUE_PORT"`
+}
+
+func setOlympusDefaults(o *OlympusConfig) *OlympusConfig {
+	if o == nil {
+		o = &OlympusConfig{}
+	}
+	overrideDefaultStr(&o.StorageType, "memory")
```
marwan-at-work: i find these super weird, the way we had it with `envy` was actually cleaner. 
I have two suggestions: 
1. Use the `default` struct tags in the struct defintion of configs.
2. Use the TOML file itself to set all the defaults, which makes sense. If you ever wanna change a default, you can set an env var. 

I like 2 better than 1. Do you see a problem in this?

pkg/config/parse.go
rchakra3: The env variable is ATHENS_LOG_LEVEL, so that helps with the conversion to a camelcase property

pkg/config/olympus.go
rchakra3: I would prefer 2 for sure.
It just forces the existence of a config file with all the required values defined, which is what the defaults help with

pkg/config/parse.go
marwan-at-work: @rchakra3 why not just say `evnconfig: "ATHENS_LOG_LEVEL"` so that it's clear exactly what the corresponding ENV VAR is :)

pkg/config/parse.go
```diff
@@ -0,0 +1,117 @@
...
+	return &config, nil
+}
+
+func setDefaults(config *Config) {
+	if config.MaxConcurrency == 0 {
+		config.MaxConcurrency = runtime.NumCPU()
+	}
+
+	overrideDefaultStr(&config.GoEnv, "development")
+	overrideDefaultStr(&config.GoBinary, "go")
```
marwan-at-work: just as a random example, the TOML file already sets config.GoBinary to the string "go", so why do we need to do this?

pkg/config/olympus.go
marwan-at-work: @rchakra3 what do you mean by forcing the existence of a config file? This function only sets a default if I'm not mistaken. 

In any case, the config file is going to be mandatory, the server shouldn't work without it. We can even include it as part of the Go Binary if we _really_ want to.

rchakra3:
@marwan-at-work I think I've addressed all your comments!

@arschles I know you had concerns about documentation not lining up - I'll update the docs in #517 once this gets merged because this PR by itself has absolutely no change in behavior.

pkg/config/storage.go
```diff
@@ -0,0 +1,66 @@
...
+// StorageConfig provides configs for various storage backends
+type StorageConfig struct {
+	CDN   *CDNConfig   `validate:""`
+	Disk  *DiskConfig  `validate:""`
+	GCP   *GCPConfig   `validate:""`
+	Minio *MinioConfig `validate:""`
+	Mongo *MongoConfig `validate:""`
+}
+
+func setStorageTimeouts(s *StorageConfig, defaultTimeout int) {
```
marwan-at-work: can we not get rid of this function and just use the `default` struct tag instead?

config.example.toml
```diff
@@ -0,0 +1,147 @@
...
+    # Defaults to ":6379"
+    # Env override: ATHENS_REDIS_QUEUE_PORT
+    RedisQueueAddress     = ":6379"
+    # Flag to turn off Proxy Filter middleware
+    # Defaults to true
+    # Env override: PROXY_FILTER_OFF
+    FilterOff = true
+    # Username for basic auth
+    # Env override: BASIC_AUTH_USER
+    BasicAuthUser = "user"
```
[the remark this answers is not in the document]
rchakra3: done!

robjloranger:
Let's get this in if your happy @marwan-at-work, there are some more config changes not yet worked on so we can avoid making Rohan keep this updated.

Issue #309, opened 2018-07-19
We should have a private module filter for the proxy


arschles:
We have an exclude list which tells the proxy to not return a module that matches a pattern on the list now. We also need to implement a list of patterns that the proxy can use to determine if a module is "private" - this means that it shouldn't ask for the module from upstream (the registry) and instead should directly fetch them from the VCS (with `go get`)

marpio:
@fedepaol relevant issue: https://github.com/gomods/athens/issues/96 and PR https://github.com/gomods/athens/pull/212

arschles:
@fedepaol please do! and unlike last time, this one is not going to go away or become obsolete 😄

fedepaol:
I gave a quick look at https://github.com/gomods/athens/pull/212 . What I would try to do is to extend the domain of the current filter (now +/- ) with a p maybe?

Something like
- a
+ a/b/c
p a/b/c/d

Would mean that a/b/c/d will never be looked for in olympus an fetched directly from the vcs.
Does it make sense? To me it feels better than having a different filter file.

Can I try to proceed with this implementation?

fedepaol:
New question, sorry @arschles :-( 

I am looking at https://github.com/gomods/athens/blob/db7a318a6f1944640b4474f3df3a83332d74f271/cmd/proxy/actions/app_proxy.go#L21 and https://github.com/gomods/athens/blob/db7a318a6f1944640b4474f3df3a83332d74f271/cmd/proxy/actions/cache_miss_handler.go#L25 

My understanding of the current status of the proxy is that when there is a miss in the local storage, alignments (notify and fetch from the registry) are queued asynchronously but the request that caused the miss gets a 404.

Is that correct? 
If so, what part should I implement with this issue? Just leaving the thing as they are (trying to asynchronously fill the local cache directly from the vcs) or something different?

arschles:
Hey Federico - I’m traveling today and going to be offline tomorrow morning
so hopefully someone else can give more detail.

We’re going to begin filling caches synchronously soon, so if there’s any
way you can design the filter so that it can be applied later after the
synchronous implementation gets merged, that would be the best option.

There are a few PRs open that show how the async => sync work is coming but
I’m on my phone so I don’t have them handy

fedepaol:
Grazie!
Also, does my suggestion of extending the current configuration instead of
adding a new one make sense (
https://github.com/gomods/athens/issues/309#issuecomment-406823918)?

[...]

Issue #138, opened 2018-05-16
Add tests for proxy routes


marpio:
At the moment there are not tests for cmd/proxy/actions package.

arschles:
I'm going to take this tomorrow 😄

arschles:
@fedepaol I remember you mentioning that you have a PR that added some tests for proxy routes. was it #456?

nclandrei:
Hi there, guys! I really want to start contributing to Athens so I thought this PR was a good start. 

I see the PR suggested by @fedepaol hasn’t been merged yet. Is this still available? Can I start working on a PR for it? 😁

nclandrei:
Awesome, thanks! :) Awaiting reply then!

fedepaol:
Sorry for not being helpful !

nclandrei:
No problem at all, @fedepaol! 🙂

robjloranger:
@nclandrei that PR is very out of date so feel free to start another, just comment here if you do and we can close the old one.

xiaoyunyang:
I'd like to take this on if no one else is working on this right now.

michalpristas:
Hey @xiaoyunyang feel free to start, in case of any questions just ping us on Athens slack channel on Gophers slack

xiaoyunyang:
@michalpristas Thanks. I'm targeting submitting a PR by the end of this week.

arschles:
@xiaoyunyang I'm going to close this issue now. If you'd still like to write more tests, please feel free to comment here or submit a PR. If you comment here, I'll still get the notification and we can continue the discussion. Cheers! 😄

PR #456, opened 2018-08-10
Added support for private repos

fedepaol:
Meant to fix https://github.com/gomods/athens/issues/309

Extended the filter file described in https://github.com/gomods/athens/pull/212 with a new domain value P

Now we can have 

    - a
    + a/b/c
    P a/b/c/d

Where P means that the package does not need to be looked for in olympus. 
The current behaviour of the middleware function is 

    - -> returns 404
    + -> redirect to olympus
    P -> do the proxy stuff


cmd/proxy/main.go
```diff
@@ -4,10 +4,12 @@ import (
 	"log"
 
 	"github.com/gomods/athens/cmd/proxy/actions"
+	"github.com/gomods/athens/pkg/module"
 )
 
 func main() {
-	app, err := actions.App()
+	mf := module.NewFilter()
```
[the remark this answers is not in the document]
fedepaol: By not injecting the filter in the constructor I won't be able to drive the behaviour of the tests. Can you suggest me another approach? 
Shipping the filter file is sub optimal.

fedepaol:
Really going to bed now :-)

cmd/proxy/main.go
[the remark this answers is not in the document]
marwan-at-work: @fedepaol Not sure if I misunderstood, but I just meant that you should call `module.NewFilter()` inside `cmd/proxy/actions/app.go` instead of `cmd/proxy/main.go`. Would that not work?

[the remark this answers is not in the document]
fedepaol: That would obviously work, but by passing the filter to the constructor, I can provide a custom tailored one that behaves as I expect in the tests: https://github.com/gomods/athens/pull/456/files#diff-c11aa63db187809541291d77069b2dc9R24

If I don't do that, the filter is created empty and I would need to ship a `filter.conf` containing the different rules, that will be loaded in *all* the tests (unit and e2e), and that will be the same for all the scenarios.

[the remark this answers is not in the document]
marwan-at-work: I understand now. @fedepaol do you need to spin up the whole App? Can you not just test the middleware by itself?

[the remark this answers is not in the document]
fedepaol: I think it's the way testing with buffalo works. Looking at https://gobuffalo.io/en/docs/testing it says it needs a suite (https://github.com/gobuffalo/suite) to work.

The suite was already in place (see https://github.com/gomods/athens/blob/master/cmd/proxy/actions/actions_test.go#L14) 

I could test the single middleware function but I am afraid I would not have all the buffalo testing framework support (such as faking requests and asserting the responses)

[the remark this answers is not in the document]
marwan-at-work: @fedepaol let's find out if we can just test the middleware _and_ be able to fake requests/responses. I really hate for us to set up the entire *buffalo.App  (with storage, workers, and everything) _*just*_ so we can test one middleware that has nothing to do with all of those dependencies. It defeats the purpose of unit testing :) 

I'm not a buffalo expert so CCing @arschles and @robjloranger to see if that's possible.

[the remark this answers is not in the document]
fedepaol: Relevant to this, https://github.com/gomods/athens/issues/138 needs to be implemented and will suffer of the same problems.
So the suite and the app setup (**which was already in place before implementing my middleware test** and runned along with the other unit tests) will be for testing all the proxy routes and not _just one middleware_ .

Let's wait for the @arschles  and @robjloranger  , then I'll try to mock the buffalo context in order to make the test work.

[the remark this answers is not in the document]
robjloranger: I think we can test just the middleware effectively. I had a quick look at the middlewares that ship with buffalo, and it looks pretty straight forward, see [buffalo/csrf_test.go](https://github.com/gobuffalo/buffalo/blob/master/middleware/csrf/csrf_test.go) for an idea. I should have time tonight if you want to pair up.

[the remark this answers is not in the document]
fedepaol: Sorry I could not pair yesterday, had to get some sleep :-) 
Will take inspiration from the test you just linked to provide a custom middleware test.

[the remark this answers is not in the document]
fedepaol: Done. The only quirk is https://github.com/gomods/athens/pull/456/commits/a244bac6f0b02e28d0276a82fc9f62a44f807a49#diff-bd90c74522e398d450dbe3611f594105R21 to prevent a null pointer crash when the middleware function was tracing

cmd/proxy/actions/filter.go
```diff
@@ -0,0 +1,59 @@
+package actions
+
+import (
+	"net/http"
+	"net/url"
+
```
marpio: please remove empty line

cmd/proxy/actions/filter.go
```diff
@@ -0,0 +1,59 @@
...
+			defer sp.Finish()
+			mod, err := paths.GetModule(c)
+
+			if err != nil {
+				// if there is no module the path we are hitting is not one related to modules, like /
+				return next(c)
+			}
+
+			// not checking the error. Not all requests include a version (i.e. list requests)
+			version, _ := paths.GetVersion(c)
```
michalpristas: on err please go next(c)

cmd/proxy/actions/filter.go
```diff
@@ -0,0 +1,59 @@
...
+			}
+
+			rule := mf.Rule(mod)
+			switch rule {
+			case module.Exclude:
+				return c.Render(http.StatusForbidden, nil)
+			case module.Private:
+				return next(c)
+			case module.Include:
+				newURL := redirectToOlympusURL(c.Request().URL)
```
michalpristas: please add a TODO: to spinup cache filling worker

pkg/module/filter.go
```diff
@@ -65,13 +64,17 @@ func (f *Filter) AddRule(path string, rule FilterRule) {
 
 // ShouldProcess evaluates path and determines if module should be communicated or not
 func (f *Filter) ShouldProcess(path string) bool {
```
michalpristas: this i believe is not used anywhere. please remove if so

michalpristas: with 3 states it is not so easy to tell if processing should or shouldn't be done. let's make caller do the call.

marpio: It is used (i.e. cache_miss_fetcher and cache_miss_reporter) but it's mostly dead code.

michalpristas: we should not care about the dead code. we surely dont need cachemiss reported, this an go away, cache miss fetcher we might want to reuse at some point so i would just comment it out with a TODO that it needs to reflect new Rules logic. also it should not be needed there at all.
so action steps: 
- remove cache miss reporter
- remove check from cache_miss_fetcher

michalpristas: #461

pkg/module/filterRule.go
```diff
@@ -10,4 +10,6 @@ const (
 	Include
```
marpio: Would it make sense to rename it to `Public`?

marpio: We can exclude everything - public as well as private modules. The difference between Include and Private is that in the first case we will fetch it from Olympus and not from private VCS but the name doesn't really reflect that.

michalpristas: my concern is that you can explicitly say
```
+ github.com/a/b
```
which will Include this even though `github.com/a` was excluded. 
in this case, it is more intuitive to read it as include github.com/a/b then github.com/a/b is public
i understand this as an operation

pkg/module/filter.go
```diff
@@ -168,7 +172,7 @@ func newRule(r FilterRule) ruleNode {
 }
 
 func getConfigLines() ([]string, error) {
-	configName := env.IncludeExcludeFileName()
+	configName := env.FilterConfigurationFileName()
```
marwan-at-work: let's include `const op errors.Op = "module.getConfigLines"` and return `errors.E` in the rest of the body of this function :)

cmd/proxy/actions/filter.go
michalpristas: Discussed offline: will keep it here as we want List endpoint follow the same flow

pkg/module/filterRule.go
marpio: I understand your point but now we can say:
P github.com/a/b
which introduces the same problem.
So maybe IncludePrivate and IncludePublic?
Either way it's just an comment and can be resolved later.

michalpristas: Yep and it is perfectly fine, `P github.com/a/b` treats github repo as a private, so it skips Olympus and goes to VCS directly. maybe we can rename Private to Direct so it reflects behavior.
i think this is also a valid scenario.
but these are the namings, the core feature behind those names implemented by Federico is good and we can start a discussion on this in a separate thread.

cmd/proxy/actions/filter.go
fedepaol: Yes, but we will have to change this logic then , since filling the cache while redirecting to olympus makes no sense.

fedepaol:
I think I have done everything that was requested here

pkg/module/filter.go
arschles: @fedepaol now that #461 is merged, can you create an issue for removing this method?

I'd like to get this PR merged sooner rather than later, so if we have that issue, I'm cool with merging.

cmd/proxy/actions/cache_miss_fetcher.go
```diff
@@ -13,7 +13,7 @@ import (
 
 const (
 	// OlympusGlobalEndpoint is a default olympus DNS address
-	OlympusGlobalEndpoint = "olympus.gomods.io"
+	OlympusGlobalEndpoint = "http://localhost:3001"
```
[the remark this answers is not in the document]
arschles: Or you can create an issue to do this as a follow-up for now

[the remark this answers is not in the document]
fedepaol: It's already configurable, that's the default value.
The filter uses `GetOlympusEndpoint()` defined here https://github.com/gomods/athens/blob/5f02ada292e77a61710b89e2c0e6cd3002c3d1a4/cmd/proxy/actions/cache_miss_fetcher.go#L78 `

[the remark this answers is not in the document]
arschles: ah ok, I missed that - thanks for letting me know!

cmd/proxy/actions/filter.go
```diff
@@ -0,0 +1,60 @@
...
+			}
+
+			rule := mf.Rule(mod)
+			switch rule {
+			case module.Exclude:
+				return c.Render(http.StatusForbidden, nil)
+			case module.Private:
+				return next(c)
+			case module.Include:
+				// TODO : spin up cache filling worker and serve the request using the cache
```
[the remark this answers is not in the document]
marwan-at-work: @arschles why do we need a worker instead of just calling `next(c)`?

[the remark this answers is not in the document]
arschles: I'm not sure if we decided for sure, but iirc we were talking about redirecting to olympus and downloading into the local storage in the background? If I'm remembering right, that would be what the worker is for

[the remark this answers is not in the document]
fedepaol: opened https://github.com/gomods/athens/issues/471 , if we want to next() we'll need to change the download handlers to hit olympus instead of the vcs

pkg/module/filter.go
arschles: Issue to follow-up with: https://github.com/gomods/athens/issues/472

arschles:
@michalpristas @marpio I'm gonna dismiss your reviews and merge this PR because I think your comments have been addressed:

- for https://github.com/gomods/athens/pull/456/files#r209499480, we have #472 to follow-up with
- for https://github.com/gomods/athens/pull/456/files#r209554785, there might be some agreement (?) but regardless, I created #473 to continue discussion if we need it

fedepaol:
Then we are already ready. Just treat every module as private.


Il mar 14 ago 2018, 01:23 Marwan Sulaiman <notifications@github.com> ha
scritto:

cmd/proxy/actions/filter.go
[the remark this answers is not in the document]
arschles: I created https://github.com/gomods/athens/issues/474 to continue discussion

Issue #330, opened 2018-07-25
Proposal: get rid of the storage.Connector


arschles:
I'm not sure about this one. Do we need a separate `Connect()` function, or is it ok to connect automatically when we create a new `Saver`, `Lister`, etc...?

robjloranger:
I like just connecting on creation, this gives the caller a ready to use `Saver` etc

kteb:
I'll work on this one.

michalpristas:
Closing #457  is in

PR #457, opened 2018-08-11
Remove storage connector

kteb:
This is a way to see the impact of the following proposal: #330 

As only mango & rdbms were using it. For me it make sense to remove it, and integrate it to the creation of the connection.


pkg/storage/mongo/mongo.go
```diff
@@ -16,18 +16,19 @@ type ModuleStore struct {
...
-// that satisfies the Backend interface.  You must call
-// Connect() on the returned store before using it.
-func NewStorage(url string) *ModuleStore {
-	return &ModuleStore{url: url}
+// NewStorage returns a connected Mongo backed storage
+// that satisfies the Backend interface.
+func NewStorage(url string) (*ModuleStore, error) {
+	ms := &ModuleStore{url: url}
+	err := ms.connect()
+	return ms, err
```
arschles: @kteb could you return `nil, err` if the call to `connect` failed?

robjloranger: :+1:

pkg/storage/rdbms/test_suite.go
```diff
@@ -14,12 +14,12 @@ type TestSuite struct {
 // NewTestSuite creates a common test suite
 func NewTestSuite(model *suite.Model) (storage.TestSuite, error) {
 	conn := model.DB
-	rdbmsStore := NewRDBMSStorageWithConn(conn)
+	rdbmsStore, err := NewRDBMSStorageWithConn(conn)
```
robjloranger: can you check the error here and return it with a nil TestSuite instead? Then nil on last return.

pkg/storage/mongo/test_suite.go
```diff
@@ -33,12 +33,12 @@ func newTestStore() (*ModuleStore, error) {
 		return nil, err
 	}
 
-	mongoStore := NewStorage(muri)
+	mongoStore, err := NewStorage(muri)
 	if mongoStore == nil {
```
marpio: Maybe just check the error instead?

robjloranger:
Just curious, how did you merge the changes from master? It's making it look like you are adding them in this PR.

cmd/proxy/.env
```diff
@@ -1 +1,3 @@
 ATHENS_MONGO_STORAGE_URL=mongodb://127.0.0.1:27017
+GO_ENV=test_postgres
```
michalpristas: I know this is not the main focus of this PR. but can you please include
`TMPDIR=/tmp/` which is needed if the contributor does not have buffalo installed and 
`PATH=${PATH}:./bin` which is needed if somebody installs buffalo using our script

pkg/storage/mongo/mongo.go
```diff
@@ -16,18 +16,24 @@ type ModuleStore struct {
...
-func NewStorage(url string) *ModuleStore {
-	return &ModuleStore{url: url}
+// NewStorage returns a connected Mongo backed storage
+// that satisfies the Backend interface.
+func NewStorage(url string) (*ModuleStore, error) {
+	ms := &ModuleStore{url: url}
+
+	err := ms.connect()
+	if err != nil {
+		return nil, err
```
michalpristas: can you please
```
const op errors.Op = "fs.NewStorage"
return nil, errors.Op(op, err)
```

pkg/storage/rdbms/rdbms.go
```diff
@@ -11,27 +11,26 @@ type ModuleStore struct {
...
-// connectionName
-func NewRDBMSStorageWithConn(connection *pop.Connection) *ModuleStore {
-	return &ModuleStore{
+// that satisfies the Storage interface.
+func NewRDBMSStorageWithConn(connection *pop.Connection) (*ModuleStore, error) {
+	ms := &ModuleStore{
 		conn: connection,
 	}
+	err := ms.connect()
+	return ms, err
```
michalpristas: can you please
```
const op errors.Op = "fs.NewRDBMSStorageWithConn"
return nil, errors.Op(op, err)
```

scripts/test_unit.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
 
 # test_unit.sh
+
+source cmd/proxy/.env
```
michalpristas: if you include changes requested up there (.env) please make sure that `.env` is sourced before buffalo download starts. ideally at the beginning of `set-dev-env`
also there is a PR in focusing on this so you might get rid of it entirely as well #449

robjloranger: I think it's an unintentional result of a rebase, not a direct change from this PR

kteb:
As it's kind of a big mess (because of my rebase, I don't want to take credit of other people work) I thought maybe to close this PR and recreate a new one without this issue  what do you think ?

pkg/storage/rdbms/rdbms.go
```diff
@@ -10,28 +11,31 @@ type ModuleStore struct {
...
+	}
+	return ms, nil
 }
 
 // NewRDBMSStorageWithConn  returns a connected RDBMS Module Storage
-// that satisfies the Storage interface. You must call
-// Connect() on the returned store before using it.
-// connectionName
+// that satisfies the Storage interface.
 func NewRDBMSStorageWithConn(connection *pop.Connection) *ModuleStore {
```
kteb: Not sure if this method should really stay here, it's only used in a test (in test_suite.go) not it sure if it's really needed

cmd/proxy/.env
arschles: @michalpristas would it be ok for us to do this in a follow up?

pkg/storage/mongo/test_suite.go
arschles: +1, I feel like that would read nicer. @kteb can you do an `if err != nil` here instead?

pkg/storage/rdbms/rdbms.go
arschles: @kteb could the test use some other kind of module store? if so, it's up to you whether you'd like to remove it in this PR or wait for a follow-up. I usually do things like that in a follow up for what it's worth

cmd/proxy/.env
michalpristas: this was addressed in other PRs. please ignore my previous comment

pkg/storage/mongo/test_suite.go
```diff
@@ -33,12 +33,12 @@ func newTestStore() (*ModuleStore, error) {
 		return nil, err
 	}
 
-	mongoStore := NewStorage(muri)
-	if mongoStore == nil {
+	mongoStore, err := NewStorage(muri)
+	if err != nil {
 		return nil, fmt.Errorf("Mongo storage is nil")
```
robjloranger: Just need to adjust the error message

pkg/storage/rdbms/rdbms.go
arschles: Follow-up: https://github.com/gomods/athens/issues/491

PR #460, opened 2018-08-13
Remove TMPDIR env from scripts

michalpristas:
Improving first run experience


carolynvs:
I think at this point everything that I wrote has been assimilated by the Athens collective and improved. 💯  I am so proud! 😊

PR #465, opened 2018-08-13
pkg/storage: add err ops

marwan-at-work:
last pr and I think we're all set with having error operations in all of the code that is being used today :)


pkg/storage/s3/storage.go
```diff
@@ -7,6 +7,8 @@ import (
 	"io"
 	"net/url"
 
+	"github.com/gomods/athens/pkg/errors"
+
```
michalpristas: can you remove newlines here?

PR #468, opened 2018-08-13
Tiny start to the codeowners file

### LINKED ISSUE #466: Add a codeowners file
Let's add this file so we can get Github to tell us which folks should review a PR before it can merge

Docs on this file: https://help.github.com/articles/about-codeowners/

Let's keep categories broad for now and make more specific if necessary as we learn

arschles:
Fixes #466 

After this is merged, I'll turn on "require review from codeowners" in the branch protection settings

screen shot 2018-08-13 at 11 59 43 am


.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+/ @gomods/contributors @gomods/maintainers
```
robjloranger: I was thinking about it after approving, do we want to have maintainers and contributors own everything by default and the only contributors own `/pkg`? Or did you intend to have maintainers own everything and then contributors own `/pkg`?

robjloranger: also I think the syntax is `*` for all, not `/`.

robjloranger: as well, these rules operate by 'last match precedence'. So without contributors above, the below line will make them owners of `/pkg` over maintainers.

.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
...
+/pkg @gomods/contributors
```
robjloranger: the rule for a directory _and_ it's contents should be `/pkg/`, if that's what your going for.

robjloranger:
I will take this over

robjloranger:
So with the changes made, Maintainers own everything by default and Contributors own everything _under_ `/pkg/` by default.

robjloranger:
This is good to go, if https://github.com/gomods/athens/pull/468#issuecomment-415525821 is acceptable defaults.

robjloranger:
Removed contributors as owners as owners require write access.

PR #475, opened 2018-08-14
Default to in-memory storage for the proxy

carolynvs:
This makes it easier for someone to try out or just run the proxy by executing the binary without further environment setup.

### DIFF
--- a/cmd/proxy/actions/storage.go
+++ b/cmd/proxy/actions/storage.go
@@ -18,7 +18,6 @@ import (
 
 // GetStorage returns storage backend based on env configuration
 func GetStorage() (storage.BackendConnector, error) {
-	// changing to mongo storage, memory seems buggy
 	storageType := env.StorageTypeWithDefault("memory")
 	var storageRoot string
 	var err error
--- a/cmd/proxy/actions/storage.go
+++ b/cmd/proxy/actions/storage.go
@@ -19,7 +19,7 @@ import (
 // GetStorage returns storage backend based on env configuration
 func GetStorage() (storage.BackendConnector, error) {
 	// changing to mongo storage, memory seems buggy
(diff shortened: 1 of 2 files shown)


cmd/proxy/actions/storage.go
```diff
@@ -19,7 +19,7 @@ import (
 // GetStorage returns storage backend based on env configuration
 func GetStorage() (storage.BackendConnector, error) {
 	// changing to mongo storage, memory seems buggy
```
rchakra3: nit- maybe get rid of this comment?

rchakra3: As per https://github.com/gomods/athens/issues/165 it looks like the issue was fixed!

robjloranger: Awesome, so just comment clean up :+1:

michalpristas:
I'm so looking forward to this @carolynvs 
On this, I believe we should put new community members first. I believe we are able of setting one env var to tweak behavior.

marwan-at-work:
@michalpristas 100%. I was just thinking it might be a bad experience if they're not aware that we're storing modules in their memory by default. So @carolynvs 's docs are gonna be perfect 💯

robjloranger:
this is going to be great @carolynvs, that third install option will go nicely with the little server under my desk :grinning:

carolynvs:
hehe I _may_ have some experience with deploying things under my desk to avoid IT 😇

carolynvs:
Yup! I will call out the OOM and other concerns of running with in-memory storage in follow-up doc PRs.

marwan-at-work:
@carolynvs @arschles I love all the lengthy explanations from everyone. Sometimes it's hard to be aware of all the scenarios so they really help :)

PR #477, opened 2018-08-14
Update e2e tests to not use sudo

robjloranger:
Using a system temporary directory via mktemp we can place our GOMOD_CACHE
in a place that does not require root permission to clear.

As well we use the environment variable for GO_BINARY_PATH to point to
a module compatible go version. At this time, 1.11beta3.


robjloranger:
This will also require me to add instructions for setting `GO_BINARY_PATH`, which makes things complicated for people again.

If all are apposed to the change I can just keep some stashed changes for that script to run locally. I don't like having sudo in scripts on my local filesystem.

.travis.yml
```diff
@@ -9,6 +9,7 @@ env:
     - PATH=${PATH}:./bin
     - POP_PATH=$PWD/cmd/proxy
     - GO_ENV=test_postgres
+    - GO_BINARY_PATH=go
```
robjloranger: this can just be `go` in travis because travis is using 1.11beta3

robjloranger:
Another idea re: `GO_BINARY_PATH`, what about a conditional check in the e2e script that sets `GO_BINARY_PATH` to just `go` if not set already. It could even print a warning in such a situation that the script is expecting the go version installed to support modules, and link the docs for install instructions.

scripts/test_e2e.sh
```diff
@@ -7,12 +7,12 @@ set -xeuo pipefail
...
 
-GOMOD_CACHE=$(go env GOPATH)/pkg/mod
+TMPDIR=$(mktemp -d)
+GOMOD_CACHE=$TMPDIR/pkg/mod
 export PATH=${REPO_DIR}/bin:${PATH}
 
 clearGoModCache () {
-  # The sudo is a necessary workaround until go is fixed
-  sudo rm -fr ${GOMOD_CACHE}
+  rm -fr ${GOMOD_CACHE}
```
carolynvs: Has the bug where sudo is required fixed yet? I'm still seeing on my machine that I can't remove the mod cache without sudo using go1.11rc1

```
$ go version
go version go1.11rc1 darwin/amd64
$ rm -fr ~/go/pkg/mod/
rm: /Users/carolynvs/go/pkg/mod//github.com/pkg/errors@v0.8.0/example_test.go: Permission denied
```

robjloranger: I hadn't checked, this would work around that by placing the mod cache in a user accessible temporary location. Check out ln#11

scripts/test_e2e.sh
```diff
@@ -42,10 +41,10 @@ clearGoModCache
 export GO111MODULE=on
 # Make sure that our test repo works without the GOPROXY first
 unset GOPROXY
-go run .
+$GO_BINARY_PATH run .
```
carolynvs: This script is intended to work both on a local dev machine when executed via `make test-e2e` and on travis. So GO_BINARY_PATH should be conditionally defaulted at the top so that each dev don't need to figure out that var and set it explicitly. e.g. `export GO_BINARY_PATH=${GO_BINARY_PATH:-go}`

robjloranger: awesome, that's what I was thinking too. I'm going to be working today so if you have time and are interested I'm open to commits :grin:

scripts/test_e2e.sh
robjloranger: No problem, It was more of a test to start and seemed to work so I ran with it.

scripts/test_e2e.sh
carolynvs: "open to commits" I'm not sure if you are suggesting that I submit a PR to your repo to add that line to your branch? Let me know how I can help!

robjloranger: I think you can push commits directly, maybe not. No need for the trouble of a PR. I can update with the changes this evening.

carolynvs: I am not a member or maintainer in this repo, so I can't push to your branch directly. 😊

robjloranger: :disappointed: I thought you were. No worries.

carolynvs: Yeah just tried and I can't push to your branch.

robjloranger:
@carolynvs changes up :grin:

Issue #472, opened 2018-08-13
Remove ShouldProcess from filter


fedepaol:
With https://github.com/gomods/athens/pull/461 `module.ShouldProcess` https://github.com/gomods/athens/blob/master/pkg/module/filter.go#L67 is not used anymore and must be removed

PR #480, opened 2018-08-14
Removed ShouldProcess

michalpristas:
ShouldProcess of a filter cannot determine the behavior of the caller with additional `Private` state implemented in #456 
This way we force the caller to decide

Fixes: #472

### DIFF
--- a/pkg/module/filter.go
+++ b/pkg/module/filter.go
@@ -66,7 +66,7 @@ func (f *Filter) AddRule(path string, rule FilterRule) {
 // Rule returns the filter rule to be applied to the given path
 func (f *Filter) Rule(path string) FilterRule {
 	segs := getPathSegments(path)
-	rule := f.shouldProcess(segs...)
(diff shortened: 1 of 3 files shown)


pkg/module/filter.go
```diff
@@ -63,17 +63,15 @@ func (f *Filter) AddRule(path string, rule FilterRule) {
...
-	rule := f.Rule(path)
-	// process everything unless it's excluded
-	return rule != Exclude
-}
-
 // Rule returns the filter rule to be applied to the given path
 func (f *Filter) Rule(path string) FilterRule {
 	segs := getPathSegments(path)
-	return f.shouldProcess(segs...)
+	rule := f.shouldProcess(segs...)
```
marpio: `shouldProcess` returns 3 states as well so maybe rename that too?

michalpristas: great point @marpio

PR #482, opened 2018-08-14
Updated to go1.11rc1

michalpristas:
RevInfo updated to match go upstream

> // We want to canonicalize to .info files with those fields omitted

By those fields meant: Name and Short
https://github.com/golang/go/blob/daeb0b4f538872a37787626816c47afcf71bfd2c/src/cmd/go/internal/modfetch/cache.go#L312

Looks like @latest is according to specs now, 
retrieves latest version, in case of no version pseudo version is retrieved

### DIFF
--- a/pkg/download/goget/goget.go
+++ b/pkg/download/goget/goget.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"os/exec"
-	"strings"
 	"time"
 
 	"github.com/gomods/athens/pkg/config"
@@ -78,13 +77,7 @@ func (gg *goget) Latest(ctx context.Context, mod string) (*storage.RevInfo, erro
 		return nil, errors.E(op, err)
 	}
 
-	pseudoInfo := strings.Split(lr.Version, "-")
-	if len(pseudoInfo) < 3 {
-		return nil, errors.E(op, fmt.Errorf("malformed pseudoInfo %v", lr.Version))
-	}
 	return &storage.RevInfo{
-		Name:    pseudoInfo[2],
-		Short:   pseudoInfo[2],
 		Time:    lr.Time,
 		Version: lr.Version,
 	}, nil
--- a/pkg/storage/rev_info.go
+++ b/pkg/storage/rev_info.go
@@ -11,7 +11,5 @@ import (
 // (see "Download Protocol" header)
 type RevInfo struct {
 	Version string    `json:"Version"` // version string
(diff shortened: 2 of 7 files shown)


pkg/download/goget/goget.go
```diff
@@ -78,13 +77,7 @@ func (gg *goget) Latest(ctx context.Context, mod string) (*storage.RevInfo, erro
 		return nil, errors.E(op, err)
 	}
 
-	pseudoInfo := strings.Split(lr.Version, "-")
-	if len(pseudoInfo) < 3 {
-		return nil, errors.E(op, fmt.Errorf("malformed pseudoInfo %v", lr.Version))
-	}
 	return &storage.RevInfo{
-		Name:    pseudoInfo[2],
-		Short:   pseudoInfo[2],
```
arschles: @michalpristas was it necessary to delete these?

pkg/storage/rev_info.go
```diff
@@ -11,7 +11,5 @@ import (
 // (see "Download Protocol" header)
 type RevInfo struct {
 	Version string    `json:"Version"` // version string
-	Name    string    `json:"Name"`    // complete ID in underlying repository
-	Short   string    `json:"Short"`   // shortened ID, for use in pseudo-version
```
arschles: @michalpristas does rc1 no longer require these?

michalpristas:
@robjloranger they are there just because some .info files might contain them, but they are ignored. probably to avoid parsing failures. 

> The disk might have stale .info files that have Name and Short fields set.

I would like to keep them away so we don't have anything we don't need, but I'm ok to keep them in for a while If you folks would rather do this in two steps

pkg/storage/rev_info.go
michalpristas: no, comments here: 
https://github.com/golang/go/blob/daeb0b4f538872a37787626816c47afcf71bfd2c/src/cmd/go/internal/modfetch/cache.go#L311:L312

pkg/download/goget/goget.go
michalpristas: go will slowly read old .info and update it without those fields. so i hoped as we're not in prod that we can skip this step and remove them directly

robjloranger:
awesome, sounds good @michalpristas

marwan-at-work:
@arschles i believe your comments are addressed here, if so let's merge!

arschles:
Yup, my comments are addressed here 👍

Issue #420, opened 2018-08-06
Fully Dockerized Development Environment


marwan-at-work:
Sorry if this is a duplicate issue. But it would be very helpful to onboard new contributors and anyone who just wants to interact with Athens with nothing but docker (no buffalo needed, not even Go is needed) 

Let's have our dev set up to two tiers: 

1. Just spin things up, I hate installing things. 
`docker-compose up athens-dev` -- this will spin up the minimum dependencies but as well as two docker images (one for proxy and one for athens) that will already have buffalo and a module-enabled-go binary in them. So all you have to do is ping localhost:3000 

2. I'd like to contribute and know how things work and be able to develop fast. 
We already have this, just good old manual install of Go, Buffalo, and >go1.11beta3

michalpristas:
Use case one covered in #81

marwan-at-work:
@michalpristas Thanks for showing me this issue! 81 seems to be for publishing images which will probably be optimized without having buffalo or GOPATH. Does it cover building a local docker image just for development purposes?

michalpristas:
what do you mean by having a buffalo? 
81 should result in a pair of docker images which you can download, deploy and play with

marwan-at-work:
@michalpristas I assume 81 is about production-ready docker images. They can totally be used to play with locally. But they will probably be optimized, for example it will probably be `FROM scratch` and so a user cannot open a shell into a docker image and hack on things. It might be a nicer experience to provide a local Dockerfile (never published to docker hub) that is `FROM buffalo` and also has everything a user needs to "hack on Athens". 

You're right though, a user can totally download and play with the docker images from 81, but maybe we should give them a more powerful docker image for local development? 

Either way is cool w/e me :)

michalpristas:
i think it goes hand in hand and one precedes the other. we can start with playable image and than productize it into publishable form

robjloranger:
@marwan-at-work re: multi-build, we could look at using packer from hashicorp. It allows building multiple images from one config. We could even use it to eventually build VM images.

ref: https://packer.io

bearcherian:
Docker does allow for multi-build and stopping at a stage. Using the example you had you could do the following

```
docker build --target athens-dev -t athens/dev:local . 
```

Then it builds up to the `athens-dev` stage and tags the image as `athens/dev:local`. Then we could have a line in the Makefile to spin up this image.

[...]

PR #487, opened 2018-08-15
Some improvements to test scripts

arschles:
@robjloranger talked over slack about how it's difficult to run all tests locally. We agreed that https://github.com/gomods/athens/issues/420 would be a good long-term solution, but for now this should improve things. With these changes I was able to reduce the number of test failures on master, but not fix them all.

- Not sourcing from cmd/proxy/.env, because that doesn't export any variables
- Removing mysql support (I feel like 1 database is enough)
- Pruning networks on teardown

I'm happy to split these changes up into separate PRs - it's the end of the day for me so I wanted to get everything in :smile:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,6 @@ bench:
 
 .PHONY: alldeps
 alldeps:
-	docker-compose -p athensdev up -d mysql
 	docker-compose -p athensdev up -d postgres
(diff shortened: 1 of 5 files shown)


robjloranger:
I think the build failure is because you changed the name of the proxy development database. From `test_postgres` to `test`

Issue #311, opened 2018-07-20
Implement a proxy 'admission controller'


ryancox:
Allow users of the proxy to implement a webhook that codifies their policies for allowed packages. Prior to populating the cache, the webhook would be consulted and the request would either be allowed or failed and a 403 or something similar returned.

Examples of policy rules:

- No GPL enforced via a library like https://github.com/ryanuber/go-license
- No high entropy string enforced via a library like https://github.com/securego/gosec

arschles:
Thanks for bringing this up @ryancox! It would be a great addition to the proxy. 

Related issues: 

- https://github.com/gomods/athens/issues/96
- https://github.com/gomods/athens/issues/309

robjloranger:
Please do

PR #492, opened 2018-08-15
Proxy 'admission controller'

fedepaol:
Fixes https://github.com/gomods/athens/issues/311
The hook call is enabled when `ATHENS_PROXY_VALIDATOR` is set and contains the url of a controller webhook.

If the variable is set, the middleware filter performs a post on the webhook url passing the module and the version as json:

    {"Module":"/github.com/athens-artifacts/happy-path","Version":"v.1.0.0"}

If the hook returns `200`, the validation is passed, otherwise if the hook returns `403` (forbiddent) the filter blocks the call and returns 403 itself


cmd/proxy/actions/filter.go
```diff
@@ -1,24 +1,32 @@
...
+	"github.com/gomods/athens/pkg/config/env"
+	"github.com/gomods/athens/pkg/download"
 	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/log"
 	"github.com/gomods/athens/pkg/module"
 	"github.com/gomods/athens/pkg/paths"
 )
 
-func newFilterMiddleware(mf *module.Filter) buffalo.MiddlewareFunc {
+func newFilterMiddleware(mf *module.Filter, lggr *log.Logger) buffalo.MiddlewareFunc {
```
marwan-at-work: We should have a top middleware that creates a log Entry and pass it down. This middleware should not be _the one_ that creates an Entry interface instance.

cmd/proxy/actions/filter.go
```diff
@@ -1,24 +1,32 @@
...
 	const op errors.Op = "actions.FilterMiddleware"
 
 	return func(next buffalo.Handler) buffalo.Handler {
 		return func(c buffalo.Context) error {
 			sp := buffet.SpanFromContext(c).SetOperationName("filterMiddleware")
 			defer sp.Finish()
+
+			entry := download.ContextLogEntry(c, lggr)
```
marwan-at-work: the `download` package is the wrong place to instantiate a log Entry from *log.Logger. I know it does that for its own handlers but that's potentially a special case for type safety. If we want to merge them, then we should create a middleware package where we put all of our middleware handlers there and instantiate a request-scoped Entry in its own middleware. We don't have to merge them right away, but it's up to you

cmd/proxy/actions/filter.go
```diff
@@ -30,6 +38,18 @@ func newFilterMiddleware(mf *module.Filter) buffalo.MiddlewareFunc {
...
 
+			if validatorHook, ok := env.ValidatorHook(); ok && version != "" {
+				valid, err := validate(validatorHook, mod, version)
+				if err != nil {
+					entry.SystemErr(err)
+					return c.Render(http.StatusInternalServerError, nil)
+				}
+
+				if !valid {
+					return c.Render(http.StatusForbidden, nil)
```
marwan-at-work: should this return whatever status code the webhook returned?

pkg/config/env/validate.go
```diff
@@ -0,0 +1,18 @@
+package env
+
+import "github.com/gobuffalo/envy"
+
+const noProxyValidator = "NOVALIDATOR"
```
marwan-at-work: We should just default to an empty string which would effectively mean no hook.

cmd/proxy/actions/filter.go
```diff
@@ -30,6 +38,18 @@ func newFilterMiddleware(mf *module.Filter) buffalo.MiddlewareFunc {
 			// i.e. list requests path is like /{module:.+}/@v/list with no version parameter
 			version, _ := paths.GetVersion(c)
 
+			if validatorHook, ok := env.ValidatorHook(); ok && version != "" {
```
michalpristas: can we separate this into its own middleware? then we can have FCB to enable/disable this

cmd/proxy/actions/filter.go
fedepaol: I'd prefer to keep the interface between the proxy and the hook clear, stating that we only expect 200 and 403 as results and the proxy behave according to those two results (looking at the valid result). 403 is incidentally the result the proxy returns after a 403 from the hook, but I would avoid to propagate the result back. As it is now, every other status code raises a (logged) error because we expect only those two values from the hook

fedepaol:
@marwan-at-work @michalpristas the `middleware.go` file got pretty beefy. Not sure if it deserves to be moved under pkg (even if it's proxy-centric) .

michalpristas:
@fedepaol i agree with extracting middleware into its own pkg. this will add readability which is my main concern

cmd/proxy/actions/app.go
```diff
@@ -16,8 +16,10 @@ import (
 	"github.com/gocraft/work"
 	"github.com/gomods/athens/pkg/config/env"
 	"github.com/gomods/athens/pkg/log"
+	gomodsmiddleware "github.com/gomods/athens/pkg/middleware"
```
marwan-at-work: `pkgmiddleware` or just `mw` is probably cleaner.

fedepaol: Will change!

fedepaol:
@marwan-at-work opened https://github.com/gomods/athens/issues/537

cmd/proxy/actions/app.go
```diff
@@ -16,8 +16,10 @@ import (
 	"github.com/gocraft/work"
 	"github.com/gomods/athens/pkg/config/env"
 	"github.com/gomods/athens/pkg/log"
+	mw "github.com/gomods/athens/pkg/middleware"
 	"github.com/gomods/athens/pkg/module"
 	"github.com/gomods/athens/pkg/storage"
+
```
michalpristas: remove space please

cmd/proxy/actions/app.go
```diff
@@ -129,14 +131,22 @@ func App() (*buffalo.App, error) {
...
 	app.Use(T.Middleware())
+
 	if !env.FilterOff() {
 		mf := module.NewFilter()
-		app.Use(newFilterMiddleware(mf))
+		app.Use(mw.NewFilterMiddleware(mf))
+	}
+
+	// Having the hook set means we want to use it
+	if _, ok := env.ValidatorHook(); ok {
```
michalpristas: can we inject value into the middleware?
i don't think we will change the env value after the container is up, rather than restarting containers with proper configuration

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/gomods/athens/pkg/module"
+	"github.com/gomods/athens/pkg/paths"
+	"github.com/sirupsen/logrus"
+)
+
+type middlewareFunc func(entry log.Entry) buffalo.MiddlewareFunc
+
+// LogEntryMiddleware builds a log.Entry applying the request parameter to the given
+// log.Logger and propagates it to the given MiddlewareFunc
+func LogEntryMiddleware(middleware middlewareFunc, lggr *log.Logger) buffalo.MiddlewareFunc {
```
michalpristas: can you separate those so we have 
- middleware/filter.go
- middleware/log_entry.go
- middleware/validation.go
- middleware/middleware_test.go

michalpristas:
Great @fedepaol, Thank you

PR #498, opened 2018-08-17
create in memory driver for cdn metadata

### LINKED ISSUE #334: CDN metadata in memory driver
the last piece of #147 is the in memory driver.

This could be as simple as using the existing file system implementation but passing in an `afero.NewMemMapFs` to the (./pkg/cdn/metadata/fs).NewStorage when in memory is desired.

mydiemho:
fixes #334


pkg/cdn/metadata/mem/mem.go
```diff
@@ -0,0 +1,36 @@
+package mem
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/spf13/afero"
+
```
michalpristas: please remove this space

pkg/cdn/metadata/mem/mem.go
```diff
@@ -0,0 +1,36 @@
...
+		return memStorage, nil
+	}
+
+	memFs := afero.NewMemMapFs()
+	tmpDir, err := afero.TempDir(memFs, "", "")
+	if err != nil {
+		return nil, errors.E(op, fmt.Errorf("could not create temp dir for 'In Memory' CDN metadata (%s)", err))
+	}
+
+	return fs.NewStorage(memFs, tmpDir), nil
```
marpio: Please set the pkg level memStorage variable to the result of fs.NewStorage(...) before the return. Otherwise each call to mem.NewStorage will create a new instance and the behaviour will differ from other storages.

pkg/cdn/metadata/mem/mem.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/spf13/afero"
+
+	"github.com/gomods/athens/pkg/cdn/metadata"
+	"github.com/gomods/athens/pkg/cdn/metadata/fs"
+	"github.com/gomods/athens/pkg/errors"
+)
+
+var (
+	l          sync.Mutex
+	memStorage metadata.Storage
```
arschles: Would you be able to avoid these package-level variables? it looks like you can create a new memory fs (`NewMemMapFs`) on each call to `NewStorage`. What do you think?

marpio: The problem with that is that if we call NewStorage(...) and save something there we won't find it in another storage created with NewStorage(...) because it will reference different afero.Fs. with an OS file system it's not an issue but here we get a new instance on in-memory FS.

pkg/cdn/metadata/mem/mem.go
```diff
@@ -0,0 +1,36 @@
...
+	defer l.Unlock()
+
+	if memStorage != nil {
+		return memStorage, nil
+	}
+
+	memFs := afero.NewMemMapFs()
+	tmpDir, err := afero.TempDir(memFs, "", "")
+	if err != nil {
+		return nil, errors.E(op, fmt.Errorf("could not create temp dir for 'In Memory' CDN metadata (%s)", err))
```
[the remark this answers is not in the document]
michalpristas: i would gladly avoid having an additional abstraction just to avoid 5 lines of duplicate code

michalpristas:
@arschles @marpio were your comments addressed?

robjloranger:
@michalpristas I think it is safe to dismiss Aaron's change request as it was a question regarding the use of the package level variables. This prevents accidentally calling New more than once and not being able to access the stored data.

Issue #495, opened 2018-08-16
Make the proxy redirect to an upstream download protocol (i.e. Olympus) with HTTP 301s


arschles:
We voted in https://github.com/gomods/athens/issues/423#issuecomment-411862230 to make the proxy issue HTTP 301 redirects to the CLI when the proxy datastore doesn't have the given module@version.

_Note that we're going to do this for our first released version_. We can add local caching/storage after that.

arschles:
Also, this means that if you make the proxy "exclude" all modules in its exclude list, then it will act as if there is no upstream, fetch everything from the VCS (using `go get` on the backend), and store it all in its datastore.

marwan-at-work:
@arschles did you mean "include" all modules? "excluding" all modules would return status forbidden.

michalpristas:
exclude should break the communication.
Include goes upstream (whatever it is, for proxy its olympus, for olympus it's VCS)
direct goes to VCS without caring about upstream

fedepaol:
Can I take this one?

michalpristas:
This was discussed as post-MVP but sure @fedepaol

michalpristas:
@marwan-at-work you self-assigned this few days ago, are you ok with @fedepaol working on this?

marwan-at-work:
This is already done through the filter middleware and the global switch. Will close the issue and open a new one for cache filling.

PR #501, opened 2018-08-17
Proxy: use a global switch for Filter Middleware

marwan-at-work:
Fixes last part of https://github.com/gomods/athens/issues/495

### DIFF
--- a/pkg/config/env/filter.go
+++ b/pkg/config/env/filter.go
@@ -9,3 +9,11 @@ const defaultConfigurationFileName = "filter.conf"
 func FilterConfigurationFileName() string {
 	return envy.Get("ATHENS_FILTER_FILENAME", defaultConfigurationFileName)
 }
+
+// FilterOff checks PROXY_FILTER_OFF env and returns
+// true of it's equal to "true", otherwise false always.
+// It defaults to "true" until Olympus is the default
+// place to grab modules before the Proxy.
+func FilterOff() bool {
+	return envy.Get("PROXY_FILTER_OFF", "true") == "true"
+}
--- a/pkg/module/filter.go
+++ b/pkg/module/filter.go
@@ -15,6 +15,7 @@ var (
 
 // Filter is a filter of modules
 type Filter struct {
+	Off  bool
 	root ruleNode
 }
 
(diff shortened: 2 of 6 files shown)


pkg/module/filter.go
```diff
@@ -118,6 +119,7 @@ func (f *Filter) initFromConfig() {
 	lines, err := getConfigLines()
 
 	if err != nil || len(lines) == 0 {
+		f.Off = true
```
michalpristas: can we also have a ENV override? what do you think? even with a filter file provided, we could turn it off

marwan-at-work: @michalpristas I like that. I'll add one. I think the Filter should be Off by default unless provided with an actual configuration (or ENV override). But once we have Olympus working well, then maybe the Filter should be On by default unless the user explicitly turns it off by ENV only.

marpio: Could we also skip the whole middleware if the environment var is set?

marwan-at-work: @michalpristas i liked your ENV idea more than my "Off" idea that I basically removed my code 😄 -- it's much easier to just not use the middleware at all if we want it off.

pkg/config/env/filter.go
```diff
@@ -9,3 +9,11 @@ const defaultConfigurationFileName = "filter.conf"
 func FilterConfigurationFileName() string {
 	return envy.Get("ATHENS_FILTER_FILENAME", defaultConfigurationFileName)
 }
+
+// FilterOff checks PROXY_FILTER_OFF env and returns
+// true of it's equal to "true", otherwise false always.
+// It defaults to "true" until Olympus is the default
+// place to grab modules before the Proxy.
+func FilterOff() bool {
+	return envy.Get("PROXY_FILTER_OFF", "true") == "true"
```
robjloranger: Should the envy call default to `false` as the check is whether the filter is off?

like `envy.Get("PROXY_FILTER_OFF", "false") == "true"`
if `PROXY_FILTER_OFF` was `true`, result would be `true`
if `PROXY_FILTER_OFF` was 'false' result would be `false`

the way it is now, with default of true, if the user does not set `PROXY_FILTER_OFF` the result will be `true`.

robjloranger: I guess not, re-reading the comment clears that up. :D

marwan-at-work: Yes by default, it's just the Proxy, no filtering needed. But we will use Olympus by default once Olympus is ready.

marwan-at-work:
@michalpristas merging this to unblock myself and @carolynvs -- feel free to make comments if you have any and I'll open another PR :)

PR #507, opened 2018-08-17
Ensure there is only one in-mem storage instance

marpio:
At the moment each mem.NewStorage() will create own instance of afero.Fs which means if we do:
```
s1 = mem.NewStorage()
s1.Save(mod@ver)
```
then `s2.Get(mod@ver)` will return `nil` because `s2` has it's own in-mem file system.
This can be confusing because every other storage would return the module.

### DIFF
--- a/pkg/storage/mem/mem.go
+++ b/pkg/storage/mem/mem.go
@@ -31,9 +31,9 @@ func NewStorage() (storage.Backend, error) {
 		return nil, errors.E(op, fmt.Errorf("could not create temp dir for 'In Memory' storage (%s)", err))
 	}
 
-	s, err := fs.NewStorage(tmpDir, memFs)
+	memStorage, err = fs.NewStorage(tmpDir, memFs)
 	if err != nil {
 		return nil, errors.E(op, fmt.Errorf("could not create storage from memory fs (%s)", err))
 	}
-	return s, nil
+	return memStorage, nil
 }


devdinu:
@marpio Since we've `memStorage != nil and return` the issue you mentioned wouldn't happen right?

marwan-at-work:
@marpio why do we need a package level variable in the first place? sorry just saw this :)

michalpristas:
@marwan-at-work the problem is only with in memory. if you have disk you will retrieve same disk storage each time you call New, that's fine
but with memory, you will get the new instance of in-memory space each time you call new.
so you upload module
 - calls api
 - gets storage (in memory instance)
 - saves module
 - fine

then you do list
 - calls api
 - gets disk instance (new in memory instance)
 - returns {}
 - hmmmm

marwan-at-work:
In the 2nd step, when you say "get storage", are we calling NewStorage multiple times? Because I thought we call NewStorage only once when the server starts, then pass it as a closure to all the handlers. Does that still create a new instance??

michalpristas:
yes, that's true and from the view of current implementation, it works. 
but our concern was about the behavior consistency between drivers. if you call New on S3 or any other storage you will get the same backing storage every time. we wanted to achieve the same result with the in-mem as well so we don't end up with some unpredicted behavior during refactorings

marwan-at-work:
@michalpristas ah, that makes a lot of sense. Thanks for the explanation!

michalpristas:
sorry for making it so confusing :-)

PR #512, opened 2018-08-18
Proxy: remove unused arguments

marwan-at-work:
Just some clean up as I go


cmd/proxy/actions/app.go
```diff
@@ -87,7 +86,7 @@ func App() (*buffalo.App, error) {
 
 	// Automatically redirect to SSL
 	app.Use(ssl.ForceSSL(secure.Options{
-		SSLRedirect:     ENV == "production",
+		SSLRedirect:     ENV == "production" && false,
```
manugupt1: Why would you want to not to force ssl?

marwan-at-work: @manugupt1 ops :)

PR #515, opened 2018-08-19
Proxy: make Force SSL configurable

marwan-at-work:
HTTPS should not be forced on Production environments for the Proxy because the Proxy can potentially run inside Google App Engine, GKE, or other platforms where certs are provided by the platform and your app doesn't need to have its own certs.

### DIFF
--- a/pkg/config/env/ssl.go
+++ b/pkg/config/env/ssl.go
@@ -1,0 +1,13 @@
+package env
+
+import (
+	"os"
+)
+
+// ForceSSL returns true if the PROXY_FORCE_SSL
+// env is set to "true", otherwise defaults to false.
+// This is used to make sure the Proxy would redirect
+// to https on any request.
+func ProxyForceSSL() bool {
+	return os.Getenv("PROXY_FORCE_SSL") == "true"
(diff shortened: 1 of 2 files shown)


pkg/config/env/ssl.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"os"
+)
+
+// ProxyForceSSL returns true if the PROXY_FORCE_SSL
+// env is set to "true", otherwise defaults to false.
+// This is used to make sure the Proxy would redirect
+// to https on any request.
+func ProxyForceSSL() bool {
+	return os.Getenv("PROXY_FORCE_SSL") == "true"
```
rchakra3: can we name this ATHENS_FORCE_SSL? with the env conf change it makes it easier to map to config values

marwan-at-work: @rchakra3 this is potentially *not* an option for Olympus, because I imagine Olympus is going to be public and with https by default.

rchakra3: ah, that makes sense!

PR #516, opened 2018-08-19
Athens: disable buffalo logger in production

marwan-at-work:
Buffalo logger (so far) is just for development purposes, your proxy is likely gonna run in the cloud and regular info logs such as requests will be provided automatically. Otherwise, your regular log monitoring service is gonna get too cluttered with unparsable buffalo logs.


pkg/log/buffalo_logger.go
```diff
@@ -12,9 +12,13 @@ import (
 // It's also good to note that internal logs
 // from buffalo should only be allowed in development
 // as our logging-system should be handled from our codebase.
-func Buffalo() buffalo.Logger {
+// If the environment is production, then its a noop logger.
+func Buffalo(env string) buffalo.Logger {
 	l := logrus.New()
 	l.Formatter = &buffaloFormatter{}
+	if env == "production" {
```
manugupt1: Is it not possible to drain the production logs somewhere?
Cloud monitoring systems won't be able to diagnose cases where panics happen only in production but not in other modes?

manugupt1: Just to be clear, we might not want to have drain somewhere in MVP

marwan-at-work: @manugupt1 this is just buffalo logs, not application logs. Buffalo spits out internal logs such as every single request that comes in. It clutters the cloud monitoring systems. 
However, you're right about panics. We should capture those. Unfortunately buffalo swallows panics and tries to log them with missing info, but we should at least log whatever it gives us. So I've updated this pr. Thanks!

pkg/log/buffalo_logger.go
```diff
@@ -12,9 +12,13 @@ import (
...
+		l.Out = nopwriter{}
+	}
```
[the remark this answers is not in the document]
marwan-at-work: @devdinu we don't need error level from Buffalo, error level is logic based and it comes from our app-logger not buffalo-logger. However, panic makes sense so I've updated the PR. I'm also passing down the level now instead of using ENV.

cmd/olympus/actions/app.go
```diff
@@ -62,7 +63,14 @@ func App(config *AppConfig) (*buffalo.App, error) {
...
 	}
 
-	lggr := log.New(env.CloudRuntime(), env.LogLevel())
+	lvl, err := env.LogLevel()
+	if err != nil {
+		return nil, err
+	}
+	lggr := log.New(env.CloudRuntime(), lvl)
+	if ENV == "production" {
+		lvl = logrus.PanicLevel
```
arschles: could we have a way to override this level for production? For example, if you want to deploy a canary of a new release, you may want to deploy to a single node/pod/process with verbose logging and see what buffalo logs say when you turn on a small bit of traffic

PR #525, opened 2018-08-20
Powershell scripts

### LINKED ISSUE #429: Include scripts for windows
continuing from: https://github.com/gomods/athens/pull/415#issuecomment-410953621

We should add batch or powershell scripts for windows users.

Can we have make select the script based on OS?

marpio:
Fixes #429


michalpristas:
I have a suggestion: how about on top of that, we provide init.cmd which will alias all those scripts under own aliases available to use in windows cmd
e.g 
you will run your cmd, then run init.cmd and then you will have aliases provided
test-e2e = powershell scripts/ps/test-e2e.ps
....

so you can use it even from a traditional command line

marpio:
Yes, I was also thinking about that (that's why it's WIP) although, I actually wanted to provide a ps script with [switch] params, which you can call like `init.ps -test-unit -test-e2e` and so on

init.ps1
```diff
@@ -0,0 +1,152 @@
...
+	Set-Location $(Join-Path cmd proxy)
+	& buffalo dev
+}
+
+if ($docs.IsPresent) {
+	Set-Location docs 
+	& hugo
+}
+
+if ($setup_dev_env.IsPresent) {
```
michalpristas: can we place setupdevenv before build?
and dev/alldeps before tests?
so you can say `init -setupdevenv build alldeps test-unit` and you have a killer combo

scripts/ps/check_gofmt.ps1
```diff
@@ -0,0 +1,7 @@
+# check_gofmt.ps1
+# Fail if a .go file hasn't been formatted with gofmt
+$GO_FILES = Get-ChildItem -Recurse -Filter *.go  -Name | Select-String -Pattern $(Join-Path "vendor" | Join-Path -ChildPath "") -NotMatch
+$out = & gofmt -s -l $GO_FILES
+if ($out.length -gt 0) {
+    Write-Error $out
+}
```
michalpristas: m: but can you make sure each script ends with a newline

scripts/ps/check_golint.ps1
```diff
@@ -0,0 +1,7 @@
+# check_golint.ps1
+# Run the linter on everything
+
+$out = & golint -set_exit_status $(& go list ./...)
+if ($LastExitCode -ne 0) {
+    Write-Error $out
+}
```
michalpristas: one more new-line please, you can use inline code editing for that.
it looks awesome @marpio

PR #527, opened 2018-08-20
SSL Support for Mongo

michalpristas:
In Order for Mongo to support TLS these things needs to be done
- you need to have cert generated: `pem` and `crt`
- volume needs to be shared with docker image and needs to contain `pem` file `-v /etc/ssl:/etc/ssl`
- you need to create user in mongo (set `MONGO_INITDB_ROOT_USERNAME` and `MONGO_INITDB_ROOT_PASSWORD` in your image)
- env for proxy ATHENS_MONGO_CONNECTION_STRING is in a format `mongodb://user:pass@localhost:27017`
- env for proxy ATHENS_MONGO_CERT_PATH needs to point to `crt` file


marwan-at-work:
@michalpristas can we still run mongo without TLS after this PR?

michalpristas:
@marwan-at-work yes indeed, mongo continues to run without SSL in our test for example. 
But once you set CERT path, it assumes you want to use SSL and expects that everything is setup properly

pkg/eventlog/mongo/mongo.go
```diff
@@ -108,3 +115,35 @@ func (m *Log) Clear(id string) error {
...
+
+	dialInfo, err := mgo.ParseURL(m.url)
+	if err != nil {
+		return nil, err
+	}
+
+	dialInfo.Timeout = env.MongoConnectionTimeoutSecWithDefault(1)
+
+	if m.certPath != "" {
+		roots := x509.NewCertPool()
```
carolynvs: Just a question, not a request, but I'm wondering if we are going to want to allow people to bring their own custom CA's too eventually?

pkg/eventlog/mongo/mongo.go
```diff
@@ -108,3 +115,35 @@ func (m *Log) Clear(id string) error {
...
+		cert, err := ioutil.ReadFile(m.certPath)
+		if err != nil {
+			return nil, err
+		}
+
+		if ok := roots.AppendCertsFromPEM(cert); !ok {
+			return nil, fmt.Errorf("failed to parse certificate from: %s", m.certPath)
+		}
+
+		tlsConfig.InsecureSkipVerify = true
```
carolynvs: Why are we skipping verification?

pkg/storage/mongo/all_test.go
```diff
@@ -3,22 +3,19 @@ package mongo
...
 
 func (d *MongoTests) SetupTest() {
-	store, err := NewStorage("mongodb://127.0.0.1:27017")
+	ms, err := newTestStore()
 
-	d.NoError(err)
+	d.Require().NoError(err)
 
-	store.s.DB(store.d).C(store.c).RemoveAll(nil)
-	d.storage = store
```
carolynvs: How come we don't need to set `d.storage` anymore? That doesn't appear to be related to the PR.

michalpristas: yes it is not related, we used that only to test creation of the storage but we don't use it anywhere so i removed it

pkg/eventlog/mongo/mongo.go
michalpristas: good point this should be configurable, skipping for self signed which i had during testing

pkg/eventlog/mongo/mongo.go
michalpristas: i haven't thought about this. can you please elaborate, as you have more experience in this field definitely, why would this be a wanted feature. I haven't worked with custom CA so far. is it about people being paranoid?

carolynvs: Lots of companies have their own CA, and it's not a trusted root by default. It lets them issue certs inside their own infra without dealing with paying for it, or the rotation schedules demanded by Lets Encrypt.

Another reason is self-signed certs. We can let people specify their ca and use a self-signed cert without having skip verify turned on. (which addresses my comment below)

carolynvs: I'm fine with keeping skip verify on for now, and then circling back in a follow-on PR to support custom CAs. I don't want to over-complicate this PR, what you have works just fine as an intermediate step. 👍

carolynvs: :shipit:

michalpristas: added a todo and created #540

PR #528, opened 2018-08-20
Proxy: add NETRC file mount

marwan-at-work:
We need to have a way to authenticate a deployed Proxy to have access to private repos (that's kinda the whole point of the Proxy lol). 

Currently, .netrc is the easiest way possible but we can't mount it directly to the home directory because Kubernetes would remove other files in ~ causing our internal goget to be confused. 

I'd like to open another issue to use the Linux in-memory authentication as suggested [here](https://confluence.atlassian.com/bitbucketserver/permanently-authenticating-with-git-repositories-776639846.html)

### DIFF
--- a/pkg/config/env/netrc.go
+++ b/pkg/config/env/netrc.go
@@ -1,0 +1,15 @@
+package env
+
+import (
+	"os"
+)
+
+// NETRCPath tells you where the .netrc path initially resides.
+// This is so that you can mount the .netrc file to a secret location
+// in the fs system and then move it ~/.netrc. In certain deployments
+// like Kubernetes, we can't mount directly to ~ because it would then
+// clean out whatever is already there as part of the image (such as
+// .cache directory in the Go image).
+func NETRCPath() string {
+	return os.Getenv("ATHENS_NETRC_PATH")
+}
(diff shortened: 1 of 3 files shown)


marwan-at-work:
@michalpristas you're killing it 💯

pkg/config/env/netrc.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+// NETRCPath tells you where the .netrc path initially resides.
+// This is so that you can mount the .netrc file to a secret location
+// in the fs system and then move it ~/.netrc. In certain deployments
+// like Kubernetes, we can't mount directly to ~ because it would then
+// clean out whatever is already there as part of the image (such as
+// .cache directory in the Go image).
+func NETRCPath() string {
+	return os.Getenv("ATHENS_NETRC_PATH")
```
arschles: perhaps this should have used `envy`?

marwan-at-work: @arschles a little intentional since I don't want to default to any value if the env var is not set. I can potentially do `envy.Get("ATHENS_NETRC_PATH", "")` if you'd like? no biggie on my side

PR #536, opened 2018-08-20
Proxy: MVP Dockerfile

marwan-at-work:
The Proxy (and Olympus) both need a Go distribution installed to download proxies, why not just use the Go image itself? 

For MVP, I removed any buffalo build stuff for the proxy because we don't need it. You can just build a go binary directly. 

cc: @arschles because I know you have a PR with updates to the same dockerfile itself.


manugupt1:
Are you able to run the web app with these changes in the Dockerfile?

marwan-at-work:
@manugupt1 what do you by web app? Like front end? Probably not. The Proxy is just an API as of now and I think we're gonna have cool UI things after MVP. When that time comes, we can update the Dockerfile.

manugupt1:
I meant can you hit the healthz endpoint that you wrote and get 200 response from this build image?

marwan-at-work:
@manugupt1 yes of course :)

manugupt1:
This awesome on so many levels @marwan-at-work I did not realize since we were already committing the `vendor/`  files, go would suffice

cmd/proxy/Dockerfile
```diff
@@ -1,36 +1,14 @@
...
-FROM gobuffalo/buffalo:v0.11.0 as builder
-
-RUN mkdir -p $GOPATH/src/github.com/gomods/athens/cmd/proxy
-WORKDIR $GOPATH/src/github.com/gomods/athens/cmd/proxy
-
-# this will cache the npm install step, unless package.json changes
-ADD cmd/proxy/package.json .
-ADD cmd/proxy/yarn.lock .
-RUN yarn install --no-progress
+FROM golang:1.11-rc
```
michalpristas: 👍

marwan-at-work:
@carolynvs feel free to take a look and comment and I can open another PR if you have any concerns or suggestions :)

cmd/proxy/Dockerfile
```diff
@@ -1,36 +1,14 @@
...
-RUN apk add --no-cache ca-certificates
-
-WORKDIR /bin/
-
-COPY --from=builder /bin/app .
+COPY . .
 
-# Comment out to run the binary in "production" mode:
-# ENV GO_ENV=production
+RUN CGO_ENABLED=0 go build -o /bin/app ./cmd/proxy
```
carolynvs: Just curious why we need to disable cgo here since we aren't cross compiling? If we do need to cross compile later (e.g. we are trying to do a multi-stage build with scratch as the final target), then this command needs to be changed a bit more.

michalpristas: we need to try out, when i wrote this with one of the first versions of vgo, cgo_enabled=1 caused some issues with vgo get and this was just an ugly hack how to avoid failures

marwan-at-work: @carolynvs how would it be changed if we're building into a scratch target? Just curious :) AFAIK, CGO_ENABLED=0 will allow us to put that binary in a scratch container as it will be a static binary. Could be wrong tho. 

Side note, the Athens dependencies are going to make it a little bit of a pain to build into a scratch container, because we'll need a go distribution (not just the go binary), and the go binary itself needs a git binary and all the other supported VCS's I assume?

carolynvs: To be clear, I'm not advocating a scratch container! 😀 Just saying that disabling cgo isn't enough, you also need to deal with the net libs that normally go doesn't build into the binary, the standard way to deal with that is including `-a -tags netgo -installsuffix netgo` in the build command.

carolynvs: Yeah I think busybox still has enough libs available that it would work, while scratch is completely empty. You can also run into this if you cross compile a linux binary that you want to use on say... arch and debian IIRC.

PR #538, opened 2018-08-21
Remove separate docker volume prune command

### LINKED ISSUE #535: Remove docker volume prune from Makefile
Down targets in make file run `docker volume prune` which removes _all_ unused volumes (and asks for confirmation). It will be better to run [`docker-compose down`](https://docs.docker.com/v17.09/compose/reference/down/) with `-v` flag which deletes volumes related to the project.

>     -v, --volumes           Remove named volumes declared in the `volumes`
>                                     section of the Compose file and anonymous volumes
>                                     attached to containers.

ebati:
Fixes #535

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,10 +67,8 @@ dev:
 
 .PHONY: down
 down:
-	docker-compose -p athensdev down
-	docker volume prune
+	docker-compose -p athensdev down -v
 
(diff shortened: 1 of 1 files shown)


michalpristas:
Thanks @ebati for the PR ;-) code deletion is always nice

PR #544, opened 2018-08-21
Deploy docker image during CI

### LINKED ISSUE #81: Build and publish docker images for the Athens server
We'll need to do this on the road to #57 and #80

carolynvs:
🚨 I am testing this out on my own repo and registry. Will remove the WIP once it's working as intended.
___

master builds publish using the git describe output  as the immutable tag and canary for the mutable tag. Examples:

    gomods/proxy:abc1234
    gomods/proxy:canary

tagged builds publish using the tag as the immutable tag, which is usually going to be a semver value and with latest for the mutable tag.
Examples:

    gomods/proxy:v1.0.0
    gomods/proxy:latest

branch builds publish using the git describe output as the immutable tag and the branch name for the mutable tag. Examples:

    gomods/proxy:def4567
    gomods/proxy:beta-1

Suggestions welcome on exactly where to document this for others. @jeremyrickard is writing install docs that are going to have people use the images, but it would be nice to also tell people what the tags mean.

TODO:
 - [x] Test out the travis config on my own repo
 - [ ] ~Cache the docker layers between travis builds~ I think this is best done in a follow-up.

Closes #81

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ env:
     - ATHENS_MONGO_CONNECTION_STRING=mongodb://127.0.0.1:27017
     - CODE_COV=1
     - ATHENS_DIR=${GOPATH}/src/github.com/gomods/athens
+    - REGISTRY=carolynvs/ # Temporary workaround until we have access to the gomods/ registry
 
 before_script:
   - mkdir -p $ATHENS_DIR && rsync -azr . $ATHENS_DIR && cd $ATHENS_DIR
@@ -30,3 +31,13 @@ after_success:
     else
       echo codecov not enabled;
     fi
+
+before_deploy:
+  - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+
+deploy:
+  provider: script
+  script: make docker-push
+  on:
+    repo: gomods/athens
+    all_branches: true
--- a/scripts/push-docker-images.sh
+++ b/scripts/push-docker-images.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+# Push our docker images to a registry
+set -xeuo pipefail
+
+REGISTRY=${REGISTRY:-gomods/}
+
+# Use the travis variables when available because travis clones different than what is on a local dev machine
+# VERSION = the tag if present, otherwise the short commit hash
+# BRANCH = the current branch, empty if not on a branch
+if [[ "${TRAVIS-}" == "true" ]]; then
+    VERSION=${TRAVIS_TAG:-${TRAVIS_COMMIT::7}}
+    BRANCH=${TRAVIS_BRANCH}
+else
+    TAG=$(git describe --tags --exact-match 2> /dev/null || true)
+    COMMIT=$(git rev-parse --short=7 HEAD)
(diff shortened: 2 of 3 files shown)


scripts/push-docker-images.sh
```diff
@@ -0,0 +1,33 @@
...
+REGISTRY=${REGISTRY:-gomods/}
+
+VERSION=$(git describe --tags --abbrev=7 --dirty)
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
+REPO_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." >/dev/null && pwd )/"
+
+# mutable tag is the docker image tag that we will reuse between pushes, it is not a stable tag like a SHA or version.
+MUTABLE_TAG=${MUTABLE_TAG:-} # defaulted below based on the branch
+if [[ "$MUTABLE_TAG" == "" ]]; then
+    case "BRANCH" in
```
mydiemho: should this be `$BRANCH` instead since `BRANCH` is a variable?

scripts/push-docker-images.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+# Push our docker images to a registry
+set -xeuo pipefail
+
+# Manually set the registry to a docker registry where you have push access to test this script.
+REGISTRY=${REGISTRY:-gomods/}
```
mydiemho: is there a pros/cons over using value without `/`?  I feel like I will miss the forwarding `/` when I'm setting it manually

scripts/push-docker-images.sh
```diff
@@ -0,0 +1,52 @@
...
+    echo "defaulting BRANCH using git..."
+    BRANCH=`git rev-parse --abbrev-ref HEAD`
+    if [[ "$BRANCH" == "HEAD" ]]; then
+        BRANCH=${VERSION} # Make our branch var work like travis_branch
+    fi
+fi
+
+# mutable tag is the docker image tag that we will reuse between pushes, it is not a stable tag like a SHA or version.
+MUTABLE_TAG=${MUTABLE_TAG:-} # defaulted below based on the branch
+if [[ -z "$MUTABLE_TAG" ]]; then
```
mydiemho: I prefer your original implementation using case statement.  I thought it was simpler that way

.travis.yml
```diff
@@ -13,6 +13,7 @@ env:
     - ATHENS_MONGO_CONNECTION_STRING=mongodb://127.0.0.1:27017
     - CODE_COV=1
     - ATHENS_DIR=${GOPATH}/src/github.com/gomods/athens
+    - REGISTRY=carolynvs/ # Temporary workaround until we have access to the gomods/ registry
```
carolynvs: As soon as we have travis's environment variables updated with our godmods account creds, we can switch this back to `gomods/`.

scripts/push-docker-images.sh
carolynvs: It's the convention that I've seen other repositories use. This let's you set REGISTRY to "" and then the docker image is just "proxy:latest" which is nice for local work.

Once this is in, we can follow-up with documentation for how to build and use the docker image locally in our DEVELOPMENT.md doc, in addition to explaining what the various docker image tags mean.

PR #546, opened 2018-08-21
put git and go in runtime image

joshgav:
The `goget` impl of the Download protocol requires both go 1.11 and git on the system. I updated the Dockerfile accordingly.

@marwan-at-work you removed much of the proxy Dockerfile, but without it my builds fail. Can you PTAL? Thanks!

### DIFF
--- a/cmd/olympus/actions/app.go
+++ b/cmd/olympus/actions/app.go
@@ -1,7 +1,9 @@
 package actions
 
 import (
+	"fmt"
 	stdlog "log"
+	"strings"
 
 	"github.com/gobuffalo/buffalo"
 	"github.com/gobuffalo/buffalo/middleware"
@@ -56,6 +58,9 @@ var (
 // This is the nerve center of your application.
 func App(config *AppConfig) (*buffalo.App, error) {
 	port := env.Port(":3001")
+	if hasColon := strings.HasPrefix(port, ":"); hasColon == false {
+		port = fmt.Sprintf(":%s", port)
+	}
 
 	w, err := getWorker(config.Storage, config.EventLog)
 	if err != nil {
--- a/cmd/proxy/Dockerfile
+++ b/cmd/proxy/Dockerfile
@@ -1,13 +1,28 @@
-FROM golang:1.11-rc
+FROM gobuffalo/buffalo:v0.12.6 as buffalo-builder
+
+RUN mkdir -p $GOPATH/src/github.com/gomods/athens/cmd/proxy
+WORKDIR $GOPATH/src/github.com/gomods/athens/cmd/proxy
+
+# this will cache the npm install step, unless package.json changes
+ADD cmd/proxy/package.json .
+ADD cmd/proxy/yarn.lock .
+RUN yarn install --no-progress
 
 RUN mkdir -p $GOPATH/src/github.com/gomods/athens
 WORKDIR $GOPATH/src/github.com/gomods/athens
 
-COPY . .
+ADD . .
+RUN cd cmd/proxy && buffalo build -s -o /bin/app
 
-RUN CGO_ENABLED=0 go build -o /bin/app ./cmd/proxy
+
+FROM golang:1.11-rc-alpine
+RUN apk add --no-cache bash git openssh ca-certificates
+
+COPY --from=buffalo-builder /bin/app /bin/app
 
 ENV GO_ENV=production
+# Bind the app to 0.0.0.0 (all interfaces) so it can be seen from outside the container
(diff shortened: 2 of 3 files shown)


cmd/proxy/Dockerfile
```diff
@@ -1,13 +1,30 @@
-FROM golang:1.11-rc
+FROM gobuffalo/buffalo:v0.12.6 as buffalo-builder
+
+RUN mkdir -p $GOPATH/src/github.com/gomods/athens/cmd/proxy
+WORKDIR $GOPATH/src/github.com/gomods/athens/cmd/proxy
+
+# this will cache the npm install step, unless package.json changes
+ADD cmd/proxy/package.json .
```
manugupt1: Why do we need yarn, package.json?

We do not have a web app for MVC.

cmd/proxy/Dockerfile
```diff
@@ -1,13 +1,30 @@
...
 RUN mkdir -p $GOPATH/src/github.com/gomods/athens
 WORKDIR $GOPATH/src/github.com/gomods/athens
 
-COPY . .
+ADD . .
+RUN cd cmd/proxy && buffalo build -s -o /bin/app
+
+
+FROM golang:1.11-rc-alpine
+RUN apk add --no-cache bash git openssh ca-certificates
```
manugupt1: If you just use golang:1.11-rc image, then you wont need to install git. I think golang:1.11-rc is ubuntu whereas alpine is closer to scratch.

https://hub.docker.com/_/golang/

cmd/proxy/Dockerfile
manugupt1: MVP**

joshgav: I see, we don't want any HTML/JS assets in either the proxy or Olympus? Do we still want to serve a basic HTML page from Go for verification purposes?

If we don't need HTML/JS, perhaps we can use `go build` instead of `buffalo build` - that's what @marwan-at-work was proposing I think.

cmd/proxy/Dockerfile
joshgav: Smaller is nice, but if we prefer Ubuntu I can change. Before it was straight Alpine. Thanks!

cmd/proxy/Dockerfile
manugupt1: Right! We can always hit the `healthz` endpoint to verify if everything is working

https://github.com/gomods/athens/blob/master/cmd/proxy/actions/health.go
https://github.com/gomods/athens/blob/d749465e35e5183cde0c1cfd89327ca3f1eadca6/cmd/proxy/actions/app_proxy.go#L18

As to the second part, the routers are still part of the `buffalo build`. There is a PR that 
removes pop dependency from buffalo build in proxy and allows buffalo build and buffalo test to run without setting up a database
https://github.com/gomods/athens/pull/511

I am not sure if we can really do go build as of now.

cmd/proxy/Dockerfile
manugupt1: yeah! I like smaller is nice and more secure :) 
I do not know what we prefer tbh! I will let @gomods/maintainers decide about that

marwan-at-work:
@joshgav why does your build fail? the `golang:1.11-rc` has **both** Go and Git. I've deployed it to GCP and it's been working. Curious what might be your edge case.

cmd/olympus/actions/app.go
```diff
@@ -56,6 +58,9 @@ var (
 // This is the nerve center of your application.
 func App(config *AppConfig) (*buffalo.App, error) {
 	port := env.Port(":3001")
+	if hasColon := strings.HasPrefix(port, ":"); hasColon == false {
```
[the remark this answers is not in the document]
devdinu: We could move this to config package validate and read int port from config, and return ":"+port after the merge of #453.

michalpristas:
From a discussion on slack, I guess we can settle on Debian.

cmd/olympus/actions/app.go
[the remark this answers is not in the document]
marwan-at-work: @devdinu I'd rather have the ":" mandatory to stay consistent. If there's a use case where someone just cannot pass a ":" then let's open an issue and check out that scenario :)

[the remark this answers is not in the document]
michalpristas: I'm with @marwan-at-work on this one

marwan-at-work:
Closing for now. @joshgav let me know if things don't work for you and we'll go from there. Thanks!

PR #550, opened 2018-08-22
Athens: add .DS_Store to gitignore

marwan-at-work:
The worst https://en.wikipedia.org/wiki/.DS_Store

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ tmp
 cmd/olympus/bin
 cmd/proxy/bin
 .idea
+.DS_Store


.gitignore
```diff
@@ -27,3 +27,4 @@ tmp
 .vs
 cmd/olympus/bin
 cmd/proxy/bin
+.DS_Store
```
michalpristas: what code editor do you use? 🆕 line

marwan-at-work: @michalpristas  I use VS-Code but it's been very slow and not working for me :( which is why you see .idea somewhere in my other pr lol.

marwan-at-work:
@michalpristas updated the pr. let me know if it's good to go :)

PR #553, opened 2018-08-22
Proxy: add cache-control middleware

marwan-at-work:
Certain Download Protocol handlers must be careful not to cache anything down the network. In particular, /list and /@latest because they always depend on upstream and are subject to change in real time. 

This PR adds a middleware to set custom Cache-Control header and adds it to the two handlers that need it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ tmp
 .vs
 cmd/olympus/bin
 cmd/proxy/bin
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -27,4 +27,3 @@ tmp
 .vs
 cmd/olympus/bin
 cmd/proxy/bin
-.idea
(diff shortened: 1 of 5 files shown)


pkg/download/handler.go
```diff
@@ -44,9 +45,14 @@ func RegisterHandlers(app *buffalo.App, opts *HandlerOpts) {
 	if opts == nil || opts.Protocol == nil || opts.Engine == nil || opts.Logger == nil {
 		panic("absolutely unacceptable handler opts")
 	}
+	noCacheMw := middleware.CacheControl("private, no-store")
```
manugupt1: This is a question, should we update to what MDN says about turning off completely?

and have 
```
Cache-Control: no-cache, no-store, must-revalidate
```

Ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

marwan-at-work: @manugupt1 "no-store" is usually good enough, but I'll keep things consistent with the MDN docs :) -- thanks for the call out!

.gitignore
```diff
@@ -27,3 +27,4 @@ tmp
 .vs
 cmd/olympus/bin
 cmd/proxy/bin
+.idea
```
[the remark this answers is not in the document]
michalpristas: remove before merging ;-)

PR #554, opened 2018-08-22
Fix running the athens proxy on windows

### LINKED ISSUE #532: Fix running the athens proxy on windows
This is a placeholder, Jeremy will fill in the stacktraces and error messages.

marpio:
Fixes #532 
Apparently on Windows `go mod` requires some additional ENV vars which we didn't pass so I'm using the parent environment and overriding the relevant ones.


pkg/module/go_get_fetcher.go
```diff
@@ -86,15 +86,10 @@ func getSources(goBinaryName string, fs afero.Fs, gopath, repoRoot, module, vers
...
-	gopathEnv := fmt.Sprintf("GOPATH=%s", gopath)
-	cacheEnv := fmt.Sprintf("GOCACHE=%s", filepath.Join(gopath, "cache"))
-	disableCgo := "CGO_ENABLED=0"
-	enableGoModules := "GO111MODULE=on"
-
 	cmd := exec.Command(goBinaryName, "mod", "download", fullURI)
 	// PATH is needed for vgo to recognize vcs binaries
 	// this breaks windows.
-	cmd.Env = []string{"PATH=" + os.Getenv("PATH"), gopathEnv, cacheEnv, disableCgo, enableGoModules}
+	cmd.Env = prepareEnv(gopath)
```
michalpristas: @marwan-at-work removed this while ago. 
marwan do you remember why?

pkg/module/go_get_fetcher.go
```diff
@@ -120,6 +115,38 @@ func getSources(goBinaryName string, fs afero.Fs, gopath, repoRoot, module, vers
...
+		"GOPATH",
+		"GOCACHE",
+		"CGO_ENABLED",
+		"GO111MODULE",
+		"GOPROXY",
+	}
+	parentEnv := os.Environ()
+	shouldRemove := func(parentEnvVar string) bool {
+		for _, n := range names {
+			if strings.Contains(parentEnvVar, n) {
```
michalpristas: why Contains? shouldnt it be `==`?

pkg/module/go_get_fetcher.go
```diff
@@ -120,6 +115,38 @@ func getSources(goBinaryName string, fs afero.Fs, gopath, repoRoot, module, vers
...
+func prepareEnv(gopath string) []string {
+	names := []string{
+		"GOPATH",
+		"GOCACHE",
+		"CGO_ENABLED",
+		"GO111MODULE",
+		"GOPROXY",
+	}
+	parentEnv := os.Environ()
+	shouldRemove := func(parentEnvVar string) bool {
```
michalpristas: can you name it shouldKeep or something like that and switch the logic down there?

pkg/module/go_get_fetcher.go
marpio: The environment var is something like 'GOPROXY=http://something' so it won't match with ==

michalpristas: oh i see. what about HasPrefix n+'='
i want to avoid false positives

marpio: But HasPrefix would make more sense probably.

marpio: Hehe you have been faster.

pkg/module/go_get_fetcher.go
```diff
@@ -86,15 +86,10 @@ func getSources(goBinaryName string, fs afero.Fs, gopath, repoRoot, module, vers
 
 	fullURI := fmt.Sprintf("%s@%s", uri, version)
 
-	gopathEnv := fmt.Sprintf("GOPATH=%s", gopath)
-	cacheEnv := fmt.Sprintf("GOCACHE=%s", filepath.Join(gopath, "cache"))
-	disableCgo := "CGO_ENABLED=0"
-	enableGoModules := "GO111MODULE=on"
-
 	cmd := exec.Command(goBinaryName, "mod", "download", fullURI)
 	// PATH is needed for vgo to recognize vcs binaries
```
robjloranger: Comment is no longer relevant with this change

robjloranger:
I would approve but don't have a windows machine to verify, just the one comment which should have been requesting changes as the `PATH` environment variable is no longer being set. The variable removal should be fine as we are no longer using vgo.

marpio:
@marwan-at-work you're completely right. The problem was that `go mod` didn't give me any error, just empty `pkg/mod` so I thought it will take ages before I find the missing env var.

pkg/module/go_get_fetcher.go
```diff
@@ -120,6 +112,25 @@ func getSources(goBinaryName string, fs afero.Fs, gopath, repoRoot, module, vers
...
+func prepareEnv(gopath string) []string {
+	pathEnv := fmt.Sprintf("PATH=%s", os.Getenv("PATH"))
+	gopathEnv := fmt.Sprintf("GOPATH=%s", gopath)
+	cacheEnv := fmt.Sprintf("GOCACHE=%s", filepath.Join(gopath, "cache"))
+	disableCgo := "CGO_ENABLED=0"
+	enableGoModules := "GO111MODULE=on"
+	cmdEnv := []string{pathEnv, gopathEnv, cacheEnv, disableCgo, enableGoModules}
+	// add Windows specific ENV VARS
+	// if this one is missing we will get an error from go
+	if val, ok := os.LookupEnv("USERPROFILE"); ok {
```
manugupt1: Is it a good idea to say to panic on windows if the env vars do not exist?

pkg/module/go_get_fetcher.go
```diff
@@ -120,6 +112,25 @@ func getSources(goBinaryName string, fs afero.Fs, gopath, repoRoot, module, vers
...
+		cmdEnv = append(cmdEnv, fmt.Sprintf("USERPROFILE=%s", val))
+	}
+	// if this one is missing go won't give any error but the pkg/mod dir will be empty
```
michalpristas: let end up with an error then, with empty packages proxy is not fulfilling its purpose

pkg/module/go_get_fetcher.go
robjloranger: For that one yes, or in some other way shutdown as it won't work without it.

marpio: But should we do this here? Maybe some check at the top of stack?

manugupt1: yep! I might have something in init() of proxy that determine if we have sane defaults to run the app?

manugupt1: Maybe a different PR?

jeremyrickard:
🙌 sweeeet

marwan-at-work:
Merging as everything seems addressed :)

PR #555, opened 2018-08-22
Athens: add .idea to gitignore

marwan-at-work:
just starting to use Goland as im struggling with vscode performacne, and it adds this .idea folder so I'm not sure if it's normal for people to add it or ignore it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ tmp
 .vs
 cmd/olympus/bin
 cmd/proxy/bin
+.idea


.gitignore
```diff
@@ -27,3 +27,4 @@ tmp
 .vs
 cmd/olympus/bin
 cmd/proxy/bin
```
devdinu: You could also add *.swp

marwan-at-work: @devdinu what's a .swp file? is this related to intellij? Thanks!

devdinu: folks using vim editor, its the temporary swap files.

PR #558, opened 2018-08-24
pkg/storage: ensure mongo returns KindNotFound

marwan-at-work:
In mongo.Zip, KindNotFound is not included which ended up making certain builds fail because we told the donload.Protocol it was an unexpected error, instead of a NotFound error so it ended up returning 500 to cmd/go instead of fetching the module from Upstream.

### DIFF
--- a/pkg/storage/mongo/getter.go
+++ b/pkg/storage/mongo/getter.go
@@ -3,8 +3,8 @@ package mongo
 import (
 	"context"
 	"io"
-	"strings"
 
+	"github.com/globalsign/mgo"
 	"github.com/globalsign/mgo/bson"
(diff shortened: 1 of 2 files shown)


pkg/storage/mongo/getter.go
```diff
@@ -57,7 +59,11 @@ func (s *ModuleStore) Zip(ctx context.Context, module, vsn string) (io.ReadClose
 	fs := s.s.DB(s.d).GridFS("fs")
 	f, err := fs.Open(zipName)
 	if err != nil {
-		return nil, errors.E(op, err)
+		kind := errors.KindUnexpected
```
michalpristas: is ok to return error not found for list instead of [], what DP expects?

PR #560, opened 2018-08-24
KindNotFound for all storages

michalpristas:
Making behavior consistent on top of that

### DIFF
--- a/pkg/storage/fs/lister.go
+++ b/pkg/storage/fs/lister.go
@@ -2,6 +2,7 @@ package fs
 
 import (
 	"context"
+	"os"
 
 	"github.com/gomods/athens/pkg/errors"
 	opentracing "github.com/opentracing/opentracing-go"
@@ -15,7 +16,12 @@ func (l *storageImpl) List(ctx context.Context, module string) ([]string, error)
 	loc := l.moduleLocation(module)
 	fileInfos, err := afero.ReadDir(l.filesystem, loc)
 	if err != nil {
-		return nil, errors.E(op, err)
+		kind := errors.KindUnexpected
+		if _, ok := err.(*os.PathError); ok {
+			kind = errors.KindNotFound
+		}
+
+		return nil, errors.E(op, err, kind)
 	}
 	ret := []string{}
 	for _, fileInfo := range fileInfos {
--- a/pkg/storage/fs/lister.go
+++ b/pkg/storage/fs/lister.go
@@ -17,7 +17,7 @@ func (l *storageImpl) List(ctx context.Context, module string) ([]string, error)
 	fileInfos, err := afero.ReadDir(l.filesystem, loc)
 	if err != nil {
 		kind := errors.KindUnexpected
-		if _, ok := err.(*os.PathError); ok {
+		if os.IsNotExist(err) {
 			kind = errors.KindNotFound
 		}
 
--- a/pkg/storage/fs/lister.go
+++ b/pkg/storage/fs/lister.go
@@ -21,7 +21,7 @@ func (l *storageImpl) List(ctx context.Context, module string) ([]string, error)
 			kind = errors.KindNotFound
 		}
 
-		return nil, errors.E(op, err, kind)
+		return nil, errors.E(op, errors.M(module), err, kind)
 	}
 	ret := []string{}
 	for _, fileInfo := range fileInfos {
--- a/pkg/storage/gcp/getter.go
+++ b/pkg/storage/gcp/getter.go
@@ -31,7 +31,7 @@ func (s *Storage) Info(ctx context.Context, module, version string) ([]byte, err
 	infoBytes, err := ioutil.ReadAll(infoReader)
 	infoReader.Close()
 	if err != nil {
-		return nil, err, errors.M(module), errors.V(version))
+		return nil, errors.E(op, err, errors.M(module), errors.V(version))
 	}
 	return infoBytes, nil
 }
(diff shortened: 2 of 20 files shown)


pkg/storage/minio/getter.go
```diff
@@ -49,7 +49,13 @@ func (v *storageImpl) Zip(ctx context.Context, module, vsn string) (io.ReadClose
 	const op errors.Op = "minio.Zip"
 	sp, ctx := opentracing.StartSpanFromContext(ctx, "storage.minio.Zip")
 	defer sp.Finish()
+
 	zipPath := fmt.Sprintf("%s/source.zip", v.versionLocation(module, vsn))
+	_, err := v.minioClient.StatObject(v.bucketName, zipPath, minio.StatObjectOptions{})
+	if err != nil {
+		return nil, errors.E(op, err, errors.KindNotFound)
```
marpio: Could you add `errors.M(mod), errors.V(vsn)` ?

pkg/storage/minio/lister.go
```diff
@@ -29,7 +30,12 @@ func (l *storageImpl) List(ctx context.Context, module string) ([]string, error)
 			dict[ver] = struct{}{}
 		}
 	}
+
 	ret := []string{}
+	if len(dict) == 0 {
+		return ret, errors.E(op, fmt.Errorf("Module %s not found", module), errors.KindNotFound)
```
marpio: Could you use `errors.M(module)` instead of `fmt.Errorf("Module %s not found", module)`?

pkg/storage/fs/lister.go
```diff
@@ -15,7 +16,12 @@ func (l *storageImpl) List(ctx context.Context, module string) ([]string, error)
 	loc := l.moduleLocation(module)
 	fileInfos, err := afero.ReadDir(l.filesystem, loc)
 	if err != nil {
-		return nil, errors.E(op, err)
+		kind := errors.KindUnexpected
+		if _, ok := err.(*os.PathError); ok {
+			kind = errors.KindNotFound
+		}
+
+		return nil, errors.E(op, err, kind)
```
marpio: Could you add `errors.M(module)`?

pkg/storage/mongo/lister.go
```diff
@@ -27,6 +28,10 @@ func (s *ModuleStore) List(ctx context.Context, module string) ([]string, error)
 	}
 
 	versions := make([]string, len(result))
+	if len(result) == 0 {
+		return versions, errors.E(op, fmt.Errorf("Module %s not found", module), errors.KindNotFound)
```
marpio: Could you use `errors.M(module)` instead?

michalpristas:
revisited all errors returned from storage, associated module and version with each of them because:
- log search will be easier for related blobs
- the error might not give the detail about the blob

pkg/storage/fs/lister.go
```diff
@@ -15,7 +16,12 @@ func (l *storageImpl) List(ctx context.Context, module string) ([]string, error)
 	loc := l.moduleLocation(module)
 	fileInfos, err := afero.ReadDir(l.filesystem, loc)
 	if err != nil {
-		return nil, errors.E(op, err)
+		kind := errors.KindUnexpected
+		if _, ok := err.(*os.PathError); ok {
```
marwan-at-work: would ~`os.IsNotExist` work here?

pkg/storage/minio/getter.go
```diff
@@ -49,7 +49,13 @@ func (v *storageImpl) Zip(ctx context.Context, module, vsn string) (io.ReadClose
 	const op errors.Op = "minio.Zip"
 	sp, ctx := opentracing.StartSpanFromContext(ctx, "storage.minio.Zip")
 	defer sp.Finish()
+
 	zipPath := fmt.Sprintf("%s/source.zip", v.versionLocation(module, vsn))
+	_, err := v.minioClient.StatObject(v.bucketName, zipPath, minio.StatObjectOptions{})
```
marwan-at-work: is there no explicit check for NotFound? What if this was a connection error or something?

michalpristas: yes, but at the point you read the blob. you can see that at the other places here. but in case of zip we don't want to read it, we need client to read it

pkg/storage/fs/lister.go
michalpristas: yes, it can, but at this point, i wanted to check errors from Open specifically which are wrapped into PathError. 
What IsNotExist does is that it compares error msg of the error. But i'm ok with isNotExist if you prefer this way

marwan-at-work: @michalpristas as long as PathError _always_ means NotFound, I'm cool with either one. Thanks!

pkg/storage/gcp/getter.go
```diff
@@ -43,20 +43,20 @@ func (s *Storage) GoMod(ctx context.Context, module, version string) ([]byte, er
...
 	modReader, err := s.bucket.Open(ctx, config.PackageVersionedName(module, version, "mod"))
 	if err != nil {
-		return nil, errors.E(op, err)
+		return nil, errors.E(op, err, errors.M(module), errors.V(version))
 	}
 	modBytes, err := ioutil.ReadAll(modReader)
 	modReader.Close()
 	if err != nil {
-		return nil, errors.E(op, fmt.Errorf("could not get new reader for mod file: %s", err))
+		return nil, errors.E(op, fmt.Errorf("could not get new reader for mod file: %s", err), errors.M(module), errors.V(version))
```
robjloranger: I know this is old but you could probably remove the formatted error for consistency, the other calls no longer do this.

PR #561, opened 2018-08-24
Makefile: Add minio to dev target

### LINKED ISSUE #476: Tidy up Makefile
continuing my verbose response for #452  https://github.com/gomods/athens/issues/452#issuecomment-412725709

---

The Makefile should not call the target `dev` during `setup-dev-env`, it is insufficient for running unit tests or buffalo tests. <strike>The target `dev` is not used anywhere, so could be removed.</strike> see: https://github.com/gomods/athens/issues/476#issuecomment-412730391

The target `down` is identical to `dev-teardown`, since `dev-teardown` is analogous to `dev` I propose we remove `down`, rename `alldeps` to `dev` and update docs to reflect.

https://github.com/gomods/athens/blob/5a445aacf8bafc32bb40d9245121979ad8059a6e/Makefile#L15-L17
https://github.com/gomods/athens/blob/5a445aacf8bafc32bb40d9245121979ad8059a6e/Makefile#L69-L71
https://github.com/gomods/athens/blob/5a445aacf8bafc32bb40d9245121979ad8059a6e/Makefile#L74-L76

robjloranger:
This adds minio to the makefile target `dev`. This is because unit tests require minio to run.

closes #476

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,10 +60,11 @@ alldeps:
 	docker-compose -p athensdev up -d minio
 	docker-compose -p athensdev up -d jaeger
 	echo "sleeping for a bit to wait for the DB to come up"
-	sleep 5	
+	sleep 5
 
 .PHONY: dev
 dev:
+	docker-compose -p athensdev up -d minio
 	docker-compose -p athensdev up -d mongo
 	docker-compose -p athensdev up -d redis
 
--- a/Makefile
+++ b/Makefile
@@ -67,10 +67,6 @@ dev:
 	docker-compose -p athensdev up -d mongo
 	docker-compose -p athensdev up -d redis
 
-.PHONY: down
-down:
-	docker-compose -p athensdev down -v
-
 .PHONY: dev-teardown
 dev-teardown:
 	docker-compose -p athensdev down -v
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -46,7 +46,7 @@ it easy to get up and running:
 That's it! After the `make dev` command is done, everything will be up and running and you can move
 on to the next step.
 
-If you want to stop everything at any time, run `make down`.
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -60,17 +60,14 @@ alldeps:
 	docker-compose -p athensdev up -d minio
 	docker-compose -p athensdev up -d jaeger
 	echo "sleeping for a bit to wait for the DB to come up"
-	sleep 5	
+	sleep 5
 
 .PHONY: dev
 dev:
+	docker-compose -p athensdev up -d minio
```
michalpristas: so now the only thing differing alldeps and dev is jeager? does it make sense

robjloranger: Not really, but you can't run unit tests without minio so maybe we should just rename `alldeps` to `dev` and remove the old `dev`.

robjloranger:
whomp whomp, this is already covered in docs that one should run `make alldeps` before unit tests.. https://github.com/gomods/athens/blob/master/DEVELOPMENT.md#run-unit-tests

Makefile
michalpristas: it makes sense to me as well, @marwan-at-work had some strong opinions against it

marwan-at-work: Actually now that I think about it, we don't even need redis in dev 😄 -- which we can replace with minio, but from the slack discussion it seemed that peeps are happy making minio not a dependency of tests and therefore just mongo in `dev` would be enough in the future. We can add minio here as a temporary thing until tests are cleaned up. 

The only reason I want dev to be minimal is for *fast* iteration on hitting cacheFill every single time. In otherwords, cleaning the storage which we can achieve in other ways (would be more efficient)

robjloranger: Let's revisit when we get around to fixing up tests, now that our Docker builds are moving.

PR #569, opened 2018-08-26
No conflicts

### LINKED ISSUE #273: Add build step to prevent merging conflicts
conflicts = the artifacts marking the conflicts. ie:
```
<<<<<< HEAD
```

A grep in bash should do the trick, like `^[<>=]{7}` would check for lines beginning with any of the `<<<<<<<`, `=======`, `>>>>>>>`

This way the build would fail when these are hiding somewhere in a toml or other 'diff collapsed by default' file.

robjloranger:
This new script returns exit code 1 when any files other than '*.go,go.mod,go,sum' contain merge conflict artifacts. `<<<<<<<HEAD`, `=======` or `>>>>>>>branch-name`.

It skips the mentioned files as those will be caught during linting and dependency checks.

closes #273

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ Run `make test-e2e` to run the end-to-end tests.
 
 The first time you run the tests,
 you must run `make setup-dev-env` first, otherwise you will see errors like the one below:
-<<<<<<<HEAD
+
 ```
 error connecting to storage (no reachable servers)
 ```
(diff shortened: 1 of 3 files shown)


marpio:
Wouldn't the build fail anyway if .go files contain merge conflict artifacts? I guess this shouldn't compile.

michalpristas:
@marpio this should check files like *.toml, *.yml and docs

marpio:
Oh, I've missed the *.go exclude. :+1:

robjloranger:
@arschles it's included in the `make verify` target which CI is using.

arschles:
ohhhh, got it @robjloranger - looks great! I'm waiting on CI, then will merge

PR #572, opened 2018-08-27
Proxy: return 200 on / regardless of prefix

marwan-at-work:
This is a little annoying, but some Ingress Controllers (such as GCP Load Balancer) can not send custom headers and therefore if the proxy is running behind a prefix as well as some authentication mechanism, we should allow the plain / to return 200. Which some argue that / should always return 200 regardless of auth or prefixes.


robjloranger:
Should we then remove the `/healthz` endpoint? Or does it make sense to keep both in the proxy?

marwan-at-work:
@robjloranger I'd say let's keep it for now. Because that one can be the one that actually checks dependencies and returns a "true" health check.

Issue #308, opened 2018-07-19
Deduplicate cache fill requests to avoid cache stampede in proxy


Kunde21:
Large and heavily-used projects will have users with automated update mechanisms, leading to a flood of requests for the newest tag.  Proxy & Olympus need to be able to identify duplicate requests and only fill cache from VCS once.  

Initially, we can simply block all duplicate requests until the module is downloaded (#290), but this will become more complicated when working with streams (#281).

arschles:
Totally agree. Taking inspiration from https://github.com/golang/groupcache, I have a rough proposal (which I'm pretty sure needs work!).

The basic idea is that we build a distributed lease on which any process can watch for release events. Then, when a server receives a `go get module@version` request:

1. Check for `module@version` in storage. If it exists, return it and stop. If not, continue
2. Attempt to acquire lease for `module@version`
3. If it's acquired, fetch `module@version`, return it to client, fill storage, release lease and return
4. If it's not acquired, subscribe to the release event for `module@version`. When it fires, go to 1

These are leases and not locks, so they should have timeouts. So a few notes:

- If we get to (4) and the release event fires, `module@version` might not have been storage (if the process that had the lease failed)
- We should have a maximum number of retries until failure because (4) -> (1) could happen forever if things keep failing

Kunde21:
I'm hesitant to put the fetch logic on the request side of the interaction.  If we keep the fetch on the storage side, we can keep the retry logic there.  That also allows us to register failed requests to a background worker that performs the retry with back-off and/or fill from elsewhere in the athens system.  

As a wrapper around the storage backend, the "cache-warmer" would implement `storage.Lister` and `storage.Getter` interfaces with ( "or timeout" refers to received context cancellation/timeout):

1. Check for `module@version` cache fill requests.  If it exists, return result when fill request completes (or timeout).
2. If cache fill request doesn't exist, check for `module@version` in storage backend.  If it exists, return the module in storage. 
3. If module doesn't exist in storage, register cache fill request and return result when fill request completes (or timeout).
4. Cache-warmer attempts to fetch module (up to `N` retries - configurable).  If successful, return module and put in storage.
5. On failure, return error to all requests waiting on that module.  Cache fill request can be dropped, added to back-off list with deadline, added to async fill queue, etc

All of this can live between the storage backend interface and the selected backend, giving us flexibility to change how we fill the cache and return results.  

[...]

PR #573, opened 2018-08-27
Athens: introduce Single Flight

marwan-at-work:
This PR introduces single flight, but I had to do some refactoring along side this PR to make sure we don't accrue potential technical debt. 

Therefore, this PR
1. Introduces a new Stasher interface. Its job is to "stash" things. We've been calling this cacheFill but it's technically incorrect since there's no TTL. So the method Stash is short and descriptive enough of the process: which is to move something from Upstream and stashes it into its own Storage. 

2. Separate the Worker Pools into two: one for processing Protocol requests and one for processing "go mod download" requests. This is because we can serve a higher number of protocol requests than a number of module downloads. 

3. Introduce the concept of Single Flight when stashing a module: make sure you don't stash the same thing twice at the same time. 

Fixes https://github.com/gomods/athens/issues/308


cmd/olympus/actions/app.go
```diff
@@ -3,6 +3,8 @@ package actions
 import (
 	stdlog "log"
 
+	"github.com/gomods/athens/pkg/stash"
+
```
michalpristas: newline pls

cmd/proxy/actions/app_proxy.go
```diff
@@ -18,11 +23,41 @@ func addProxyRoutes(
...
+	// by the functionality: a download.Protocol that just takes care
+	// of "go getting" things, and another Protocol that just takes care
+	// of "pooling" requests etc.
+
+	// In our case, we'd like to compose both interfaces in a particular
+	// order to ensure logical ordering of execution.
+
+	// Here's the order of an incoming request to the download.Protocol:
+
+	// 1. The downloadpool gets hit first, and manages concurrent requests
```
michalpristas: I find it overly complicated, do you find it readable?

marwan-at-work: @michalpristas I 100% agree that it's more complex. I also agree that it's *less* readable than what we currently have, but that comes with the complexity part. 

My first option was just to lump all of this code into the `download.go` file and that would have been much much harder to read because you'd have **two worker pools* and a *singleflight* all under the same struct. I don't mind returning to that option if you want? Otherwise, let me know if you have other ideas, totally happy to change things :) 

Here's what's required: 
1. 2 pools (one for download.Protocol, one for go mod download) 
2. one singleflight for the operation of "go mod download" and "storage.Save"

michalpristas: I imagined something simpler e.g (and I am totally not thinking this through, just writing an idea as it goes)
- the request comes to get/list/whatever - no handling of concurrency just let it come/fail
- from request, you call X to spin up worker
- X returns a channel C which will be closed once the module is prepared in Store
- X checks store for module.v, closes C if it does
- X checks whether work for module.v is already queued
-  if not it creates a job with a channel passed into goroutine and stored in a local `map[module.v]chan`. Returns chan
- if c,ok := map[module.v]; ok; return c
- spinned up job performs go get, storage.Save, close(C)
- request waits on <-C, and then serves stored module

marwan-at-work: @michalpristas 
1. The request does come to the handler and the handler doesn't know anything about underlying concurrency, including all the download protocols that are not concurrent (i.e. goget)
2. What is X, is it all 3 requirements together? 
3. From the rest of those points, there's a missing requirement which is separating the download.Protocol Pool and the goget/storage.Save Pool. How would you split those up? 
4. I'm kind of doing everything you're saying, but more separated out than all in one file: X. I'm happy to put them all in X if you'd like :)

pkg/stash/stash.go
```diff
@@ -0,0 +1,46 @@
...
+	"context"
+	"time"
+
+	"github.com/gomods/athens/pkg/download"
+	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/storage"
+)
+
+// Stasher has the job of taking a module
+// from its upstream Sownload Protcool and
```
michalpristas: typo

cmd/proxy/actions/app_proxy.go
michalpristas: I dismissed my blocking review. if folks are ok with this design I will not block and will think this through if needed once I'm mentally capable of thinking.

marwan-at-work: @michalpristas no I 100% agree with you. I think we shouldn't rush this PR since it adds a lot of complexity. I'll mark this as do not merge until we have a bunch of people weighing in and maybe proposing different approaches. One idea is that I can have some of the Athens peeps at GopherCon doing a group review and brainstorm some other stuff and bringing it back here. I even don't mind opening another PR (or someone else opening another PR) that does the same thing differently so we can really compare. Thank you!

michalpristas: Hey if you manage to get people together and brainstorm it would be great ;-)

pkg/download/stasher/stasher.go
```diff
@@ -0,0 +1,132 @@
...
+		info, err = p.s.Info(ctx, mod, ver)
+	}
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+
+	return info, nil
+}
+
+func (p *protocol) Stash(mod, ver string) error {
```
marpio: Why do we need this func?

marwan-at-work: @marpio good catch, we don't need it. I originally made it like that but it made wrapping it with "pooling" and "singleflight" harder. I'll remove it.

pkg/download/stasher/stasher.go
```diff
@@ -0,0 +1,115 @@
+package stasher
```
marpio: Could we rename this pkg? Much of the difficulties I had reading this was due to this name. Like, it contains an impl. of a **protocol** which 'stashes' but the name is exactly as the `stash.Stasher` interface.

marwan-at-work: @marpio very good point. Any idea what we should call it instead? 
I wanted to call it the "main" protocol, but obviously that name is taken lol.

marpio: Yeah, naming is hard ;-)
But maybe something like this:
Separate the DP implementations into two categories
 - one that provides the core functionality which is to get the  required bits (mod, info, zip) from upstream. This would be i.e. implementation in the `download/goget` pkg or, in future, one that uses Olympus.
- and the wrappers/addons which provide some additional functionality like stash, worker pool, retry etc. So something like `protocol/addons` with functions like `func WithStash(p download.Protocol, ...)download.Protocol`

cmd/proxy/actions/app_proxy.go
```diff
@@ -18,11 +23,41 @@ func addProxyRoutes(
...
 		return err
 	}
-	p := download.New(gg, s, env.GoGetWorkers())
+	st := singleflight.New(
+		stashpool.New(
+			stash.New(gg, s),
+			env.GoGetWorkers(),
+		),
+	)
+	p := downloadpool.New(
```
marpio: I don't entirely understand why we need the `downloadpool`. Could you expand on that? Like, what would happen if we wouldn't use it at all  and just pass `p := stasher.New(gg, s, st)` to the handler opts?

cmd/proxy/actions/app_proxy.go
```diff
@@ -18,11 +23,41 @@ func addProxyRoutes(
...
+		stasher.New(gg, s, st),
+		env.GoGetWorkers(),
```
marpio: Shouldn't this be `env.ProtocolWorkers()` ?

pkg/download/stasher/stasher.go
marwan-at-work: @marpio i've seen this pattern with gRPC and I like it. But I still don't know what to name the "main" one that uses goget as an upstream and storage as a first stop and a stasher as an intermediary step :)

marwan-at-work: @marpio I can have the addons package include the main one and just call it "WithStasher"?

cmd/proxy/actions/app_proxy.go
marwan-at-work: @marpio Because we need to be able to control how many concurrent connections to storage we can open as well as how much data we copy from the storage through the proxy and back to the client.

pkg/download/stasher/stasher.go
marpio: Yes, this sounds good!

cmd/proxy/actions/app_proxy.go
marwan-at-work: @marpio totally, the connections is not the issue but how much copying do you do from those connections. In other words, the memory part is the issue that I ran into :)

michalpristas:
I've spent some time over the piece of paper today and this is my proposal this will be very long and confusing
you will have
stasher.go
- implements Stasher
- New() this will return a func(func (context.Context, mod, ver string) (*storage.Version, error))Stasher
- basically returns a func which accepts dp.Version and returns a Stasher 
- provides basic Stash method which calls dp.Version
Example for basic stasher
```go
func New() func(func(context.Context, string, string) (*storage.Version, error)) Stasher {
  return func(f func(context.Context, string, string) (*storage.Version, error)) Stasher {
	return &stasher{s, f}
  }
}
```

Example for wrapper stasher
```go
func New(baseStasher Stasher) func(func(context.Context, string, string) (*storage.Version, error)) Stasher {
  return func(f func(context.Context, string, string) (*storage.Version, error)) Stasher {
    return &stasher{baseStasher, s}
  }
}
```

`goget.go` - will implement DP 
- New(func(func(context.Context, string, string) (*storage.Version, error)) Stasher) 
   - this will create a goget getter and instantiate own stasher with its own Version method
- when you call gg.List => gg.dp.list()
- when you call gg.Get => gg.stasher.Stash() ... similar to your stash.go

app_proxy.go will look like this
```go
stasher := stasher.Deduplicated(
    stasher.Pooled(
        stasher.New()
    )
)
gg := goget.New(stasher)
downloadpool.New(gg, N)
```

I dont think it is readable as a breeze but it separates Stashers from DP a bit in a sense that Stash calls are happening inside goget. If you dont want to use stasher you just provide a Stash builder method which calls dp.Version as is.
Other implementations of stasher (Deduplicated, Pooled), are treated as wrappers [other possible naming is WithDeduplication, WithPool]

so file structure may be like this
```
- download\
    - goget\
    - stasher\
        - pooled.go
        - deduplicated.go
        - stash.go // basic DP => Stasher
        - raw.go  // no stashing direct call
    - pool
        - pool.go
```

marwan-at-work:
@michalpristas I'm gonna apply @marpio's comments first. Then see how your comments fit in to the PR. Quick Q: what does `Deduplicated` do? And do we still have a stash.Stasher interface as well?

marwan-at-work:
@michalpristas could you please write an example of a download.Protocol method (i.e. GoMod or Zip) stash a module with your approach? Thanks!!!

marwan-at-work:
@michalpristas @marpio I just updated the code with @marpio 's suggestions, feel free to take another look to keep Michal's suggestions in the context of the latest changes. 

Thankssss

michalpristas:
@marwan-at-work Deduplicated means singleflight. the approach where you remember input-output
Also will try to write some sample, but i will do it probably on sunday. i am on vacation for 3 day starting today and will not have much time during these days

marwan-at-work:
@michalpristas i'd prefer to call it Singleflight because it's a more descriptive/well-known term :) 

but yeah that sounds good! Take your time :) 

@fedepaol i agree, it's looking more reasonable to read after @marpio 's suggestions.

marwan-at-work:
I think in a separate PR (or even this one), I can pass the GoGetFetcher instead of a download.Protocol to the stash.Stasher. This might make it even more readable. I think @marpio suggested that in another PR.

marwan-at-work:
removing the do not merge since there's been a bunch of :eyes: on it already. But won't merge until it looks good to everyone

cmd/proxy/actions/app_proxy.go
```diff
@@ -18,11 +20,36 @@ func addProxyRoutes(
...
+	// 1. The singleflight picks up the first request and latches duplicate ones.
+	// 2. The singleflight passes the stash to its parent: stashpool.
+	// 3. The stashpool manages limiting concurrent requests and passes them to stash.
+	// 4. The plain stash.New just takes a request from upstream and saves it into storage.
 	gg, err := goget.New()
 	if err != nil {
 		return err
 	}
-	p := download.New(gg, s, env.GoGetWorkers())
+	st := stash.New(gg, s, stash.WithSingleflight, stash.WithPool(env.GoGetWorkers()))
```
marpio: @marwan-at-work do you think it would make sense to split the DP interface into smaller ones or just pass functions into `stash.New(...` and `addons.WithStasher(...)`? I'm asking because it's a little confusing to me when I see that `stash.New` takes DP interface because I know it only needs one function - `Version`
The same with `addons.WithStasher` - it wasn't clear to me why it needs the DP at all. Then I looked up and it's obviously needed for `List` and `Latest` but again, I had to look up deeper to see what's going on.
Let me know what you think.

marwan-at-work: @marpio updated the PR. Instead of creating yet another type, we can just use the `module.Fetcher` interface. However, that interface can use some refactoring to make this PR cleaner. But I don't want to go that far. In another PR, I would make the module.Fetcher.Fetch method just immediately return a `*storage.Version` instead of having it be two steps (Fetch and Ref)

marpio: :+1:  I was also thinking about pulling `module.Fetcher' out of goget

michalpristas:
@marwan-at-work @marpio I created not really buildable PR #624 
It is an example of the proposal I wrote earlier.

cmd/proxy/actions/app_proxy.go
```diff
@@ -18,12 +21,37 @@ func addProxyRoutes(
...
+	fs := afero.NewOsFs()
+	mf, err := module.NewGoGetFetcher(goBin, fs)
 	if err != nil {
 		return err
 	}
-	p := download.New(gg, s, env.GoGetWorkers())
-	opts := &download.HandlerOpts{Protocol: p, Logger: l, Engine: proxy}
+	st := stash.New(mf, s, stash.WithSingleflight, stash.WithPool(env.GoGetWorkers()))
+	dp := download.New(s, st, goBin, fs)
+	dp = addons.WithPool(dp, env.ProtocolWorkers())
```
marpio: To be consistent here we could either do this:
`dp := download.New(s, st, goBin, fs, addons.WithPool(env.ProtocolWorkers())`
or
`st := stash.WithPool(stash.WithSingleflight(stash.New(mf, s), env.GoGetWorkers())`

marwan-at-work:
@michalpristas @marpio the PR is updated. Let me know if there's anything still missing or different from what you have suggested. I bet there's further refactoring we could do but this PR is already bigger than what I would have liked it to be. Thank you both :)

pkg/download/addons/with_pool.go
```diff
@@ -0,0 +1,121 @@
...
+	"io"
+
+	"github.com/gomods/athens/pkg/download"
+	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/storage"
+)
+
+type withpool struct {
+	dp download.Protocol
+	ch chan func()
```
michalpristas: can you rethink the name? what is ch doing signaling done or accepting work?
from looking at the code below I can read that, but it would be nice to know this straight away

pkg/download/addons/with_pool.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (p *withpool) List(ctx context.Context, mod string) ([]string, error) {
+	const op errors.Op = "pool.List"
+	var vers []string
+	var err error
+	done := make(chan struct{}, 1)
+	p.ch <- func() {
+		vers, err = p.dp.List(ctx, mod)
+		done <- struct{}{}
```
michalpristas: close(done)

pkg/download/addons/with_pool.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (p *withpool) Info(ctx context.Context, mod, ver string) ([]byte, error) {
+	const op errors.Op = "pool.Info"
+	var info []byte
+	var err error
+	done := make(chan struct{}, 1)
+	p.ch <- func() {
+		info, err = p.dp.Info(ctx, mod, ver)
+		done <- struct{}{}
```
michalpristas: same: close(done), struct{}{} is ugly

pkg/download/protocol.go
```diff
@@ -26,95 +34,121 @@ type Protocol interface {
...
-// New takes an upstream Protocol and storage
-// it always prefers storage, otherwise it goes to upstream
-// and fills the storage with the results.
-func New(dp Protocol, s storage.Backend, workers int) Protocol {
-	ch := make(chan *job)
-	p := &protocol{dp: dp, s: s, ch: ch}
-	p.start(workers)
+// New returns a full implementation of the download.Protocol
+// that the proxy needs.
+func New(opts *Opts, wrappers ...Wrapper) Protocol {
```
michalpristas: from this, it's not clear which wrapper of passed wrappers is outer most, 
decorators are one option, adding outer most info to docs another one

pkg/download/protocol.go
```diff
@@ -26,95 +34,121 @@ type Protocol interface {
...
+	stasher   stash.Stasher
+	goBinPath string
+	fs        afero.Fs
 }
 
-func (p *protocol) listen() {
-	for j := range p.ch {
-		j.done <- p.fillCache(j.mod, j.ver)
+func (p *protocol) List(ctx context.Context, mod string) ([]string, error) {
+	const op errors.Op = "goget.List"
```
michalpristas: protocol.List

pkg/download/protocol.go
```diff
@@ -26,95 +34,121 @@ type Protocol interface {
...
+	return lr.Versions, nil
 }
 
-func (p *protocol) request(mod, ver string) error {
-	j := &job{
-		mod:  mod,
-		ver:  ver,
-		done: make(chan error),
+func (p *protocol) Latest(ctx context.Context, mod string) (*storage.RevInfo, error) {
+	const op errors.Op = "goget.Latest"
```
michalpristas: protocol.Latest

pkg/download/protocol.go
```diff
@@ -26,95 +34,121 @@ type Protocol interface {
...
 
-	return nil
+	return &lr, nil
 }
 
-func (p *protocol) Latest(ctx context.Context, mod string) (*storage.RevInfo, error) {
-	const op errors.Op = "protocol.Latest"
-	info, err := p.dp.Latest(ctx, mod)
+func (p *protocol) Info(ctx context.Context, mod, ver string) ([]byte, error) {
+	const op errors.Op = "stasher.Info"
```
michalpristas: protocol.Info

pkg/download/protocol.go
```diff
@@ -123,10 +157,10 @@ func (p *protocol) Latest(ctx context.Context, mod string) (*storage.RevInfo, er
 }
 
 func (p *protocol) GoMod(ctx context.Context, mod, ver string) ([]byte, error) {
-	const op errors.Op = "protocol.GoMod"
+	const op errors.Op = "stasher.GoMod"
```
michalpristas: protocol.GoMod

pkg/download/protocol.go
```diff
@@ -140,10 +174,10 @@ func (p *protocol) GoMod(ctx context.Context, mod, ver string) ([]byte, error) {
 }
 
 func (p *protocol) Zip(ctx context.Context, mod, ver string) (io.ReadCloser, error) {
-	const op errors.Op = "protocol.Zip"
+	const op errors.Op = "stasher.Zip"
```
michalpristas: protocol.Zip

pkg/stash/with_pool.go
```diff
@@ -0,0 +1,51 @@
+package stash
+
+import (
+	"github.com/gomods/athens/pkg/errors"
+)
+
+type withpool struct {
+	s  Stasher
+	ch chan func()
```
michalpristas: same comment as with download pool, what's the purpose of the chan

pkg/stash/with_singleflight.go
```diff
@@ -0,0 +1,56 @@
...
+	sf.subs = map[string][]chan error{}
+
+	return sf
+}
+
+type withsf struct {
+	s Stasher
+
+	mu   sync.Mutex
+	mp   map[string]struct{}
```
michalpristas: what is mp?

pkg/stash/stasher.go
```diff
@@ -0,0 +1,62 @@
...
+type stasher struct {
+	f module.Fetcher
+	s storage.Backend
+}
+
+func (s *stasher) Stash(mod, ver string) error {
+	const op errors.Op = "stasher.Stash"
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*10)
+	defer cancel()
+	v, err := s.fetchModule(ctx, mod, ver)
```
michalpristas: can we check if mod-ver is present.
what if call to singleflight.Stash comes for m,v at the time when it is already fetched and subscribers are notified.
after subs are notified, lock is released and stash continue with m,v with fetching because it finds out that nobody is working on m,v and it reported a cache miss
maybe

pkg/download/addons/with_pool.go
marwan-at-work: since this is a worker pool, I think `jobCh` makes sense. Would that be ok? I can also add some internal docs for it :)

pkg/stash/with_singleflight.go
marwan-at-work: I'll keep mp for short since it's unexported. But I'll add a comment explaining why it's there.

marwan-at-work: it's a way to check if a request is currently in-flight.

marwan-at-work: Actually I can remove it and just use the singleflight map as a way to check an inflight request...I'll update this PR and you can tell me if it's more/less readable and I can revert that commit accordingly.

pkg/stash/stasher.go
marwan-at-work: @michalpristas that shouldn't happen (AFAIK). Because any request that comes in will *either* be part of the singleflight or will be gotten from storage when the download.Protocol checks. 
I can write a test (in another PR) that ensures a Stash is never hit more than once.

marwan-at-work:
@michalpristas comments addressed. Feel free to take a look and do further suggestions :) thank you a ton!

PR #576, opened 2018-08-27
Update documentation about running unit tests

artmello:


DEVELOPMENT.md
```diff
@@ -89,11 +89,10 @@ In order to run unit tests, services they depend on must be running first:
 make alldeps
 ```
 
-and database created:
+and database environment variable needs to be set:
 
 ```console
-buffalo db create
-buffalo db migrate up
+export ATHENS_MONGO_CONNECTION_STRING=mongodb://127.0.0.1:27017
```
marpio: Do we need that? It is defined in the `.env` file here: https://github.com/gomods/athens/blob/master/cmd/proxy/.env so buffalo dev should use it.

artmello: Honestly, I don't know. I didn't execute the `buffalo dev` (maybe I should?) step and after `make alldeps` to `make test-unit`. I got errors during test execution like: 
```
error getting mongo uri (missing mongo connection string: could not find ENV var with ATHENS_MONGO_CONNECTION_STRING)
```

robjloranger: I was wondering about that, I think I mentioned before somewhere.. I forget where, but the config package should probably default to the localhost mongo URI when not set explicitly.

marpio: @artmello I think we can delete the whole database section. https://github.com/gomods/athens/pull/578 should fix the problem you encountered.

robjloranger:
Also see https://github.com/gomods/athens/issues/579 for info regarding the missing variable.

PR #580, opened 2018-08-27
Set a default value for ATHENS_MONGO_CONNECTION_STRING env var

### LINKED ISSUE #579: pkg/config: add MongoConnectionString default
## What

I propose we set `mongodb://127.0.0.1:27017` as the default value for `ATHENS_MONGO_CONNECTION_STRING`.

We would change:
https://github.com/gomods/athens/blob/b62ebddb1bc09c9e7cdbc60ebe8b11aa37d0bd6e/pkg/config/env/mongo.go#L11-L19

To become:
```go
func MongoConnectionString() string { 
    return envy.Get("ATHENS_MONGO_CONNECTION_STRING", "mongodb://127.0.0.1:27017") 
}
```

All uses of the function call would need to be updated to reflect the change in returned values. No longer includes an error.

We could also remove the following:
https://github.com/gomods/athens/blob/b62ebddb1bc09c9e7cdbc60ebe8b11aa37d0bd6e/.travis.yml#L13
https://github.com/gomods/athens/blob/b62ebddb1bc09c9e7cdbc60ebe8b11aa37d0bd6e/scripts/test_unit.sh#L5-L8
https://github.com/gomods/athens/blob/b62ebddb1bc09c9e7cdbc60ebe8b11aa37d0bd6e/cmd/olympus/.env#L1
https://github.com/gomods/athens/blob/b62ebddb1bc09c9e7cdbc60ebe8b11aa37d0bd6e/cmd/proxy/.env#L1

Possibly something in `./pkg/charts` will need to be updated, but I think #578 fixed that.

## Why

It seems to hang up new contributors that there are these undocumented variables that need to be set. We could document them, but instead of having contributors and users set up a bunch of variables we could make use of our config package and handle all defaults for them.

artmello:
- ATHENS_MONGO_CONNECTION_STRING will default to mongodb://127.0.0.1:27017 if not set

Fixes #579


pkg/config/env/mongo.go
```diff
@@ -10,7 +10,7 @@ import (
 
 // MongoConnectionString returns Athens Mongo Storage connection string defined by ATHENS_MONGO_CONNECTION_STRING
 func MongoConnectionString() (string, error) {
-	env, err := envy.MustGet("ATHENS_MONGO_CONNECTION_STRING")
+	env, err := envy.MustGet("ATHENS_MONGO_CONNECTION_STRING", "mongodb://127.0.0.1:27017")
```
robjloranger: This will just be `envy.Get` which only returns a string.

pkg/config/env/mongo.go
```diff
@@ -10,7 +10,7 @@ import (
...
 	if err != nil {
```
robjloranger: Then this error check will no longer be needed. And the function signature updated.

PR #588, opened 2018-08-30
Athens: update mystery uuid version

marwan-at-work:
Our Gopkg.lock has gobuffalo/uuid@v2.0.1 yet buffalo/uuid does not have a v2.0.1 in its tags. It either got deleted or dep somehow made a mistake. Furthermore, or go.mod file had a lot of junk from before 1.11 so it included test dependencies that we didn't need.


marwan-at-work:
Closing for now. For some reason my Go1.11 binary is not translating correct dependency versions :/

PR #589, opened 2018-08-30
Update modules to be correct

marwan-at-work:
Update dependencies and remove unused ones. Turns out copying Go.mod from Gopkg.toml is inconsistent with building go.mod from scratch. This one *looks* more correct as I compared versions one to one between the two files.


michalpristas:
LGTM

PR #593, opened 2018-08-30
[walkthrough]: adjust instructions to clone athens on gopath

krancour:


krancour:
@robjloranger ya... that was a mistake. Fixing.

Issue #595, opened 2018-08-30
Make port configurable for the proxy


arschles:
It's configurable for olympus, and defaulted to `3001`, I feel like we should make it configurable for athens, and defaulted to `3000`

manugupt1:
@arschles What do you think about also creating a non-standard default port for the proxy and olympus as 3000 and 3001 are default for most web apps.

PR #596, opened 2018-08-30
Making port configurable for the proxy

arschles:
We found at the GopherSource workshop at GopherCon that some folks already had processes running on port 3000. We needed to change the port but weren't able to without a code change, so I think making the proxy's port configurable would have been really helpful.

Fixes https://github.com/gomods/athens/issues/595


robjloranger:
for follow up #610

PR #598, opened 2018-08-30
[walkthrough]: add powershell walkthrough

lpdyck:
Since the proxy now works on windows, this re-adds the powershell commands sections, replaces a `CMD` call with native powershell commands, and adds missing quotes to the `rm` command for powershell


docs/content/walkthrough.md
```diff
@@ -69,18 +66,17 @@ INFO[0000] Starting application at 127.0.0.1:3000
...
 
-<!--
 **PowerShell**
 ```console
 $ mkdir $(go env GOPATH)\src\github.com\gomods
 $ cd $(go env GOPATH)\src\github.com\gomods
 $ git clone https://github.com/gomods/athens.git
 $ cd athens
+$ $env:GO111MODULE = "off"
 $ start -NoNewWindow go "run .\cmd\proxy"
```
marpio: This should be `Start-Process -NoNewWindow go run .\cmd\proxy`

lpdyck: You're right. Better to use `Start-Process` than the alias.

lpdyck:
@arschles It did get tested, and I just retested it all this afternoon, which led to some touching up.

Issue #597, opened 2018-08-30
Force HTTPS on the gomods site


arschles:
Right now, if you go to `http://docs.gomods.io`, it won't force you to HTTPS. See attached screenshot.

<img width="976" alt="screen shot 2018-08-30 at 4 04 36 pm" src="https://user-images.githubusercontent.com/70865/44881904-78ed5580-ac6e-11e8-8799-0dc5e4b55252.png">

Sjeanpierre:
Hi @arschles I submitted a PR for this issue which should correct the behavior described on the production URL using redirects. 

When testing this solution on the staging site, it does not appear to work due to the static url value I have set for the redirect. Would you prefer we do a more global redirect which would solve this issue no matter what URL the docs site is running on?

robjloranger:
closed via #600

PR #600, opened 2018-08-30
Redirect http traffic to https for docs

Sjeanpierre:
The commits in this PR should correct issue #597 without taking a more extreme route of not accepting http traffic at all. This fix may not appear to work in Netlify staging environments due to the base URL of the production site being the only one with the redirect.

Information for this commit was gathered from

- https://stackoverflow.com/questions/49712714/can-netlify-redirect-traffic-from-http-to-https-without-forcing-ssl
- https://www.netlify.com/docs/redirects/#structured-configuration

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -6,3 +6,9 @@
     publish = "docs/public/"
     command = "hugo -v"
     environment = { HUGO_VERSION = "0.37" }
+
+[redirects]
+  from = "http://docs.gomods.io/*"
+  to = "https://docs.gomods.io/:splat"
+  status = 301
+  force = true    
--- a/netlify.toml
+++ b/netlify.toml
@@ -7,7 +7,7 @@
     command = "hugo -v"
     environment = { HUGO_VERSION = "0.37" }
 
-[redirects]
+[[redirects]]
   from = "http://docs.gomods.io/*"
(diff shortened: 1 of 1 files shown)


Sjeanpierre:
@arschles according to the following guidance from SO I wanted to be as least invasive as possible as I am not yet certain if their exists content on the site that does not have the right links in place.

> Once you force TLS using Netlify they will set the STS (Strict-Transport-Security) header in your page response headers. I will not go through the explanation but you can read about it here.
> 
> The main thing to know is:
> 
>     Once a supported browser receives this header that browser will prevent any communications from being sent over HTTP to the specified domain and will instead send all communications over HTTPS
> 
> Note that Netlify's settings (using the force TLS checkbox) are to have the visiting browsers enforce this for 1 year past visit date! So, if you have anything that fails on https, your site is going to have some issues until you are able to fix them.

robjloranger:
I'm not certain this is necessary, I would need access to our netlify as a collaborator to check. I use the service for my personal site and do not have any special files or config. It forces https.

robjloranger:
This has been resolved via the Netlify web configuration. Thanks so much for helping out @Sjeanpierre. Let us know if you want help finding another issue to work on.

PR #605, opened 2018-08-31
[protocol]: fix texts and examples in the protocol doc

dapchen:


robjloranger:
I'm not certain there was anything wrong with this as it was. The repository and module were just for example purposes. I'll wait for others opinions on whether these should be live examples or fictional.

I **do** think we should keep the multiple versions in the list example, this helps illustrate what may be returned. So if we want to switch to using a working repo/module then we should pick one with multiple versions. Maybe one of the `athens-artifacts` repos.

marpio:
I think this is just an example and we do not have to use a real repo for that. What we could fix though, is the version list
from:
```
v1.0
v1.1
v2.0
```
to a valid semver MAJOR.MINOR.PATCH i.e.
```
v0.1.0
v0.1.1
v1.0.0
v1.0.1
v1.2.0
```

michalpristas:
acidburn 🔥

docs/content/intro/protocol.md
```diff
@@ -7,24 +7,26 @@ Athens builds on top of Go CLI which specifies a set of endpoints with which it
 
 The original vgo research paper on Download protocol can be found here: https://research.swtch.com/vgo-module
 
-Each of these endpoints sits on top of a module. Let's assume module `htp` authored by `acidburn`.
+Each of these endpoints sits on top of a module. Let's assume module `assert` authored by `arschles`.
 
-So for each of endpoints mentioned bellow we will assume address `acidburn/htp/@v/{endpoint}` (e.g `acidburn/htp/@v/list`)
+So for each of endpoints mentioned bellow we will assume address `arschles/assert/@v/{endpoint}` (e.g `arschles/assert/@v/list`)
```
dapchen: The changes at line 10, 12, and 18 are to be consistent with the texts and the examples. I am okay to change to a different endpoint as long as the texts and examples are consistent.

dapchen:
@robjloranger @marpio Thanks for the comments. I updated the examples.

docs/content/intro/protocol.md
robjloranger: I think we should change this back to a fake repo, only because the versions listed do not exist for `arschels/assert` and it would be nice to show multiple version for our example.

dapchen: Makes sense. I changed back to fake endpoints `acidburn/htp/@v/{endpoint}` in my latest commit.

PR #607, opened 2018-08-31
Fix list

### LINKED ISSUE #590: Go get returns 500 after the repository is deleted
I created `github.com/arschles/throwaway1`, did a `go get` on it with `GOPROXY` set to `https://microsoftgoproxy.azurewebsites.net`, and then deleted the repository. The next `go get` returns this:

```console
arschles ~/Desktop/hellogc $ GO111MODULE=on GOPROXY=https://microsoftgoproxy.azurewebsites.net go get github.com/arschles/throwaway1@v1
go get github.com/arschles/throwaway1@v1: unexpected status (https://microsoftgoproxy.azurewebsites.net/github.com/arschles/throwaway1/@v/list): 500 Internal Server Error
```

I think this violates immutable builds? What do folks think?

marpio:
Fixes #590 

At the moment the DP List function uses `go` to get the list of available versions for a module. This means that if we:
1. `GO111MODULE=on GOPROXY=http://localhost:3000 go get github.com/user/repo`
2. delete  `github.com/user/repo` from GH

Then the next `GO111MODULE=on GOPROXY=http://localhost:3000 go get github.com/user/repo` will fail even though the proxy has it in it's storage
This is **not** what we want because it violates reproducible builds.

This PR solves it by merging the list we get from `go` with the one we get from the `storage`

Unfortunately I had to copy & paste some code from cmd/go because the list has to be sorted in a certain way.


pkg/download/download.go
```diff
@@ -75,7 +77,25 @@ func (p *protocol) request(mod, ver string) error {
...
+	const op errors.Op = "protocol.List"
+	goList, goErr := p.dp.List(ctx, mod)
+	strList, sErr := p.s.List(ctx, mod)
+	if goErr != nil && sErr != nil {
+		var errs error
+		errs = multierror.Append(errs, goErr, sErr)
+		return nil, errors.E(op, errs)
+	}
+	if goErr != nil && sErr == nil {
+		semver.SortVersions(strList)
```
marpio: It's not guaranteed that storage returns a sorted versions list. Should storage.List take care of this? What do you think?

marwan-at-work: @marpio I think Go does not require us to sort versions. Also, I'm maintaining a copy of the Go internal semver stuff here github.com/marwan-at-work/vgop (including parse of go.mod file etc) but I don't think we need to do any sorting any way.

pkg/download/download.go
```diff
@@ -75,7 +77,39 @@ func (p *protocol) request(mod, ver string) error {
...
+		return strList, nil
+	}
+	if goErr == nil && sErr != nil {
+		return goList, nil
+	}
+
+	combinedList := union(goList, strList)
+	semver.SortVersions(combinedList)
```
marwan-at-work: same, we can remove

pkg/download/download.go
```diff
@@ -75,7 +77,39 @@ func (p *protocol) request(mod, ver string) error {
 }
 
 func (p *protocol) List(ctx context.Context, mod string) ([]string, error) {
-	return p.dp.List(ctx, mod)
+	const op errors.Op = "protocol.List"
+	goList, goErr := p.dp.List(ctx, mod)
+	strList, sErr := p.s.List(ctx, mod)
+	if goErr != nil && sErr != nil {
```
marwan-at-work: We should return if either of errors are nil, not both. Otherwise, the results will not be consistent to the user.

pkg/download/download.go
marpio: Your completely right - go sorts the list from proxy anyway.

pkg/download/download.go
marpio: @marwan-at-work could you expand on that?
There are 4 cases:
1. `go list` failed (i.e. repo has been deleted like in the issue #590 ) and storage has the versions -> versions from storage
2. `go list` succeeded and storage failed (i.e. is empty) -> versions from `go list`
3. both `go list` and storage returned their lists -> combined lists storage + `go list`
4. `go list` and storage failed -> return error

Or did you mean we should differentiate between ErrNotFound and unexpected error? We could do that for the storage but I don't know how to do that for `go list`...

pkg/download/download.go
```diff
@@ -75,7 +77,39 @@ func (p *protocol) request(mod, ver string) error {
...
+		return goList, nil
+	}
+
+	combinedList := union(goList, strList)
+	semver.SortVersions(combinedList)
+	return combinedList, nil
+}
+
+// union concatenates two version lists and removes duplicates
+func union(list1, list2 []string) []string {
```
michalpristas: can we check nil for list[1,2] in case of nil init with []string and merge so we can get rid of awkward err checking up there?

pkg/download/download.go
marwan-at-work: @marpio 
1. Can we gracefully handle that error? If a repo is deleted, `go list -m` should return a 404 figuratively speaking. But if let's say github is down, then that's an unexpected error. Do we want to ignore upstream if upstream is unavailable? Obviously we ignore upstream if the "client" already knows what version it wants to get from Athens. Because `/list` never gets called. 
2. If storage fails, we should fail. always. 
3. this is happy path. We're good. 
4. Since storage failed, this one fails as well. Don't think we need to combine the errors, just take the storage error: that's more important. But up to you.

marwan-at-work:
@marpio The download.Protocol list method should handle "KindNotFound" to make sure it doesn't error out back to the client and just take the upstream list results. 

Furthermore, we should make sure for *all* the storages, that we appropriately return "KindNotFound" or just an empty list, whichever you prefer.

marpio:
>Furthermore, we should make sure for all the storages, that we appropriately return "KindNotFound" or just an empty list, whichever you prefer.

All storages return KindNotFound.

marwan-at-work:
@marpio in this case, we should update the PR to include that check on the /list side if it's not there already.

pkg/download/download.go
marwan-at-work: My answer is that, we should be able to ask `go list -m` questions as such: 
- Athens: why did you error out? 
- Go: repository not found 
- Athens: ok, this is not an error, only take from storage. 

- Athens: why did you error out? 
- Go: GitHub is down
- Athens: ok this is an error, we should report it to our metrics and tell the user to try again. Because we might give them an outdated list. Of course, /list is now down but /info /mod and /zip still work fine. 

And so on and so forth. The biggest win from this is that Athens remains reproducible. If we just simply ignore upstream errors, then one second /list returns a full list, the next second returns only the storage lists. And we'll never be able to make Athens stable in that regard. 

Parsing the `go list -m` errors is gonna be ugly, see this comment: 
https://github.com/gomods/athens/issues/590#issuecomment-418020244

I hope the go command can in the future return structured errors.

marpio: Definitely makes sense. Thank you Marwan!

pkg/download/go_lister.go
```diff
@@ -0,0 +1,75 @@
...
+type Lister func(string) (*storage.RevInfo, []string, error)
+
+type listResp struct {
+	Path     string
+	Version  string
+	Versions []string `json:",omitempty"`
+	Time     time.Time
+}
+
+// NewGoLister bla
```
arschles: @marpio would you mind filling out the godocs here? we can save this for a follow-up if you want too

pkg/download/list_merge_test.go
```diff
@@ -0,0 +1,142 @@
+package download
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"io/ioutil"
+	"testing"
+
+	athenser "github.com/gomods/athens/pkg/errors"
```
arschles: I like this name - maybe members of the athens community should be called "athenians" but also we could call ourselves "athensers"

😆

pkg/download/list_merge_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	for i := range a {
+		if a[i] != b[i] {
+			return false
+		}
+	}
+
+	return true
+}
```
arschles: ❤️ these tests!!!!!!!!!!!!!!!!!!!!!!

marpio:
@marwan-at-work 
I've made the changes we discussed. I have also altered the storage.List behavior - it returns empty slice and not an error. It made the code little bit easier to read and it kind of made sense to me since no stored versions is a perfectly valid state.

PR #608, opened 2018-08-31
Provide binary release

### LINKED ISSUE #581: Provide a binary release
At the moment we provide a docker image, which is great but sometimes `wget && tar -xvf && ./proxy` is just enough.
We could use https://developer.github.com/v3/repos/releases/#create-a-release

marpio:
This PR provides support for binary releases (Windows, Linux, MacOs) of Athens-proxy (Olympus can be easily added)
It does so using goreleaser https://goreleaser.com

It requires a Github token with repo scope set in Travis
@arschles it would be great if you could create one and set it in Travis

Fixes #581

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,26 @@
+builds:
+  -
+    goos:
+      - linux
+      - darwin
+      - windows
+    goarch:
+      - amd64
+    main: ./cmd/proxy/main.go
+    binary: athens-proxy
+    env:
+      - CGO_ENABLED=0
+archive:
+  name_template: '{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ if .Arm
+    }}v{{ .Arm }}{{ end }}'
+  format: tar.gz
+checksum:
+  name_template: 'checksums.txt'
+snapshot:
+  name_template: SNAPSHOT-{{ .Commit }}
+changelog:
+  sort: asc
+  filters:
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -0,0 +1,33 @@
...
+    env:
+      - CGO_ENABLED=0
+archive:
+  name_template: '{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ if .Arm
+    }}v{{ .Arm }}{{ end }}'
+  format: tar.gz
+  files:
+  - licence*
+  - LICENCE*
+  - license*
```
DonMcNamara: This might be a dumb question, but was the duplication of license intentional? (it is there 4 times, twice uppercase and lowercase)

marpio: Not dumb at all! Copy & paste error. Thanks!

robjloranger:
@marpio do you mean an OAuth token? I can create and add one for you.

robjloranger:
All set, `GITHUB_TOKEN` is present in travis. We should revisit at some point with a 'bot' account under our org, this is good for now though.

robjloranger:
I'm not certain how we will test this before merge. The Travis variables are not available to pull requests coming from other forks. Maybe we have to merge into a new branch here in the Athens repo first.

marpio:
Additionally the deployment step is skipped so I guess we have no other choice than merge this into a new branch.

PR #614, opened 2018-09-01
Clean up error operation names to match function names.

DonMcNamara:
Update the error operation names to match function names.


pkg/download/goget/goget.go
```diff
@@ -125,7 +125,7 @@ func (gg *goget) list(op errors.Op, mod string) (*listResp, error) {
 }
 
 func (gg *goget) GoMod(ctx context.Context, mod string, ver string) ([]byte, error) {
-	const op errors.Op = "goget.Info"
+	const op errors.Op = "goget.GoMod"
```
manugupt1: Please add package name also. for example download.goget.GoMod
There is a WIP PR that does that, but I am happy to rebase it on top of your changes.

https://github.com/gomods/athens/pull/604/files

pkg/middleware/validation.go
```diff
@@ -15,7 +15,7 @@ import (
 // NewValidationMiddleware builds a middleware function that performs validation checks by calling
 // an external webhook
 func NewValidationMiddleware(entry log.Entry, validatorHook string) buffalo.MiddlewareFunc {
-	const op errors.Op = "actions.ValidationMiddleware"
+	const op errors.Op = "actions.NewValidationMiddleware"
```
manugupt1: This can be NewValidationMiddleware. If you are doing that can you add a new errors.Op in the middleware function that is returned with `const op errors.Op = "actions.ValidationMiddleware` and update the span function with the correct op

pkg/storage/mongo/mongo.go
```diff
@@ -31,7 +31,7 @@ func NewStorage(connectionString string) (*ModuleStore, error) {
 // NewStorageWithCert returns a connected Mongo backed storage
 // that satisfies the Backend interface.
 func NewStorageWithCert(connectionString, certPath string) (*ModuleStore, error) {
-	const op errors.Op = "fs.NewStorage"
+	const op errors.Op = "fs.NewStorageWithCert"
```
manugupt1: thanks for catching this!

michalpristas:
Please resolve conflicts, goget got renamed to fetcher

pkg/download/goget/goget.go
arschles: @manugupt1 can we do this and the below change in a follow-up? or, if it's not too bad for you to rebase, is that ok?

manugupt1: Yes, GoGet has been removed so we do not need this change anymore.
Resolving merge conflicts  should fix this automatically

pkg/middleware/validation.go
manugupt1: yep! Sure

pkg/download/goget/goget.go
marwan-at-work: Just FYI we don't need to make this *sound* like a stack trace with exact package-file-method operations.

Since it's just a string we can make it more human readable and easy to look up and understand when debugging in cloud logs.

In other words this can be `downloadProtocol.GoMod` or `gogetProtocol.GoMod` -- or simply just `goget.GoMod` because those strings are likely unique and easy to scan in your eye but also easy to do global look ups when you're searching the code base 

So I think we should keep it short and sweet, but also understandable.

manugupt1: Yep! I do not really want it to sound like a stack trace. The reason I like this has got less to do with the length but for the following reasons:

1) It has a convention, so it is easy to ask others to stick to it
2) It is easier to filter logs, so if you want to filter logs you can separate by the delimiter and check it out with a simple regex.

manugupt1: Oh also! I am saying something like package.object.func rather than package.file.func

marwan-at-work: @manugupt1 i think the convention is difficult to figure out, not always do you have something like `package.object.func` sometimes you just have `package.func` or `package.globalVar` etc. 

The nice thing about function operations is that it's meant for humans, yet it can still be filterable. So keeping things short and understandable like `goget.GoMod` is both easier for humans to read and easier for humans to filter. 

Most likely, you're gonna read an array of operations so reading some thing like 
```["download.GoMod", "stasher.Pool", "goget.GoMod", "mongo.Get"]``` 
is much easier on the eye than : 

```["download.Handlers.GoMod", "stasher.pool.Pool", "download.goget.GoMod", "storage.Mongo.Get"]```

**Comment by DonMcNamara** [comment]
Hi all, sorry for not responding earlier. I was away. I'm hoping to work on this over the next day or two. Thanks!

**Comment by DonMcNamara** [comment]
I've updated this PR. Now that  `goget.go` has moved/changed, the changes in this PR are mostly cosmetic -- the old version of `goget.go` had some incorrectly copy/pasted operation names.

File: pkg/module/disk_ref.go [hunk 4]
```diff
@@ -50,7 +50,7 @@ func newDiskRef(fs afero.Fs, root, module, version string) *diskRef {
 // ClearFiles deletes all data from the given fs at path root
 // This function must be called when zip is closed to cleanup the entire GOPATH created by the diskref
 func ClearFiles(fs afero.Fs, root string) error {
-	const op errors.Op = "clearFiles"
+	const op errors.Op = "ClearFiles"
```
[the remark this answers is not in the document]
marwan-at-work: will update this in a follow up pr.

PR #624, opened 2018-09-02
[DO NOT MERGE] Show case of proposed single-flight implementation

michalpristas:


pkg/download/stasher/basic.go
```diff
@@ -0,0 +1,56 @@
...
+// Stasher has the job of taking a module
+// from its upstream Download Protcool and
+// stashing to its Storage Backend. Both
+// interfaces must be defined on the type previously.
+type Stasher interface {
+	Stash(mod, ver string) error
+}
+
+// DPVersionFn is a version func of download protocol
+type DPVersionFn func(context.Context, string, string) (*storage.Version, error)
```
marwan-at-work: this is basically what the module.Fetcher should *really* be. The module.Fetcher is a two step process for no reason. 

We should just have 

```golang
type Fetcher interface {
  Fetch(context.Context, string, string) (*storage.Version, error)
}
```

marwan-at-work:
@michalpristas my pr is basically getting very close to what you're describing here. I thought I'd make a second PR to do what you're doing. But I'll just push to this pr

pkg/download/stasher/basic.go
michalpristas: yes, we can even omit Version

michalpristas: i believe you already did that in your updated version, i worked with the one before the update

michalpristas:
my main concern is readability of this

```go
fs := afero.NewOsFs()
mf, err := module.NewGoGetFetcher(goBin, fs)
if err != nil {
    return err
}
st := stash.New(mf, s, stash.WithSingleflight, stash.WithPool(env.GoGetWorkers()))
dp := download.New(s, st, goBin, fs)
dp = addons.WithPool(dp, env.ProtocolWorkers())
opts := &download.HandlerOpts{Protocol: dp, Logger: l, Engine: proxy}
```

vs

```go
stasherFactory := stasher.Deduplicated(stasher.Pooled(stasher.Basic(s), env.GoGetWorkers()))
gg, err := goget.New(stasherFactory)
if err != nil {
	return err
}
p := pool.New(gg, env.GoGetWorkers())
opts := &download.HandlerOpts{Protocol: p, Logger: l, Engine: proxy}
download.RegisterHandlers(app, opts)
```

and that wrappers are not entirely implemented as decorators but passed as args.
But your PR is much more readable comparing to what it was before

PR #629, opened 2018-09-04
fix for #626

### LINKED ISSUE #626: Docs Inclusion : How to quit from the Athens Proxy
Although the Docs gives a good explanation as to how the Athens proxy is to be run ( in the Walk-through ), it doesn't tell how to quit from it. A small yet important thing for new people trying it out

Mention it in the docs, in the Walk Through section

sahitpj:
Adds the "how to stop proxy" step to the walkthrough

Resolves #626


docs/content/walkthrough.md
```diff
@@ -165,6 +165,17 @@ The 🦁 says rawr!
 No additional output is printed because Go found **github.com/athens-artifacts/samplelib@v1.0.0** in the Go Module
 cache and did not need to request it from the proxy.
 
+In order to quit from the proxy at any point of time, we can press Ctlr-C
+
```
robjloranger: This will not work, only if the user runs the proxy with `go run` or similar. In this document the user is instructed to start the proxy in the background.
L#62: `$ GO111MODULE=off go run ./cmd/proxy &`
L#76: `$ Start-Process -NoNewWindow go "run .\cmd\proxy"`

We will need to give instructions for both bash and powershell on how to list running processes for the session and how to kill them.

For example in bash, `ps` will list processes and their IDs. Then `kill [pid]` will end the running background process, where `pid` is the process id.

sahitpj:
heyy @robjloranger Made the requested changes

PR #630, opened 2018-09-05
Refactor error e

sauvaget:
Based on a discussion with Michal Pristas on the #athens Slack Channel


robjloranger:
I like this overall, it's nice and clean. I am wondering, however, if we should consider it complete. For example in `pkg/storage/fs` there are a few functions not yet modified, `Exists`, `Info`, `GoMod`, `Zip`.

Should we be refactoring the entire code base? Or is this meant to just be a starting point?

I can't think of a better way but am undecided about the use of this new return in the `Exists` functions specifically. For example when there is an error checking for a module, but the error is not that the module was not found, we return true and the error. While the caller should be checking the error first, it feels wrong to be saying "Yes it exists, but there was an error checking if it exists."

arschles:
whoa, I have no idea on this one. I'm torn like @marwan-at-work is. @sauvaget is this the start of a larger refactor, or were you thinking of keeping it to the places you changed here?

sauvaget:
Hey, I still have some functions to go over. My fork wasn't on the current HEAD so I missed a few things. @arschles I received the task from @michalpristas in the Slack Channel. I can put it on hold for a little if it needs more discussing?

arschles:
@sauvaget this seems ok to keep pursuing. I apologize for letting this go so long, and that there are now conflicts to resolve. When you do resolve them, let me/us know and we'll get it merged!

arschles:
@sauvaget I'm going to close this PR since it's been over a month and has a lot of conflicts. You can absolutely feel free to reach out on slack (`#athens`) or re-open this PR if you're still interested in this, or contribute any other way you like. We hope you come back 😄

Issue #633, opened 2018-09-06
Minimize Docker Image footprint


gcstang:
**Describe the bug**
Currently the Docker image for gomods/proxy is over 1GB in size and although the image is based on the full 1.11 Go distribution it could be minimized by utilizing the Alpine version instead.

**Expected behavior**
Smaller docker image

**Environment (please complete the following information):**
 - Go version : 1.11
 - proxy version : latest image id: 027e580061f7

PR #635, opened 2018-09-06
changed base to use 1.11-alpine and addded GO111MODULE=on to environment

gcstang:
Satisfies proxy change to 1.11-alpine

https://github.com/gomods/athens/issues/633


cmd/proxy/Dockerfile
```diff
@@ -1,10 +1,11 @@
-FROM golang:1.11
+FROM golang:1.11-alpine
 
 RUN mkdir -p $GOPATH/src/github.com/gomods/athens
 WORKDIR $GOPATH/src/github.com/gomods/athens
 
 COPY . .
 
+ENV GO111MODULE=on
```
marwan-at-work: you can just have this be in the next line: 
`RUN GO111MODULE=on CGO_ENABLED=0 go build -mod=vendor -o /bin/app ./cmd/proxy` 

No need to hash a whole step for it.

Kunde21:
Subtle bug:  The `golang:1.11-alpine` container doesn't have vcs systems installed (git/svn/etc), so they need to be installed in the build for `go get` to work properly.

gcstang:
Specifically which ones are you looking to have installed?
There wasn't any in the original Dockerfile so I'll have to add them to this PR.

gcstang:
updated to add all VCS's in original stretch image.

Kunde21:
@manugupt1 That's added in the [golang build](https://github.com/docker-library/golang/blob/e06b00cbe4974436ee00eaec215d7fbb123cbc24/1.11/alpine3.8/Dockerfile#L3), so it's covered.

marwan-at-work:
@gcstang with this update + adding the VCS binaries, can you tell me what the difference is in image size? Thanks!

gcstang:
@marwan-at-work size is now 630MB down from original of 1.03GB

arschles:
@marwan-at-work this looks ready for review. if you do another review and things look good, would you mind merging?

robjloranger:
I don't have time to test but the changes look good 👍

arschles:
@marwan-at-work are you ok with merging this now?

marwan-at-work:
@arschles I think 600MB is still big and not that different from 1Gig. Especially that we have to install the VCS deps ourselves in the Dockerfile. I say we should stick with the original golang image so we don't have to maintain anything. 

The only time we should introduce this complexity is if the gain is super high: say 10MB image vs 1Gig image. Which I believe we can still achieve if we just extracted the `go mod download` part (potentially just the cmd/go binary and not the linker etc) as well as the VCS binaries which im not sure how big they are.

manugupt1:
@marwan-at-work By default, Alpine is much secure than Debian. My sense is that even if it is not size, we should just adopt alpine for security reasons. Further, by not having other packages, we are reducing the security vulnerabilities that the container can be exposed to a minumum. If there is a vcs, that we have missed we can always add it.

marwan-at-work:
@manugupt1 if we're going for it, then let's just grab the go binary and the vcs only in a busy box image. That'd be the most win for both securities and size.

manugupt1:
@marwan-at-work I thought alpine was based on busybox, Am I missing something here?
https://alpinelinux.org/about/

marwan-at-work:
@manugupt1 yes but the image we're using is `golang:1.11-alpine` which is 600MB. If we just use alpine or busybox and add the cmd/go binary without the distribution then that removes all the packages for security/size

manugupt1:
I do not see why that is a bad idea at all.

marwan-at-work:
@manugupt1 hahah it's not, im saying that'd be the appropriate way to do it.

Kunde21:
Got a chance to look at this today.  One problem is the repository is being copied into the container, adding a lot of data that's not needed after the build.
Moving to a multi-stage dockerfile (copying `/bin/app` to a fresh `golang:alpine` container) reduces the size to ~440MB.

Kunde21:
If we want to really cut it tight, copying the `go` and `/bin/app` binaries into a fresh alpine container gets us down to  ~150MB.

gcstang:
I'm able to get this down to as was mentioned ~150MB but it seems there is a build tool that maybe isn't in the project or I haven't found to create this image.  

Would someone know where it is or be able to give it to me in order to run this properly?

Currently I have to 
1. fork/clone
2. cd ~/athens 
3. cp cmd/proxy/Dockerfile .
4. docker build -t athens .

However it seems there are steps missing since the URL shows a partial page with missing assets this occurs when I build with or without it stripped down so assets are in the same place as normal.

gcstang:
after making this multistage size is now down to 162MB

cmd/proxy/Dockerfile
```diff
@@ -1,11 +1,22 @@
...
+FROM alpine
+
+ENV PATH="/usr/local/go/bin:${PATH}"
+COPY --from=builder /bin/app /bin/app
+COPY --from=builder /go/src/github.com/gomods/athens/cmd/proxy/locales /go/src/github.com/gomods/athens/cmd/proxy/locales
+COPY --from=builder /go/src/github.com/gomods/athens/cmd/proxy/assets /go/src/github.com/gomods/athens/cmd/proxy/assets
+COPY --from=builder /usr/local/go/bin/go /usr/local/go/bin/go
+
+RUN apk update &&  \
+		apk add --no-cache bzr git mercurial openssh-client subversion procps
```
manugupt1: Can you add an ENV GO111MODULE=on to the alpine image as it is turned off by default

manugupt1: Perfect! thank you 🙇 
Can you also resolve conflicts please and update the branch to master?

Kunde21:
@gcstang From the repo root, you can run `make proxy-docker` to build the container.

cmd/proxy/Dockerfile
```diff
@@ -1,11 +1,23 @@
-FROM golang:1.11
+FROM golang:1.11-alpine AS builder
 
 RUN mkdir -p $GOPATH/src/github.com/gomods/athens
 WORKDIR $GOPATH/src/github.com/gomods/athens
```
Kunde21: `WORKDIR` will create the directory if it doesn't exist, so the preceding `RUN` line is excess.

gcstang:
Several changes:
1. add and run create default config, pre-req bash installed
2. moved go binary to /bin/go
3. go pre-req created GOROOT path at /usr/local/go

gcstang:
Please add during merge:
Co-authored-by: Kunde21

robjloranger:
re: co authoring
It needs to be at the *bottom* of the commit message, with at least one empty new line before it. Each entry goes on it's own line. The name can be anything but the email must be associated with their account. For example:
```
messages etc

Co-authored-by: Chad <public@email.com>
Co-authored-by: Gil <public@email.com>
```

@Kunde21, @gcstang  do you have a public email associated with you GH acct? You can use your GH no reply address if you prefer.

gcstang:
ok thank you robjloranger

cmd/proxy/Dockerfile
```diff
@@ -1,15 +1,30 @@
...
+	apk add --no-cache bash && \
+	GO111MODULE=on CGO_ENABLED=0 go build -mod=vendor -o /bin/app ./cmd/proxy && \
+	./scripts/create_default_config.sh
+
+FROM alpine
+
+ENV GO111MODULE=on
+
+COPY --from=builder /bin/app /bin/app
+COPY --from=builder /go/src/github.com/gomods/athens/config.toml /go/src/github.com/gomods/athens/config.toml
```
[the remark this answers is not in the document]
gcstang: I can move the config.toml to /bin/ but then the parameter will be -config_file=/bin/config.toml if you're ok with that then I'll make the change.  The current code expects a full path to the file.

cmd/proxy/Dockerfile
```diff
@@ -1,15 +1,30 @@
...
+RUN apk update &&  \
+		apk add --no-cache bzr git mercurial openssh-client subversion procps && \
+		mkdir -p /usr/local/go
 
 ENV GO_ENV=production
 
 EXPOSE 3000
 
-ENTRYPOINT /bin/app -config_file=config.toml
+ENTRYPOINT /bin/app -config_file=/go/src/github.com/gomods/athens/config.toml
```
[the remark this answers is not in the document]
gcstang: yes I can change this to support the preferred method for entrypoint here: https://docs.docker.com/engine/reference/builder/#entrypoint
I'll make this change after I receive an answer on prior request and submit it all at once.

cmd/proxy/Dockerfile
```diff
@@ -1,15 +1,30 @@
...
+ENV GO111MODULE=on
+
+COPY --from=builder /bin/app /bin/app
+COPY --from=builder /go/src/github.com/gomods/athens/config.toml /go/src/github.com/gomods/athens/config.toml
+COPY --from=builder /go/src/github.com/gomods/athens/cmd/proxy/locales /go/src/github.com/gomods/athens/cmd/proxy/locales
+COPY --from=builder /go/src/github.com/gomods/athens/cmd/proxy/assets /go/src/github.com/gomods/athens/cmd/proxy/assets
+COPY --from=builder /usr/local/go/bin/go /bin/go
+
+RUN apk update &&  \
+		apk add --no-cache bzr git mercurial openssh-client subversion procps && \
```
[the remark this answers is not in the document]
gcstang: I wouldn't know, believe whomever does the project testing would know this information.  This is simply the line from the original image that was being used and supplied the required VCS applications.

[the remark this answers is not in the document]
marwan-at-work: @gcstang that's good to know, can you put a link to that exact line? So we can always compare in case Go introduces a new VCS dependency or changes something. Thank you!

cmd/proxy/Dockerfile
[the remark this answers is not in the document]
marwan-at-work: yes, no need to create a whole path for the config. You can either put it inside /bin or /config. Up to you.

cmd/proxy/Dockerfile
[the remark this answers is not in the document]
gcstang: It's part of the buildpacks-stretchscm release that golang uses in this file: https://github.com/docker-library/buildpack-deps/blob/1845b3f918f69b4c97912b0d4d68a5658458e84f/stretch/scm/Dockerfile

gcstang:
updated with requests from @marwan-at-work

cmd/proxy/Dockerfile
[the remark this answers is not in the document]
tnextday: https://github.com/golang/go/blob/f082dbfd4f23b0c95ee1de5c2b091dad2ff6d930/src/cmd/go/internal/get/vcs.go#L90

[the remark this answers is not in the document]
marwan-at-work: I'm not familiar with anything other than Git, but can we: 
1. double check that the array @tnextday mentioned is all installed in this dockerfile? 
2. Add either/both links to the Dockerfile itself so it's easy in the future to compare if things get outdated (which they will) 

Thank you!

cmd/proxy/Dockerfile
```diff
@@ -1,15 +1,30 @@
...
+
+COPY --from=builder /bin/app /bin/app
+COPY --from=builder /go/src/github.com/gomods/athens/config.toml /config/config.toml
+COPY --from=builder /go/src/github.com/gomods/athens/cmd/proxy/locales /go/src/github.com/gomods/athens/cmd/proxy/locales
+COPY --from=builder /go/src/github.com/gomods/athens/cmd/proxy/assets /go/src/github.com/gomods/athens/cmd/proxy/assets
+COPY --from=builder /usr/local/go/bin/go /bin/go
+
+RUN apk update &&  \
+		apk add --no-cache bzr git mercurial openssh-client subversion procps && \
+		mkdir -p /usr/local/go
```
marwan-at-work: why do we need this?

cmd/proxy/Dockerfile
```diff
@@ -1,15 +1,30 @@
...
 
 ENV GO_ENV=production
 
 EXPOSE 3000
 
-ENTRYPOINT /bin/app -config_file=config.toml
+ENTRYPOINT ["app", "-config_file=/config/config.toml"]
```
marwan-at-work: shouldn't this be "/bin/app"?

cmd/proxy/Dockerfile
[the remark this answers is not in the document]
tnextday: Also need to add the [fossil](https://pkgs.alpinelinux.org/packages?name=fossil&branch=v3.8) package.

cmd/proxy/Dockerfile
gcstang: this installs the VCS tools and GO is looking for it's default GOROOT which is /usr/local/go otherwise it fails to run.

cmd/proxy/Dockerfile
gcstang: bin is on the PATH so it's not needed.

cmd/proxy/Dockerfile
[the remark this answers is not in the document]
gcstang: added

cmd/proxy/Dockerfile
tnextday: I think we should not use the generic name `app`, it is better to use names such as `athens` or `athens-proxy`.
`ENTRYPOINT` is usually used to execute pre-launch scripts. `CMD` is more suitable for this.

cmd/proxy/Dockerfile
tnextday: Copy go binary to /usr/local/go/bin/ and set PATH env ,
The compatibility of this will be better.

Kunde21:
@robjloranger Yeah, Kunde21@gmail.com is my public face.

cmd/proxy/Dockerfile
```diff
@@ -1,15 +1,34 @@
...
+	GO111MODULE=on CGO_ENABLED=0 go build -mod=vendor -o /bin/app ./cmd/proxy && \
+	./scripts/create_default_config.sh
+
+FROM alpine
+
+ENV GO111MODULE=on
+
+COPY --from=builder /bin/app /bin/app
+COPY --from=builder /go/src/github.com/gomods/athens/config.toml /config/config.toml
+COPY --from=builder /go/src/github.com/gomods/athens/cmd/proxy/locales /go/src/github.com/gomods/athens/cmd/proxy/locales
```
tnextday: Resource files should not be copied to the go source directory and should be copied to the working directory.
Athens uses the gobuffalo package. Normally, it should be published using `buffalo build`, which will package the resources.
If we don't use `buffalo build` then we need to copy the resources ourselves.

Here is the location of the settings:

https://github.com/gomods/athens/blob/c740fcb26cc0192391c10b4b0185d62b78c1410b/cmd/proxy/actions/app.go#L36
https://github.com/gomods/athens/blob/c740fcb26cc0192391c10b4b0185d62b78c1410b/cmd/proxy/actions/render.go#L9

This position is relative to PWD, which is the WORKDIR in the dockerfile.

marwan-at-work: @tnextday couple notes: 
1. buffalo build is not playing well with modules right now, so `go build` is sufficient for the Proxy (I think buffalo just released a fix very recently) 
2. We don't really have any resources. In my opinion we should remove all this stuff in another pr.

cmd/proxy/Dockerfile
marwan-at-work: @tnextday agreed. However, I think I'll approve and merge this for now and submit a change afterwards. @gcstang has gracefully kept up with all of our comments and it looks good enough for this pr. Thank you both!

PR #645, opened 2018-09-11
Using the buffalo built-in logger

### LINKED ISSUE #537: Propagate Log.Entry down to all the middleware functions / handlers
With https://github.com/gomods/athens/pull/492 and https://github.com/gomods/athens/pull/456 we have two middleware functions in addition to the type safe one implemented as `LogEntryHandler` .

Since the `log.Entry` parameters are bound to the request that traverses all the middleware functions (and the handler) it would be nice to have a way to instantiate it in the top level middleware and propagate it down to all the subsequent calls.

arschles:
Instead of trying to thread log entries down the stack, we can use the built-in buffalo logger and pass that down the stack as we already do. The buffalo logger is backed by logrus too.

@fedepaol and I discussed this on slack today. I think it can fix #537 if people are ok with it.

cc/ @marwan-at-work - I know you were pretty deep in conversation about this topic

Ref #537

### DIFF
--- a/pkg/download/latest.go
+++ b/pkg/download/latest.go
@@ -6,7 +6,6 @@ import (
 	"github.com/gobuffalo/buffalo"
 	"github.com/gobuffalo/buffalo/render"
 	"github.com/gomods/athens/pkg/errors"
-	"github.com/gomods/athens/pkg/log"
 	"github.com/gomods/athens/pkg/paths"
 )
 
@@ -14,18 +13,20 @@ import (
 const PathLatest = "/{module:.+}/@latest"
 
 // LatestHandler implements GET baseURL/module/@latest
-func LatestHandler(dp Protocol, lggr log.Entry, eng *render.Engine) buffalo.Handler {
+func LatestHandler(dp Protocol, eng *render.Engine) buffalo.Handler {
 	const op errors.Op = "download.LatestHandler"
 	return func(c buffalo.Context) error {
 		mod, err := paths.GetModule(c)
 		if err != nil {
-			lggr.SystemErr(errors.E(op, err))
+			c.Logger().Warn(errors.E(op, err))
+			// lggr.SystemErr(errors.E(op, err))
 			return c.Render(500, nil)
 		}
 
 		info, err := dp.Latest(c, mod)
 		if err != nil {
-			lggr.SystemErr(errors.E(op, err))
+			c.Logger().Warn(errors.E(op, err))
+			// lggr.SystemErr(errors.E(op, err))
 			return c.Render(errors.Kind(err), eng.JSON(errors.KindText(err)))
 		}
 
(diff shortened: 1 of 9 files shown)


pkg/download/latest.go
```diff
@@ -6,26 +6,27 @@ import (
...
 
 // LatestHandler implements GET baseURL/module/@latest
-func LatestHandler(dp Protocol, lggr log.Entry, eng *render.Engine) buffalo.Handler {
+func LatestHandler(dp Protocol, eng *render.Engine) buffalo.Handler {
 	const op errors.Op = "download.LatestHandler"
 	return func(c buffalo.Context) error {
 		mod, err := paths.GetModule(c)
 		if err != nil {
-			lggr.SystemErr(errors.E(op, err))
+			c.Logger().Warn(errors.E(op, err))
```
marwan-at-work: This won't work because `Warn` is a hard coded. How do we deduce the log level based on the error.Error struct with this?

pkg/middleware/validation.go
```diff
@@ -32,7 +31,11 @@ func NewValidationMiddleware(entry log.Entry, validatorHook string) buffalo.Midd
 			if version != "" {
 				valid, err := validate(validatorHook, mod, version)
 				if err != nil {
-					entry.SystemErr(err)
+					// Not sure if we can create a `SystemErr` function instead of this?
```
marwan-at-work: yes, we should figure this out because I have Athens deployed in GCP and I won't be able to correctly observe things :)

pkg/download/latest.go
fedepaol: Agree. If we use the level explicitly instead of relying on the error we are stepping back from what we have right now.

arschles: what do you think of a `SystemErr` function in `github.com/gomods/athens/pkg/log` to do this?

pkg/middleware/validation.go
arschles: would you be cool with [this](https://github.com/gomods/athens/pull/645/files#r217569404) if it had basically the same code as `lggr.SystemErr` in it?

arschles:
@marwan-at-work do you think that https://github.com/gobuffalo/buffalo/issues/1171 is a dealbreaker? any others?

pkg/download/latest.go
fedepaol: I think it would work. It's what you proposed down here https://github.com/gomods/athens/pull/645/files#diff-7522a4893602f93e19a3c36e780166e2R36

arschles:
@marwan-at-work @fedepaol sorry for letting this sit so long. Do you think I should resolve conflicts and keep trying to make this work, or I should close this and figure out another way to thread log entries down the stack (or do something else)?

marwan-at-work:
@arschles we should either pass things through closures or through the regular context (as opposed to buffalo context). the buffalo context logger won't be able to give us the SystemErr without type casting and so we might as well use the regular context. But if you wanna make closures work (to keep things type-safe) that's a win :)

arschles:
I am going to close this without merging since it touches so much code (and now has so many conflicts). I'll create an issue with what @fedepaol said [above](https://github.com/gomods/athens/pull/645#issuecomment-427249684) so we can follow up.

PR #649, opened 2018-09-12
Fix end 2 end test

### LINKED ISSUE #518: Fix test_e2e.sh - set GOPATH to the temp dir
At the moment 
https://github.com/gomods/athens/blob/493945c13a740e790ef9d04b638598f3445d5d7d/scripts/test_e2e.sh#L14-L16
doesn't clear the cache because the cache is under `GOPATH/pkg/mod` Because of that go doesn't use the proxy which is the point of the test.

To fix that we need to **temporary** set `GOPATH` to 
https://github.com/gomods/athens/blob/493945c13a740e790ef9d04b638598f3445d5d7d/scripts/test_e2e.sh#L10

marpio:
Fixes #518

### DIFF
--- a/scripts/test_e2e.sh
+++ b/scripts/test_e2e.sh
@@ -6,8 +6,13 @@
 set -xeuo pipefail
 
 REPO_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )/.."
+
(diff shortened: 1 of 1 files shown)


marpio:
@robjloranger I've removed sudo

PR #650, opened 2018-09-12
Added Docker sub-section to Proxy section in Walkthrough

nunix:
added a first extract on how to use the Docker image of the Proxy and added it to the Proxy section in order to be aligned with the whole walkthrough process.

If the "way" it's explained is OK, then I will create a next PR with the full setup.


robjloranger:
Hey @nunix, just wanted to touch base on this one. Did you intent to close this PR and open a second one? Or will you make further changes here with the goal of merging, and then create a second PR with other changes? Just want to clarify your first comment here. :smile:

nunix:
Hi Robbie,
hmm good question, I definitively want/will make more changes. So I can open a new one I guess.
This one can be closed, and if you already have comments on this one, please let me know as I will incorporate them for my next PR then.
Thank you 😊

robjloranger:
I'll save them for the next one, I forget what they were. Just spelling I think. We don't have to close this, unless you want to start fresh. I was confused about what you wanted to do with it.

docs/content/walkthrough.md
```diff
@@ -86,9 +86,41 @@ the proxy is using in-memory storage, which is only suitable for trying out the
 for a short period of time, as you will quickly run out of memory and the storage
 doesn't persist between restarts.
 
+### With Docker
+In order to run Athens localy in our dev machine with Docker (proxy enabled), we need first to create a directory that will store the persitant modules.
```
robjloranger: I think this line might be more clear as something like `In order to run the Athens Proxy using docker, we need to first...`.

nunix: updated it, thanks a lot, it is better to read indeed.

PR #663, opened 2018-09-14
redirect other domain requests to docs.gomods.io

### LINKED ISSUE #651: Visiting https://gomods.io produces a certificate warning
When visiting https://gomods.io, users are presented with a certificate warning because a self-signed certificate for `bryankellar.com` is being served. Visiting the HTTP version at http://gomods.io yields the following error page from a cPanel instance:

default_web_site_page

The same issue is present with the `www` subdomain as well. It's possible that this is a DNS misconfiguration.

**To Reproduce**
- Visit https://gomods.io , https://www.gomods.io, to see the TLS issues and http://gomods.io, http://www.gomods.io in any browser for the cPanel error showed above.
- `openssl s_client -connect gomods.io:443` can help verify that the right cert is being served

**Expected behavior**
I expect to be presented with an official landing page, or redirected to https://docs.gomods.io

robjloranger:
I used a 303 status code as the pages currently hosted should always be requested at `docs.gomods.io` not any of the other possible domains. `gomods.io | gomods-athens.netlify.com`.

will close #651

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -8,16 +8,16 @@
     environment = { HUGO_VERSION = "0.37" }
 
 [[redirects]]
-  from = https://gomods.io/*
-  to = https://docs.gomods.io/:splat
+  from = "https://gomods.io/*"
+  to = "https://docs.gomods.io/:splat"
   status = 303
 
 [[redirects]]
-  from = http://gomods.io/*
-  to = https://docs.gomods.io/:splat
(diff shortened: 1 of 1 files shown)


robjloranger:
Well it's not working yet but may take some time to propagate. At the least the redirect from the `gomods-athens.netlify.com` should work. 🤞

Issue #646, opened 2018-09-11
The master build seems to be failing


arschles:
**Describe the bug**

Travis CI builds for the master branch seem to be failing. Here is a recent failure: https://travis-ci.org/gomods/athens/builds/427422588?utm_source=email&utm_medium=notification

They aren't failing on PR builds, though.
**Error Message**

```console
Error: file not found
       1: cmd/logger/logger.go:294:logger.LogIf()
       2: cmd/fs-v1.go:963:cmd.(*FSObjects).listDirFactory.func1()
       3: cmd/tree-walk.go:145:cmd.doTreeWalk()
       4: cmd/tree-walk.go:253:cmd.startTreeWalk.func1()
```

This log output is [here](https://travis-ci.org/gomods/athens/builds/427422588#L1390-L1394)

**To Reproduce**

Run tests on the master branch, or merge a PR so that a build against master kicks off

**Expected behavior**

I expected master branch builds to pass. They should pass on every build against master.

**Environment (please complete the following information):**
 - OS: Travis CI linux (iirc this is an Ubuntu 16 64 bit distro)
 - Go version : 1.11
 - Buffalo Version : 0.12.4
 - Olympus or proxy version : tip of master

**Additional context**

This looks like an issue specific to master because PR builds pass before we merge them, so I'm suspecting that we're seeing a problem with configuration that only pops up for master builds.

penDerGraft:
Noticed a couple of things looking at this locally. 

1. The `make setup-dev-env` command modifies `go.mod` and `go.sum`, by adding golint and its indirect deps (assuming that's because of [this line here](https://github.com/gomods/athens/blob/master/scripts/get_dev_tools.sh#L7)? )
2. The builds started failing after the removal of the dep files [here](https://travis-ci.org/gomods/athens/builds/423637774)

Wondering if in the changeover somehow golint didn't get added to the .mod and .sum files. I think it's possible that just running `make setup-dev-env` locally and committing the modified `go.mod` and `go.sum` files will fix the build. 

Side note that I still don't think this addresses the error above, but it doesn't look like that section of the build is actually failing. Everything seems green until it tries to deploy.

marwan-at-work:
Fixed by 
https://github.com/gomods/athens/pull/679
https://github.com/gomods/athens/pull/678
https://github.com/gomods/athens/pull/676
https://github.com/gomods/athens/pull/675
https://github.com/gomods/athens/pull/674
https://github.com/gomods/athens/pull/635

PR #665, opened 2018-09-14
add golint (and its deps) to mod and sum files

penDerGraft:
Based on discussion in #646 it seems like these deps got excluded from these files somehow and are being introduced during the build steps. I think adding these should fix the build. 

I'm still new to modules so feel free to close the PR if this is unexpected behavior.


penDerGraft:
At second glance, I think there are additional build errors. May still want to have this PR though since the files are getting modified locally as well.

manugupt1:
I have a couple of questions

1. Did you not get checksum mismatch error?
https://github.com/gobuffalo/suite/issues/13

I suspect this bug is related to the proxy only though.

2.  Should we not update the vendor folder if we are adding and we have a vendor folder?

penDerGraft:
Thanks for the feedback! @manugupt1 Yep, running into that now. Going to dig into this to see if that is resolvable.

penDerGraft:
Closing since #674 seems to solve the same issue.

PR #666, opened 2018-09-15
Add the Host config to setup

### LINKED ISSUE #655: First log message always "Starting application at http://127.0.0.1:3000" despite config
**Describe the bug**
I was playing around with the configuration and changed the port of the proxy to 3002. Despite doing this I still got the "Starting application at http://127.0.0.1:3000" log message at the start. It turns out that no matter what the configuration is this message will always display. This occurs for both Olympus and proxy.

Looking into this further it appears to be because the buffalo option of `Addr` but not `Host` is not given. 
<pre>
	app := buffalo.New(buffalo.Options{
		Env: ENV,
		PreWares: []buffalo.PreWare{
			cors.Default().Handler,
		},
		SessionName: "_athens_session",
		Logger:      blggr,
		Addr:        conf.Proxy.Port,
	})
</pre>

**To Reproduce**
Change the Port variable in the configuration file and start the server. 

**Expected behavior**
I expected the log to be consistent with where the server actually binded to.

 - OS: darwin 64
 - Go version : 1.11
 - Buffalo Version : v0.12.4
 - Olympus or proxy version : commit 57b11e848969918f611db498592b7108a274ab00

**Discussion**
I can provide the PR fixing this but I was wondering how we should handle the "Host" buffalo option. Should it be an extra configuration? Or should something else be done?

zkry:
Fixes #655
@arschles @rchakra3 
Here's my first go at this. I wasn't quite sure what to do with the port option and the ':' prefix as changing that would leave previously created configs broken. Should I update this PR so the port doesn't have the ':' and update the setting up of the options accordingly?


cmd/olympus/actions/app.go
```diff
@@ -106,7 +105,8 @@ func App(conf *config.Config) (*buffalo.App, error) {
 	blggr := log.Buffalo(bLogLvl)
 
 	app := buffalo.New(buffalo.Options{
-		Addr: port,
+		Addr: conf.Olympus.Port,
+		Host: "http://127.0.0.1" + conf.Proxy.Port,
```
rchakra3: I feel we could remove the colon from the Port config and just use the Host option. What do you think?

Also - I think we should be listening on 0.0.0.0 instead of the loopback by default

zkry: Ok, that makes sense to change the host to 0.0.0.0. As for the other comment, I'm not sure I completely understand. 
In the config file should I change the `Port = ":3000"` to `Port = "3000"` (for all port related options?) and then change the initialization to:
```
	app := buffalo.New(buffalo.Options{
		Env: ENV,
		PreWares: []buffalo.PreWare{
			cors.Default().Handler,
		},
		SessionName: "_athens_session",
		Logger:      blggr,
		// --- Addr:  line removed ---
		Host: "http://0.0.0.0:" + conf.Proxy.Port,
	})
```
It seems though that doing this though doesn't make it bind to the configured port as the 'Host:' config option doesn't seem to be used when setting up the buffalo app.

marwan-at-work: Yep, let's keep it as port until someone files an issue with a specific reason as to why we need to include the host.

zkry:
Great👍. I'll get this one merged right away. Thanks

michalpristas:
Thanks @zkry getting this in. good job there

PR #667, opened 2018-09-15
Patch points of test flag leakage.

### LINKED ISSUE #653: testing flags leaking into cmd/proxy
**Describe the bug**
```
Daves-MBP(~/devel/athens) % ./proxy --help
Usage of ./proxy:
  -config_file string
        The path to the config file (default "../../config.toml")
  -httptest.serve string
        if non-empty, httptest.NewServer serves on this address and blocks
  -test.bench regexp
        run only benchmarks matching regexp
  -test.benchmem
        print memory allocations for benchmarks
  -test.benchtime d
        run each benchmark for duration d (default 1s)
  -test.blockprofile file
        write a goroutine blocking profile to file
  -test.blockprofilerate rate
        set blocking profile rate (see runtime.SetBlockProfileRate) (default 1)
  -test.count n
        run tests and benchmarks n times (default 1)
  -test.coverprofile file
        write a coverage profile to file
  -test.cpu list
        comma-separated list of cpu counts to run each test with
  -test.cpuprofile file
        write a cpu profile to file
  -test.failfast
        do not start new tests after the first test failure
  -test.list regexp
        list tests, examples, and benchmarks matching regexp then exit
  -test.memprofile file
        write an allocation profile to file
  -test.memprofilerate rate
        set memory allocation profiling rate (see runtime.MemProfileRate)
  -test.mutexprofile string
        write a mutex contention profile to the named file after execution
  -test.mutexprofilefraction int
        if >= 0, calls runtime.SetMutexProfileFraction() (default 1)
  -test.outputdir dir
        write profiles to dir
  -test.parallel n
        run at most n tests in parallel (default 8)
  -test.run regexp
        run only tests and examples matching regexp
  -test.short
        run smaller test suite to save time
  -test.testlogfile file
        write test action log to file (for use only by cmd/go)
  -test.timeout d
        panic test binary after duration d (default 0, timeout disabled)
  -test.trace file
        write an execution trace to file
  -test.v
        verbose: print additional output
  -testify.m string
        regular expression to select tests of the testify suite to run
```

**Expected behavior**

`testing` package flags should not appear in the final binary.

zkry:
Fixes #653.
@michalpristas 

After looking into this there were two points of leakage.

One was the GetConfLogErr function in `pkg/config/parse.go`. This was a helper function used in several places. To fix this I wrote it out in the long form in the files it was used once and copied it to `pkg/module/filter_test.go` where it was used several times.

The other point of leakage was from the buffalo/suite.Model object being into the `test_suite.go` files of fs, mem, minio, and mongo. This I was a lot less sure on how to fix. It seemed to me that the `*suite.Model` could be removed from the TestSuite structs with no problem though I wasn't sure.

### DIFF
--- a/cmd/olympus/actions/actions_test.go
+++ b/cmd/olympus/actions/actions_test.go
@@ -23,7 +23,10 @@ type ActionSuite struct {
 }
 
 func Test_ActionSuite(t *testing.T) {
-	conf := config.GetConfLogErr(testConfigFile, t)
+	conf, err := config.GetConf(testConfigFile)
+	if err != nil {
+		t.Fatalf("Unable to parse config file: %s", err.Error())
+	}
 	app, err := App(conf)
 	if err != nil {
 		t.Fatalf("Failed to initialize app: %s", err)
--- a/cmd/proxy/actions/actions_test.go
+++ b/cmd/proxy/actions/actions_test.go
@@ -17,7 +17,10 @@ type ActionSuite struct {
 }
 
 func Test_ActionSuite(t *testing.T) {
-	conf := config.GetConfLogErr(testConfigFile, t)
+	conf, err := config.GetConf(testConfigFile)
+	if err != nil {
+		t.Fatalf("Unable to parse config file: %s", err.Error())
+	}
 	app, err := App(conf)
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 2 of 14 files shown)


marwan-at-work:
@zkry could you address the CI failure. Thanks!

zkry:
No problem. Sounds good, I'll get this merged right away. Thanks 🙂

PR #668, opened 2018-09-17
Add HTTP_PROXY environment variable to go_get_fetcher

tnextday:
Pass the `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` environment variables to the go command in `go_get_fetcher`.


tnextday:
@marwan-at-work  In China, many websites are inaccessible because of [GFW](https://en.wikipedia.org/wiki/Great_Firewall), such as gloang.org. So we need to use a proxy to download the package.

tnextday:
@marpio I have fixed it.

psy-core:
Does this feature available in the docker latest image? how can I use it?

marwan-at-work:
@psy-core it does. Just include this environment variable to the athens process/container.

psy-core:
> @psy-core it does. Just include this environment variable to the athens process/container.

Thanks a lot! :-)

Issue #612, opened 2018-08-31
Park Olympus Until Designs Are Finalized


marwan-at-work:
I'm not sure what this issue would entail exactly, but the current Olympus directory is quite outdate and not being used at all. It also causes some builds to break when you do some refactoring for the Proxy. 

Should we remove Olympus and rewrite it once the designs/paper is complete and ready for dev? 

Or should we just update the CI/CD to ignore the `olympus` directory for now?

Or should we just leave things the way they are?

robjloranger:
I'm n favor of parking for now. It seems we are using only the proxy at the moment, and in the design doc from Brian there is question of whether we even need Olympus.

ref: https://docs.google.com/document/d/1F0cVKxX7kEzxjrr09moqFzRQ7FoNc-Z2igfsEIkp7mc/edit?usp=sharing

marwan-at-work:
cc @arschles

komuw:
we could move olympus to a new repo under`https://github.com/gomods` until it's final fate is decided.

robjloranger:
@komuw this would just be not working on Olympus until we set the design. No need to move the code.

komuw:
> It also causes some builds to break when you do some refactoring for the Proxy.

if builds break because of code that you should not be maintaining(olympus), then you actually have to start maintaining that code.

arschles:
I saw up above in the thread that we could omit it from the build, and I'm cool with that. what do folks think about also adding something to the cmd/olympus README?

robjloranger:
I like it, just skip in the build for now and leave a note until we are ready to move forward on it.

arschles:
As of #772, we're going to focus on the global proxy and not on the registry (which is what Olympus was). Since this issue was for parking Olympus until designs are finalized, that's obsolete now. I'm going to close this and open a new issue for removing Olympus code

PR #675, opened 2018-09-19
Olympus: stop docker push

marwan-at-work:
Stops creating Olympus images when merging to master. 
Updates https://github.com/gomods/athens/issues/612


marwan-at-work:
@michalpristas will do!

PR #680, opened 2018-09-19
Minio conf + tests refactoring

### LINKED ISSUE #671: Minio Port conflict with vscode extensions server
When we run `make alldeps` to do some `docker-compose` magic to bring up the services, Minio tries to bind to port `9000` which is being used by vscode for Its extension server. Which In turn will cause,
```
Starting athensdev_minio_1 ... error

ERROR: for athensdev_minio_1  Cannot start service minio: driver failed programming external connectivity on endpoint athensdev_minio_1 (b060df5b66108f320faf15e484a1036ea543b257a4c720c3e0839b17279cfa6a): Error starting userland proxy: listen tcp 0.0.0.0:9000: bind: address already in use
```
If we kill the process on port `9000 `
```
07:02:52 ▶ fuser -k 9000/tcp
9000/tcp:             4710
```
In vscode
vscode

**Environment**
##### OS 
NAME="Ubuntu"
VERSION="18.04.1 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.1 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic

Linux drogon 4.15.0-33-generic Ubuntu SMP Wed Aug 15 16:00:05 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
 ##### Athens Commit Hash
c740fcb26cc0192391c10b4b0185d62b78c1410b

**Additional context**
If this is faced by many people, The port which is bound for minio can be changed to improve the UX.

marpio:
Changes the minio default port to 9001

Tests refactoring and cleanup:
  - it wasn't possible to execute pkg modulestorage tests multiple times because the cleanup func couldn't find the config file
  - also on windows the delete test for file system didn't work because `version :="delete" + time.Now().String()` doesn't produce a valid path

Fixes #671


config.example.toml
```diff
@@ -180,7 +180,7 @@ EnableCSRFProtection = false
     [Storage.Minio]
         # Endpoint for Minio storage
         # Env override: ATHENS_MINIO_ENDPOINT
-        Endpoint = "minio.example.com"
+        Endpoint = "127.0.0.1:9000"
```
michalpristas: 9001?

.travis.yml
```diff
@@ -20,7 +20,7 @@ before_script:
   - mkdir -p $ATHENS_DIR && rsync -azr . $ATHENS_DIR && cd $ATHENS_DIR
   - make setup-dev-env
   - wget "https://dl.minio.io/server/minio/release/linux-amd64/minio"
-  - chmod +x minio && nohup ./minio server . &
+  - chmod +x minio && nohup ./minio server --address ":9001" . &
```
michalpristas: can you add a comment why we're moving from default port

config.example.toml
marpio: Since it's an example config i've put the default minio port here similar to the mongo part.
I've only changed it in our docker-compose file and the test config so it doesn't conflict with other services on a dev machine.

michalpristas: yes i understand intentions.
but latest change here: https://github.com/gomods/athens/pull/678/files
uses example as a default for proxy docker

marpio: @michalpristas I'll revert it to the old value (minio.example.com) so it stays as generic as possible. As per slack discussion - we should probably use the config.test.toml in our docker img.

scripts/ps/create_default_config.ps1
```diff
@@ -1,6 +1,6 @@
 $repoDir = Join-Path $PSScriptRoot ".." | Join-Path -ChildPath ".."
 if (-not (Join-Path $repoDir config.toml | Test-Path)) {
-    $example = Join-Path $repoDir config.example.toml
+    $example = Join-Path $repoDir config.test.toml
```
michalpristas: there are few differences in example and test:
timeout (each) in test is set to 1, it is definitely too low for test env.

PR #681, opened 2018-09-19
Use config for trace url

marwan-at-work:
This gets rid of the last `env` config (I think) 
For now we use Jaegar is the only trace exporter, but I'd like to make it modular so that users can input their own exporter if they like, such as stack driver.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -69,7 +69,6 @@ Timeout = 300
 # Env override: ATHENS_ENABLE_CSRF_PROTECTION
 EnableCSRFProtection = false
 
-
 [Proxy]
     # StorageType sets the type of storage backend the proxy will use.
     # Possible values are memory, disk, mongo, gcp, minio
@@ -127,6 +126,10 @@ EnableCSRFProtection = false
     # .cache directory in the Go image).
     # Env override: ATHENS_NETRC_PATH
     NETRCPath = ""
+
+    # TraceExporterURL is the URL to which Athens populates distributed tracing 
+    # inforamtion such as Jaegar. 
+    TraceExporterURL = ""
 
 [Olympus]
     # StorageType sets the type of storage backend Olympus will use.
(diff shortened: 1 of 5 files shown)


pkg/config/parse.go
```diff
@@ -25,6 +25,7 @@ type Config struct {
 	CloudRuntime         string `validate:"required" envconfig:"ATHENS_CLOUD_RUNTIME"`
 	FilterFile           string `envconfig:"ATHENS_FILTER_FILE"`
 	EnableCSRFProtection bool   `envconfig:"ATHENS_ENABLE_CSRF_PROTECTION"`
+	TraceExporterURL     string `envconfig:"TRACE_EXPORTER"`
```
timraymond: Should this also have the `ATHENS_` prefix for consistency? I remember that `envconfig` has a way to globally set a prefix, so it could be that the other two config options shouldn't be hardcoding the prefix.

config.example.toml
```diff
@@ -128,6 +127,10 @@ EnableCSRFProtection = false
     # Env override: ATHENS_NETRC_PATH
     NETRCPath = ""
 
+    # TraceExporterURL is the URL to which Athens populates distributed tracing 
+    # inforamtion such as Jaegar.
```
timraymond: typo: information

pkg/config/parse.go
marwan-at-work: Yeah probably. I'll add it.

config.example.toml
```diff
@@ -128,6 +127,11 @@ EnableCSRFProtection = false
     # Env override: ATHENS_NETRC_PATH
     NETRCPath = ""
 
+    # TraceExporterURL is the URL to which Athens populates distributed tracing 
+    # information such as Jaegar.
```
michalpristas: /s/Jaegar/Jaeger

marwan-at-work:
@michalpristas fixed the typo. Let me know if it's all good. Thanks!

Issue #526, opened 2018-08-20
Ensure goget is part of tracing


marwan-at-work:
Once https://github.com/gomods/athens/issues/392 is resolved, let's ensure that concurrent go-gets are properly traced as part of the entire request. This is because the request Context is not being passed to the cacheFill/storageFill operations which is a good thing because if the user abandons a request, we want to continue doing the cacheFill/storageFill. However, I'm not sure if we want to copy values from the old context to the new one (or if OpenCensus doesn't care about CancelledContexts)

arschles:
@manugupt1 was this done as part of https://github.com/gomods/athens/pull/604?

marwan-at-work:
This is a follow up to 604. I can work on it.

PR #682, opened 2018-09-19
Trace stash operations

marwan-at-work:
Fixes https://github.com/gomods/athens/issues/526

### DIFF
--- a/pkg/stash/stasher.go
+++ b/pkg/stash/stasher.go
@@ -6,13 +6,15 @@ import (
 
 	"github.com/gomods/athens/pkg/errors"
 	"github.com/gomods/athens/pkg/module"
+	"github.com/gomods/athens/pkg/observ"
 	"github.com/gomods/athens/pkg/storage"
+	"go.opencensus.io/trace"
 )
 
 // Stasher has the job of taking a module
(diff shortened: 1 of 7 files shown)


pkg/stash/stasher.go
```diff
@@ -35,10 +37,16 @@ type stasher struct {
...
-func (s *stasher) Stash(mod, ver string) error {
+func (s *stasher) Stash(ctx context.Context, mod, ver string) error {
 	const op errors.Op = "stasher.Stash"
-	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*10)
+	_, span := observ.StartSpan(ctx, op.String())
+	defer span.End()
+
+	// create a new context that ditches whatever deadline the caller passed
+	// but keep the tracing info so that we can properly trace the whole thing.
+	ctx, cancel := context.WithTimeout(trace.NewContext(context.Background(), span), time.Minute*10)
```
manugupt1: This is awesome 💃 

I would have put this into an observ package so that we can look at it always something like 
`observ.ContextFromSpan(*span) observabilityContext.` but it dis upto you!

manugupt1: I figured as much! that is why I think it does not really matter.

PR #683, opened 2018-09-20
Make example config the default for dev

marwan-at-work:
We should make onboarding very easy, and asking people to create their own config off the bat is very difficult. 

The config.example.toml works out of the box (and we should always keep it so).

### DIFF
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -10,7 +10,7 @@ import (
 )
 
 var (
-	configFile = flag.String("config_file", filepath.Join("..", "..", "config.toml"), "The path to the config file")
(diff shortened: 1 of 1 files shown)


cmd/proxy/main.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 var (
-	configFile = flag.String("config_file", filepath.Join("..", "..", "config.toml"), "The path to the config file")
+	configFile = flag.String("config_file", filepath.Join("..", "..", "config.example.toml"), "The path to the config file")
```
michalpristas: dont' you need to change dockerfiles so it does not do `copy config.example.toml config.toml` ?

marwan-at-work: @michalpristas the Dockerfile overrides this default by passing the flag explicitly.

PR #684, opened 2018-09-20
Git ignore .env file

marwan-at-work:
Now that we have a config file, dotenv is not necessary to be checked. 

I'm thinking more about the development flow and how we should tweak config things while writing code locally. And might open an issue or two for the new config flow.


michalpristas:
can you double check docs, so we're sure we're not telling peeps to source env or something like that before mering

PR #691, opened 2018-09-20
Rename config.example to config.dev and rm config.test

marpio:
Renamed the config.example.toml -> config.dev.toml
I've also deleted config.test.toml since config.dev.toml can be used instead


cmd/olympus/actions/actions_test.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 var (
-	testConfigFile = filepath.Join("..", "..", "..", "config.test.toml")
```
michalpristas: i'm thinking about keeping it test in test and .dev. as a default in a code

marwan-at-work: @michalpristas I don't mind us deleting `config.test.toml` unless there's something about the test.toml file that is different from the regular one? We can always introduce it later on if the case turns out to be true.

marpio: At the moment they are identical.

michalpristas: if they're identical then one of them is redundant

marwan-at-work: yep, so let's merge this and if we ever need a special config values for tests we can put this file back, but i'd argue that we can just pass those special vals directly in the tests.

docs/content/walkthrough.md
```diff
@@ -59,7 +59,7 @@ $ mkdir -p $(go env GOPATH)/src/github.com/gomods
 $ cd $(go env GOPATH)/src/github.com/gomods
 $ git clone https://github.com/gomods/athens.git
 $ cd athens
-$ GO111MODULE=off go run ./cmd/proxy -config_file=config.example.toml &
+$ GO111MODULE=off go run ./cmd/proxy -config_file=config.dev.toml &
```
marwan-at-work: ops just realized we didn't need this flag anymore. will pr it after.

Issue #637, opened 2018-09-07
could not find assets/application.852998aa3b53a0897008.css


elvizlai:
**Describe the bug**
docker run

**Error Message**
```console
could not find assets/application.852998aa3b53a0897008.css
```

**To Reproduce**

**Expected behavior**

**Environment (please complete the following information):**
 - OS: [e.g. Linux 64bit]
 - Go version :
 - Buffalo Version :
 - Olympus or proxy version :

**Additional context**

robjloranger:
@elvizlai Thanks for the bug report.

Is this for the proxy or olympus?

In both cases this is a known issue, there is no front end for either the proxy or olympus at this time. So no assets are present.

It would be great to somehow disable the application from trying to find them, perhaps we missed something during the removal? Are you interested in trying to take that on?

Small note: in the future if there are portions of the bug report template that you feel are not related to your issue, you can just remove them entirely.

elvizlai:
the proxy report this err log.
it looks like when packing go bin using `github.com/gobuffalo/packr` missing the generated css.

manugupt1:
@gomods/maintainers we can close this!

marpio:
Assets have been removed. Closing.

PR #693, opened 2018-09-20
Proxy remove assets and config scripts

marwan-at-work:
The Proxy is not using any of these assets so there's no need in maintaining them. 

My idea is that the Proxy should be a pure API and we can create a separate UI server for it after beta.


manugupt1:
😍 this! I totally agree!
Can you verify this and close it as well? https://github.com/gomods/athens/issues/637

cmd/proxy/actions/home.go
```diff
@@ -5,5 +5,5 @@ import (
 )
 
 func proxyHomeHandler(c buffalo.Context) error {
-	return c.Render(200, proxy.HTML("index.html"))
+	return c.Render(200, proxy.JSON("Welcome to The Athen Proxy"))
```
manugupt1: We should probably remove this handler as well, since you are asking for an API.

The `healthz` endpoint should be enough to verify that

marwan-at-work: @manugupt1 We can keep this instead of returning a 404 on the home page which I know is ok for APIs usually.

cmd/proxy/Dockerfile
```diff
@@ -8,19 +8,14 @@ WORKDIR $GOPATH/src/github.com/gomods/athens
...
-	./scripts/create_default_config.sh
+RUN GO111MODULE=on CGO_ENABLED=0 go build -mod=vendor -o /bin/athens-proxy ./cmd/proxy
 
 FROM alpine
 
 ENV GO111MODULE=on
 
 COPY --from=builder /bin/athens-proxy /bin/athens-proxy
 COPY --from=builder /go/src/github.com/gomods/athens/config.dev.toml /config/config.toml
-COPY --from=builder /go/src/github.com/gomods/athens/cmd/proxy/locales /go/src/github.com/gomods/athens/cmd/proxy/locales
```
michalpristas: could you please align handling of config? 
in olympus we do copy from workdir to /config in builder
and then copy from `/config/config.dev.toml` to `/config.toml`
while here we go directly to git repo dir

marwan-at-work: @michalpristas this is one reason I'd like us to park Olympus. It's very outdated code at this point and wouldn't work anyway. I updated it tho :) .

michalpristas: thanks, i would do it in a separate PR if you wouldnt though

PR #710, opened 2018-10-01
Use go mod download -json

marpio:
We can use the `-json` flag on the `go mod download` cmd to get the zip, mod and info paths directly from go so that we don't have to do that. Also the encoding of the uppercase paths isn't needed anymore since go does it for us.


pkg/module/go_get_fetcher.go
```diff
@@ -22,6 +22,18 @@ type goGetFetcher struct {
 	goBinaryName string
 }
 
+type goModule struct {
+	Path     string // module path
```
michalpristas: can you annotate with `json:`

pkg/module/all_test.go
```diff
@@ -10,8 +10,8 @@ import (
 
 const (
 	// these values need to point to a real repository that has a tag
-	repoURI = "github.com/arschles/assert"
```
michalpristas: intentional?

pkg/module/go_get_fetcher.go
```diff
@@ -86,38 +120,39 @@ func Dummy(fs afero.Fs, repoRoot string) error {
 	return nil
 }
 
-// given a filesystem, gopath, repository root, module and version, runs 'vgo get'
+// given a filesystem, gopath, repository root, module and version, runs 'go mod download -json'
 // on module@version from the repoRoot with GOPATH=gopath, and returns a non-nil error if anything went wrong.
-func getSources(goBinaryName string, fs afero.Fs, gopath, repoRoot, module, version string) error {
-	const op errors.Op = "module.getSources"
+func downloadModule(goBinaryName string, fs afero.Fs, gopath, repoRoot, module, version string) (*goModule, error) {
```
michalpristas: do we need to return pointer?

pkg/module/zip_read_closer.go
```diff
@@ -0,0 +1,50 @@
...
+	return ClearFiles(rc.fs, rc.goPath)
+}
+
+func (rc *zipReadCloser) Read(p []byte) (n int, err error) {
+	return rc.zip.Read(p)
+}
+
+// ClearFiles deletes all data from the given fs at path root
+// This function must be called when zip is closed to cleanup the entire GOPATH created by the diskref
+func ClearFiles(fs afero.Fs, root string) error {
```
michalpristas: no need to be exposed

pkg/module/all_test.go
marpio: yes, i've changed it to NYTimes/gizmo because it's an example of a module name which needs to be encoded (!n!y!times) and I wanted to cover that as well.

pkg/module/all_test.go
```diff
@@ -10,6 +10,7 @@ import (
 
 const (
 	// these values need to point to a real repository that has a tag
+	// github.com/NYTimes/gizmo is a example of a path that needs to be encoded so we can cover that case as well
```
michalpristas: 👍

pkg/module/zip_read_closer.go
marpio: it's used here as well:
https://github.com/gomods/athens/blob/7ee25d59d6930be4e963c332c949b411a6915ebd/pkg/download/upstream_lister.go#L61

pkg/module/all_test.go
marwan-at-work: maybe we should create our own upper case repo in github.com/athens-artifacts and write a test with it. Happy to do that as a next pr

marpio: Great idea @marwan-at-work !

PR #719, opened 2018-10-02
Clean up store before test.

DonMcNamara:
I had a failing test due to some stale data in the mongo store. I'm not exactly sure how it happened.

This PR will clean up the stores before the tests are run to avoid any stale data causing failure.


michalpristas:
Thank you @DonMcNamara

PR #720, opened 2018-10-02
Proxy: Add HGRC file mount

### LINKED ISSUE #716: Support .hgrc files to access private Mercurial repos
Authorize proxy to access a private Mercurial repo.

Proxy supports .netrc file to access a private repo.
Mercurial has its own RC file - `.hgrc` - which can be used to store credentials.
We should add support for that similar to `.netrc`

Ref: https://github.com/gomods/athens/pull/528

t-tomalak:
Fixes #716 

Added support for .hgrc file also changed previous implementation, and now should be more flexible for other dot files. 

Fixed one issue in proxy.md file where wrong env param for NETRC was given.


cmd/proxy/actions/app.go
```diff
@@ -41,7 +41,11 @@ func App(conf *config.Config) (*buffalo.App, error) {
 
 	// mount .netrc to home dir
 	// to have access to private repos.
-	initializeNETRC(conf.Proxy.NETRCPath)
+	initializeDotFile(conf.Proxy.NETRCPath)
```
marwan-at-work: dotFiles can be many things (.env, .gitignore, etc). So let's call this something more clear of the task: moving an auth file to the root directory. So something like: `initializeAuthFile`.

cmd/proxy/actions/dotfile.go
```diff
@@ -0,0 +1,30 @@
...
+	if err != nil {
+		log.Fatalf("could not read %s: %v", path, err)
+	}
+
+	hdir, err := homedir.Dir()
+	if err != nil {
+		log.Fatalf("could not get homedir: %v", err)
+	}
+	rcp := filepath.Join(hdir, filepath.Base(path))
+	ioutil.WriteFile(rcp, fileBts, 0666)
```
michalpristas: ignoring error we should at least log what's wrong. it will help a person troubleshooting the issue.
 also please mention in the doc that any .netrc/hgrc file in home will be replaced by what is in `path`

t-tomalak: I don't saw that missed error ( it was also missed previously), will add that.

About the documentation, we don't know in that function which file will be replaced, and this description like this is added in config.dev.toml. Should I maybe add it also to proxy.md I mean this part with override?

cmd/proxy/actions/app.go
t-tomalak: Should I put this function in cmd/proxy/actions/auth.go?

cmd/proxy/actions/dotfile.go
michalpristas: we know that files copied using these methods will be replaced so i would start with that and add a warning to each of them. right at the place, you modified proxy.md

cmd/proxy/actions/app.go
michalpristas: or init.go, both are ok i guess

cmd/proxy/actions/dotfile.go
manugupt1: Is it a hard req to have 0666 as permission, anyone from the world can read it.
I think the best is 0600 or 0660. Can you check what is recommended by hgrc or change the file permission.

t-tomalak: It seems like previously for .netrc file got to much permission (https://www.labkey.org/Documentation/wiki-page.view?name=netrc), will fix that after investigating minimal permission for .hgrc

cmd/proxy/actions/app.go
```diff
@@ -41,7 +41,11 @@ func App(conf *config.Config) (*buffalo.App, error) {
 
 	// mount .netrc to home dir
 	// to have access to private repos.
-	initializeNETRC(conf.Proxy.NETRCPath)
+	initializeAuthFile(conf.Proxy.NETRCPath)
```
manugupt1: This is like a super tiny nit:

you can probably do 

`initializeAuthFile(conf.Proxy.NetRCPath, conf.HGRCPath)`

and change the function signature to 

`func initializeAuthFile(files []string)`

michalpristas:
@t-tomalak looking good, will merge after you bring this even with master and address manus comment. either by doing it or saying that this is fine. 
for me, it is fine as it is

cmd/proxy/actions/app.go
t-tomalak: I think it's fine as it is, the loop inside initializeAuthFile is no needed for now when we have two files, maybe in future when there will be more file initialized we could think of that

t-tomalak:
@michalpristas Should I merge master again as is there new changes on it?

Issue #714, opened 2018-10-01
Tracing exporter for Datadog


marpio:
**Is your feature request related to a problem? Please describe.**
Similar to #642 we could add support for Datadog

**Describe the solution you'd like**
It should be possible to export traces to Datadog.
https://opencensus.io/guides/exporters/supported-exporters/go/datadog/

bhaveshpraveen:
Hey, I'd like to work on this.

manugupt1:
Awesome! Feel free to shoot questions to me

Cheers
Manu
________________________________

PR #723, opened 2018-10-03
Tracing Exporter for Datadog #714

bhaveshpraveen:
Tracing Exporter for Datadog #714 and Support for other tracers

**How is the fix applied?**

Every trace exporter using opencensus has two mandatory methods, `NewExporter()` and `Flush()`(`Stop()` in case of datadog). Different Services could have different names for the above two functions. I wanted to keep the calls consistent:-
- So I created an interface `Exporter`. Whenever a new trace exporter has to be added, it has to conform to the `Exporter` interface which has two mandatory methods `RegisterTraceExporter()` and `Flush()`. 
- `RegisterTraceExporter()` would create and register the trace exporter. `Flush()` makes sure that all traces will be written before the process ends.
- So each service(Datadog, Jaeger, etcetera) would have its own`RegisterTraceExporter()` and `Flush()` methods.  
- The type of trace service (datadog, jaeger) is determined by the the call to `RetriveTracer()` which uses the environment variable `ATHENS_TRACE_EXPORTER`. `RetrieveTracer()` returns a `struct` of a particular service(datadog, jaejer) which conforms to `Exporter` interface.

So the code for running a new service would look something like this:-

```
// error handling has been ignored in this snippet for brevity
exporter, err := observ.RetrieveTracer(conf.TraceExporter)

err = exporter.RegisterTraceExporter(conf.TraceExporterURL, Service, ENV)
defer exporter.Flush()	
```

Previously there was support only for Jaejer service, now any service can be freely added.
The code for adding support to new trace exporters(datadog in the example) would look something like this
```
type datadogTraceExporter struct {
	exporter (*(datadog.Exporter))
}
func (d datadogTraceExporter) RegisterTraceExporter(...) error {
    // Code to register the service.
    // This might be different for different services
}
func (d datadogTraceExporter) Flush (){
    // Code to flush the export tracer
}
```
And finally add the service to `RetrieveTracer()`
```
func RetrieveTracer(TraceExporter string) (Exporter, error) {

	const op errors.Op = "RegisterTracer"

	switch TraceExporter {
        // To add a new service, `jaejer`
        case "jaeger":
		return jaegerTraceExporter{}, nil

       // Support for other services
	default:
		return nil, errors.E(op, "Exporter not specified. Traces won't be exported")
	}
}
```

**Fixes issue:  #714**


pkg/observ/observ.go
```diff
@@ -17,16 +18,62 @@ type observabilityContext struct {
 	spanCtx context.Context
 }
 
-// RegisterTraceExporter returns a jaeger exporter for exporting traces to opencensus.
+// jaegerTraceExporter is used to hold the trace exporter for Jaeger
+type jaegerTraceExporter struct {
+	exporter (*(jaeger.Exporter))
```
marwan-at-work: remove extra parens

pkg/observ/observ.go
```diff
@@ -17,16 +18,62 @@ type observabilityContext struct {
...
+}
+
+// datadogTraceExporter is used to hold the trace exporter for Datadog
+type datadogTraceExporter struct {
+	exporter (*(datadog.Exporter))
```
marwan-at-work: ditto: `*datadog.Exporter` should suffice

pkg/observ/observ.go
```diff
@@ -17,16 +18,62 @@ type observabilityContext struct {
...
+}
+
+// Exporter might be a JaejerExporter, DataDogExporter, etc
```
marwan-at-work: s/JaejerExporter/JaegerExporter

pkg/observ/observ.go
```diff
@@ -17,16 +18,62 @@ type observabilityContext struct {
...
+// All Exporters *must* implement these two functions.
```
marwan-at-work: no need for this line of comment. The compiler enforces interface implementations.

pkg/observ/observ.go
```diff
@@ -17,16 +18,62 @@ type observabilityContext struct {
...
+// Flush method stops the trace exporter, this method needs to be
```
marwan-at-work: Put this on top of Flush(), since it's documentation specific for that method.

pkg/observ/observ.go
```diff
@@ -17,16 +18,62 @@ type observabilityContext struct {
...
+	RegisterTraceExporter(URL, service, ENV string) error
+	Flush()
+}
+
+// RetrieveTracer determines the type of TraceExporter service for exporting traces from opencensus
+// User can choose from multiple tracing services (datadog, jaegar)
+// TODO: Extend beyond jaeger and datadog
+func RetrieveTracer(TraceExporter string) (Exporter, error) {
+
+	const op errors.Op = "RegisterTracer"
```
marwan-at-work: let's call this `"RetrieveTracer"` since you changed the name of the function.

pkg/observ/observ.go
```diff
@@ -17,16 +18,62 @@ type observabilityContext struct {
...
+
+	switch TraceExporter {
+	case "jaeger":
+		return jaegerTraceExporter{}, nil
+	case "datadog":
+		return datadogTraceExporter{}, nil
+	default:
+		return nil, errors.E(op, "Exporter not specified. Traces won't be exported")
```
marwan-at-work: let's differentiate between an empty string, and an unsupported exporter. This way users know that if they put something like "stackdriver", that it's still not supported.

pkg/observ/observ.go
```diff
@@ -17,16 +18,62 @@ type observabilityContext struct {
...
+	d.exporter.Stop()
+}
+
+// RegisterTraceExporter creates a jaeger exporter for exporting traces to opencensus.
+// Currently uses the 'TraceExporter' variable in the config file.
 // It should in the future have a nice sampling rate defined
-// TODO: Extend beyond jaeger
-func RegisterTraceExporter(URL, service, ENV string) (*(jaeger.Exporter), error) {
+func (j jaegerTraceExporter) RegisterTraceExporter(URL, service, ENV string) error {
 	const op errors.Op = "RegisterTracer"
```
marwan-at-work: let's call this `jaeger.RegisterTraceExporter`

pkg/observ/observ.go
```diff
@@ -40,16 +87,37 @@ func RegisterTraceExporter(URL, service, ENV string) (*(jaeger.Exporter), error)
 	})
 
 	if err != nil {
-		return nil, errors.E(op, err)
+		return errors.E(op, err)
 	}
 
 	// And now finally register it as a Trace Exporter
-	trace.RegisterExporter(je)
+	trace.RegisterExporter(j.exporter)
```
marwan-at-work: let's move this duplication to a private function multiple implementations can use: 

```golang
func registerExporter(ex trace.Exporter) {
  trace.RegisterExporter(ex)
  if ENV == "development" {...}
}
```

manugupt1:
I like the changes so far. I have not tested it on datadog. But  a question that I have is that does not data-dog have an API key that you would add in the env var to actually send out traces.

For more reference, check this out  : https://docs.datadoghq.com/api/?lang=python#authentication

pkg/observ/observ.go
bhaveshpraveen: Hey, how do you reckon I solve this? I was thinking of creating a new interface method to register the tracer.

marwan-at-work: @bhaveshpraveen just like the function I have above. The `trace.Exporter` argument will accept both datadog and other exporters. 

Now that I took a second look, you actually dont need the Exporter interface in the observ package. The `RetrieveExporter` can just be `RegisterExporter` and it does all the registration for you, and just returns a `func()` that handles the "flushing" or "stopping"

bhaveshpraveen: @marwan-at-work : I think I understand what you're saying
So, the code would look something like this?
```
function_to_flush, err := RegisterExporter(...)
function_to_flush()
```

bhaveshpraveen:
@manugupt1 , @marwan-at-work Please check if its alright.

cmd/proxy/actions/app.go
```diff
@@ -75,12 +75,18 @@ func App(conf *config.Config) (*buffalo.App, error) {
...
+	// RegisterExporter returns a function that needs to be run for all services(datadog, jaeger)
+	// exporting traces from opencensus.
+	flushExporter, err := observ.RegisterExporter(conf.TraceExporter,
+		conf.TraceExporterURL,
+		Service, ENV)
 	if err != nil {
 		lggr.Infof("%s", err)
 	} else {
-		defer exporter.Flush()
+		defer flushExporter()
```
manugupt1: Why did you choose to return the function flush instead of the exporter object.

The reason I am asking is because if you read the code, you are registering an exporter and would expect an exporter, not a function that flushes it and ends the object.

My major concerns are
A) To understand what flushExporter does, you will have to go and look into the package code
B) It is harder to extend the object in proxy app.

pkg/observ/observ.go
```diff
@@ -17,16 +18,56 @@ type observabilityContext struct {
...
+// RegisterExporter returns the 'Flush' function for that particular tracing service
+// TODO: Extend beyond jaeger and datadog
+func RegisterExporter(TraceExporter, URL, service, ENV string) (func(), error) {
+
+	const op errors.Op = "RegisterExporter"
+
+	switch TraceExporter {
+	case "jaeger":
+		ex, err := registerJaegerExporter(URL, service, ENV)
+		return jaegerFlush(ex), errors.E(op, err)
```
manugupt1: Why are you always returning an error. 
https://github.com/gomods/athens/commit/0853f073f268e0b44eb75b2b45f3811b51054101

If I am reading the code correctly, error will never be nil.

I tried this :
`	fmt.Println("----------------", errors.E("manu", nil))`
and it returned 
`Internal Server Error`

cmd/proxy/actions/app.go
marwan-at-work: @manugupt1 that was my suggestion. No need to create 1 interface and N structs. We can just return a `func()` that performs flushing. This PR will need a bunch of cleaning up to make it look a little bit more readable. Will add my comments as a review.

pkg/observ/observ.go
```diff
@@ -17,16 +18,56 @@ type observabilityContext struct {
 	spanCtx context.Context
 }
 
-// RegisterTraceExporter returns a jaeger exporter for exporting traces to opencensus.
+// RegisterExporter determines the type of TraceExporter service for exporting traces from opencensus
+// User can choose from multiple tracing services (datadog, jaegar)
+// RegisterExporter returns the 'Flush' function for that particular tracing service
+// TODO: Extend beyond jaeger and datadog
+func RegisterExporter(TraceExporter, URL, service, ENV string) (func(), error) {
+
```
marwan-at-work: remove space

pkg/observ/observ.go
marwan-at-work: `return ex.Flush, nil` -- no need to create a function that returns a function that calls a function :)

pkg/observ/observ.go
```diff
@@ -17,16 +18,56 @@ type observabilityContext struct {
...
+	const op errors.Op = "RegisterExporter"
+
+	switch TraceExporter {
+	case "jaeger":
+		ex, err := registerJaegerExporter(URL, service, ENV)
+		return jaegerFlush(ex), errors.E(op, err)
+	case "datadog":
+		ex, err := registerDatadogExporter(URL, service, ENV)
+		return datadogFlush(ex), errors.E(op, err)
```
marwan-at-work: same as above

pkg/observ/observ.go
```diff
@@ -43,13 +84,30 @@ func RegisterTraceExporter(URL, service, ENV string) (*(jaeger.Exporter), error)
 		return nil, errors.E(op, err)
 	}
 
-	// And now finally register it as a Trace Exporter
-	trace.RegisterExporter(je)
+	trace.RegisterExporter(ex)
```
marwan-at-work: like i said in my previous comments, let's put this line and the following 3 lines in a function so that we don't repeat it in every implementation.

pkg/observ/observ.go
```diff
@@ -43,13 +84,30 @@ func RegisterTraceExporter(URL, service, ENV string) (*(jaeger.Exporter), error)
...
+		trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
+	}
+
+	return ex, nil
+}
+
+// RegisterDatadogTracerExporter creates a datadog exporter.
+// Currently uses the 'TraceExporter' variable in the config file.
+func registerDatadogExporter(URL, service, ENV string) (*(datadog.Exporter), error) {
+
```
marwan-at-work: remove empty line

pkg/observ/observ.go
```diff
@@ -17,16 +18,56 @@ type observabilityContext struct {
...
+	}
+}
+
+// registerJaegerExporter creates a jaeger exporter for exporting traces to opencensus.
+// Currently uses the 'TraceExporter' variable in the config file.
 // It should in the future have a nice sampling rate defined
-// TODO: Extend beyond jaeger
-func RegisterTraceExporter(URL, service, ENV string) (*(jaeger.Exporter), error) {
-	const op errors.Op = "RegisterTracer"
+func registerJaegerExporter(URL, service, ENV string) (*(jaeger.Exporter), error) {
```
marwan-at-work: remove unnecessary parens

pkg/observ/observ.go
```diff
@@ -17,16 +18,56 @@ type observabilityContext struct {
...
+		return nil, errors.E(op, "Exporter not specified. Traces won't be exported")
+	default:
+		return nil, errors.E(op, "Exporter not supported. Please open PR or an issue at github.com/gomods/athens")
+	}
+
+}
+
+// Flush method stops the trace exporter, this method needs to be
+// called for most of the TraceExporters using opencensus
+func jaegerFlush(j *(jaeger.Exporter)) func() {
```
marwan-at-work: remove function

pkg/observ/observ.go
```diff
@@ -17,16 +18,56 @@ type observabilityContext struct {
...
+	return func() {
+		j.Flush()
+	}
+}
+
+// Flush method stops the trace exporter, this method needs to be
+// called for most of the TraceExporters using opencensus
+func datadogFlush(d *(datadog.Exporter)) func() {
```
marwan-at-work: remove function

pkg/observ/observ.go
```diff
@@ -43,13 +84,30 @@ func RegisterTraceExporter(URL, service, ENV string) (*(jaeger.Exporter), error)
...
+	if ENV == "development" {
+		trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
+	}
+
+	return ex, nil
+}
+
+// RegisterDatadogTracerExporter creates a datadog exporter.
+// Currently uses the 'TraceExporter' variable in the config file.
+func registerDatadogExporter(URL, service, ENV string) (*(datadog.Exporter), error) {
```
marwan-at-work: remove unnecessary parens.

cmd/proxy/actions/app.go
manugupt1: Looking through your review, it makes sense!
Can you do the changes that @marwan-at-work  asked for?

cmd/proxy/actions/app.go
```diff
@@ -75,12 +75,18 @@ func App(conf *config.Config) (*buffalo.App, error) {
 		app = app.Group(prefix)
 	}
 
-	// Register exporter to export traces
-	exporter, err := observ.RegisterTraceExporter(conf.TraceExporterURL, Service, ENV)
+	// RegisterExporter is used for registering and exporting traces from opencensus
```
manugupt1: will register an exporter where we will export our traces to.
Remove opencensus here.

cmd/proxy/actions/app.go
```diff
@@ -75,12 +75,18 @@ func App(conf *config.Config) (*buffalo.App, error) {
...
+	// The error from the RegisterExporter would be nil if the tracer was specified by the user and
+	// and the trace exporter was created successfully.
+	// RegisterExporter returns a function that needs to be run for all services(datadog, jaeger)
```
manugupt1: Can you update the comment as such
RegisterExporter returns the function that all traces are flushed to the exporter and the exporter needs to be stopped. The function should be called when the exporter is no longer needed.

cmd/proxy/actions/app.go
```diff
@@ -75,12 +75,18 @@ func App(conf *config.Config) (*buffalo.App, error) {
...
+	// exporting traces from opencensus.
+	flushExporter, err := observ.RegisterExporter(conf.TraceExporter,
```
manugupt1: let us rename it as 
flushTraces for now.

pkg/observ/observ.go
```diff
@@ -43,13 +84,30 @@ func RegisterTraceExporter(URL, service, ENV string) (*(jaeger.Exporter), error)
...
-	trace.RegisterExporter(je)
+	trace.RegisterExporter(ex)
+	if ENV == "development" {
+		trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
+	}
+
+	return ex, nil
+}
+
+// RegisterDatadogTracerExporter creates a datadog exporter.
```
manugupt1: r should be lowercased

pkg/observ/observ.go
```diff
@@ -43,13 +84,30 @@ func RegisterTraceExporter(URL, service, ENV string) (*(jaeger.Exporter), error)
...
+// Currently uses the 'TraceExporter' variable in the config file.
+func registerDatadogExporter(URL, service, ENV string) (*(datadog.Exporter), error) {
+
+	ex := datadog.NewExporter(
+		datadog.Options{
+			TraceAddr: URL,
```
manugupt1: Nice!

bhaveshpraveen:
@manugupt1 @marwan-at-work: I've updated the code.

pkg/observ/observ.go
```diff
@@ -17,16 +18,35 @@ type observabilityContext struct {
...
+// RegisterExporter determines the type of TraceExporter service for exporting traces from opencensus
+// User can choose from multiple tracing services (datadog, jaegar)
+// RegisterExporter returns the 'Flush' function for that particular tracing service
+// TODO: Extend beyond jaeger and datadog
+func RegisterExporter(TraceExporter, URL, service, ENV string) (func(), error) {
+	const op errors.Op = "RegisterExporter"
+	switch TraceExporter {
+	case "jaeger":
+		ex, err := registerJaegerExporter(URL, service, ENV)
+		return ex.Flush, errors.E(op, err)
```
marwan-at-work: like @manugupt1 said, `errors.E` returns a non nil error. Therefore, we should either check the err and return nil, or just have that happen inside the registerX funcs and here you just do `return registerX(..)`

bhaveshpraveen: @marwan-at-work  Should I change it to this?
```
switch TraceExporter {
	case "jaeger":
		ex, err := registerJaegerExporter(URL, service, ENV)
		if err != nil {
			return nil, errors.E(op, err)
		}
		return ex.Flush, nil
```

bhaveshpraveen: Updated!!

michalpristas:
I'm late to the party, but very nice PR @bhaveshpraveen

bhaveshpraveen:
@michalpristas Thank you! Wouldn't have been possible without the support from @manugupt1 and @marwan-at-work

PR #727, opened 2018-10-04
Refactor storage tests

marwan-at-work:
This PR makes it so that we can finally run `go test ./...` without needing any dependencies. 
However, if you're working on a storage implementation, and would like to test it, you can easily do so by having an environment variable. So that `go test ./...` works both ways. 

I've also simplified our tests and removed unused code.


pkg/config/parse_test.go
```diff
@@ -284,59 +284,6 @@ func TestParseExampleConfig(t *testing.T) {
 	restoreEnv(envVarBackup)
 }
 
-// TestConfigOverridesDefault validates that a value provided by the config is not overwritten during parsing
```
marwan-at-work: I removed this test because it's inaccurate: we test that env vars don't get overwritten. This test fails if you actually do have an env var for minio set up...so it's unstable and inaccurate. We should just test that config works as intended: toml gets parsed, and env vars override toml. Which we do in the rest of this file.

pkg/fixtures/mongo.go
```diff
@@ -1,55 +0,0 @@
...
-	m.Require().NoError(m.DB.DropDatabase())
-}
-
-// NewMongo creates a new Mongo test fixture
-func NewMongo(url string) *Mongo {
-	return &Mongo{
-		url:    url,
-		DBName: names.NameSep("-") + "-athens-testing",
-	}
-}
```
arschles: would we want to leave this around for future integration-level tests? or are they replaced by the compliance tests?

pkg/storage/fs/fs_test.go
```diff
@@ -1,14 +1,45 @@
...
 
-func (d *FsTests) TestLocationFuncs() {
-	r := d.Require()
-	storage := d.storage.(*storageImpl)
-	moduleLoc := storage.moduleLocation(module)
-	r.Equal(filepath.Join(d.rootDir, module), moduleLoc)
-	versionedLoc := storage.versionLocation(module, version)
-	r.Equal(filepath.Join(d.rootDir, module, version), versionedLoc)
+func TestBackend(t *testing.T) {
+	fs := afero.NewOsFs()
```
arschles: what do you think about making this and the below a memory FS, so that these tests don't touch the system at all?

pkg/fixtures/mongo.go
marwan-at-work: @arschles they are replaced by compliance tests. Basically, if the Mongo URL Env Var is set, then mongo db will get tested, otherwise mongo gets skipped. Same is the case for Minio (and soon all the other storages)

pkg/storage/fs/fs_test.go
marwan-at-work: @arschles totes, but I'd like us to keep the Benchmarks to compare FS vs Memory. Benchmarks don't run during CI/CD as of now, so we should be ok.

pkg/fixtures/mongo.go
arschles: Awesome, thanks for clarifying @marwan-at-work

PR #743, opened 2018-10-04
Changing the PR template

arschles:
So that when you write `Fixes #123` GH will close the issue after merge. As far as I could tell, it wouldn't do that with the old PR template


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,6 +7,6 @@ Describe the issue you have been trying to solve.
 
 Mention briefly how you have applied the fix.
 
-**Fixes issue:  #[]**
+Fixes #[]
```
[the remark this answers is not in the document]
arschles: oh, good point. @robjloranger @michalpristas I changed it to `Fixes #123` to hopefully make it clearer that this should be an issue number

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,6 +7,6 @@ Describe the issue you have been trying to solve.
 
 Mention briefly how you have applied the fix.
 
-**Fixes issue:  #[]**
+Fixes #123
```
carolynvs: I recommend changing this to `Fixes #`. Otherwise it would be quite easy to miss that the issue they said they are fixing is actually 123. :-)

arschles: @carolynvs thanks - I like that! fixed in https://github.com/gomods/athens/pull/743/commits/1b2c9a1da6a726b75073c0e850cd928708c66bf2

PR #750, opened 2018-10-05
Implement storage.Backend for S3 storage

### LINKED ISSUE #704: Implement storage.Backend for S3 storage
S3 is missing implementation

- Let `s3` storage join the fun in `/pkg/storage/storage_tests`

t-tomalak:

Missing s3 integration

**How is the fix applied?**

Integrated S3 storage that is compatiblewith storage.Backend interface. For unit test minio is used to mock AWS service. Also fixed tests for minio that where skiped after #727 because of missing env variable

Fixes #704


pkg/storage/s3/getter.go
```diff
@@ -0,0 +1,103 @@
...
+	if !exists {
+		return nil, errors.E(op, errors.M(module), errors.V(version), errors.KindNotFound)
+	}
+
+	infoReader, err := s.open(ctx, config.PackageVersionedName(module, version, "info"))
+	if err != nil {
+		return nil, errors.E(op, err, errors.M(module), errors.V(version))
+	}
+	infoBytes, err := ioutil.ReadAll(infoReader)
+	infoReader.Close()
```
marpio: Please do defer inforeader.close() before ioutil.readall(...)

pkg/storage/s3/getter.go
```diff
@@ -0,0 +1,103 @@
...
+	if !exists {
+		return nil, errors.E(op, errors.M(module), errors.V(version), errors.KindNotFound)
+	}
+
+	modReader, err := s.open(ctx, config.PackageVersionedName(module, version, "mod"))
+	if err != nil {
+		return nil, errors.E(op, err, errors.M(module), errors.V(version))
+	}
+	modBytes, err := ioutil.ReadAll(modReader)
+	modReader.Close()
```
marpio: defer close

pkg/storage/s3/checker.go
```diff
@@ -0,0 +1,40 @@
...
+
+	pkgName := config.PackageVersionedName(module, version, "mod")
+	hoParams := &s3.HeadObjectInput{
+		Bucket: aws.String(s.bucket),
+		Key:    aws.String(pkgName),
+	}
+
+	_, err := s.s3API.HeadObjectWithContext(ctx, hoParams)
+	if err != nil {
+		if err.(awserr.Error).Code() == s3ErrorCodeNotFound {
```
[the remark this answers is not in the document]
t-tomalak: Yeah, i was thinking that it should have that but couldn't find it

cmd/proxy/actions/storage.go
```diff
@@ -50,6 +51,14 @@ func GetStorage(storageType string, storageConfig *config.StorageConfig) (storag
 			return nil, errors.E(op, "Invalid CDN Storage Configuration")
 		}
 		return gcp.New(context.Background(), storageConfig.GCP, storageConfig.CDN)
+	case "s3":
+		if storageConfig.S3 == nil {
+			return nil, errors.E(op, "Invalid S3 Storage Configuration")
+		}
+		if storageConfig.CDN == nil {
```
michalpristas: i'm wondering why are we failing on nil,
we use CDN config as this 
```go
func (s *Storage) BaseURL() *url.URL {
	return s.cdnConf.CDNEndpointWithDefault(s.baseURI)
}
```
if we move this nil check to BaseURL
```gofunc (s *Storage) BaseURL() *url.URL {
	if(s.cdnConf == nil){
		return s.baseUri
	}
	return s.cdnConf.CDNEndpointWithDefault(s.baseURI)
}
```

it should be also fine, and allows me to shrink my config file a bit

thought? @t-tomalak @rchakra3

pkg/config/s3.go
```diff
@@ -0,0 +1,14 @@
...
+
+// S3Config specifies the properties required to use S3 as the storage backend
+type S3Config struct {
+	TimeoutConf
+	Region         string `validate:"required" envconfig:"AWS_REGION"`
+	Key            string `validate:"required" envconfig:"AWS_ACCESS_KEY_ID"`
+	Secret         string `validate:"required" envconfig:"AWS_SECRET_ACCESS_KEY"`
+	Token          string `envconfig:"AWS_SESSION_TOKEN"`
+	Bucket         string `validate:"required" envconfig:"ATHENS_S3_BUCKET_NAME"`
+	Endpoint       string
```
michalpristas: no ENV vars. is this on purpose?

pkg/storage/s3/s3_test.go
```diff
@@ -0,0 +1,66 @@
...
+func getStorage(t testing.TB) *Storage {
+	url := os.Getenv("ATHENS_MINIO_ENDPOINT")
+	if url == "" {
+		t.SkipNow()
+	}
+
+	backend, err := New(
+		&config.S3Config{
+			Endpoint:       url,
+			Key:            "minio",
```
michalpristas: same thing, s3 storage with minio config. confusing

pkg/storage/s3/saver.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/gomods/athens/pkg/observ"
+	moduploader "github.com/gomods/athens/pkg/storage/module"
+)
+
+// Save implements the (github.com/gomods/athens/pkg/storage).Saver interface.
+func (s *Storage) Save(ctx context.Context, module, version string, mod []byte, zip io.Reader, info []byte) error {
+	const op errors.Op = "s3.Save"
+	ctx, span := observ.StartSpan(ctx, op.String())
+	defer span.End()
+	err := moduploader.Upload(ctx, module, version, bytes.NewReader(info), bytes.NewReader(mod), zip, s.upload, s.cdnConf.TimeoutDuration())
```
michalpristas: should be s.S3Config.TimeoutDuration

pkg/storage/s3/s3_test.go
```diff
@@ -0,0 +1,66 @@
...
+	compliance.RunTests(t, backend, backend.clear)
+}
+
+func BenchmarkBackend(b *testing.B) {
+	backend := getStorage(b)
+	compliance.RunBenchmarks(b, backend, backend.clear)
+}
+
+func (s *Storage) clear() error {
+	minioClient, _ := minio.New(os.Getenv("ATHENS_MINIO_ENDPOINT"), "minio", "minio123", false)
```
michalpristas: using minio for S3 tests
look at GCP for reference

pkg/storage/s3/s3_test.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+		if err := minioClient.RemoveObject("gomods", object.Key); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func getStorage(t testing.TB) *Storage {
+	url := os.Getenv("ATHENS_MINIO_ENDPOINT")
```
michalpristas: same

scripts/test_unit.sh
```diff
@@ -6,6 +6,8 @@ if [ -z ${GO_ENV} ]; then
     export GO_ENV="test"
 fi
 
+export ATHENS_MINIO_ENDPOINT="127.0.0.1:9001"
```
michalpristas: this should not be here, at least not with s3 pack.

michalpristas:
Great @t-tomalak in case i miss the update feel free to ping me directly so i'm not blocking

scripts/test_unit.sh
t-tomalak: so should I create issue for that and make pull request with that? Because without this, storage tests for minio are ommited.

pkg/config/s3.go
t-tomalak: Yeah they where but they where only used for test and shouldn't be accessed by user. But I will propose better solution for that.

cmd/proxy/actions/storage.go
t-tomalak: Droped that dependency and now it's not required

pkg/storage/s3/s3.go
```diff
@@ -0,0 +1,79 @@
...
+	bucket   string
+	baseURI  *url.URL
+	uploader s3manageriface.UploaderAPI
+	s3API    s3iface.S3API
+	s3Conf   *config.S3Config
+	cdnConf  *config.CDNConfig
+}
+
+// New creates a new AWS S3 CDN saver
+func New(s3Conf *config.S3Config, cdnConf *config.CDNConfig, options ...func(*aws.Config)) (*Storage, error) {
```
michalpristas: what's the purpose of the options here?

t-tomalak: Used mostly for tests to configure additional fields, like endpoint and disabled SSL. They should be on default when working with normal S3

michalpristas: if it's only for test, can test prepare config it would like to have in Arrange step and clean after itself afterward if needed?
i'm not a fan of tweaking implementation for sake of tests

t-tomalak: Yeah but then I would need to rollback this uneccessary fields that where before in config.S3Config. I don't see the way how to set that options before creating session to S3, I could create antoher `New` method eg. `NewWithConfig/NewWithOptions` and then pass this additional settings but that will create method solely for tests.

pkg/storage/s3/deleter.go
```diff
@@ -0,0 +1,46 @@
...
+func (s *Storage) remove(ctx context.Context, path string) error {
+	const op errors.Op = "s3.Delete"
+	ctx, span := observ.StartSpan(ctx, op.String())
+	defer span.End()
+	delParams := &s3.DeleteObjectInput{
+		Bucket: aws.String(s.bucket),
+		Key:    aws.String(path),
+	}
+
+	_, err := s.s3API.DeleteObjectWithContext(ctx, delParams)
```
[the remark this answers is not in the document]
t-tomalak: changed more of this kind of if's with first `_` paramter

pkg/storage/s3/s3.go
```diff
@@ -0,0 +1,79 @@
...
+	const op errors.Op = "s3.New"
+	u, err := url.Parse(fmt.Sprintf("http://%s.s3.amazonaws.com", s3Conf.Bucket))
```
marwan-at-work: should this not be `https`?

cmd/proxy/actions/storage.go
```diff
@@ -50,6 +51,11 @@ func GetStorage(storageType string, storageConfig *config.StorageConfig) (storag
 			return nil, errors.E(op, "Invalid CDN Storage Configuration")
 		}
 		return gcp.New(context.Background(), storageConfig.GCP, storageConfig.CDN)
+	case "s3":
+		if storageConfig.S3 == nil {
+			return nil, errors.E(op, "Invalid S3 Storage Configuration")
```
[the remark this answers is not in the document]
t-tomalak: We do simillar message for others. Now only this one will have different message

pkg/config/parse_test.go
```diff
@@ -150,6 +150,13 @@ func TestStorageEnvOverrides(t *testing.T) {
 				Timeout: globalTimeout,
 			},
 		},
+		S3: &S3Config{
```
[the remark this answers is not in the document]
t-tomalak: If I understand corectly this test checks if values can be set via env variables and it is checking that. Same thing is applied for other config. I found other issue in parse_test.go and fixed that, I forgot to add compare for config in `compareStorageConfigs` method.

cmd/proxy/actions/storage.go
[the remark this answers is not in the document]
marwan-at-work: you can keep this one then and I'll update all of them together in another PR. Thanks for the note!

PR #751, opened 2018-10-05
Docs move try-out from homepage to a new page

### LINKED ISSUE #741: Move the "try out athens" section from the homepage to a separate page

#740 seems to make the homepage clearer (preview it [here](https://deploy-preview-740--gomods-athens.netlify.com/)), and I think the "Try it now" section takes away from the clear "why you should stay on the docs page" message by adding noise.

I think we should move this section to a new page and link to it from the home page with some text like this:

"Like what you hear? Try Athens Now (this would be a link to the new page)!

I can't think of another way to reduce noise on the homepage

**Additional context**

I noticed this when I was was working on #740

gabrielSchneider100:

Make the homepage clearer #741 .

**How is the fix applied?**

Created a new file try-out.md and moved Try Out section from the homepage to that file.

Fixes #741

### DIFF
--- a/docs/content/_index.md
+++ b/docs/content/_index.md
@@ -8,58 +8,7 @@ date: 2018-02-11T16:52:23-05:00
 
 ## Welcome to Athens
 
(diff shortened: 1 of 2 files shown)


arschles:
@gabrielSchneider100 congratulations on your first PR merge. Ping in the slack channel if you're looking for a second one 😄

Issue #762, opened 2018-10-08
Update helm charts to include tracing from jaeger


manugupt1:
**Is your feature request related to a problem? Please describe.**
Currently, helm charts do not have any way of letting the users know about tracing. We would want Jaeger to be used as a tracing library since it is open source.

This is because helm charts have been used to deploy the public proxy afaik.

**Describe the solution you'd like**
Update helm charts to include jaeger as the tracing library

t-tomalak:
Is anyone on this? I would like to take on this. I never done anything with helm but I probably need few answers to make it done.
1. Do we wanna use jaeger/all-in-one image?
2. Do we wanna any customization of ports on this image?
3. Should we also add in deployment.yml this two env variables? First one will be `ATHENS_TRACE_EXPORTER` and this is just set to `jaeger` probably. Second one `ATHENS_TRACE_EXPORTER_URL` but I don't know what value it should had. Wild guess would be service name so probably also `jaeger`. Could someone help? :slightly_smiling_face:

manugupt1:
@t-tomalak 
1. Yes, we want to have jaeger/all-in-one image because that simplifies a lot of things.

For 2 and 3, I think we just specify the ports that the image maps to with the full URL. I would do what you feel like and we can take it from there.

manugupt1:
This has been fixed by @t-tomalak! Thank you!

PR #768, opened 2018-10-09
Add jaeger tracer to helm charts

t-tomalak:

Update helm charts to include jaeger 

**How is the fix applied?**

Added new deployment for jaeger traccer also added missing environment variables to Proxy deployment. I tested deployment into minikube and everything deployed correctly.  It's my first time working with helm so probably some tweaks to that configuration will be needed. Probably what and how some of jaeger ports should be exposed and also what url we should provide to properly comunicated Proxy with Jaeger. 

Also added `export GO111MODULE=on` to `scripts/test_unit.sh` because when working from GOPATH and when this env is not set I recieved error about `mod` flag. Is that ok or I should remove setter for that env?

**Fixes issue:  #762**


scripts/test_unit.sh
```diff
@@ -6,6 +6,8 @@ if [ -z ${GO_ENV} ]; then
     export GO_ENV="test"
 fi
 
+export GO111MODULE=on
```
manugupt1: Why do we need this?

t-tomalak: I added it in PR description but will copy/paste it here 

> Also added export GO111MODULE=on to scripts/test_unit.sh because when working from GOPATH and when this env is not set I recieved error about mod flag. Is that ok or I should remove setter for that env?

Is there a reason why we don't added it here? Because we do similar thing in test_e2e.sh

manugupt1: Oh got it! thanks! I realized I do not have GOPATH set so never landed into this bug!

PR #789, opened 2018-10-17
Upgrade Buffalo

### LINKED ISSUE #783: Buffalo now supports modules!
can we change our builds now that Buffalo supports modules?

https://blog.gobuffalo.io/buffalo-v0-13-0-released-eac17c4cc4fa

steakunderscore:

Upgrade Buffalo to v0.13

**How is the fix applied?**

Upgraded the module. Ran `buffalo fix`. Changed to `GO111MODULE=on` everywhere.

Fixes #783 ( Perhaps :thinking: )


steakunderscore:
:stop_sign: This is blocked by a [bug in Buffalo](https://github.com/gobuffalo/buffalo/pull/1383) that is waiting to be released.

t-tomalak:
Now the issue how buffalo handle trailing slash by default(https://blog.gobuffalo.io/buffalo-v0-13-0-released-eac17c4cc4fa), because it changed between versions:
This is old code:
https://github.com/gobuffalo/buffalo/blob/4d7b03aa7e68260a1fc6f8b8cdf327b3f6511123/app.go#L39
And new code:
https://github.com/gobuffalo/buffalo/blob/1b45ee310884d2d35ad6413d649a3ca720918692/app.go#L48

And by that buffalo disabled option to remove that trailing slash. It seems that with it in `c.Request().URL.Path` or `c.Header().Get("Location")` everything works fine for `athens` so we should just add `/` here:
https://github.com/gomods/athens/blob/532e93e11979dd2aeb280900c8c9e7bb15edb78f/pkg/middleware/middleware_test.go#L69

steakunderscore:
Great break down @t-tomalak. This is the same conclusion I came to.

Personally I'm fine with changing the test, but I don't understand the project well enough to know there wont be any side effects. Perhaps @michalpristas you could provide some guidance here.

michalpristas:
Hey @steakunderscore after a quick review here. I think it is ok to modify this particular test if buffalo requires it
What is returned is redirect URI so it does not even need to be in a format of download protocol. 
Maybe we can add test that both URI formats (with and without trailing slash) are resolved correctly and will respond with expected response.
Go tool will call A-level proxy by URI wihout a trailing slash which cause redirect to B-Level proxy with a trailing slash, so we need to make sure it does the right thing in both cases.

pkg/middleware/middleware_test.go
```diff
@@ -66,7 +66,7 @@ func Test_FilterMiddleware(t *testing.T) {
 	// Public, expects to be redirected to the global registry endpoint
```
michalpristas: Please add one more test here saying:
```go
// Public, expects to be redirected to the global registry endpoint - test with a trailing slash
res := w.Request("/github.com/gomods/athens/@v/list/").Get()
r.Equal(303, res.Code)
r.Equal(conf.Proxy.GlobalEndpoint+"/github.com/gomods/athens/@v/list/", res.HeaderMap.Get("Location"))
```
difference is in trailing slash in first request

arschles:
@steakunderscore this looks great. I'm so pumped that we're upgrading Buffalo!!!!! 🎉

PR #797, opened 2018-10-18
Document trying proxy outside the GOPATH

### LINKED ISSUE #780: Fix "try it out" page so documented commands produce documented output
**Describe the bug**
Trying out Athens https://docs.gomods.io/try-out/ and it looks like `proxy` looks for `../../config.dev.toml` so it needs to be run from `$GOPATH/src/github.com/gomods/athens/`

**Error Message**
```
$ go get -u github.com/gomods/athens/cmd/proxy
$ proxy &
[1] 18689
$ 2018/10/13 17:23:43 open ../../config.dev.toml: no such file or directory
[1]+  Exit 1                  proxy
```

**To Reproduce**
1. `cd ~`
2. `go get -u github.com/gomods/athens/cmd/proxy`
3. `proxy &`

**Expected behavior**
As here: https://docs.gomods.io/try-out/
```
$ go get -u github.com/gomods/athens/cmd/proxy
# the source is downloaded to GOPATH/src/github.com/gomods/athens/
$ proxy &
[1] 25243
INFO[0000] Starting application at 127.0.0.1:3000
```

 - OS: macOSX 10.14.1 Beta
 - Go version : go version go1.11.1 darwin/amd64
 - Buffalo Version : ?
 - Olympus or proxy version : master 4b6da424807b9baabe7239bd709818046ce77899

**Additional context**
Would it be better to change proxy so it could run from anywhere with sensible defaults, or to add this line to the docs (between `go get` and `proxy &`):
```
$ cd $GOPATH/src/github.com/gomods/athens/cmd/proxy
```

Or both?

I can PR a fix for the docs and I could look into running the proxy without a config file if others think that's a worthwhile idea.

voutasaurus:

Try it out documentation is not working due to relative path of config file. Also it does not show the repo working outside the GOPATH.

**How is the fix applied?**

Updated "try it out" documentation to run proxy outside the GOPATH and match expected output.

Also fixes minor formatting issues.

Fixes #780

### DIFF
--- a/docs/content/try-out.md
+++ b/docs/content/try-out.md
@@ -10,8 +10,17 @@ To quickly see Athens in action, follow these steps:
 
 First, make sure you have [Go 1.11 installed](https://gophersource.com/setup/),
 that GOPATH/bin is on your path, and that you have enabled the [Go
-Modules](https://github.com/golang/go/wiki/Modules) feature by exporting
-GO111MODULE=on.
+Modules](https://github.com/golang/go/wiki/Modules) feature.
+
+**Bash**
+```bash
+export GO111MODULE=on
+```
+
+**PowerShell**
+```powershell
+$env:GO111MODULE = "on"
+```
 
 Next, use git and Go to install and run the Athens proxy in a background process.
 
--- a/docs/content/try-out.md
+++ b/docs/content/try-out.md
@@ -8,33 +8,34 @@ menu: shortcuts
 
 To quickly see Athens in action, follow these steps:
 
-First, make sure you have [Go 1.11 installed](https://gophersource.com/setup/) and that GOPATH/bin is on your path.
+First, make sure you have [Go 1.11 installed](https://gophersource.com/setup/),
+that GOPATH/bin is on your path, and that you have enabled the [Go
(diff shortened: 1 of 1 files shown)


docs/content/try-out.md
```diff
@@ -8,33 +8,34 @@ menu: shortcuts
 
 To quickly see Athens in action, follow these steps:
 
-First, make sure you have [Go 1.11 installed](https://gophersource.com/setup/) and that GOPATH/bin is on your path.
+First, make sure you have [Go 1.11 installed](https://gophersource.com/setup/),
+that GOPATH/bin is on your path, and that you have enabled the [Go
+Modules](https://github.com/golang/go/wiki/Modules) feature by exporting
+GO111MODULE=on.
```
arschles: what do you think about showing the commands for bash and powershell on how to export `GO111MODULE=on`?

that'd be important in case someone does check out the source into their GOPATH

voutasaurus: Was considering whether it made the page too busy. That said in retrospect putting it at the end of a long sentence it seems less prominent than it should be (given how new modules are). I'll update it now with your suggestion!

PR #799, opened 2018-10-19
Update test_unit.ps1 so that it matches the bash version.

### LINKED ISSUE #798: Update test_unit.ps1 so that it matches the bash version.
**Describe the bug**
Currently the powershell test_unit.ps1 script doesn't add the default ENV VARs for minio and mongo so the storage tests are skipped even if minio and mongo are running.

**To Reproduce**
Execute test_unit.ps1

**Expected behavior**
The storage tests should execute if the services are up.

marpio:

Currently the powershell test_unit.ps1 script doesn't add the default ENV VARs for minio and mongo so the storage tests are skipped even if minio and mongo are running.

**How is the fix applied?**

Just like in the bash script we can set the env vars if they don't exists

Fixes #798


t-tomalak:
I never used golang with powershell but I made one more improvment in test_unit.sh for linux. https://github.com/gomods/athens/pull/768#discussion_r224075511 @marpio could you verify that in windows there is same issue?

t-tomalak:
As we had https://github.com/gomods/athens/pull/805 merged which resolves same issue, it seems like only one thing to resolve is `GO111MODULE` check in PowerShell, or we can close that

marpio:
Thanks @t-tomalak ! I've added the `GO111MODULE` check.

PR #839, opened 2018-10-31
change env var for port to ATHENS_PORT

### LINKED ISSUE #610: Add ATHENS_PORT configuration
Currently, as of PR #596, we are allowing a custom port to be set via the environment variable `PORT`. @marwan-at-work is 88% positive that platforms like GAE will yell at us if we try to set a `PORT` variable other than the one assigned.

We should add a configuration option for something like `ATHENS_PORT`, which could default to `3000` and be overridden by this new environment variable.

We may or may not decide to keep both `PORT` and `ATHENS_PORT`, although I'm not sure of a reason to do so. The current configuration is defined in [pkg/config/env/port.go](https://github.com/gomods/athens/blob/master/pkg/config/env/port.go)

natezyz:
Instead of looking for PORT, config looks for ATHENS_PORT now. Port number is also defaulted to 3000 if ATHENS_PORT is not found in the env.

Fixes #610

### DIFF
--- a/pkg/config/proxy.go
+++ b/pkg/config/proxy.go
@@ -4,7 +4,7 @@ package config
 type ProxyConfig struct {
 	StorageType    string `validate:"required" envconfig:"ATHENS_STORAGE_TYPE"`
 	GlobalEndpoint string `envconfig:"ATHENS_GLOBAL_ENDPOINT"` // This feature is not yet implemented
-	Port           string `validate:"required" envconfig:"PORT"`
+	Port           string `validate:"required" envconfig:"ATHENS_PORT"`
 	FilterOff      bool   `validate:"required" envconfig:"PROXY_FILTER_OFF"`
 	BasicAuthUser  string `envconfig:"BASIC_AUTH_USER"`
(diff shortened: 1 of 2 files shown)


pkg/config/proxy.go
```diff
@@ -4,7 +4,7 @@ package config
 type ProxyConfig struct {
 	StorageType    string `validate:"required" envconfig:"ATHENS_STORAGE_TYPE"`
 	GlobalEndpoint string `envconfig:"ATHENS_GLOBAL_ENDPOINT"` // This feature is not yet implemented
-	Port           string `validate:"required" envconfig:"PORT"`
+	Port           string `validate:"required" envconfig:"ATHENS_PORT" default:":3000"`
```
manugupt1: This is really nice. 
I have a few changes to request:
If you have a default value, then the validation should not be required as far as I know. Can you correct me if I am wrong?
Can you add also add a test for this as this is in a critical code section?

natezyz: Makes sense. When I first saw the validate, I thought it might be a way to hook in some kind of field format validation. Doesn't seem to be the case so it can be removed. Will add the test.

Issue #830, opened 2018-10-29
CI/CD Not Building Proxy Image Correctly


arkadyb:
Hi,

Id like to use athens for corp level go modules proxy, but cant seem to find a way get it authenticated against our private repos in http://github.com/[CompanyName]
Is there some way to use my GitHub deploy key in netrc file? Or should i be using some other auth type? Thanks for help!

michalpristas:
Hey @arkadyb, thanks for raising an issue.

please check this doc page https://docs.gomods.io/install/install-on-kubernetes/#give-athens-access-to-private-repositories-via-github-token-optional
and let me know if this answers your question

michalpristas:
Have you restarted the proxy?

fedepaol:
Also @arkadyb , could you manually throw a .netrc file in your home to check if that works?
The .netrc file should [be filled like this ](https://gist.github.com/cactus/3925422)

michalpristas:
also, one more important question comes to mind. which OS are you using?
our netrc implementations are not windows friendly

arkadyb:
@fedepaol, ive tried to run on local (macos) and Docker. .netrc file has been created at home, but i still face same issues, i get back 500 error when athens tries to access my private repos. Does it metter repo belongs to org and not my private user?

arkadyb:
I was finally able to run it. Problem is - when u run docker image from dockerhub, it does not seem to create required .netrc file event when i pass respective git hub security token. i was able to build image from Dockefile in repo and this one works.

marwan-at-work:
@arkadyb what is the image that used from docker hub? Also, how did you pass the GITHUB_TOKEN or NETRC_PATH variables to the image? Can you please provide a reproduction example. Thanks

arkadyb:
@marwan-at-work , docker hub image - https://hub.docker.com/r/gomods/proxy/
as per command it is: `docker run -d ... -e ATHENS_GITHUB_TOKEN=[my-github-token] --name athens-proxy --restart always -p 3000:3000 gomods/proxy:latest`

arkadyb:
To repro simply use:
`
export ATHENS_STORAGE=~/athens-storage
mkdir -p $ATHENS_STORAGE
docker run -d -v $ATHENS_STORAGE:/var/lib/athens \
   -e ATHENS_DISK_STORAGE_ROOT=/var/lib/athens \
   -e ATHENS_GITHUB_TOKEN=[my-github-token] \
[...]

PR #841, opened 2018-10-31
adding log lines to the docker push script

arschles:

I'm trying to figure out if there's any issues with the code we use to push docker images to dockerhub in CI

**How is the fix applied?**

I'm adding some debugging lines to the `./scripts/push-docker-images.sh` script

Ref #830

### DIFF
--- a/scripts/push-docker-images.sh
+++ b/scripts/push-docker-images.sh
@@ -32,12 +32,21 @@ if [[ "${MUTABLE_TAG:-}" == "" ]]; then
     fi
 fi
 
+echo "Building Athens image for registry ${REGISTRY}"
+echo "Version ${VERSION}"
+echo "Mutable tag ${MUTABLE_TAG}"
+
 REPO_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." >/dev/null && pwd )/"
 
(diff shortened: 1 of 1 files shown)


scripts/push-docker-images.sh
```diff
@@ -32,12 +32,21 @@ if [[ "${MUTABLE_TAG:-}" == "" ]]; then
     fi
 fi
 
+echo "Building Athens image for registry ${REGISTRY}"
+echo "Version ${VERSION}"
+echo "Mutable tag ${MUTABLE_TAG}"
+
 REPO_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." >/dev/null && pwd )/"
 
+echo "docker build -t ${REGISTRY}proxy:${VERSION} -f ${REPO_DIR}cmd/proxy/Dockerfile ${REPO_DIR}"
```
carolynvs: How about doing a `set -x` at the top of the script instead? I'm concerned that we are going to get drift between the logged lines and the actual commands run which would be extra hard to debug.

marpio: I think we're already do that.
https://github.com/gomods/athens/blob/657c9d04a35197a23667a064804deda38bcc0256/scripts/push-docker-images.sh#L4

marpio:
@arschles Do we need that? The script does `set -x` and so everything is logged and we can see it in the travis log.

Issue #809, opened 2018-10-24
Can we pass loggers down the stack via closures?


arschles:
**Is your feature request related to a problem? Please describe.**

#537 calls for passing log contexts down the stack (i.e. from handlers down to storage drivers).

I took a try at doing that in #645 but it wasn't too great.

**Describe the solution you'd like**

@fedepaol and @marwan-at-work talked about solving this by passing closures down the stack (see [here](https://github.com/gomods/athens/pull/645#issuecomment-427249684)), so I wanted to create this issue to remind us that's a possibility

**Describe alternatives you've considered**

**Additional context**

I added help wanted and hacktoberfest labels to this issue because I think it would be a good project for someone wanting to get familiar with the codebase. And since this can be done iteratively, we can break the refactor up into small pieces.

penDerGraft:
@arschles I'd like to work on at least a part of this. Is there a way I could get a little more context into the idea here?

penDerGraft:
@chriscoffee we may want to wait until the new entry is actually being used `pkg/download`. I should be able to make a PR for that tonight.

PR #844, opened 2018-11-01
add new log context middleware to store log entries in the context

penDerGraft:

The current, request-scoped log entry is not available throughout the stack without passing it as a param in multiple places. Also, the code that creates the entry also lives in multiple places at the moment. 

**How is the fix applied?**

Created new middleware that will create a log entry and store it in the buffalo.Context. I also added some helpers in `pkg/log` that own the storing/retrieving logic. 

This PR more or less implements the suggested approach in #809, but it uses the `buffalo.Context` instead of `context.Context`. I couldn't find a straightforward way to create a new `context.WithValue` without replacing the `buffalo.Context` everywhere. Happy to hear suggestions though. 

Also, I just added and tested the middleware. It isn't actually being used anywhere yet, so it adds to the duplication at the moment. I figured I could get some feedback on the approach and then apply it as incrementally as possible.

Ref #809


pkg/log/log_context.go
```diff
@@ -0,0 +1,31 @@
...
+	ctx.Set(logEntryKey, e)
+}
+
+// EntryFromContext returns an Entry that has been stored in the buffalo context.
+// If there is no value for the key or the type assertion fails, it returns a new
+// entry from the provided logger
+func EntryFromContext(ctx buffalo.Context, l *Logger) Entry {
+	d := ctx.Data()
+	v := d[logEntryKey]
+	if v == nil {
```
marwan-at-work: if there's no logger in the context, let's give them back a logger that does nothing. We should create a function in the log pacakge called `NoOpLogger` and call it on this line and line 27 below.

pkg/middleware/log_entry_test.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"bytes"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/sirupsen/logrus"
+
```
marwan-at-work: remove space

pkg/middleware/log_entry_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/stretchr/testify/assert"
+
```
marwan-at-work: ditto

pkg/log/log_context.go
```diff
@@ -0,0 +1,31 @@
+package log
+
+import (
+	"github.com/gobuffalo/buffalo"
+	"github.com/sirupsen/logrus"
+)
+
+const logEntryKey string = "log-entry-context-key"
```
marwan-at-work: no need to specify the type just `const logEntryKey = "..."`

pkg/middleware/log_entry.go
```diff
@@ -24,3 +24,21 @@ func LogEntryMiddleware(middleware middlewareFunc, lggr *log.Logger, validatorHo
 		}
 	}
 }
+
+// LogContextMiddleware builds a log.Entry, setting the request fields
+// and storing it in the context to be used throughout the stack
+func LogContextMiddleware(lggr *log.Logger) buffalo.MiddlewareFunc {
```
marwan-at-work: Can you remove the `LogEntryMiddleware` and rename this one to be `LogEntryMiddleware` and make sure the Filter is using this one properly. I'll update the Download Protocol logEntry in another PR :)

pkg/log/log_context.go
michalpristas: we can also use custom type for safety so it is extractable only for us

penDerGraft: @michalpristas I had the same thought but the `Set` method accepts `(string, interface{})`. I didn't know of a way around that.

pkg/middleware/log_entry_test.go
penDerGraft: Auto-formatter 🤦‍♂️

pkg/middleware/log_entry.go
penDerGraft: @marwan-at-work For sure. I didn't actually apply the middleware or remove any working code because I wanted to validate the approach first. I actually made the changes in `pkg/download`, but didn't commit them for that reason. If you're good with this strategy though, I'm happy to commit those changes (here or in a separate PR), but if you'd rather take care of it, that's great too 👍

pkg/log/log_context.go
marwan-at-work: @penDerGraft it should accept it because a constant with no type and quotes is just a string as well.

pkg/log/log_context.go
penDerGraft: Great suggestion. Are you thinking `NoOpLogger` would return a `Logger` instance, or should it create an actual `entry`?

pkg/log/log_context.go
michalpristas: ooh this is buffalo context, not the classic one. please ignore me

pkg/log/log_context.go
marwan-at-work: it should return a `*Logger` instance which would satisfy the Entry interface so it works here.

pkg/log/log_context.go
marwan-at-work: Ops. I didn't read @michalpristas's comments. Yes, this is a buffalo middleware :) -- if we end up removing buffalo for the API part some time in the future then we'd use the regular convention of making a special type for context keys.

arschles:

michalpristas:
Looks very promising. Thanks @penDerGraft

cmd/proxy/actions/app.go
```diff
@@ -129,7 +129,8 @@ func App(conf *config.Config) (*buffalo.App, error) {
 
 	// Having the hook set means we want to use it
 	if vHook := conf.ValidatorHook; vHook != "" {
-		app.Use(mw.LogEntryMiddleware(mw.NewValidationMiddleware, lggr, vHook))
+		app.Use(mw.LogEntryMiddleware(lggr))
```
marwan-at-work: Let's move this outside to be the top level middleware. AKA the first thing that an http request hits, so that all middlewares and http requests have access to the log.Entry.

michalpristas: @marwan-at-work should be addressed now right?

marwan-at-work: actually i see this is already done...gonna give this one more review and merge if peeps are happy with it

PR #845, opened 2018-11-01
Remove redis from toml file

marwan-at-work:
Removed Redis Address from TOML file since we don't use it.


michalpristas:
can you also remove it from:
- `docker-compose.yaml`
- `charts/proxy/values.yaml` -it is commented out there
- `charts/proxy/deployment.yaml` - also commented out

so we can complete this removal

marwan-at-work:
@michalpristas thanks for the note. done!

PR #846, opened 2018-11-01
config: flatten Proxy struct

marwan-at-work:
The config file used to have "Olympus" and "Proxy" configs inside of it. Now that we only have one proxy, let's flatten out the fields inside the Proxy struct since the entire config file is dedicated to the proxy. 

When giving coworkers a tour of Athens, one thing they were confused about is why certain fields in the config file were nested while others weren't...which led me to the Olympus discussion. Avoiding that should hopefully make our big config file more straight forward. 

I believe there's still more work to be done in simplifying our config but one step at at time.


arschles:

PR #847, opened 2018-11-01
Add support for DigitalOcean Spaces storage

### LINKED ISSUE #835: Implement storage.Backend for DO Spaces
Digital Ocean Spaces https://www.digitalocean.com/products/spaces/ is a s3 compatible blob storage with nice pricing. It would be great to be able to use that as modules storage.

Spaces is s3 compatible and can be used with the minio client we're already using. We could probably use the minio implementation. The only thing to add would be configuration.

**Additional context**
https://developers.digitalocean.com/documentation/spaces/

t-tomalak:

Add support for DigitalOcean Spaces

**How is the fix applied?**

Added Region variable to Minio configuration which is needed for DigitalOcean Spaces but not necessary for Minio storage.

Fixes #835


t-tomalak:
Hi, I don't think there is even good DO Spaces SDK in Go. Also DigitalOcean at their site, as example how to use their storage is using `minio-go` client https://developers.digitalocean.com/documentation/spaces/

marwan-at-work:
huh
>he API is interoperable with Amazon's AWS S3 API allowing you to interact with the service while using the tools you already know.

Does that mean we can just use our s3 package?

t-tomalak:
Minio also interpolate with S3 API, that's why we are using it for tests in s3. But in s3 config we don't provide endpoint and some other stuff which are not needed. That's why is easier to adopt minio configuration

t-tomalak:
I don't think is good place for that discussion but it may be a good idea if we review and decide on one implementation to use to "rule them all". I mean maybe we could remove `storage/minio` or leave there only `minio.go` and wrap in it `s3.New` method which already allow to customize it's configuration `func New(s3Conf *config.S3Config, cdnConf *config.CDNConfig, options ...func(*aws.Config)) (*Storage, error)`.

marwan-at-work:
@t-tomalak just double checking: S3/Spaces/Minio are *all* compatible with each other?

t-tomalak:
@marwan-at-work Yes, but for us they differentiate on what you need to provide in configurations

marpio:
Since the ability to stream is pretty important to us, we should evaluate how it's handled by each package.

arschles:

pkg/config/config_test.go
```diff
@@ -222,6 +223,7 @@ func TestParseExampleConfig(t *testing.T) {
 			Secret:    "minio123",
 			EnableSSL: false,
 			Bucket:    "gomods",
+			Region:    "",
```
manugupt1: We can remove this as `""` is zero-value.

manugupt1: thanks! let us wait for one more review and merge it.

Issue #851, opened 2018-11-01
Add doc for how to set up the local docs development environment


arschles:
**Is your feature request related to a problem? Please describe.**

We don't have docs on how to set up the local development environment to develop docs. That'd be nice because it would make docs authoring a bit easier. Maybe it'd encourage folks to write more docs too? No idea

**Describe the solution you'd like**

We should have a document on how to set up the docs server locally (inception!) - it's setting up [hugo](https://gohugo.io). I think @nunix also has a dockerized environment we can write up as well.

**Describe alternatives you've considered**

None. I think having a doc on how to do this is really important for the community.

**Additional context**

We talked about this in the November 1, 2018 meeting

marwan-at-work:
i'd like to work on this

PR #857, opened 2018-11-02
Add a Dockerfile for Hugo

### LINKED ISSUE #856: Create a Docker image for Hugo

In today's dev meeting (November 1, 2018) we discussed having documentation on how to run the documentation site locally. #851 talks about that.

Following up on #851, @nunix, @chriscoffee, @marwan-at-work and some other folks were asking if there's a docker image that contains [hugo](https://gohugo.io) (which we use to build the docs site, and has a local server for docs).

Providing such a docker image to folks would allow folks with just docker to get up and running immediately, without installing hugo (even though that's not too hard, and we'd document how to do that too)

There are some hugo docker images around, but as far as I can tell, there's no "official" hugo image. So I think we should create our own image and publish it to the `gomods` dockerhub organization.

We could use one of the other hugo images we've found.

**Additional context**

This is following up on #851 and some discussion in the slack channel

chriscoffee:

Following yesterdays development meeting they discussed running `hugo` locally. There's no official Dockerfile so lets go and create one!

> In today's dev meeting (November 1, 2018) we discussed having documentation on how to run the documentation site locally. #851 talks about that.

**How is the fix applied?**

I've added a simple Dockerfile tied down to the current version of alpine and hugo. The hugo version can be bumped with the environment variable `HUGO_VERSION` although this might become a chore so I could write a helper script to bump this automatically.

I also changed the Makefile to build the container which we can link up to CI/CD. I can also do this as part of this ticket if needed?

I added some simple documentation to `DEVELOPMENT.md` and can either update the information in #855 or ask @marwan-at-work kindly to do it.

As it's only a local development environment I haven't included the ability to alter whether hugo watches files and whether the base url would change. I have also altered the bind address to use `0.0.0.0` if this is a problem for people we can add a proxy or something.

Fixes #856

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -1,0 +1,21 @@
+FROM alpine:3.8
+
+ENV HUGO_VERSION=0.50
+ENV HUGO_BINARY=hugo_${HUGO_VERSION}_Linux-64bit.tar.gz
+
+ADD https://github.com/gohugoio/hugo/releases/download/v${HUGO_VERSION}/${HUGO_BINARY} /tmp
+
+RUN tar -xf /tmp/${HUGO_BINARY} -C /tmp \
+    && mv /tmp/hugo /usr/local/bin/hugo \
+    && rm -rf /tmp/hugo_${HUGO_VERSION}_linux_amd64 \
+    && rm -rf /tmp/${HUGO_BINARY} \
+    && rm -rf /tmp/LICENSE.md \
+    && rm -rf /tmp/README.md \
+    && apk upgrade --update \
+    && apk add --no-cache git asciidoctor libc6-compat libstdc++ ca-certificates
+
+WORKDIR /src
+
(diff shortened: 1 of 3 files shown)


chriscoffee:
We need to think about how we build this too. If we're going to put it in the same build script then I dunno if we should use the same tags as the proxy. We'd also need to probably only build when docs got commited to, so check with something like `git diff --name-only HEAD~10..HEAD docs` or something better written.

We could even split the hugo-server off into it's own repository

Any perference on how we do deploys of this?

docs/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM alpine:3.8@sha256:621c2f39f8133acb8e64023a94dbdf0d5ca81896102b9e57c0dc184cadaf5528
```
manugupt1: Do we need a particular SHA version? Can we remove it?

chriscoffee: Sorry it's just habit. If we trust alpine, which we probably should as it's an official build then not locking it down to the digest is fine. I'll get it removed.

PR #860, opened 2018-11-02
Config cleanup

### LINKED ISSUE #813: Clean Up Config File
Now that Olympus is out of the code base and we have a good idea of what the Proxy looks like, I'd like to do some cleaning up of the Config.go file: 

- [x] Remove CSRF middleware/option (the Proxy is just an API)
- [x] Remove unnecessary functions in the config parser: `setRuntimeDefaults` and `deleteInvalidStorageConfigs`
- [x] Include one Timeout instead of many? For simplicity sake. 
- [ ] Simplify tests: we don't need to test every single field in the config, we just wanna test behavior: env > toml. That's it. 

Please let me know if you'd like to change any of the above or add more to the list :)

marpio:
The config pkg contains some unnecessary/unused code.
**How is the fix applied?**

 * removed `setRuntimeDefaults` and `deleteInvalidStorageConfigs` functions
 * removed `CDNConfig` and introduced `AzureConfig`
 * removed `BaseURL()` methods on GCP and Azure since they are artefacts from the ancient time of Olympus.

Fixes parts of #813


pkg/config/azure.go
```diff
@@ -0,0 +1,9 @@
+package config
+
+// AzureConfig specifies the properties required to use Azure as the storage backend
+type AzureConfig struct {
```
marwan-at-work: is this part of this PR?

marwan-at-work:
@marpio nevermind i just read your PR description. Should have done that first 😄 -- will give it another go

pkg/config/azure.go
manugupt1: @marpio Overall, the PR looks good to me. Can you answer Marwan's question?

marwan-at-work:
i'd like to merge this but it seems the new azure-pipelines is failing. can we merge anyway @arschles ?

marpio:
@marwan-at-work I've added the new azure pipelines windows build. It's still not ready though, so I disabled it for PRs for now.
I will merge this so you can do the rest of the cleanup.
/cc @arschles

PR #866, opened 2018-11-05
add build details to binary

### LINKED ISSUE #842: Build the commit SHA / Tag into the built Athens binary

#830 has some doubt whether we're building and pushing the right docker image on CI builds. We can kinda sorta try and figure out whether there's a problem with trial and error and hand-inspecting the script that pushes, but it's not ideal and it's hard to be sure/

It would be ideal if the binary could tell us what commit it was built from. That way we could go into GH and find the code that built it. Added bonus if we can also build with the tag, if there was one. We can then print out the SHA/tag on the homepage of the proxy (and/or other pages) and maybe build a `version` command into the CLI too

Nothing much. I've seen this done in lots of other Go programs so it's tried and true.

**Additional context**

#830 got me thinking about this.

robjloranger:
We currently have no way of knowing for sure which version or commit a binary is built from.

This PR adds a small new package that provides build date, commit/version. The proxy command now has a `-version` flag which returns this information and exits.

Fixes #842


marpio:
Note: we should pass the version info here: https://github.com/gomods/athens/blob/54f92c55d610ab31a51837fa2a3379d926519bfe/scripts/push-docker-images.sh#L37

robjloranger:
I'm not sure what I broke here, I was in a rush this morning. I'll fix it up after work.

cmd/proxy/main.go
```diff
@@ -2,19 +2,27 @@ package main
...
 	"path/filepath"
 
 	"github.com/gomods/athens/cmd/proxy/actions"
+	"github.com/gomods/athens/pkg/build"
 	"github.com/gomods/athens/pkg/config"
 )
 
 var (
 	configFile = flag.String("config_file", filepath.Join("..", "..", "config.dev.toml"), "The path to the config file")
+	version    = flag.Bool("version", true, "Print version information and exit")
```
marpio: I think the default value should be false otherwise if you run the proxy without any flags it will print the version and exit.

pkg/build/build.go
```diff
@@ -0,0 +1,45 @@
...
+// or using the build script in ./scripts.
+package build
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// details represents known data for a given build
+type details struct {
```
marpio: Please add JSON tags

pkg/build/build.go
```diff
@@ -0,0 +1,45 @@
...
+	Version string
+	Commit  string
```
marpio: Just my opinionbut I think Version will be enough - if there is no tag than use commit hash otherwise tag.

scripts/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+#
+# build.sh runs a few commands to generate build time details
+# which are stored in variables in ./pkg/build
+
+commit=$(git rev-list -1 HEAD)
+version=$(git describe --tags)
+date=$(date -u +%Y-%m-%d-%H:%M:%S-%Z)
```
marpio: I'm not sure we need this b/c it can be derived from version but ifso, we should use UTC time.

pkg/build/build.go
```diff
@@ -0,0 +1,45 @@
...
+// String returns build details as a string with formatting
+// suitable for console output.
+//
+// i.e.
+// Build Details:
+//         Version:        v0.1.0-155-g1a20f8b
+//         Commit SHA:     1a20f8b6a36136183f8533ae850a582716bbd577
+//         Date:           2018-11-05-14:33:14-UTC
+func String() string {
+	return fmt.Sprintf("Build Details:\n\tVersion:\t%s\n\tCommit SHA:\t%s\n\tDate:\t\t%s", version, commitSHA, buildDate)
```
marpio: Is there a cross-platform 'new line ?

scripts/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+#
+# build.sh runs a few commands to generate build time details
+# which are stored in variables in ./pkg/build
+
+commit=$(git rev-list -1 HEAD)
+version=$(git describe --tags)
```
marpio: This should be in sync with the push-docker-image.sh

scripts/build.sh
marpio: Nevermind, didn't know -u eq --utc

robjloranger: It doesn't, `%Z` inserts the time zone of the machine at build time. Should we convert to UTC?

cmd/proxy/main.go
robjloranger: That's probably why the ci is timing out to lol

pkg/build/build.go
robjloranger: from what i can find, `fmt` deals with `\n` OS specific under the hood. I'm just going to go source diving to check.

robjloranger: most I could find was in regards to scanning:
```
In all the scanning functions, a carriage return followed
	immediately by a newline is treated as a plain newline
	(\r\n means the same as \n).
```

robjloranger: I asked in gophers slack general channel and it seems people are just using `\n`. The standard library does this as well. I will resolve for now and we can open an issue if it causes problems.

robjloranger:
I'm thinking the formatted output can be improved a bit. Maybe something to specify that the version may be a tag or a commit SHA.

current:
```console
Build Details:
        Version:        f4d9699
        Date:           2018-11-06-03:01:10-UTC
```

manugupt1:
This is so awesome! I did not realize this existed. We can do an endpoint in a later PR?

cmd/proxy/Dockerfile
```diff
@@ -8,7 +8,7 @@ WORKDIR $GOPATH/src/github.com/gomods/athens
 
 COPY . .
 
-RUN GO111MODULE=on CGO_ENABLED=0 go build -mod=vendor -o /bin/athens-proxy ./cmd/proxy
+RUN ./scripts/build.sh
```
marpio: @robjloranger What do you think about using docker build-time variables?
We could pass the Version obtained here: https://github.com/gomods/athens/blob/master/scripts/push-docker-images.sh
So instead of https://github.com/gomods/athens/blob/f53ab925034a5c42cbf77d1bd2c53e6fa671f69b/scripts/push-docker-images.sh#L37
we could do 
`docker build --build-arg VERSION=${VERSION} -t ${REGISTRY}proxy:${VERSION} -f ${REPO_DIR}cmd/proxy/Dockerfile ${REPO_DIR}`
and use the Version in the Dockerfile
This way we wouldn't need to get the version twice.
Let me know what you think of it!

robjloranger: I like it, I'll make the change. No sense doubling effort.

robjloranger: All set, as a bonus now we can check the image history and see the version as well. Build args are visible there.

marpio: not sure, but I think you need to define the variable in the Dockerfile
`ARG VERSION=default_value`

marpio: also shouldn't it be `./scripts/build.sh ${VERSION}`? and then in `build.sh` `VERSION=$1`?
or `VERSION=${VERSION} ./scripts/build.sh`?

robjloranger: I'm not certain, I'll have to look into it tonight.

robjloranger: I'm going to have to do more research tomorrow. Playing with it tonight I can't get the script to execute so I'm doing something wrong there. If I run `ls` instead, I am in the directory I expect so it's not that.

chriscoffee: It's because it doesn't exist in `PATH`, if you ran it with `CMD` it would work. Either do that or sort the path.

eg. this works, but so would setting `ENV PATH`
```
Step 4/13 : CMD ./scripts/build.sh
 ---> Running in 78205bcd36e3
Removing intermediate container 78205bcd36e3
```

arschles: if we ever release raw binaries, would they too have the version in them? no biggie if not, just wondering

marpio: @arschles we use goreleaser to release raw binaries and it takes care of the version as well.

robjloranger: I dropped the script and declaring `ARG VERSION` fixed the issue passing a value in. Build date is generated in the Dockerfile now so all docker builds will have a date and when we build it will have a version as well.

manugupt1:
Is it possible to do the same in a different PR so that it is easier for people who want to review docs can focus on reviewing docs and code and vice versa?

robjloranger:
Sounds good to me @manugupt1

PR #867, opened 2018-11-05
Filter: fixing after refactoring

michalpristas:

After we went away from workers and redesigned data flow, filters stayed one step behind.
This PR brings them even and adds a bit different meaning to filtering keywords considering new data flow.

**How is the fix applied?**

Change is following
- Include:
  - Treats module as it should be without using filter functionality used to reset parent behavior change
- Exclude:
  - Results in exclusion of the module entirely
- Direct:
  - Goes to the upstream proxy without consulting a local storage
  - Before: went to VCS  directly without consulting Olympus as a backup, treated module as a private repo

Filtering middleware moved after auth. So we don't tell whether we treat or we don't particular modules for unauthenticated users.

Removed handling of pseudo versions, it is treated just as a module rule tells it to

### DIFF
--- a/cmd/proxy/actions/app.go
+++ b/cmd/proxy/actions/app.go
@@ -119,11 +119,6 @@ func App(conf *config.Config) (*buffalo.App, error) {
 
 	initializeAuth(app)
 
-	// Having the hook set means we want to use it
-	if vHook := conf.ValidatorHook; vHook != "" {
-		app.Use(mw.LogEntryMiddleware(mw.NewValidationMiddleware, lggr, vHook))
-	}
-
 	user, pass, ok := conf.BasicAuth()
 	if ok {
 		app.Use(basicAuth(user, pass))
@@ -135,6 +130,11 @@ func App(conf *config.Config) (*buffalo.App, error) {
 			lggr.Fatal(err)
 		}
 		app.Use(mw.NewFilterMiddleware(mf, conf.GlobalEndpoint))
+	}
+
+	// Having the hook set means we want to use it
+	if vHook := conf.ValidatorHook; vHook != "" {
+		app.Use(mw.LogEntryMiddleware(mw.NewValidationMiddleware, lggr, vHook))
 	}
 
 	if err := addProxyRoutes(app, store, lggr, conf.GoBinary, conf.GoGetWorkers, conf.ProtocolWorkers); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/middleware/filter.go
```diff
@@ -50,10 +44,6 @@ func NewFilterMiddleware(mf *module.Filter, registryEndpoint string) buffalo.Mid
 	}
 }
 
-func isPseudoVersion(version string) bool {
-	return strings.HasPrefix(version, "v0.0.0-")
-}
-
-func redirectToRegistryURL(registryEndpoint string, u *url.URL) string {
+func redirectToUpstreamURL(registryEndpoint string, u *url.URL) string {
```
marpio: registryEndpoint -> upstreamEndpoint

cmd/proxy/actions/app.go
```diff
@@ -137,6 +129,14 @@ func App(conf *config.Config) (*buffalo.App, error) {
 		app.Use(basicAuth(user, pass))
 	}
 
+	if !conf.FilterOff() {
```
marpio: Wouldn't we validate a module even if it's on the `Exclude` list if the `ValidationMiddleware` is added **before** the `FilterMiddleware`?

michalpristas: that;s right

michalpristas: @marpio changed the order, please check if it makes sense for you now

pkg/module/filterRule.go
```diff
@@ -4,12 +4,13 @@ package module
 type FilterRule int
 
 const (
-	// Default filter rule does not alter default behavior
+	// Default filter rule does not alter default/parent behavior
```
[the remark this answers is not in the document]
michalpristas: nono default is important for behavior, not for configuration. default means that it has a behavior of a parent for example, for configuration `- github.com/a`
checking module module[github.com/a/b] will yield default which means look to its parent.
maybe logic can be rewritten, but i would keep it as a separate PR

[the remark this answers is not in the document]
manugupt1: ahh.. okay! so like a tree.

manugupt1:
@fedepaol Does this look okay to you? Once you give a 👍 , will merge it in.

Issue #800, opened 2018-10-19
Ability to go to a another proxy if not found


gcstang:
**Is your feature request related to a problem? Please describe.**
Need a way to point to the public repository and if it’s not found default to another i.e. like an internal ?
Or if not found in private point to public repository.

**Describe the solution you'd like**
When you setup a GOPROXY modules should resolve in your private or a public repository.
If you have a private repository then you should be able to have some sort of fall back to another repository. 

Currently:
Developer -> Public Proxy

Possible other Scenarios:
1. Developer -> Private Proxy -> Public Proxy (this could be accomplished in Athens)
2. Developer -> Public Proxy -> Private Proxy  (most likely this would require a local go change)

If this has already been accomplished then documentation should be added to the project.

marwan-at-work:
@gcstang what do you mean by repository? Are you talking about a backend? An actual github repository? VCS vs backend?

gcstang:
by repository I mean a VCS

marwan-at-work:
@gcstang can you provide an example? 
If say you have a dependency called `github.com/userA/pkgB`. This exist in only one domain: github. Why would we go to another VCS if it's not found in github? 

Even so, it wouldn't be possible because the `go` command only knows to go to github.com

morya:
I think it would be a very useful feature.

if working like this.

Athens-A deployed inside company
Athens-B some big website dedicated as public available.
C Final-vcs github and others

So go cmd talks to A, A try B if it doesn’t have it, and B will talk to github if needed.
All working together as a big distributed network.

The client only need to find the right sever to trust.

arschles:
@gcstang do you have any other thoughts on this? Specifically I'm interested in option (2). It looks like you'd like to tell Athens to check a public proxy first, then fall back to a private proxy or VCS?

If so, can you write down a use case?

gcstang:
@arschles basically what I was trying to accomplish by this train of thought was not having to change a proxy all the time on local machines, I'm not sure if that's possible.

First Example: (private -> public)
I'm working on my machine and my proxy internally has one fqdn int.proxy.domain.com:4000  to the company I work at on their network or VPN.  When I leave there and need to use the public repository I have to change to that GOPROXY setting i.e. proxy.domain.com:4000 . It would be nice if we could instead setup the proxy maybe running on a local docker instance to have a fallback instead.  

Second Example: (public -> private)
[...]

PR #868, opened 2018-11-05
Redirect to upstream on not found error

michalpristas:

We have upstream in place but we don't redirect to it.

**How is the fix applied?**

Introducing middleware which checks error returned and if is 404 it redirects to a configured upstream proxy. If upstream not defined it returns 

Doublechecking redirect URI for ""
once when registering callback, the second time (redundancy but it makes intention clear) in middleware itself 

Just found out that it
Fixes: #800


marpio:
@michalpristas  Is it meant as a replacement for https://github.com/gomods/athens/blob/f53ab925034a5c42cbf77d1bd2c53e6fa671f69b/pkg/middleware/filter.go#L38-L39

michalpristas:
@marpio not a replacement
the thing you linked says: redirect whatever, dont even check what you have here
this middleware says: check if we have it locally, if not let user check another proxy

use case is as follows:
you have a client clien does go get
it hits first level proxy this one cannot see for whatever reason the package on a VCS, maybe it can see only internal ones or VCS is down 
so it says try this proxy as my backup
client tries to connect to upstream proxy which knows about the package

michalpristas:
one more point, this is a very minor thing and probably we don't even need it

robjloranger:
@arschles, @michalpristas that is likely just from #899 which I am writing on a fix now.

robjloranger:
FYI @michalpristas I updated the branch to include the latest changes that fix the failing build.

michalpristas:
will try that @marwan-at-work 
I was offline for a while. hopefully, I will find time to get this right

arschles:
@michalpristas @robjloranger @marwan-at-work @marpio seems like this is no longer needed, since we're synchronously filling storage with code from the VCS or upstream proxy. Do I have that right?

manugupt1:
@michalpristas are you still on this?

arschles:
@michalpristas can we close this?

marpio:
@michalpristas closing for now. Feel free to re-open if you want to continue on that.

PR #871, opened 2018-11-06
Update docs for those struggling to set up buffalo

marwan-at-work:
We should mention that you don't need to install Buffalo to run Athens locally. #allYouNeedIsGo

Later this week, I will also add the "all you need is Docker" part. 

Fixes https://github.com/gomods/athens/issues/872

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -87,3 +87,11 @@ docker run -it --rm \
         gomods/hugo
         
 ```
+
+# Linting 
+
+In our CI/CD pass, we use golint, so feel free to install and run it locally beforehand: 
+
+```
+go get golang.org/x/lint/golint
+```
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -4,8 +4,7 @@ The proxy is built on the [Buffalo](https://gobuffalo.io/) framework. We chose
 this framework to make it as straightforward as possible to get your development environment up and running.
 However, **you do not need to install buffalo to run Athens**. 
 
-Buffalo provides nice features like a file watcher for your server, so if you'd like to install Buffalo, download [v0.12.4](https://github.com/gobuffalo/buffalo/releases/tag/v0.12.4) or later to get started on Athens,
-so be sure to download the CLI and put it into your `PATH`.
+Buffalo provides nice features like a file watcher for your server, so if you'd like to install Buffalo, download [v0.12.4](https://github.com/gobuffalo/buffalo/releases/tag/v0.12.4) or later to get started on Athens and be sure to download the CLI and put it into your `PATH`.
 
 Athens uses [Go Modules](https://golang.org/cmd/go/#hdr-Modules__module_versions__and_more) for dependency management. You will need Go [v1.11](https://golang.org/dl) or later to get started on Athens.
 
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -22,11 +20,33 @@ GO_BINARY_PATH=go1.11.X
...
+```console
+Starting application at 127.0.0.1:3000
+```
+
 ### Dependencies
 
 # Services that Athens Needs
 
-Athens relies on several services (i.e. databases, etc...) to function properly. We use [Docker](http://docker.com/) images to configure and run those services.
+Athens relies on several services (i.e. databases, etc...) to function properly. We use [Docker](http://docker.com/) images to configure and run those services. **However, Athens does not require any storage dependencies by default**. The default storage is in memory, you can opt-in to using the `fs` which would also require no dependencies. But if you'd like to test out Athens against a real storage backend (such as MongoDB, Minio, S3 etc), continue reading this section:
```
arschles: I thought the default storage is disk?

marwan-at-work: @arschles we had a discussion back in the day and ended up with Memory

DEVELOPMENT.md
```diff
@@ -2,17 +2,15 @@
...
-You'll need Buffalo [v0.12.4](https://github.com/gobuffalo/buffalo/releases/tag/v0.12.4) or later to get started on Athens,
+Buffalo provides nice features like a file watcher for your server, so if you'd like to install Buffalo, download [v0.12.4](https://github.com/gobuffalo/buffalo/releases/tag/v0.12.4) or later to get started on Athens,
 so be sure to download the CLI and put it into your `PATH`.
 
 Athens uses [Go Modules](https://golang.org/cmd/go/#hdr-Modules__module_versions__and_more) for dependency management. You will need Go [v1.11](https://golang.org/dl) or later to get started on Athens.
 
 See our [Contributing Guide](CONTRIBUTING.md) for tips on how to submit a pull request when you are ready.
 
-# Initial Development Environment Setup
-Athens relies on having a few tools installed locally. Run `make setup-dev-env` to install them.
```
Lyoness: maybe we keep `make setup-dev-env`, as it has `go get golang.org/x/lint/golint` (love the linter) but omit the [`get_buffalo.sh` script ](https://github.com/gomods/athens/blob/master/scripts/get_dev_tools.sh#L9)

DEVELOPMENT.md
arschles: oh wait - are you talking about https://github.com/gomods/athens/blob/master/config.dev.toml#L61 here? if so, ignore me and sorry to confuse this 😄

arschles:
P.S. I'm really excited for #allYouNeedIsDocker 🎉

DEVELOPMENT.md
marwan-at-work: @Layoness, our CI actually uses `make setup-dev-env` to install Buffalo and run an end-to-end test with it (see .travis.yml). So I'm a bit hesitant in changing the CI/CD behavior. 

I can either have a new make command that go gets the linter, or maybe just mention it in the README that we use `golang.org/x/lint/golint` and that users should install/run it before submitting a PR (though that's not necessary)

Lyoness: Ah. I hadn't thought of that.  the `go get` buffalo was failing for me locally with an error message identical to https://github.com/gobuffalo/buffalo/issues/1439, until I added a `-u`, so removing it as a instruction in DEVELOPMENT.md and asking users to install themselves is 👍, like you've done with asking users to install docker-compose as a dependency.

Issue #749, opened 2018-10-05
Add more documentation around "why does it matter"


arschles:
**Is your feature request related to a problem? Please describe.**

After #740 is merged, there will be a "why does it matter" section on the homepage that briefly explains why Athens matters for better dependency management, but it leaves out vendor directory problems, auth, and more (see [here](https://github.com/gomods/athens/pull/740/files#r222852546) for the full list by @marwan-at-work)

**Describe the solution you'd like**

I think we should create a full new page (maybe under the "intro" section) that has details on each reason why Athens matters. We should do a quick TL;DR of each of those on the homepage and then link to that new page.

**Describe alternatives you've considered**

We could put everything on the home page, but then it's not really an elevator pitch. Probably more like a walk-up-the-stairs pitch 😆 

**Additional context**

Follow-up on [this thread](https://github.com/gomods/athens/pull/740/files#r222852546)

arschles:
I"d like to take this one

chriscoffee:
@arschles What do you feel is missing here? I'd like to help out today/tonight, then we can get v0.1.0 out 🎆 🍰

marwan-at-work:
I'll pick this up real quick 😄

PR #873, opened 2018-11-06
doc: why does it matter

marwan-at-work:
Fixes https://github.com/gomods/athens/issues/749

Dependent on this PR: https://github.com/gomods/athens/pull/864

### DIFF
--- a/docs/content/_index.md
+++ b/docs/content/_index.md
@@ -15,6 +15,10 @@ On this site, we document Athens in detail. We'll teach you what it does, why it
 
 Athens provides a server for [Go Modules](https://github.com/golang/go/wiki/Modules) that you can run. It serves public code and your private code for you, so you don't have to pull directly from a version control system (VCS) like GitHub or GitLab.
 
+#### Why does it matter? 
+
+There are many reasons why you'd want a proxy server such as security and performance. [Take a look](/intro/why) at a few of them
+
 #### How Do I Use It?
 
 Athens is easy to run yourself. We give you a few options:
--- a/docs/content/_index.md
+++ b/docs/content/_index.md
@@ -15,16 +15,6 @@ On this site, we document Athens in detail. We'll teach you what it does, why it
 
 Athens provides a server for [Go Modules](https://github.com/golang/go/wiki/Modules) that you can run. It serves public code and your private code for you, so you don't have to pull directly from a version control system (VCS) like GitHub or GitLab.
 
-#### Why Does it Matter?
-
-Previously, the Go community has had lots of problems with libraries disappearing or changing without warning. It's easy for package maintainers to make changes to their code that can break yours - and much of the time it's an accident! Could your build break if one of your dependencies did this?
-
-- Commit `abdef` was deleted
-- Tag `v0.1.0` was force pushed
-- The repository was deleted altogether
-
- Since your app's dependencies come directly from GitHub, any of those above cases can happen to you and your builds can break when they do - oh no! Athens solves these problems by copying code from VCS's into _immutable_ storage.
-
 #### How Do I Use It?
 
 Athens is easy to run yourself. We give you a few options:
--- a/docs/content/intro/why.md
+++ b/docs/content/intro/why.md
@@ -21,7 +21,7 @@ The Go community has had lots of problems with libraries disappearing or changin
 
 ### Logic 
 
-The fact that the Go command line can now ping _your own_ server to download dependencies, that means you can program whatever logic you want around providing such dependencies. Things like Access Control (discussed below), adding custom versions, custom forks, and custom packages.
+The fact that the Go command line can now ping _your own_ server to download dependencies, that means you can program whatever logic you want around providing such dependencies. Things like Access Control (discussed below), adding custom versions, custom forks, and custom packages. For example, Athens provides a [Validation Hook](https://github.com/gomods/athens/blob/master/config.dev.toml#L87) that will get called for every module download to determine whether a module should be downloaded or not. Therefore, you can extend Athens with your own logic such as scanning a module path or code for red flags etc. 
 
 
 ### Performance 
--- a/docs/content/intro/why.md
+++ b/docs/content/intro/why.md
@@ -26,12 +26,12 @@ The fact that the Go command line can now ping _your own_ server to download dep
 
 ### Performance 
 
-Downloading stored dependencies from Athens is _significantly_ faster than downloading dependencies from Version Control Systems. This is because `go get` by default uses VCS to download modules such as `git clone`, while `go get` with GOPROXY enabled will use HTTP to download zip archives. THerefore, depending on your computer and internet connection speed, it takes 10 seconds to download the CockroachDB source tree as a zip file from GitHub but almost four minutes to git clone it. 
+Downloading stored dependencies from Athens is _significantly_ faster than downloading dependencies from Version Control Systems. This is because `go get` by default uses VCS to download modules such as `git clone`, while `go get` with GOPROXY enabled will use HTTP to download zip archives. Therefore, depending on your computer and internet connection speed, it takes 10 seconds to download the CockroachDB source tree as a zip file from GitHub but almost four minutes to git clone it. 
 
 ### Access Control 
 
 Worse than packages disappearing, packages can be malicious. To make sure no such malicious package is ever installed by your team or company, you can have your proxy server return a 500 when the Go command line asks for an excluded module. This will cause the build to fail because Go expects a 200 HTTP response code. With Athens, you can achieve this through the [filter file](/configuration/filter.md). 
 
 
-### No Vendor Directory
+### Vendor Directory Becomes Optional
 With immutability, performance, and a robust proxy server, there's no longer an absolute need for each repository to have its vendor directory checked in to its version control. The go.sum file ensures that no package is manipulated after the first install. Furthermore, your CI/CD can install all of your dependencies on every build with little time. 
--- a/docs/content/intro/why.md
+++ b/docs/content/intro/why.md
@@ -1,0 +1,34 @@
+---
+title: "Why Does It Matter?"
+date: 2018-11-06T13:58:58-07:00
+weight: 4
+LastModifierDisplayName: "Marwan"
+LastModifierEmail: "marwan.sameer@gmail.com"
+
+---
+
+### Immutability
+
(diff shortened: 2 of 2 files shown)


docs/content/intro/why.md
```diff
@@ -0,0 +1,34 @@
...
+LastModifierDisplayName: "Marwan"
+LastModifierEmail: "marwan.sameer@gmail.com"
+
+---
+
+### Immutability
+
+Previously, the Go community has had lots of problems with libraries disappearing or changing without warning. It's easy for package maintainers to make changes to their code that can break yours - and much of the time it's an accident! Could your build break if one of your dependencies did this?
+
+- Commit `abdef` was deleted
```
manugupt1: Let's not have this now as we do not support it?

docs/content/intro/why.md
```diff
@@ -0,0 +1,34 @@
...
+- The repository was deleted altogether
+
+ Since your app's dependencies come directly from GitHub, any of those above cases can happen to you and your builds can break when they do - oh no! Athens solves these problems by copying code from VCS's into _immutable_ storage.
+
+ This way, you don't need to upload anything manually to Athens storage. The first time Go asks Athens for a dependency, Athens will go get it from VCS (github, bitbucket etc). But once that module has been retrieved, it will be forever persisted in its storage backend and will never go back to VCS or that same version again. This is how Athens achieves module immutability. Keep in mind, you are in charge of that storage backend. 
+
+
+### Logic 
+
+The fact that the Go command line can now ping _your own_ server to download dependencies, that means you can program whatever logic you want around providing such dependencies. Things like Access Control (discussed below), adding custom versions, custom forks, custom packages etc.
```
manugupt1: The last sentence is incomplete. Can you add like etc. is possible or something else?

docs/content/intro/why.md
```diff
@@ -0,0 +1,34 @@
...
+The fact that the Go command line can now ping _your own_ server to download dependencies, that means you can program whatever logic you want around providing such dependencies. Things like Access Control (discussed below), adding custom versions, custom forks, custom packages etc. 
+
+
+### Performance 
+
+Downloading stored dependencies from Athens is _significantly_ faster than downloading dependencies from Version Control Systems. For example, it takes 10 seconds to download the CockroachDB source tree as a zip file from GitHub but almost four minutes to git clone it. 
+
+### Access Control 
+
+Worse than packages disappeaing, packages can be malicious. Therefore, you can make sure that no program inside your company or team will ever install github.com/some-user/malicious-package. This is because Go will come to your server and ask for that same exact package, and instead of fetching it from github, you can just return a 500 to the Go command line causing the build to fail. With Athens, you can achieve this through the [filter file](/configuration/filter.md).
```
manugupt1: Worse than packages disappearing packages are malicious packages. To make sure that no program .will ever install..., the proxy can be used.

manugupt1: Also, check the spelling for disappearing!

docs/content/intro/why.md
```diff
@@ -0,0 +1,37 @@
...
+date: 2018-11-06T13:58:58-07:00
+weight: 4
+LastModifierDisplayName: "Marwan"
+LastModifierEmail: "marwan.sameer@gmail.com"
+
+---
+
+### Immutability
+
+Previously, the Go community has had lots of problems with libraries disappearing or changing without warning. It's easy for package maintainers to make changes to their code that can break yours - and much of the time it's an accident! Could your build break if one of your dependencies did this?
```
marpio: I wouldn't say `Previously` - the problem persists as of now.

docs/content/intro/why.md
```diff
@@ -0,0 +1,37 @@
...
+
+- Commit `abdef` was deleted
+- Tag `v0.1.0` was force pushed
+- The repository was deleted altogether
+
+ Since your app's dependencies come directly from GitHub, any of those above cases can happen to you and your builds can break when they do - oh no! Athens solves these problems by copying code from VCS's into _immutable_ storage.
```
marpio: `directly from GitHub` - maybe directly from VCS?

docs/content/intro/why.md
```diff
@@ -0,0 +1,37 @@
...
+
+ This way, you don't need to upload anything manually to Athens storage. The first time Go asks Athens for a dependency, Athens will go get it from VCS (github, bitbucket etc). But once that module has been retrieved, it will be forever persisted in its storage backend and will never go back to VCS or that same version again. This is how Athens achieves module immutability. Keep in mind, you are in charge of that storage backend.
```
marpio: `and will never go back to VCS or that same version again.` 
or -> for
Who will never go back to VCS? Maybe "and the proxy will never..."

docs/content/intro/why.md
```diff
@@ -0,0 +1,37 @@
...
+ This way, you don't need to upload anything manually to Athens storage. The first time Go asks Athens for a dependency, Athens will go get it from VCS (github, bitbucket etc). But once that module has been retrieved, it will be forever persisted in its storage backend and will never go back to VCS or that same version again. This is how Athens achieves module immutability. Keep in mind, you are in charge of that storage backend. 
+
+### Logic 
+
+The fact that the Go command line can now ping _your own_ server to download dependencies, that means you can program whatever logic you want around providing such dependencies. Things like Access Control (discussed below), adding custom versions, custom forks, and custom packages. 
+
+
+### Performance 
+
+Downloading stored dependencies from Athens is _significantly_ faster than downloading dependencies from Version Control Systems. For example, it takes 10 seconds to download the CockroachDB source tree as a zip file from GitHub but almost four minutes to git clone it.
```
marpio: `four minutes to git clone it.`
Maybe "to go get it." since thats what matters and maybe not everyone knows that go uses git clone to fetch a mod.

docs/content/intro/why.md
```diff
@@ -0,0 +1,37 @@
...
+The fact that the Go command line can now ping _your own_ server to download dependencies, that means you can program whatever logic you want around providing such dependencies. Things like Access Control (discussed below), adding custom versions, custom forks, and custom packages. 
+
+
+### Performance 
+
+Downloading stored dependencies from Athens is _significantly_ faster than downloading dependencies from Version Control Systems. For example, it takes 10 seconds to download the CockroachDB source tree as a zip file from GitHub but almost four minutes to git clone it. 
+
+### Access Control 
+
+Worse than packages disappearing, packages can be malicious. To make sure no such malicious is ever installed by your team or company, you can have your proxy server return a 500 when the Go command line asks for an excluded module. This will cause the build to fail because expects a 200 HTTP response code. With Athens, you can achieve this through the [filter file](/configuration/filter.md).
```
marpio: `To make sure no such malicious is ever`
no such malicious **module( or package)** is ever

`This will cause the build to fail because expects a 200 HTTP response code.`
because `go` expects

marpio: Also we could mention the validation hook
https://github.com/gomods/athens/blob/9f8e7bc14594438a5d6df239623e4342d41ec12a/cmd/proxy/actions/app.go#L139-L141

docs/content/intro/why.md
michalpristas: also * depends on your connection speed

docs/content/intro/why.md
manugupt1: We can't fetch by commit hashes right? I am a bit confused?

marwan-at-work: @manugupt1 we can if they are pseudo versions. This is a bit unrelated.

docs/content/intro/why.md
marwan-at-work: Got it, I'll actually put that in the Logic reason then.

docs/content/intro/why.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Logic 
+
+The fact that the Go command line can now ping _your own_ server to download dependencies, that means you can program whatever logic you want around providing such dependencies. Things like Access Control (discussed below), adding custom versions, custom forks, and custom packages. For example, Athens provides a [Validation Hook](https://github.com/gomods/athens/blob/master/config.dev.toml#L87) that will get called for every module download to determine whether a module should be downloaded or not. Therefore, you can extend Athens with your own logic such as scanning a module path or code for red flags etc. 
+
+
+### Performance 
+
+Downloading stored dependencies from Athens is _significantly_ faster than downloading dependencies from Version Control Systems. This is because `go get` by default uses VCS to download modules such as `git clone`, while `go get` with GOPROXY enabled will use HTTP to download zip archives. THerefore, depending on your computer and internet connection speed, it takes 10 seconds to download the CockroachDB source tree as a zip file from GitHub but almost four minutes to git clone it.
```
marpio: THerefore, depending...
Therefore, depending...

docs/content/intro/why.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Access Control 
+
+Worse than packages disappearing, packages can be malicious. To make sure no such malicious package is ever installed by your team or company, you can have your proxy server return a 500 when the Go command line asks for an excluded module. This will cause the build to fail because Go expects a 200 HTTP response code. With Athens, you can achieve this through the [filter file](/configuration/filter.md). 
+
+
+### No Vendor Directory
+With immutability, performance, and a robust proxy server, there's no longer an absolute need for each repository to have its vendor directory checked in to its version control. The go.sum file ensures that no package is manipulated after the first install. Furthermore, your CI/CD can install all of your dependencies on every build with little time.
```
arschles: I think we should say that projects don't have to use a vendor directory if they don't want to, so we're not perceived as trying to replace the vendor directory (even though we can totally replace it 😉 )

What do you think?

arschles: Also maybe we should mention that not having a vendor directory can be beneficial for libraries that are intended exclusively to be imported (i.e. not buildable into binaries)?

docs/content/_index.md
```diff
@@ -15,16 +15,6 @@ On this site, we document Athens in detail. We'll teach you what it does, why it
...
-#### Why Does it Matter?
-
-Previously, the Go community has had lots of problems with libraries disappearing or changing without warning. It's easy for package maintainers to make changes to their code that can break yours - and much of the time it's an accident! Could your build break if one of your dependencies did this?
-
-- Commit `abdef` was deleted
-- Tag `v0.1.0` was force pushed
-- The repository was deleted altogether
-
- Since your app's dependencies come directly from GitHub, any of those above cases can happen to you and your builds can break when they do - oh no! Athens solves these problems by copying code from VCS's into _immutable_ storage.
-
```
arschles: could we have a link to the new `/intro/why` document in here?

PR #874, opened 2018-11-07
extract log entry from context in pkg/download

### LINKED ISSUE #809: Can we pass loggers down the stack via closures?

penDerGraft:

Use the log entry stored in the buffalo context via middleware in pkg/download so that we're not creating log entries from the request in multiple places. #844 creates the middleware but the download package is still creating it's own entry for the protocol handlers. This PR updates that logic. 

**How is the fix applied?**

Use the existing `LogEntryHandler` to extract the log entry from the buffalo context and pass it to the protocol handlers.

Fixes #809

### DIFF
--- a/pkg/download/handler.go
+++ b/pkg/download/handler.go
@@ -5,7 +5,6 @@ import (
 	"github.com/gobuffalo/buffalo/render"
 	"github.com/gomods/athens/pkg/log"
 	"github.com/gomods/athens/pkg/middleware"
-	"github.com/sirupsen/logrus"
 )
 
 // ProtocolHandler is a function that takes all that it needs to return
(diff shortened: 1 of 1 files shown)


michalpristas:
Merging in as we discussed that Log package should be part of Beta and is not affected by code freeze

PR #884, opened 2018-11-09
using the proper env var name for the mongo URL

arschles:

When you `helm install` the proxy into a Kubernetes cluster, the proxy doesn't start up. That's because the helm chart sets `ATHENS_MONGO_CONNECTION_STRING` but the code looks for `ATHENS_MONGO_STORAGE_URL`. The result is that each proxy pod never starts serving because it tries to connect to a mongoDB server running on localhost.

**How is the fix applied?**

I changed `ATHENS_MONGO_CONNECTION_STRING` to `ATHENS_MONGO_STORAGE_URL` in the helm chart.

@marwan-at-work could we get this into the `v0.2.0` release by chance?


manugupt1:
agree with Michal

PR #888, opened 2018-11-09
Adding a docker compose service for local runs

arschles:

There are a few steps to set up the local environment to run the proxy. It could be easier

**How is the fix applied?**

I added the app to the docker compose file, so you should be able to run `docker-compose up dev`

This is just a start, but should make it easier to run athens locally in docker compose, with all services available.

I don't think this works right now b/c the networks are messed up between mongo & the app. After that's working, other services should be added

cc/ @marwan-at-work because you talked about this before. ~I'm tagging this with do not merge for now since I think you have something more comprehensive~ (Edit from Aaron on Dec. 5, 2018: I think this is a good foundation for https://github.com/gomods/athens/issues/420)


docker-compose.yml
```diff
@@ -1,9 +1,26 @@
...
+      - ATHENS_MONGO_STORAGE_URL=mongodb://127.0.0.1:27017
+      - TIMEOUT=2
+      # - ATHENS_STORAGE_TYPE=mongo
+      - ATHENS_STORAGE_TYPE=memory
+    ports:
+      - 3000:3000
+    depends_on:
+      - mongo
+    networks:
+      - mongo
```
marwan-at-work: do we need a mongo network? docker-compose creates a network by default making all the services reachable

arschles: @marwan-at-work right you are! I'm gonna push this change shortly

DEVELOPMENT.md
```diff
@@ -10,57 +10,67 @@ Athens uses [Go Modules](https://golang.org/cmd/go/#hdr-Modules__module_versions
...
+1. Access Athens on `http://localhost:3000`
 
-That's it! After the `make dev` command is done, everything will be up and running and you can move
-on to the next step.
+That's it! Athens will take a few seconds to configure and connect to all the services it uses. When it's done, everything will be up and running and ready to use.
 
 If you want to stop everything at any time, run `make down`.
 
-Note that `make dev` only runs the minimum amount of dependencies needed for things to work. If you'd like to run all the possible dependencies run `make alldeps` or directly the services available in the `docker-compose.yml` file. Keep in mind, though, that `make alldeps` does not start up Athens or Oympus, but **only** their dependencies.
+>`make down` will stop all the resources that `make dev` started. If you want to completely delete all the unused Docker resources on your system, run `docker system prune`. This command may delete other unused resources that you didn't create with `make dev`
```
chriscoffee: A bit nit picking but do we need the part about pruning. Won't `docker-compose down` tear down any resources and their networks? Thay we don't need to suggest people run `docker system prune` will `unused containers, networks, images (both dangling and unreferenced)` which could get rid of resources they need elsewhere.

DEVELOPMENT.md
```diff
@@ -10,57 +10,67 @@ Athens uses [Go Modules](https://golang.org/cmd/go/#hdr-Modules__module_versions
...
 
-Athens relies on several services (i.e. databases, etc...) to function properly. We use [Docker](http://docker.com/) images to configure and run those services. **However, Athens does not require any storage dependencies by default**. The default storage is in memory, you can opt-in to using the `fs` which would also require no dependencies. But if you'd like to test out Athens against a real storage backend (such as MongoDB, Minio, S3 etc), continue reading this section:
+By default, Athens doesn't require any external services to be running. By default, it uses in-memory storage and you can also opt into on-disk storage, and neither method requires any external services.
+
+However, Athens does support other storage drivers, and takes advantage of other external services that can be helpful for more permanent and "production-like" deployments. Here are some of the services that Athens can take advantage of:
+
+- [Minio]((https://minio.io/)) storage driver
+- [MongoDB](https://www.mongodb.com/) storage driver
+- [Jaeger](https://www.jaegertracing.io/) for distributed tracing
+
```
chriscoffee: We support datadog too! Are we counting this as a _service_ that Athens can take advantage of?

DEVELOPMENT.md
arschles: Yea, it does. The problem comes in if they're not in the athens directory, but to be honest that's mostly my problem and I let my extreme solution leak into the docs 😆 . I'll take this out.

arschles:
@chriscoffee I addressed your comments - can you check out what you think again?

DEVELOPMENT.md
arschles: Ooh good call. I added that

chriscoffee:
> @chriscoffee I addressed your comments - can you check out what you think again?

image

DEVELOPMENT.md
```diff
@@ -10,57 +10,68 @@ Athens uses [Go Modules](https://golang.org/cmd/go/#hdr-Modules__module_versions
...
 
-### Go version
-Athens is developed on Go1.11+.
+## Go Version
 
-To point Athens to a different version of Go set the following environment variable
-```
-GO_BINARY_PATH=go1.11.X
-or whichever binary you want to use with athens
+You'll need Go version 1.11 or higher to build and run Athens. Since it uses the `go` CLI internally, you can tell it which CLI to use via the `GO_BINARY_PATH` environment. Set it to the binary of choice like this:
```
robjloranger: `GO_BINARY_PATH` environment variable

DEVELOPMENT.md
robjloranger: Should we list other storages as well, like GCP AWS etc

DEVELOPMENT.md
```diff
@@ -89,10 +100,10 @@ docker run -it --rm \
 
 Then open [http://localhost:1313](http://localhost:1313/).
 
-# Linting
+# Run the Linter on the Code
 
 In our CI/CD pass, we use golint, so feel free to install and run it locally beforehand:
 
-```
+```go
```
robjloranger: should probably be console

arschles:
Closing in favor of #1082

PR #889, opened 2018-11-09
Checking the number of bytes written to Mongo GridFS

arschles:

There was an `io.Copy` call to send source code zip files to Mongo's GridFS, but we weren't checking the actual number of bytes sent by the copy.

**How is the fix applied?**

This checks the return number of bytes, in addition to the error. If there was no error, but 0 bytes (or a negative number was returned) were written, then this also errors.

### DIFF
--- a/pkg/storage/mongo/saver.go
+++ b/pkg/storage/mongo/saver.go
@@ -23,9 +23,13 @@ func (s *ModuleStore) Save(ctx context.Context, module, version string, mod []by
 	}
 	defer f.Close()
 
-	_, err = io.Copy(f, zip) // check number of bytes written?
+	numBytesWritten, err := io.Copy(f, zip)
 	if err != nil {
 		return errors.E(op, err, errors.M(module), errors.V(version))
+	}
+	if numBytesWritten <= 0 {
+		e := fmt.Errorf("copied %d bytes to Mongo GridFS", numBytesWritten)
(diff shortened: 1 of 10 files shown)


pkg/storage/mongo/saver.go
```diff
@@ -23,10 +24,14 @@ func (s *ModuleStore) Save(ctx context.Context, module, version string, mod []by
 	}
 	defer f.Close()
 
-	_, err = io.Copy(f, zip) // check number of bytes written?
+	numBytesWritten, err := io.Copy(f, zip)
 	if err != nil {
 		return errors.E(op, err, errors.M(module), errors.V(version))
 	}
+	if numBytesWritten <= 0 {
+		e := fmt.Errorf("copied %d bytes to Mongo GridFS", numBytesWritten)
```
robjloranger: I think since this only fires when there were `<=0` bytes written we could just write than in.

michalpristas: looks good, i wonder if we can get content length from vcs response and use it here for validation, not needed to be done in this PR (separate ok) just a thought

michalpristas: oh we're not using API calls so we cannot but we can do Stat to find out filesize and pass it down,
in case we start using API calls we can do the same with content length

PR #900, opened 2018-11-11
Fix docker builds in Travis

### LINKED ISSUE #899: Docker build is failing on travis
The new change to the proxy dockerfile is failing when built during CI. It does build fine when done locally.

I'm thinking of moving the call to `date` out on the argument passed to the go linker and storing in a env var or build arg.

I'm working on this.

robjloranger:
fixes issue with `date` command being evaluated inline go linker flag on Travis.

closes #899

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_script:
   - chmod +x minio && nohup ./minio server --address ":9001" . &
 
 script:
-  - make verify test-unit test-e2e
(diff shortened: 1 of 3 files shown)


robjloranger:
After the successful test build in https://travis-ci.org/gomods/athens/builds/453468377, I consider this fix effective and we can merge this in to fix the broken docker build

PR #901, opened 2018-11-11
Changed HUGO_VERSION variable to ARG instead of ENV

nunix:
this is a small change, however it will allow the creation of images based on a specific GoHugo version if needed.
The build command should be something like:
`docker build --build-arg HUGO_VERSION=0.51 -t repo/image-name .`

I tested the build as written above and it worked fine, but please feel free to test it further.

**How is the fix applied?**

Fixes #


manugupt1:
Hi @nunix Thanks for this. This is awesome 🙇 If you have additional improvements, please open a new PR. I am merging this now.

PR #903, opened 2018-11-11
Fix CI build

carolynvs:

The CI build on master is broken.

**How is the fix applied?**

* Do not use an invalid version (with a whitespace) in the default VERSION value in our dockerfile.
* Do a docker build on PR builds so that we find problems before we merge.

Addresses broken build https://travis-ci.org/gomods/athens/builds/453641688#L1457

### DIFF
--- a/cmd/proxy/Dockerfile
+++ b/cmd/proxy/Dockerfile
@@ -8,7 +8,7 @@ WORKDIR $GOPATH/src/github.com/gomods/athens
 
 COPY . .
 
-ARG VERSION="Not Specified"
+ARG VERSION="unset"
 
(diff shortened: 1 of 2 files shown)


cmd/proxy/Dockerfile
```diff
@@ -8,7 +8,7 @@ WORKDIR $GOPATH/src/github.com/gomods/athens
 
 COPY . .
 
-ARG VERSION="Not Specified"
+ARG VERSION="unset"
 
 RUN DATE="$(date -u +%Y-%m-%d-%H:%M:%S-%Z)" && GO111MODULE=on CGO_ENABLED=0 go build -mod=vendor -ldflags "-X github.com/gomods/athens/pkg/build.version=$VERSION -X github.com/gomods/athens/pkg/build.buildDate=$DATE" -o /bin/athens-proxy ./cmd/proxy
```
marwan-at-work: can we remove the first `&&` here, it's unnecessary or worse doesn't register DATE for the next command?

carolynvs: I think it is available in the next command, here's my expert test:

```console 
~$ FARTS=on && echo $FARTS
on
```

But yes, I think I can remove it. I'll wait for a green build then tweak.

marwan-at-work: @carolynvs LOL, thank you!

PR #909, opened 2018-11-12
Rename docker image

carolynvs:

We renamed the project to Athens but didn't rename the docker image.

**How is the fix applied?**

* I've manually retagged our docker image from `gomods/proxy` to `gomods/athens` for the release v0.2.0
* I've updated the CI to push using the new tag going forward.
* I've switched from using the stable tag to using latest, to fit our documentation and docker standards.
* I've updated the helm chart to match and also bumped the tag used in the chart with our latest release (v0.2.0).
* I've updated the documentation.

N/A

### DIFF
--- a/scripts/push-docker-images.sh
+++ b/scripts/push-docker-images.sh
@@ -18,11 +18,11 @@ else
     BRANCH=${BRANCH:-$(git symbolic-ref -q --short HEAD || echo "")}
 fi
 
-# MUTABLE_TAG is the docker image tag that we will reuse between pushes, it is not a stable tag like a commit hash or tag.
+# MUTABLE_TAG is the docker image tag that we will reuse between pushes, it is not an immutable tag like a commit hash or tag.
 if [[ "${MUTABLE_TAG:-}" == "" ]]; then
     # tagged builds
     if [[ "$VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
-        MUTABLE_TAG="stable"
+        MUTABLE_TAG="latest"
     # master build
     elif [[ "$BRANCH" == "master" ]]; then
         MUTABLE_TAG="canary"
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ docker: proxy-docker
 
 .PHONY: proxy-docker
 proxy-docker:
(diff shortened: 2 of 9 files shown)


scripts/push-docker-images.sh
```diff
@@ -18,11 +18,11 @@ else
     BRANCH=${BRANCH:-$(git symbolic-ref -q --short HEAD || echo "")}
 fi
 
-# MUTABLE_TAG is the docker image tag that we will reuse between pushes, it is not a stable tag like a commit hash or tag.
+# MUTABLE_TAG is the docker image tag that we will reuse between pushes, it is not an immutable tag like a commit hash or tag.
 if [[ "${MUTABLE_TAG:-}" == "" ]]; then
     # tagged builds
     if [[ "$VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
-        MUTABLE_TAG="stable"
+        MUTABLE_TAG="latest"
```
michalpristas: check #858

carolynvs: the reason why I suggest changing it to latest is this:
```$ docker pull gomods/athens
Using default tag: latest
Error response from daemon: manifest for gomods/athens:latest not found
```

The latest tag is a convention in docker. IMO not pushing a latest tag isn’t a viable option, though we can also tag things with stable too if we feel strongly about it. We always have the `:v0.2.0` immutable tags that I hope people use most of the time so the latest tag is really just to help with doc examples and people trying stuff out.

robjloranger: I agree against using latest. It is an arbitrary tag which points to the most recent image **without a tag**

carolynvs: @robjloranger I'm sorry that this got merged before you had an opportunity to comment. This doesn't have to be a final decision, it was mostly just us trying to get a beta out the door. If you have a suggestion for an alternative, how about opening up a proposal issue and we can make this a more intentional change?

The switch to the stable tag ended up breaking our releases and doc, so it's probably a good idea to not make changes like this quickly and without more thorough review going forward.

robjloranger: No problem, I think I had the tab open half the day and commented not knowing it had been merged

PR #913, opened 2018-11-13
mongo: remove drop dups option and return error if module exists

### LINKED ISSUE #907: DropDups is silently ignored after MongoDB >2.8
After some discussion in [#athens](https://gophers.slack.com/archives/C9LRAQN8N/p1542027520801700) we figured out that dropDupes is [no longer available after 3.0](https://docs.mongodb.com/manual/release-notes/3.0-compatibility/#remove-dropdups-option) for `createIndex()`, `ensureIndex()`, and `createIndexes.`

The latest stable version of MongoDB is 4.0.

Should we consider dropping this or continue to support it to allow some backwards compatibility? And should it be true?

For those wanting to know what DropDups does:
`DropDups   bool     // Drop documents with the same index key as a previously indexed one`

penDerGraft:

Remove the silently ignored option "DropDups" from the mongo configuration.

**How is the fix applied?**

The `Save` method in mongo now returns an error when the module exists. I used the logic from the GCP implementation.

Fixes #907


manugupt1:
thanks @penDerGraft for this! This is awesome 🙇

PR #821, opened 2018-10-25
Use Azurite to test Azure Storage

chriscoffee:
**What is the problem I am trying to address?**
At the moment we skip the Azure Storage tests in our travis builds. We could use Azurite https://github.com/Azure/Azurite to test the storage just like we test s3 using minio.

**How is the fix applied?**

Whilst we wait for the interface to be implemented so we can enable the tests #703. I've updated the compose file to include azurite and also sorted the docker-compose file so it doesn't have a mixture of  array/dicts for environment variable defintions and as a list for command. Also sorted alphabetically.

- [x] Add Azurite to docker-compose
- [ ] Add tests against local Azure Blob Storage

Signed-off-by: Chris M <millscj01@gmail.com>

Fixes #820

PR #917, opened 2018-11-15
Implement storage.Backend for azureblob

### LINKED ISSUE #703: Implement storage.Backend for Azure Blob storage
Azure Blob Storage is missing implementation

- Rename `azurecdn` to `azureblob`
- Let `azure` storage join the fun in `/pkg/storage/storage_tests`

scriptonist:
Towards #703 

**Mention briefly how you have applied the fix.**

storage.Backend interface is now implemented for azureblob. This a **WIP** PR intended for early feedback. **No** tests have been written.I believe It that can be done after merging  https://github.com/gomods/athens/pull/821. Please review the code and provide feedback. Thanks 🙏 

Fixes #703


pkg/storage/azureblob/saver.go
```diff
@@ -0,0 +1,25 @@
...
+	"io"
+
+	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/observ"
+	moduploader "github.com/gomods/athens/pkg/storage/module"
+)
+
+// Save implements the (github.com/gomods/athens/pkg/storage).Saver interface.
+func (s *Storage) Save(ctx context.Context, module, version string, mod []byte, zip io.Reader, info []byte) error {
+	fmt.Printf("Saving %v\n", module)
```
marpio: Please remove this

pkg/storage/azureblob/azureblob.go
```diff
@@ -53,18 +55,60 @@ func New(conf *config.AzureConfig) (*Storage, error) {
...
-	defer span.End()
-	err := moduploader.Upload(ctx, module, version, bytes.NewReader(info), bytes.NewReader(mod), zip, s.cl.UploadWithContext, s.conf.TimeoutDuration())
+// BlobExists checks if a particular blob exists in the container
+func (c *azureBlobStoreClient) BlobExists(ctx context.Context, path string) (bool, error) {
+	// TODO: Any better way of doing this ?
+	blobURL := c.containerURL.NewBlockBlobURL(path)
+	_, err := blobURL.GetProperties(ctx, azblob.BlobAccessConditions{})
+	if err != nil {
+		serr := err.(azblob.StorageError)
+		if serr.Response().StatusCode == 404 {
```
marpio: Can you do something like `if serr, ok := err.(azblob.StorageError); !ok {...` and also use `http.StatusNotFound` instead of 404?

config.dev.toml
```diff
@@ -211,3 +211,14 @@ TraceExporter = ""
...
+        # Storage Account name for Azure Blob
+        AccountName = "AZURE_BLOB_ACCOUNT_NAME"
+
+        # Account Key to use with the storage account
+        AccountKey = "AZURE_BLOB_ACCOUNT_KEY"
+
+        # Name of container in the blob storage
+        ContainerName = "AZURE_BLOB_CONTAINER_NAME"
+
+        Timeout = 300
```
marpio: Remove this pls - we only have one top level timeout now.

pkg/storage/azureblob/azureblob.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		var serr azblob.StorageError
+		var ok bool
+
+		if serr, ok = err.(azblob.StorageError); !ok {
+			return false, fmt.Errorf("Error in casting to azure error type")
+		}
+		if serr.Response().StatusCode == http.StatusNotFound {
+			return false, nil
+		}
```
marpio: Sorry for being nit-picky but what do you think of this?
```
serr, ok = err.(azblob.StorageError)
if !ok {
    return false, fmt.Errorf("Error in casting to azblob.StorageError type")
}
if serr.Response().StatusCode == http.StatusNotFound {
    return false, nil
}
```

scriptonist: I'm sorry I didn't get you, please feel free to help me understand the issue. I don't consider myself a go ninja, so any help to improve the code is really appreciated appreciated 🙏

pkg/storage/azureblob/azureblob.go
```diff
@@ -0,0 +1,151 @@
...
+func (c *azureBlobStoreClient) BlobExists(ctx context.Context, path string) (bool, error) {
+	// TODO: Any better way of doing this ?
+	blobURL := c.containerURL.NewBlockBlobURL(path)
+	_, err := blobURL.GetProperties(ctx, azblob.BlobAccessConditions{})
+	if err != nil {
+		var serr azblob.StorageError
+		var ok bool
+
+		if serr, ok = err.(azblob.StorageError); !ok {
+			return false, fmt.Errorf("Error in casting to azure error type")
```
marpio: Can you use `errors.E(op, err)` ?

pkg/storage/azureblob/azureblob.go
```diff
@@ -0,0 +1,151 @@
...
+			blobs = append(blobs, blob.Name)
+		}
+	}
+	return blobs, nil
+}
+func (c *azureBlobStoreClient) DeleteBlob(ctx context.Context, path string) error {
+	blobURL := c.containerURL.NewBlockBlobURL(path)
+	_, err := blobURL.Delete(ctx, azblob.DeleteSnapshotsOptionNone, azblob.BlobAccessConditions{})
+	if err != nil {
+		return err
```
marpio: Can you use errors.E(op, err) ?

pkg/storage/azureblob/azureblob.go
```diff
@@ -0,0 +1,151 @@
...
+	return true, nil
+
+}
+
+// ReadBlob returns an io.ReadCloser for the contents of a blob
+func (c *azureBlobStoreClient) ReadBlob(ctx context.Context, path string) (io.ReadCloser, error) {
+	blobURL := c.containerURL.NewBlockBlobURL(path)
+	downloadResponse, err := blobURL.Download(ctx, 0, 0, azblob.BlobAccessConditions{}, false)
+	if err != nil {
+		return nil, err
```
marpio: errors.E(op, err)

pkg/storage/azureblob/azureblob.go
```diff
@@ -0,0 +1,151 @@
...
+		var ok bool
+
+		if serr, ok = err.(azblob.StorageError); !ok {
+			return false, fmt.Errorf("Error in casting to azure error type")
+		}
+		if serr.Response().StatusCode == http.StatusNotFound {
+			return false, nil
+		}
+
+		return false, err
```
marpio: errors.E(op, err)

pkg/storage/azureblob/azureblob.go
marpio: It's not really an issue, your code will work, just thought it's more readable/less code that way.

pkg/storage/azureblob/azureblob.go
```diff
@@ -0,0 +1,163 @@
...
+			ContentType: contentType,
+		}
+	}
+	bufferSize := 1 * 1024 * 1024 // Size of the rotating buffers that are used when uploading
+	maxBuffers := 3               // Number of rotating buffers that are used when uploading
+
+	uploadStreamOpts := azblob.UploadStreamToBlockBlobOptions{
+		BufferSize:       bufferSize,
+		MaxBuffers:       maxBuffers,
+		BlobHTTPHeaders:  httpHeaders(contentType),
```
t-tomalak: Why we use func here? could we just pass `azblob.BlobHTTPHeaders{ContentType: contentType}`?

pkg/storage/azureblob/azureblob.go
```diff
@@ -0,0 +1,163 @@
...
+		Metadata:         emptyMeta,
```
t-tomalak: If we don't set any value in them, this properties ( `Metadata`, `AccessConditions` ) can be omitted because they will be set to their coresponding `Nil Values`.

pkg/storage/azureblob/azureblob.go
```diff
@@ -0,0 +1,155 @@
...
+	GetTimeout() time.Duration
+}
+
+type azureBlobStoreClient struct {
+	containerURL *azblob.ContainerURL
+	timeout      time.Duration
+}
+
+func newBlobStoreClient(accountURL *url.URL, accountName, accountKey, containerName string, timeout time.Duration) (*azureBlobStoreClient, error) {
+	const op errors.Op = "azure.newBlobStoreClient"
```
michalpristas: s/azure/azureblob

michalpristas: same for every op here

pkg/storage/azureblob/deleter.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// Delete implements the (./pkg/storage).Deleter interface and
+// removes a version of a module from storage. Returning ErrNotFound
+// if the version does not exist.
+func (s *Storage) Delete(ctx context.Context, module string, vsn string) error {
+	const op errors.Op = "azureblob.Delete"
+	ctx, span := observ.StartSpan(ctx, op.String())
+	defer span.End()
+	return errors.E(op, fmt.Errorf("Not Implemented"), errors.M(module))
```
michalpristas: TODO for me: check once it's there

pkg/storage/azureblob/getter.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Info implements the (./pkg/storage).Getter interface
+func (s *Storage) Info(ctx context.Context, module string, version string) ([]byte, error) {
+	const op errors.Op = "azureblob.Info"
+	ctx, span := observ.StartSpan(ctx, op.String())
+	defer span.End()
+
+	exists, err := s.Exists(ctx, module, version)
+	if err != nil || !exists {
+		return nil, errors.E(op, errors.M(module), errors.V(version), errors.KindNotFound)
```
michalpristas: we should not swallow errors, errors are valuable for investigating issues in prod, you can inject it like this
`return nil, errors.E(op, errors.M(module), errors.V(version), errors.KindNotFound, err)`

michalpristas: same for everywhere

pkg/storage/azureblob/getter.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	infoReader, err := s.client.ReadBlob(ctx, config.PackageVersionedName(module, version, "info"))
+	if err != nil {
+		return nil, errors.E(op, err, errors.M(module), errors.V(version))
+	}
+	defer infoReader.Close()
```
michalpristas: please don't use defer for close, we are ignoring errors returned by Close() this way.
close when done with reading and if err, then return err

pkg/storage/azureblob/getter.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	if !exists {
+		return nil, errors.E(op, errors.M(module), errors.V(version), errors.KindNotFound)
+	}
+
+	modReader, err := s.client.ReadBlob(ctx, config.PackageVersionedName(module, version, "mod"))
+	if err != nil {
+		return nil, errors.E(op, err, errors.M(module), errors.V(version))
+	}
+	defer modReader.Close()
```
michalpristas: same with defer

pkg/storage/azureblob/saver.go
```diff
@@ -0,0 +1,23 @@
...
+func (s *Storage) Save(ctx context.Context, module, version string, mod []byte, zip io.Reader, info []byte) error {
+	const op errors.Op = "azureblob.Save"
+	ctx, span := observ.StartSpan(ctx, op.String())
+	defer span.End()
+	err := moduploader.Upload(ctx, module, version, bytes.NewReader(info), bytes.NewReader(mod), zip, s.client.UploadWithContext, s.client.GetTimeout())
+	if err != nil {
+		return errors.E(op, err, errors.M(module), errors.V(version))
+	}
+	return nil
+}
```
michalpristas: please newline

pkg/storage/azureblob/azureblob.go
scriptonist: I'm sorry, didn't get you 🤔

scriptonist: Should I change the op to `storage.azure.azureblob`

pkg/storage/azureblob/azureblob.go
```diff
@@ -0,0 +1,155 @@
...
+	return cl, nil
+}
+
+// Storage implements (github.com/gomods/athens/pkg/storage).Saver and
+// also provides a function to fetch the location of a module
+type Storage struct {
+	client client
+}
+
+// New creates a new azure blobs storage saver
```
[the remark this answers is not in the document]
scriptonist: I was referring [this](https://github.com/gomods/athens/blob/987291ae000acc56a1f76dabc9e6d878bbf583f0/pkg/storage/s3/s3.go#L34) and copied it blindly, will fix it right away

pkg/storage/azureblob/azureblob.go
michalpristas: sorry my bad
it shoud be `const op errors.Op = "azureblob.newBlobStoreClient"`
same for every errors.Op constant in this file. other files are fine.
the shortcut `s/azure/azureblob` used in the comment means
`s`ubstitue `azure` with `azureblob`

t-tomalak: @scriptonist @michalpristas mean to change it from `azure.newBlobStoreClient` to `azureblob.newBlobStoreClient` and for other change `azure.*` => `azureblob.*`.

scriptonist: Oh Got it now, wasn't accustomed to seeing vim syntax in comments! , hence the confusion.

michalpristas: no worries ;-)

pkg/storage/azureblob/checker.go
```diff
@@ -0,0 +1,20 @@
...
+
+	"github.com/gomods/athens/pkg/config"
+	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/observ"
+)
+
+// Exists implements the (./pkg/storage).Checker interface
+// returning true if the module at version exists in storage
+func (s *Storage) Exists(ctx context.Context, module string, version string) (bool, error) {
+	const op errors.Op = "azureblob.newBlobStoreClient.Save"
```
michalpristas: i think we misunderstood again
the convention is `package.method` so in this case it was correct `azureblob.Save`

pkg/storage/azureblob/azureblob.go
```diff
@@ -0,0 +1,155 @@
...
+	GetTimeout() time.Duration
+}
+
+type azureBlobStoreClient struct {
+	containerURL *azblob.ContainerURL
+	timeout      time.Duration
+}
+
+func newBlobStoreClient(accountURL *url.URL, accountName, accountKey, containerName string, timeout time.Duration) (*azureBlobStoreClient, error) {
+	const op errors.Op = "storage.azure.azureblob"
```
michalpristas: and ops constants in this file should be as well `azureblob.{methodName}` 
in this case `azureblob.newBlobStoreClient` 
for List method
`azureblob.List`

pkg/storage/azureblob/getter.go
t-tomalak: @michalpristas but what if `ioutil.ReadAll` fail? We probably will not close `infoReader` in that case. As far as I know it is safe to `defer Close` and ignore errors if we didn't write to that resource. Second solution we could leave that `defer infoReader.Close()` but also add second close `if err := infoReader.Close(); err != nil { return nil, errors.E(op, err, errors.M(module), errors.V(version))}` after `ioutil.ReadAll` as it should be safe to close it multiple times

michalpristas: you're right that on non writable it is ok, but it would be good to know what happened.
if read/write fails you can call xx.Close() inside error handling block and return first error that occurred.
considering this is readable maybe we can just defer close with log message in case err happens

t-tomalak: I think we need to improve that case in every storage, so I would leave here as it is now and fix that for other storages at once

michalpristas: i'm fine with separate PR
will create a task later

pkg/storage/azureblob/checker.go
scriptonist: Sorry my bad.

pkg/storage/azureblob/azureblob.go
scriptonist: This was a dumb mistake 😦 , hope I don't do any new blunders 😅

pkg/storage/azureblob/getter.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+// Info implements the (./pkg/storage).Getter interface
+func (s *Storage) Info(ctx context.Context, module string, version string) ([]byte, error) {
+	const op errors.Op = "azureblob.Info"
+	ctx, span := observ.StartSpan(ctx, op.String())
+	defer span.End()
+
+	exists, err := s.Exists(ctx, module, version)
+	if err != nil || !exists {
```
t-tomalak: Could you split this two error handling  to look like in other method (`GoMod` and `Zip`)?

pkg/storage/azureblob/azureblob.go
```diff
@@ -0,0 +1,155 @@
...
+			return nil, errors.E(op, err)
+		}
+		marker = listBlob.NextMarker
+
+		for _, blob := range listBlob.Segment.BlobItems {
+			blobs = append(blobs, blob.Name)
+		}
+	}
+	return blobs, nil
+}
```
t-tomalak: Could you add new line here?

scriptonist: Sure, but I was wondering why we are doing it, can you help me out?

scriptonist:
Thanks for the review @steakunderscore @marpio @michalpristas @t-tomalak @marwan-at-work. I've gone ahead and removed the WIP status, please review it and let me know what more can be done. Thanks 🙏

pkg/storage/azureblob/azureblob.go
t-tomalak: Just code formating, nothing major

michalpristas:
will take a look throughout the day. a lot of meetings today but i will make some time for you ;-)

config.dev.toml
marpio: @scriptonist please remove the Timeout - there is only one global timeout now.

pkg/storage/azureblob/azureblob.go
```diff
@@ -0,0 +1,156 @@
...
+	const op errors.Op = "azure.ReadBlob"
+	blobURL := c.containerURL.NewBlockBlobURL(path)
+	downloadResponse, err := blobURL.Download(ctx, 0, 0, azblob.BlobAccessConditions{}, false)
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+	return downloadResponse.Body(azblob.RetryReaderOptions{}), nil
+}
+
+func (c *azureBlobStoreClient) ListBlobs(ctx context.Context, prefix string) ([]string, error) {
```
marpio: Add comment please

pkg/storage/azureblob/azureblob.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		marker = listBlob.NextMarker
+
+		for _, blob := range listBlob.Segment.BlobItems {
+			blobs = append(blobs, blob.Name)
+		}
+	}
+	return blobs, nil
+}
+func (c *azureBlobStoreClient) DeleteBlob(ctx context.Context, path string) error {
```
marpio: Add comment please

config.dev.toml
t-tomalak: bump

pkg/storage/azureblob/azureblob.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// New creates a new azure blobs storage
+func New(conf *config.AzureBlobConfig, timeout time.Duration) (*Storage, error) {
+	const op errors.Op = "azure.New"
+	u, err := url.Parse(fmt.Sprintf("https://%s.blob.core.windows.net", conf.AccountName))
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+	cl, err := newBlobStoreClient(u, conf.AccountName, conf.AccountKey, conf.ContainerName, timeout)
```
marpio: Do we need the timeout in client if we set it on storage (L 66)? We shouldn't need it in two places.

pkg/storage/azureblob/saver.go
```diff
@@ -0,0 +1,25 @@
...
+	"bytes"
+	"context"
+	"io"
+
+	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/observ"
+	moduploader "github.com/gomods/athens/pkg/storage/module"
+)
+
+// Save implements the (github.com/gomods/athens/pkg/storage).Saver interface.
```
michalpristas: you can shorten to ./pkg/storage as you use in other cases

pkg/storage/azureblob/azureblob.go
scriptonist: Oh okay, cool 😃 Fixing it now 🔧

pkg/storage/azureblob/getter.go
marpio: err will be nil in that case so no need to do that

michalpristas: it enters the block on err != nil or !exists so it may or may not be nil

marpio: @michalpristas I think the "Conversation" Tab shows different code that the "Files Changed" Tab b/c here i see
```
exists, err := s.Exists(ctx, module, version)
if err != nil {
	return nil, errors.E(op, err, errors.M(module), errors.V(version))
}
if !exists {
    return nil, errors.E(op, errors.M(module), errors.V(version), errors.KindNotFound)
}
```

marpio:
@michalpristas I will merge this PR since GH is having troubles due to it's size. We can fix some minor issues in a follow up.

PR #918, opened 2018-11-15
extend bug_report tmpl by storage inf

marpio:

We could use some additional info about the storage used with the proxy while investigating bugs.

**How is the fix applied?**

Added Storage to the Environment description

No issue for that.


manugupt1:
thanks @marpio for this!

PR #919, opened 2018-11-15
Add error checks in a few test files and fix misspelled words

cixtor:

A few error messages returned by some functions used in the unit-tests for the config package are being ignored. It may be more useful to handle these errors to make failures more obvious in the CI. Additionally, I’ve found some misspelled words in the documentation.

**How is the fix applied?**

Errors triggered by both `ioutil.TempFile()` and `os.Chmod()` will be reported using `t.Fatalf()`.


pkg/download/protocol.go
```diff
@@ -89,7 +89,7 @@ func (p *protocol) List(ctx context.Context, mod string) ([]string, error) {
 
 func (p *protocol) Latest(ctx context.Context, mod string) (*storage.RevInfo, error) {
 	const op errors.Op = "protocol.Latest"
-	ctx, span := observ.StartSpan(ctx, op.String())
+	_, span := observ.StartSpan(ctx, op.String())
```
manugupt1: I am not sure why this change is required. It decreases readbility and also introduces the possibility of errors when a new function taking a function is called.

pkg/middleware/middleware_test.go
```diff
@@ -75,7 +75,7 @@ func Test_FilterMiddleware(t *testing.T) {
 	r.Error(err, "Expected error when a file not existing on the filesystem is given")
 
 	app, err = middlewareFilterApp(filter.Name(), conf.GlobalEndpoint)
-	r.NoError(err, "app should be succesfully created in the test")
+	r.NoError(err, "app should be successfully created in the test")
```
manugupt1: This is fine!

pkg/storage/gcp/gcp_test.go
```diff
@@ -30,6 +30,8 @@ func (g *GcpTests) TestSaveGetListExistsRoundTrip() {
 		r.Equal(info, infoBts)
 
 		ziprc, err := g.store.Zip(ctx, g.module, g.version)
+		r.NoError(err)
```
manugupt1: This is fine.

pkg/download/protocol.go
marwan-at-work: Agree with manu, we need to shadow ctx here because it's super easy to miss this little detail and have a headache if we pass `ctx` to a new function and the Traces are not being recorded.

michalpristas: 👍 on this one. I think we should keep it. price of not having a correct relationship between spans is higher than this.

cixtor:
Thank you for the feedback.

I’ve followed your suggestions and reverted the removal of the “cxt” variable to keep the code readable. The purpose of the pull-request is now to add additional error checks in some unit tests, and to fix a few misspelled words in the documentation.

I’ve been suggested to change `t.FailNow()` to `r.NoError()` but this method is only available after importing “testify” and I would prefer to keep the dependency list as small as possible, unless more people think it’s a good idea to add it.

Alternatively, I could use `t.Errorf()` instead, @robjloranger what do you think?

```golang
f1, err := ioutil.TempFile(os.TempDir(), "prefix-")
if err != nil {
    t.Errorf("Cannot create temp file: %s", err)
}

if err = os.Chmod(f1.Name(), 0777); err != nil {
    t.Errorf("Cannot chmod temp file: %s", err)
}
```

robjloranger:
@cixtor ya sorry about that. I forgot we don't use testify everywhere. `t.Errorf` should do great. Just so we can quickly see from CI logs what failed the test, or running them locally.

cixtor:
I’ve modified the test to report a more information using `t.Errorf()` as suggested.

manugupt1:
Hi @cixtor, can you resolve conflicts so that Marwan and others can give a final look and merge it in. Thanks for being patient.

cixtor:
The conflicts have been resolved, @marwan-at-work feel free to review the changes again.

pkg/config/config_test.go
```diff
@@ -306,17 +306,21 @@ func Test_checkFilePerms(t *testing.T) {
 	}
 	f1, err := ioutil.TempFile(os.TempDir(), "prefix-")
 	if err != nil {
-		t.FailNow()
+		t.Errorf("Cannot create 1st temp file: %s", err)
```
marwan-at-work: `t.Errorf` does not stop the execution of the test. And if `f1` is nil, then we'd get a panic. We should use `t.Fatalf` no?

pkg/config/config_test.go
```diff
@@ -306,17 +306,21 @@ func Test_checkFilePerms(t *testing.T) {
...
 	}
 	defer os.Remove(f1.Name())
-	err = os.Chmod(f1.Name(), 0777)
+	if err = os.Chmod(f1.Name(), 0777); err != nil {
+		t.Errorf("Cannot chmod 1st temp file: %s", err)
```
marwan-at-work: ditto

pkg/config/config_test.go
```diff
@@ -306,17 +306,21 @@ func Test_checkFilePerms(t *testing.T) {
...
+	}
 
 	f2, err := ioutil.TempFile(os.TempDir(), "prefix-")
 	if err != nil {
-		t.FailNow()
+		t.Errorf("Cannot create 2nd temp file: %s", err)
```
marwan-at-work: ditto

pkg/config/config_test.go
```diff
@@ -306,17 +306,21 @@ func Test_checkFilePerms(t *testing.T) {
...
 	}
 	defer os.Remove(f2.Name())
-	err = os.Chmod(f2.Name(), 0600)
+	if err = os.Chmod(f2.Name(), 0600); err != nil {
+		t.Errorf("Cannot chmod 2nd temp file: %s", err)
```
marwan-at-work: ditto

cixtor:
@marwan-at-work I’ve changed `t.Errorf()` to `t.Fatalf()` as you suggested.

michalpristas:
@marwan-at-work i believe comments were addressed

marwan-at-work:
@cixtor do you mind fixing the conflict and will marge after.

PR #920, opened 2018-11-15
pass custom timestamp format to the log formatter

### LINKED ISSUE #880: TestCloudLogger test fails
**Describe the bug**
TestCloudLogger fails... sometimes.
https://dev.azure.com/gomods/athens/_build/results?buildId=12&view=logs

Not sure but I suppose it happens because the call to the log func
https://github.com/gomods/athens/blob/bc3563674920a7fa664f04e3287e943f65041860/pkg/log/log_test.go#L81
and the creation of the expected variable
https://github.com/gomods/athens/blob/bc3563674920a7fa664f04e3287e943f65041860/pkg/log/log_test.go#L85
happens a a different time.
Usually this isn't a problem because resolution is 1 sec but this time the first call was at `2018-11-07T21:50:59Z` and the second at `2018-11-07T21:51:00Z`

**Error Message**
https://dev.azure.com/gomods/athens/_build/results?buildId=12&view=logs

 - OS: Win Server 2016

marpio:

Logger tests fail sometimes on azure-pipelines build

**How is the fix applied?**

Added new timestampFormat param to the Logger constructor so it's possible to override the default. Could be also helpfull if we want i.e. nonoseconds resolution instead of the default one (second)

Fixes #880


marpio:
@marwan-at-work Thanks for pointing this out.
Just reading the comment:

> TODO: take cloud arg and env
> // to construct the correct JSON formatter.

I wasn't aware the time format is cloud dependent. If that's the case then you're absolutely right - we shouldn't let that override.
I will go with the first option since it was my initial solution and I also think this should be enough.I've added the rest because I thought using the default logrus format `RFC3339` can be limiting. If it does not help we can try something else.

manugupt1:
this awesome! thanks @marpio I will merge it in once CI passes.

Issue #925, opened 2018-11-15
Documentation is missing for the validation path referenced in the dev config


Xe:
I am trying to evaluate Athens for a project, I noticed that there is a [validation hook](
https://github.com/gomods/athens/blob/master/config.dev.toml#L84) in the default configuration but there is no details on what this is, means or how it is used.

My thought is that by using this, I can create a system that allows me to somehow identify the source of who is requesting a module (for example: a private module attempted to be downloaded by an attacker who does not normally have access) and prevent them from accessing things they should not be allowed to access.

PR #927, opened 2018-11-15
pkg/middleware: seamlessly passthru client Authorization header if present

Xe:
This is related to #925 and solves a usecase I would like to support.

I have this pass through the `Authorization` header contents to the validation function, and if it is more than an empty string, I set the value of that header blindly without caring about the contents of it. 

How do you want me to write a test for this?

Related to #925 

Marwan's edit: Fixes https://github.com/gomods/athens/issues/929


Xe:
Testing steps:

- I created [the following configuration file](https://gist.github.com/ce8910cf28e99f5f0323d48293627c6b)
- I loaded it into athens `$ ./proxy -config_file ./config.toml`
- I made sure the athens instance was accessible to the testing container via standard HTTP+TCP
- I started a testing docker container: `$ docker run --rm -ite GOPROXY=http://AzureDiamond:hunter2@machine.domain.tld:3000 xena/go:1.11.1` (though any docker image with the go compiler toolchain should be fine)
- In the testing container, I ran the following command: `cd ~ && mkdir app && cd app && go mod init domain.tld/projects/app && go get github.com/Xe/ln`

Results: 

- The postbin I used was able to show the `Authorization` header being passed through (see image 1)

Image 1: 
screen shot 2018-11-15 at 1 57 14 pm

marpio:
@Xe I think this will work only with Basic Authentication where you put the credentials into the URL. Otherwise Go wont send any headers to the proxy.
See: https://github.com/golang/go/issues/27132
Correct me if i'm wrong @marwan-at-work

marwan-at-work:
@marpio you're right. Basic Auth will work, which is something a user might not want...it's like you're sending someone's password to someone else :) -- Maybe we can have a separate auth header configuration for the Hook?

marwan-at-work:
After some chat, we created this issue https://github.com/gomods/athens/issues/929 -- so this PR should put this feature behind a flag, something like `PassAuthToValidation = false`

pkg/middleware/validation.go
```diff
@@ -60,7 +60,17 @@ func validate(hook, mod, ver string) (bool, error) {
...
+	if err != nil {
+		return false, errors.E(op, err)
+	}
+
+	req.Header.Add("Content-Type", "application/json")
+	if auth != "" {
+		req.Header.Add("Authorization", auth)
+	}
+
+	resp, err := http.DefaultClient.Do(req)
```
marpio: Could you use a custom http client here instead so that we can pass a timeout?

marwan-at-work: as another, more high level, option: you can do `req = req.WIthContext(ctx)` and have a timeout on the context. 

Either one is fine imo

arschles: Would it be ok with folks if we merge this and add the timeout (fwiw I like the `req.WithContext` approach better) in a follow-up?

marwan-at-work:
@arschles see my last comment. https://github.com/gomods/athens/pull/927#issuecomment-439444692

Xe:
I'm sorry, but I no longer need this functionality. If someone else wants to take this over they are free to.

marwan-at-work:
@Xe no problem at all :) thank you for all of your involvement!

Issue #760, opened 2018-10-08
Implement Distributed Deduplication Mechanisms


arschles:
**Is your feature request related to a problem? Please describe.**

#308 was for making Athens not do two `go get` calls for the same `module@version` at the same time. That meant that if N `go get mod@ver` calls came in at the same time, only one `go get` would get issues on the backend. https://github.com/gomods/athens/pull/573 went and fixed #308 by introducing "single flight" - a kind of mutual exclusion mechanism built in memory.

Now, we have an Athens that will only run `go get` once for a given `mod@ver`, but if you want to run a fleet of Athens servers (i.e. behind a load balancer in the cloud, in pods in a Kubernetes cluster, etc...), the different Athens processes don't all single-flight together. If a `go get mod@ver` request goes to athensA and another `go get mod@ver` request goes to athensB at exactly the same time, they will both still run `go get mod@ver`.

**Describe the solution you'd like**

We should build new `download.Wrapper`s (like the one [here](https://github.com/gomods/athens/pull/573/files#diff-8176b6031aa0a1b5e293d0ce973a0cdfR40)) that use distributed mutual exclusion mechanisms, so that a fleet of Athens servers can single-flight together.

I'd like us to have multiple drivers eventually. Some ideas:

- [etcd](https://coreos.com/etcd/) locks
- [Consul](https://www.consul.io/) locks
- Cloud proprietary solutions

**Describe alternatives you've considered**

I tried to think of other ways to de-duplicate `go get`s across lots of Athens servers, but can't think of others besides distributed mutual exclusion.

**Additional context**

marwan-at-work:
I'd like to work on this (I've done in implementation for it back in the day so I just need to dig it up). 

I'd like to emphasize that this should only be worked on, reviewed, and merged after beta. So that we focus our efforts on achieving the beta milestone :v:

PR #933, opened 2018-11-18
WIP: Distributed single flight

marwan-at-work:
First pass at a distributed single flight implementation for peeps to look at. 

Missing: docs. 

Note: keys are cleaned up by etcd itself after 15 minutes...This is because there's no way to tell which client is the last one that is responsible for cleaning up the key. This can be done through a distributed WaitGroup implementation, but I don't want to overly complex something that's already complex as there's no harm in letting the key garbage collect itself after an arbitrary long enough time

TODO: a lot of corner cases for errors, so I should add much more tests than just one. 

Fixes https://github.com/gomods/athens/issues/760


marwan-at-work:
lol wow github...way to force-push shame me.

pkg/stash/with_etcd.go
```diff
@@ -0,0 +1,78 @@
...
+	s Stasher
+}
+
+func (s *etcd) Stash(ctx context.Context, mod, ver string) error {
+	const op errors.Op = "etcd.Stash"
+	ctx, span := observ.StartSpan(ctx, op.String())
+	defer span.End()
+
+	mv := config.FmtModVer(mod, ver)
+	sesh, err := concurrency.NewSession(s.c)
```
michalpristas: first time i see sesh for session :-)

pkg/stash/with_etcd.go
```diff
@@ -0,0 +1,95 @@
...
+	resp, err := s.c.Get(ctx, mv)
+	if err != nil {
+		return errors.E(op, err)
+	}
+	if resp.Count == 0 {
+		lease, err := s.c.Grant(ctx, 15*60)
+		if err != nil {
+			return errors.E(op, err)
+		}
+		err = s.s.Stash(ctx, mod, ver)
```
marpio: What if this fails? Does that mean that every subsequent request for this particular mod@ver will fail until the lease expires and the key is cleared?

pkg/stash/with_etcd.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return errors.E(op, err)
+	}
+	defer mu.Unlock(ctx)
+	resp, err := s.c.Get(ctx, mv)
+	if err != nil {
+		return errors.E(op, err)
+	}
+	if resp.Count == 0 {
+		lease, err := s.c.Grant(ctx, 15*60)
```
michalpristas: can we make this configurable?

pkg/stash/with_etcd.go
michalpristas: yes and they will fail fast :)

marwan-at-work: @marpio @michalpristas very valid point. I figured that's okay...but maybe not? If not, then I can do the Distributed WaitGroup I mentioned above and this issue would be gone, but it does bring complexity higher. We can also potentially have a `WithRetries` stasher that retries fetching from upstream only when unexpected errors happen (not when "not found" errors happen)

pkg/stash/with_etcd.go
marwan-at-work: @michalpristas i think our config file is already hard to understand for users. A lot of internal details they might not care about. In this case, I'd rather do Distributed WaitGroup and put the complexity on "us" rather than the "user".

config.dev.toml
```diff
@@ -126,9 +126,17 @@ TraceExporterURL = ""
 
 # TraceExporter is the service to which the data collected by OpenCensus can be exported to.
 # Possible values are: jaeger, datadog, and stackdriver.
-# Env overide: ATHENS_TRACE_EXPORTER
+# Env overirde: ATHENS_TRACE_EXPORTER
```
chriscoffee: s/overirde/override

docker-compose.yml
```diff
@@ -33,3 +33,74 @@ services:
       - 6832:6832/udp
       - 5778:5778
       - 16686:16686
+  etcd0:
+    image: quay.io/coreos/etcd
+    ports:
+      - "2379:2379"
+    volumes:
+      - etcd0:/etcd_data
+    command:
```
chriscoffee: Can you put this over a few lines i.e:
```
command: >
  /usr/local/bin/etcd etcd0
  --data-dir /etcd_data
  -advertise-client-urls http://etcd0:2379
  -listen-client-urls http://0.0.0.0:2379
  -initial-advertise-peer-urls http://etcd0:2380
  -listen-peer-urls http://0.0.0.0:2380
  -initial-cluster etcd0=http://etcd0:2380,etcd1=http://etcd1:2380,etcd2=http://etcd2:2380

File: docker-compose.yml [hunk 6]
```diff
@@ -33,3 +33,74 @@ services:
...
+      - http://0.0.0.0:2380
+      - -initial-cluster
+      - etcd0=http://etcd0:2380,etcd1=http://etcd1:2380,etcd2=http://etcd2:2380
+  etcd1:
+    image: quay.io/coreos/etcd
+    ports:
+      - "22379:2379"
+    volumes:
+      - etcd1:/etcd_data
+    command:
```
chriscoffee: As above:
```
command: >
  ...
```

cmd/proxy/actions/app_proxy.go
```diff
@@ -2,22 +2,22 @@ package actions
...
 	"github.com/gobuffalo/buffalo"
+	"github.com/gomods/athens/pkg/config"
 	"github.com/gomods/athens/pkg/download"
 	"github.com/gomods/athens/pkg/download/addons"
 	"github.com/gomods/athens/pkg/log"
 	"github.com/gomods/athens/pkg/module"
 	"github.com/gomods/athens/pkg/stash"
 	"github.com/gomods/athens/pkg/storage"
 	"github.com/spf13/afero"
+	"strings"
```
chriscoffee: slightly nit picky but do we split import groups in athens i.e:
```
import (
       "strings"

	"github.com/gobuffalo/buffalo"
	"github.com/gobuffalo/buffalo"
	"github.com/gomods/athens/pkg/download"
	"github.com/gomods/athens/pkg/download"
	"github.com/gomods/athens/pkg/download/addons"
	"github.com/gomods/athens/pkg/download/addons"
	"github.com/gomods/athens/pkg/log"
	"github.com/gomods/athens/pkg/log"
...
```

docker-compose.yml
```diff
@@ -33,3 +33,74 @@ services:
       - 6832:6832/udp
       - 5778:5778
       - 16686:16686
+  etcd0:
+    image: quay.io/coreos/etcd
+    ports:
+      - "2379:2379"
+    volumes:
+      - etcd0:/etcd_data
```
t-tomalak: Do we need mounted volume for etcd? I think it can slow down mac/windows user a bit.

cmd/proxy/actions/app_proxy.go
```diff
@@ -46,20 +46,28 @@ func addProxyRoutes(
...
+	mf, err := module.NewGoGetFetcher(c.GoBinary, fs)
 	if err != nil {
 		return err
 	}
 
-	lister := download.NewVCSLister(goBin, fs)
-	st := stash.New(mf, s, stash.WithPool(goGetWorkers), stash.WithSingleflight)
+	lister := download.NewVCSLister(c.GoBinary, fs)
+	sf := stash.WithSingleflight
+	if c.EtcdEndpoints != "" {
```
t-tomalak: Probably we  should think of some configuration variable to select `stasher` (probably more of them to come). Something like we do with `storages` . It probably could be done in follow up PR.

pkg/stash/with_etcd.go
```diff
@@ -0,0 +1,95 @@
...
+	err = eg.Wait()
+	if err != nil {
+		return nil, errors.E(op, err)
+	}
+	return func(s Stasher) Stasher {
+		return &etcd{c, s}
+	}, nil
+}
+
+type etcd struct {
```
t-tomalak: Could we use more verbose names for struct variables? In this case something like 
`type etcd struct {	
        client *clientv3.Client
	stasher Stasher
}`

pkg/stash/with_etcd.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+		err = s.s.Stash(ctx, mod, ver)
+		stashResponse := "success"
+		if err != nil {
+			stashResponse := err.Error()
+			s.c.Put(ctx, mv, stashResponse, clientv3.WithLease(lease.ID)) // TODO: handle err
+			return err
+		}
+
+		_, err = s.c.Put(ctx, mv, stashResponse, clientv3.WithLease(lease.ID))
```
t-tomalak: Could it be writen in shortened form 
`if _, err := s.c.Put(ctx, mv, "success", clientv3.WithLease(lease.ID)); err != nil {...`? 
Also I think that `stashResponse` variable seems unnecessary

pkg/stash/with_etcd.go
```diff
@@ -0,0 +1,95 @@
...
+		lease, err := s.c.Grant(ctx, 15*60)
+		if err != nil {
+			return errors.E(op, err)
+		}
+		err = s.s.Stash(ctx, mod, ver)
+		stashResponse := "success"
+		if err != nil {
+			stashResponse := err.Error()
+			s.c.Put(ctx, mv, stashResponse, clientv3.WithLease(lease.ID)) // TODO: handle err
+			return err
```
t-tomalak: Should it be `errors.E(op,err)`?

pkg/stash/with_etcd.go
marpio: I would really :heart: to see the the distributed WaitGroup implemented but imho the `WithRetries` stasher would suffice. Could you outline any others advantages of the WaitGroup, beside this issue and being able to clear the keys more efficiently?

pkg/stash/with_etcd.go
```diff
@@ -0,0 +1,95 @@
...
+	sesh, err := concurrency.NewSession(s.c)
+	if err != nil {
+		return errors.E(op, err)
+	}
+	mu := concurrency.NewMutex(sesh, mv)
+	err = mu.Lock(ctx)
+	if err != nil {
+		return errors.E(op, err)
+	}
+	defer mu.Unlock(ctx)
```
manugupt1: Can  we handle mu.Unlock error here?
Ref: https://godoc.org/github.com/coreos/etcd/clientv3/concurrency#Mutex.Unlock

pkg/stash/with_etcd.go
```diff
@@ -0,0 +1,95 @@
...
+	if resp.Count == 0 {
+		lease, err := s.c.Grant(ctx, 15*60)
+		if err != nil {
+			return errors.E(op, err)
+		}
+		err = s.s.Stash(ctx, mod, ver)
+		stashResponse := "success"
+		if err != nil {
+			stashResponse := err.Error()
+			s.c.Put(ctx, mv, stashResponse, clientv3.WithLease(lease.ID)) // TODO: handle err
```
manugupt1: Where do you get the lease.ID from?

pkg/stash/with_etcd.go
michalpristas: so maybe we need a separate way how to treat these.
I'm usually strongly against hardcoded values, of course, everybody can tweak and build but if you found out in prod that for example 10 is enough you need to tweak, build, potentially sign, deploy + all the fancy approval needed in the corporate environment instead of enabling one simple change using config

pkg/stash/with_etcd.go
marpio: Line 71

manugupt1: oops!

pkg/stash/with_etcd.go
arschles: +1 on ❤️ distributed WaitGroup, but this looks like an awesome start to me. Are peeps ok with doing that as a follow up?

marwan-at-work: Hey peeps, so I've rethought how to do this and potentially found a much simpler way...it's sorta cheating but I think in a good way. I've updated the code so feel free to do another round of review. But in a nutshell here's what happens: 

1. Request comes in 
2. Issue a lock based on mod/ver
3. The lock then checks if the mod/ver exists through `storage.Checker` 
4. If it exists, then "someone before me already did this", we're good. 
5. If it doesn't exist, call the underlying stasher, and be the first to "stash" this. 

The only catch here, is if a previous stasher error'd out, we're only gonna return that error back to the user and not spread it around...This wouldn't qualify as singleflight anymore, but I think it's actually more a feature than a bug because we get free retries. 

In other words: 

1. User A tries, and gets an error. 
2. User B tries 1 second after, but will get a retry instead of User A's error.

cmd/proxy/actions/app_proxy.go
marwan-at-work: @t-tomalak we should think more about this. The less the user has to configure the better.

docker-compose.yml
marwan-at-work: @t-tomalak I don't think so (for now). Because we're using etcd not as a storage, but just during a runtime operation of stashing a module

docker-compose.yml
marwan-at-work: @chriscoffee what's the difference from doing `command: >` and doing it over an array of arguments?

pkg/config/config.go
```diff
@@ -35,6 +35,7 @@ type Config struct {
 	NETRCPath        string `envconfig:"ATHENS_NETRC_PATH"`
 	GithubToken      string `envconfig:"ATHENS_GITHUB_TOKEN"`
 	HGRCPath         string `envconfig:"ATHENS_HGRC_PATH"`
+	EtcdEndpoints    string `envconfig:"ATHENS_ETCD_ENDPOINTS"`
```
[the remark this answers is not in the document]
marwan-at-work: @arschles that'd be pretty cool, but I doubt it.

pkg/stash/with_etcd.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/observ"
+	"go.etcd.io/etcd/clientv3"
+	"go.etcd.io/etcd/clientv3/concurrency"
+	"golang.org/x/sync/errgroup"
+)
+
+// WithEtcd returns a distributed singleflight
+// using an etcd cluster. If it cannot connect,
+// then it will panic
```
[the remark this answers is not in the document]
marwan-at-work: @arschles good point, i've changed this to return an error but didn't update the comment

pkg/stash/with_etcd.go
marwan-at-work: done

pkg/stash/with_etcd.go
```diff
@@ -0,0 +1,85 @@
...
+	mu := concurrency.NewMutex(sesh, mv)
+	err = mu.Lock(ctx)
+	if err != nil {
+		return errors.E(op, err)
+	}
+	defer func() {
+		const op errors.Op = "etcd.Unlock"
+		lockErr := mu.Unlock(ctx)
+		if err == nil && lockErr != nil {
+			err = errors.E(op, lockErr)
```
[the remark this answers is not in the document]
marwan-at-work: @marpio this is inside a defer, so err would be returned back to the caller, and so it should be handled properly (afaik). I will make sure a test is added for this.

marwan-at-work:
@marpio no, we only lock based on mod@ver, so we can save multiple modules simultaneously -- but yes, it definitely got simpler :)

marpio:
@marwan-at-work 
Missed the mod@ver part.
Nice!

pkg/stash/with_etcd.go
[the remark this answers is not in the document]
marpio: One more question @marwan-at-work - what happens if this actually fails? Does the lock on that mod@ver stay there  forever and has to be resolved manually or does it expire after some time?

cmd/proxy/actions/app_proxy.go
michalpristas: i think @t-tomalak had multiple providers in mind. for now we have etcd but we cant force everybody to use it, similar structure as with storages would be much more valuable

cmd/proxy/actions/app_proxy.go
```diff
@@ -16,15 +18,11 @@ func addProxyRoutes(
 	app *buffalo.App,
 	s storage.Backend,
```
michalpristas: while at it can you rename to more descriptive names?

config.dev.toml
```diff
@@ -144,6 +144,14 @@ TraceExporterURL = "http://localhost:14268"
 # Env overide: ATHENS_STATS_EXPORTER
 StatsExporter = "prometheus"
 
+# EtcdEndoints are comma separated URLs that determine whether to use Etcd as a distributed lock mechanism 
+# for single-flight requests. If empty, in-memory single-flight is used. If you 
+# have a single instance of Athens, then keep this empty. Otherwise, use etcd. 
+# Athens uses a single-flight mechanism to store modules into a storage Backend 
+# so that no two concurrent requests can override each other.
+# Env override: ATHENS_ETCD_ENDPOINTS
+EtcdEndpoints = ""
```
michalpristas: probably related to previous comments about multiple distributed lock impl. 
we should aim for switchable implementation and configuration should reflect that. 
maybe some type at the root and vendor specific config in [EtcdSync] or something similar.
i would just move `EtcdEndpoints` under Etcd config for now and be bothered with multi vendors later

cmd/proxy/actions/app_proxy.go
michalpristas: can you also encapsulate into factory func maybe? as singleflight is in memory sync and etcd is distributed sync we can just call a func which returns `sf,err`

pkg/stash/with_etcd.go
```diff
@@ -0,0 +1,85 @@
...
+// WithEtcd returns a distributed singleflight
+// using an etcd cluster. If it cannot connect,
+// to any of the endpoints, it will return an error.
+func WithEtcd(endpoints []string, checker storage.Checker) (Wrapper, error) {
+	const op errors.Op = "stash.WithEtcd"
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+	c, err := clientv3.New(clientv3.Config{
+		Endpoints:   endpoints,
+		DialTimeout: time.Second * 5,
```
michalpristas: why not use c.Timeout?

pkg/stash/with_etcd.go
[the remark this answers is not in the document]
marwan-at-work: @marpio the lock has a 60 second time to live, I think we should maybe raise it to 2 minutes. I hope we won't have to make it yet another config value lol

pkg/stash/with_etcd.go
marwan-at-work: @michalpristas this is just the "init" timeout, so when clientv3 pings the Etcd backend, it only needs 5 seconds to determine whether it's healthy or not, otherwise the c.Timeout could be a minute or two and we don't want to wait that long on Athens Startup

marwan-at-work:
Closing in favor of https://github.com/gomods/athens/pull/1070

PR #940, opened 2018-11-21
Fixed more typos

deblasis:

Some more typos spotted while reading the docs :)

**How is the fix applied?**

- to _ all the way

Fixes #

### DIFF
--- a/docs/content/configuration/authentication.md
+++ b/docs/content/configuration/authentication.md
@@ -43,8 +43,8 @@ weight: 1
     $env:ATHENS_STORAGE = "$(Join-Path $pwd athens-storage)"
     $env:ATHENS_SVN = "$(Join-Path $pwd .subversion)"
     md -Path $env:ATHENS_STORAGE
-    docker run -d -v "$($env:ATHENS-STORAGE):/var/lib/athens" `
-      -v "$($env:ATHENS-SVN):/root/.subversion" `
+    docker run -d -v "$($env:ATHENS_STORAGE):/var/lib/athens" `
+      -v "$($env:ATHENS_SVN):/root/.subversion" `
       -e ATHENS_DISK_STORAGE_ROOT=/var/lib/athens `
       -e ATHENS_STORAGE_TYPE=disk `
       --name athens-proxy `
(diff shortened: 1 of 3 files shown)


docs/content/configuration/authentication.md
```diff
@@ -43,8 +43,8 @@ weight: 1
     $env:ATHENS_STORAGE = "$(Join-Path $pwd athens-storage)"
     $env:ATHENS_SVN = "$(Join-Path $pwd .subversion)"
     md -Path $env:ATHENS_STORAGE
-    docker run -d -v "$($env:ATHENS-STORAGE):/var/lib/athens" `
-      -v "$($env:ATHENS-SVN):/root/.subversion" `
+    docker run -d -v "$($env:ATHENS_STORAGE):/var/lib/athens" `
+      -v "$($env:ATHENS_SVN):/root/.subversion" `
```
manugupt1: Is this on the host? If it is on the host, this could be any envirnoment variable for ATHENS-SVN right?

docs/content/configuration/authentication.md
```diff
@@ -166,8 +166,8 @@ weight: 1
     $env:ATHENS_STORAGE = "$(Join-Path $pwd athens-storage)"
     $env:ATHENS_BZR = "$(Join-Path $pwd .bazaar)"
     md -Path $env:ATHENS_STORAGE
-    docker run -d -v "$($env:ATHENS-STORAGE):/var/lib/athens" `
-      -v "$($env:ATHENS-BZR):/root/.bazaar" `
+    docker run -d -v "$($env:ATHENS_STORAGE):/var/lib/athens" `
+      -v "$($env:ATHENS_BZR):/root/.bazaar" `
```
manugupt1: I have the same question for bzr

docs/content/configuration/authentication.md
deblasis: Environment variables can only have alphanumeric characters and underscores as far as I remember.

Anyways, the problem is that there is a mismatch with the variable definition (in this case line 44) and it's usage (in this case line 47)

docs/content/configuration/authentication.md
deblasis: same as above

PR #941, opened 2018-11-22
[WIP] use cosmos db emulator

marpio:

Azure pipelines builds are slow b/c we need to install mongodb

**How is the fix applied?**

Use Azure Cosmos DB emulator insead

No issue


azure-pipelines.yml
```diff
@@ -34,26 +34,21 @@ steps:
...
+    }
+  displayName: 'start cosmos db emulator'
 
 - powershell: |
     go test -mod=vendor -race ./...
   env:
     ATHENS_MINIO_ENDPOINT: "127.0.0.1:9001"
-    ATHENS_MONGO_STORAGE_URL: "127.0.0.1:27017"
+    ATHENS_MONGO_INSECURE: "true"
+    ATHENS_MONGO_STORAGE_URL: "mongodb://localhost:C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==@localhost:10255/admin?ssl=true"
```
manugupt1: Just a thought in mind. If we are using an emulator, we might want to check if the indices when the tests start, they should not have any documents or modules in it, so that our integration tests do not give a false success.

manugupt1:
oops! I updated it on wrong branch!

pkg/storage/mongo/mongo.go
```diff
@@ -56,7 +56,7 @@ func (m *ModuleStore) connect() error {
 
 func (m *ModuleStore) initDatabase() error {
 	// TODO: database and collection as env vars, or params to New()? together with user/mongo
-	m.d = "athens"
+	m.d = "admin"
```
arschles: What do you think about fixing up the `TODO` above instead of changing this? I know that this would break a lot of deployments that already use mongo storage (including https://athens.azurefd.net)

marpio: :+1:  Sure, I couldn't figure out how to talk to the emulator and I tried things out. It wasn't meant to stay here like this.

marpio:
I'm closing this - we should probably try again using docker.

PR #942, opened 2018-11-22
Readiness probe

### LINKED ISSUE #891: Add liveness and readiness probes for Kubernetes

The proxy can implement an endpoint that Kubernetes [liveness / readiness probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/) can use to figure out if the proxy is running and healthy.

I think we should implement the endpoint in the proxy and configure the helm chart to use it.

**Additional context**

I'm researching and preparing for #772 and found that this would be really helpful for us.

shivas:

Adding `/readiness` endpoint for readinessProbe 

**How is the fix applied?**

readiness endpoint created, charts template updated to include readiness and liveness probes

Fixes #891

### DIFF
--- a/cmd/proxy/actions/app_proxy.go
+++ b/cmd/proxy/actions/app_proxy.go
@@ -21,6 +21,7 @@ func addProxyRoutes(
 ) error {
 	app.GET("/", proxyHomeHandler)
 	app.GET("/healthz", healthHandler)
+	app.GET("/readiness", getReadinessHandler(s))
 	app.GET("/version", versionHandler)
 
 	// Download Protocol
--- a/cmd/proxy/actions/app_proxy.go
+++ b/cmd/proxy/actions/app_proxy.go
@@ -21,7 +21,7 @@ func addProxyRoutes(
 ) error {
 	app.GET("/", proxyHomeHandler)
 	app.GET("/healthz", healthHandler)
-	app.GET("/readiness", getReadinessHandler(s))
+	app.GET("/readyz", getReadinessHandler(s))
 	app.GET("/version", versionHandler)
 
 	// Download Protocol
--- a/cmd/proxy/actions/readiness.go
+++ b/cmd/proxy/actions/readiness.go
@@ -3,16 +3,11 @@ package actions
 import (
 	"github.com/gobuffalo/buffalo"
 	"github.com/gomods/athens/pkg/storage"
-	"net"
 )
 
 func getReadinessHandler(s storage.Backend) buffalo.Handler {
 	return func(c buffalo.Context) error {
 		if _, err := s.List(c, "github.com/gomods/athens"); err != nil {
-			return c.Render(500, nil)
-		}
-
-		if _, err := net.LookupIP("github.com"); err != nil {
 			return c.Render(500, nil)
 		}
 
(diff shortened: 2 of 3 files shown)


cmd/proxy/actions/app_proxy.go
```diff
@@ -21,6 +21,7 @@ func addProxyRoutes(
 ) error {
 	app.GET("/", proxyHomeHandler)
 	app.GET("/healthz", healthHandler)
+	app.GET("/readiness", getReadinessHandler(s))
```
michalpristas: can we make it `/readyz` to stick with convention?

shivas: sure, why not 👍

cmd/proxy/actions/readiness.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"github.com/gobuffalo/buffalo"
+	"github.com/gomods/athens/pkg/storage"
+	"net"
+)
+
+func getReadinessHandler(s storage.Backend) buffalo.Handler {
+	return func(c buffalo.Context) error {
+		_, err := s.List(c, "github.com/gomods/athens")
```
michalpristas: can you make it one liner:
```go
if _, err := s.List(); err != nil{
}
```

same with the one below

cmd/proxy/actions/readiness.go
```diff
@@ -0,0 +1,24 @@
...
+
+		if err != nil {
+			return c.Render(500, nil)
+		}
+
+		_, err = net.LookupIP("github.com")
```
michalpristas: List goes to local storage as well as upstream so i guess in case this test is failing the previous one fails as well
Look here: https://github.com/gomods/athens/blob/b92a642f1ac49758d21781f3e27de70b2fc140c2/pkg/download/protocol.go#L62

shivas: @michalpristas so, you suggesting removing DNS resolve test?

cmd/proxy/actions/readiness.go
```diff
@@ -0,0 +1,21 @@
...
+	"net"
+)
+
+func getReadinessHandler(s storage.Backend) buffalo.Handler {
+	return func(c buffalo.Context) error {
+		if _, err := s.List(c, "github.com/gomods/athens"); err != nil {
+			return c.Render(500, nil)
+		}
+
+		if _, err := net.LookupIP("github.com"); err != nil {
```
michalpristas: my comment got deleted:
What list does is that it checks the storage and then upstream,  so in case github is not resolvable it will fail prior to this check
https://github.com/gomods/athens/blob/b92a642f1ac49758d21781f3e27de70b2fc140c2/pkg/download/protocol.go#L62

cmd/proxy/actions/readiness.go
michalpristas: yeah as long as list stays there

cmd/proxy/actions/readiness.go
```diff
@@ -0,0 +1,21 @@
+package actions
+
+import (
+	"github.com/gobuffalo/buffalo"
+	"github.com/gomods/athens/pkg/storage"
+	"net"
```
michalpristas: add space please and separate imports
first imports from Standard library then all external ones (github...)

shivas: after removing DNS check "net" import will go away

shivas:
@michalpristas thanks!

PR #948, opened 2018-11-26
Alibaba OSS Storage

### LINKED ISSUE #775: Implement storage.Backend for Alibaba OSS
Alibaba Object Storage Service is missing implementation.

- Let OSS join the fun in `/pkg/storage/storage_tests`

**Additional context**
Related issues: #704, #703

For further reference the SDK is here https://github.com/aliyun/aliyun-oss-go-sdk

steakunderscore:

Support for Alibaba OSS can be provided by Minio. So lets mention it in the docs.

I also checked that the support works by running Athens against OSS.

**How is the fix applied?**

Update docs with how to configure Alibaba OSS.

Fixes #775

### DIFF
--- a/docs/content/configuration/storage.md
+++ b/docs/content/configuration/storage.md
@@ -14,6 +14,7 @@ The Athens proxy supports many storage types:
 1. [AWS S3](#aws-s3)
 1. [Minio](#minio)
 1. [DigitalOcean Spaces](#digitalocean-spaces)
+1. [Alibaba OSS](#alibaba-oss)
 
 All of them can be configured using `config.toml` file. You need to set a valid driver in `StorageType` value or you can set it in environment variable `ATHENS_STORAGE_TYPE` on your server.
 Also for most of the drivers you need to provide additional configuration data which will be described below.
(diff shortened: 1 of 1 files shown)


docs/content/configuration/storage.md
```diff
@@ -114,110 +115,151 @@ After this you can pass you credentials inside `config.toml` file
...
+            # Env override: ATHENS_MINIO_USE_SSL
+            EnableSSL = true
+
+            # Space name in your DO Spaces storage
+            # Env override: ATHENS_MINIO_BUCKET_NAME
+            Bucket = "YOUR_OSS_BUCKET_NAME"
+
+            # Region for Minio storage
+            # Env override: ATHENS_MINIO_REGION
+            Region = "YOUR_OSS_REGION"
```
t-tomalak: We probably could remove that from this section. Checked that and without passing region everything works. It's probably because user provide region in endpoint eg. `oss-eu-central-1.aliyuncs.com`

docs/content/configuration/storage.md
```diff
@@ -114,110 +115,151 @@ After this you can pass you credentials inside `config.toml` file
...
+
+##### Configuration:
+
+    # StorageType sets the type of storage backend the proxy will use.
+    # Env override: ATHENS_STORAGE_TYPE
+    StorageType = "minio"
+
+    [Storage]
+        [Storage.Minio]
+            # Address of DO Spaces storage
```
t-tomalak: Could we change that to  something like `# Address of Alibaba OSS`? And also all occurences of `Minio/DO Spaces` to `Alibaba OSS` in other parameters descriptions :slightly_smiling_face:

docs/content/configuration/storage.md
steakunderscore: Hmm, I'll look further into if it's actually used anywhere. I agree, if it's not used we should drop it. :+1:

docs/content/configuration/storage.md
steakunderscore: Oops :disappointed_relieved: Good spotting!

docs/content/configuration/storage.md
steakunderscore: After reading more of the docs, I agree with you, we should drop the Location for Alibaba OSS.

From the docs, the bucket region is fixed to the API endpoint region, so you need to change the endpoint to change the region.

The [PUT docs](https://www.alibabacloud.com/help/doc-detail/31959.htm) say:
> The region of the created bucket is consistent with the region of the endpoint from which the request is sent. Once the data center of the bucket is determined, all objects in this bucket are stored in the corresponding region. For more information, see Regions and endpoints .

docs/content/configuration/storage.md
```diff
@@ -14,6 +14,7 @@ The Athens proxy supports many storage types:
 1. [AWS S3](#aws-s3)
 1. [Minio](#minio)
```
manugupt1: Both Alibaba OSS and Digital Ocean use minio. Is it possible to add a note saying any minio compatible storage can be included. Can you also add a note that if a user is configuring a minio compatible storage, then a bug report should be filed?

This would cover up all the use cases.

PR #951, opened 2018-11-28
Remove timeouts from Storage documentation

t-tomalak:

As we droped timeout from `config.dev.toml` ( #928 ) for storages, we should also remove them from documentation

**How is the fix applied?**

Removed timeouts from storages configuration

Fixes:#


manugupt1:
Perfect! thanks @t-tomalak

PR #952, opened 2018-11-28
Add check if GCP bucket exists and remove GCP Bucket create

### LINKED ISSUE #932: In gcp we ignore bkt.Create "bucket already exists" error
**Describe the bug**
This drove me mad while trying to test https://github.com/gomods/athens/issues/829 .

Here https://github.com/gomods/athens/blob/master/pkg/storage/gcp/gcp.go#L46 
```
	if err != nil && !bucketExistsErr(err) {
		return nil, errors.E(op, err)
	}
```
we skip the error, assuming that if the bucket is already there it is because we created it. 
Today I learned that the name must be unique across all GCP (see https://cloud.google.com/storage/docs/naming - "bucket name considerations) and here https://stackoverflow.com/questions/25127112/google-cloud-storage-static-bucketname-not-available . 
So, if somebody else already picked up that bucket name, athens won't work and all the attempts to create / edit / delete will fail.

**Error Message**
All the calls return errors like 'fede-553@athens-test-222805.iam.gserviceaccount.com does not have storage.objects.get access to .. "

**To Reproduce**
Try to open a gcp storage with `athens-test` as bucket name.

**Expected behavior**
We should skip creating the bucket only if it exists in our project and stop ignoring http.StatusConflict error. 
Instead of checking the error, we should use https://godoc.org/cloud.google.com/go/storage#Client.Buckets to check for the existence of that bucket in the project we are targeting.

t-tomalak:

Errors from GCP are not properly verified on Bucket Create action 

**How is the fix applied?**

Removed GCP Bucket create and added simple check if bucket is available. It will be easier for user to create bucket from GCP Panel becayse names must be unique across all GCP Storages. Also will lower required permission for GCP Account used by Athens because that user will not need access to `Create New Bucket` Permission.

Fixes #932


t-tomalak:
@marpio good catch! updated

pkg/storage/gcp/gcp.go
```diff
@@ -32,38 +30,29 @@ type Storage struct {
...
+	u, err := url.Parse(fmt.Sprintf("https://storage.googleapis.com/%s", gcpConf.Bucket))
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
-	bkt := gcpBucket{storage.Bucket(bucketname)}
-	err = bkt.Create(ctx, gcpConf.ProjectID, nil)
-	if err != nil && !bucketExistsErr(err) {
+
+	bkt := s.Bucket(gcpConf.Bucket)
+	if _, err := bkt.Attrs(ctx); err != nil {
```
marwan-at-work: Let's check if `err == storage.ErrBucketNotExist` and give a friendlier message in that case: return `errors.E(op, "You must manually create a storage bucket for Athens, see https://cloud.google.com/storage/docs/creating-buckets#storage-create-bucket-console")`

t-tomalak: Added but I don't know if I should break the line with log message (now is very long line)

manugupt1: @marwan-at-work I think your comments have been addressed. Can we merge this?

PR #958, opened 2018-12-03
Add prometheus metrics collectors for http handlers

### LINKED ISSUE #816: Support for Metrics using Prometheus
Only Tracing is currently supported. It'd be nice to also track metrics.
Export metrics to prometheus 
Reference: https://opencensus.io/guides/exporters/supported-exporters/go/prometheus/

t-tomalak:

In Athens we don't collect metrics for http requests

**How is the fix applied?**

Added middleware that wraps buffalo handlers and instruments them to collect count of requests, requests duration, response size, request sizes. As first I wanted to add base http stats in follow up Pull Request there can be added other stats collectors . Some of them are:

1. Errors - count,
2. Disk usage
3. network usage
4. For workers number of in/out work items and time spent in a queue
5. Cache hit rate (because it's fun to watch)

Fixes #816


cmd/proxy/actions/app.go
```diff
@@ -6,9 +6,9 @@ import (
 
 	"github.com/gobuffalo/buffalo"
 	"github.com/gobuffalo/buffalo/render"
-	forcessl "github.com/gobuffalo/mw-forcessl"
-	paramlogger "github.com/gobuffalo/mw-paramlogger"
-	sessions "github.com/gobuffalo/x/sessions"
+	"github.com/gobuffalo/mw-forcessl"
```
robjloranger: These three imports should not be changed, unless their respective uses are also updated.

t-tomalak: I double checked that and I see that they are redundant. `github.com/gobuffalo/mw-forcessl` is `package forcessl` and `github.com/gobuffalo/mw-paramlogger` is `package paramlogger`. I can revert, but with this changes everything still works :slightly_smiling_face:

robjloranger: Oh weird, just the import path doesn't match. I was wondering why the build didn't fail.

pkg/metrics/http.go
```diff
@@ -0,0 +1,49 @@
...
+		},
+		[]string{},
+	)
+
+	return []prometheus.Collector{total, duration}
+}
+
+// Middleware is middleware that instruments buffalo handlers
+func Middleware(collectors ...prometheus.Collector) buffalo.MiddlewareFunc {
+	return func(next buffalo.Handler) buffalo.Handler {
```
michalpristas: trying to figure out if we can be buffalo independent here, but i guess it's ok, we will revisit once we're there

pkg/observ/stats.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, errors.E(op, "StatsExporter not specified. Stats won't be collected")
+	default:
+		return nil, errors.E(op, fmt.Sprintf("StatsExporter %s not supported. Please open PR or an issue at github.com/gomods/athens", statsExporter))
+	}
+
+	if err := registerViews(); err != nil {
+		return nil, errors.E(op, err)
+	}
+
+	return func() {}, nil
```
t-tomalak: Currently `func() {}` it's not needed but when we would add StackDriver or DataDog exporter it will be needed and by having that already here it will be easy to implement.

manugupt1:
Hi @t-tomalak,  codewise it looks perfect to me. This is rad.

Can you link the instructions that you used to run it. So, that I can use it to test it. You can also add it in the docs in this PR if you would like to?

Have a great holiday!

pkg/observ/stats.go
michalpristas: can you then add a comment there so we know why it is there.
i'm not a big fan of future code and what if we will need it approach.

t-tomalak:
@manugupt1 I already created #970 issue that will allow to automaticaly set up prometheus in local environment, so I don't think we need docs about how to setup prometheus. Metrics can be either checkd manually on `localhost:3000/metrics` or you can start own prometheus. You can start it by creating configuration file `prometheus.yml`
```
global:
  scrape_interval: 2s

  external_labels:
    monitor: 'athens'

scrape_configs:
  - job_name: 'athens'

    scrape_interval: 2s

    static_configs:
      - targets: ['localhost:3000']
``` 
Then run: 
`docker run --name prometheus --network=host -v prometheus.yml:/etc/prometheus/prometheus.yml -d -p 9090:9090 prom/prometheus`

And after that prometheus should be accessible under loclhost:9090 in your browser.

pkg/observ/stats.go
t-tomalak: Added, also removed that func() from `registerPrometheusExporter` method

pkg/observ/stats.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/gomods/athens/pkg/errors"
+	"go.opencensus.io/exporter/prometheus"
+	"go.opencensus.io/plugin/ochttp"
+	"go.opencensus.io/stats/view"
+)
+
+// RegisterStatsExporter determines the type of StatsExporter service for exporting stats from Opencensus
+// Currently it supports: prometheus
+func RegisterStatsExporter(app *buffalo.App, statsExporter, service string) (func(), error) {
+	const op errors.Op = "RegisterStatsExporter"
```
marpio: Can you add the pkg name like `observ.RegisterStatsExporter` ?

pkg/observ/stats.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// Currently func() prop it's not needed by Prometheus exporter.
+	// This param should be used to pass StackDriver Flush or
+	// DataDog Stop method when this Exporters will be implemented.
+	return func() {}, nil
+}
+
+// registerViews register stats which should be collected in Athens
+func registerViews() error {
+	const op errors.Op = "registerViews"
```
marpio: add pkg name please

pkg/observ/stats.go
t-tomalak: sure, I was keeping convention from second obervable packet. Should I change there also?

t-tomalak: I guess yes so I added :slightly_smiling_face:

pkg/observ/stats.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return nil
+}
+
+func metricsHandler(handler http.Handler) buffalo.Handler {
+	return buffalo.WrapHandler(handler)
+}
+
+// StatsMiddleware is middleware that instruments buffalo handlers
+func StatsMiddleware() buffalo.MiddlewareFunc {
```
michalpristas: just a small nit, can be exported first and unexported after them?

t-tomalak: Sorted! Thanks for review :slightly_smiling_face:

PR #960, opened 2018-12-04
Propagate HOME env to go command execution

### LINKED ISSUE #959: Private repo using ssh not supported since git global config is ignored
I have a private gitlab repo that I use ssh instead of https to access. To make the `go` command aware of this, I add a configuration in my global git configuration file `$HOME/.gitconfig`:

`git config --global url."gitlab@privaterepo:".insteadOf "https://privaterepo/"`

Now `go get` from my private repo works like a charm.

When I use Athens as proxy, it calls `go mod download -json` and it fails since it still tries to access my private repo using https despite my config.

The reason for this is that Athens sets environment variables in: https://github.com/gomods/athens/blob/154c993c42c6dfd43a6adf60e543f5a98b468509/pkg/module/go_get_fetcher.go#L161-L182

And `HOME` isn't one of the variables that is passed on to the command. The result is that `git` cannot find the global configuration file, and all settings are ignored.

 - OS: Linux
 - Go version : 1.11
 - Buffalo Version : -
 - Proxy version : master
 - Storage (fs/mongodb/s3 etc.) : mem or disk

leitzler:
Add the ability to manually point out a git configuration file. This is needed to be able to use private repositories that uses ssh instead of https, see #959 

There is also a `GIT_CONFIG` env but it seems like only `git config` uses that one:
https://github.com/git/git/blob/965798d1f2992a4bdadb81eba195a7d465b6454a/builtin/config.c#L601

Fixes #959

### DIFF
--- a/pkg/module/go_get_fetcher.go
+++ b/pkg/module/go_get_fetcher.go
@@ -160,6 +160,7 @@ func downloadModule(goBinaryName string, fs afero.Fs, gopath, repoRoot, module, 
 // successfully (such as GOPATH, GOCACHE, PATH etc)
 func PrepareEnv(gopath string) []string {
 	pathEnv := fmt.Sprintf("PATH=%s", os.Getenv("PATH"))
+	homeEnv := fmt.Sprintf("HOME=%s", os.Getenv("HOME"))
 	httpProxy := fmt.Sprintf("HTTP_PROXY=%s", os.Getenv("HTTP_PROXY"))
 	httpsProxy := fmt.Sprintf("HTTPS_PROXY=%s", os.Getenv("HTTPS_PROXY"))
 	noProxy := fmt.Sprintf("NO_PROXY=%s", os.Getenv("NO_PROXY"))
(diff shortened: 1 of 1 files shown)


robjloranger:
Is the env variable `HOME` available on all host architectures? I'm mostly wondering about windows.

We might need an OS switch for this one and point to `%UserProfile%` or similar.

leitzler:
You might be right, I don't have access to any Windows machine to test. The implementation follows the environment variables in specified in [git-scm](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables). Wouldn't be surprised if Windows handles it in a special way.

robjloranger:
You can use the variable `GIT_CONFIG` to point to a configuration file.

## GIT_CONFIG
```Take the configuration from the given file instead of .git/config. Using the "--global" option forces this to ~/.gitconfig. Using the "--system" option forces this to $(prefix)/etc/gitconfig```

leitzler:
See original description. Feel free to try it yourself with `fetch` (and share what version of git you use if it works).

leitzler:
Also, it wouldn’t enable more configuration than this PR. Since if HOME is passed on athens will be able to (automatically) use the specific users global configuration.

leitzler:
Looks like the official git uses `HOME` even for Windows:

Opening the config file: https://github.com/git/git/blob/d166e6afe5f257217836ef24a73764eba390c58d/builtin/config.c#L627

Expanding `HOME`: https://github.com/git/git/blob/d166e6afe5f257217836ef24a73764eba390c58d/path.c#L719

arschles:
@leitzler I think @robjloranger had a good idea [in slack](https://gophers.slack.com/archives/C9LRAQN8N/p1544053564163400) (copied here):

>Maybe we can add a test to see if the variable is found and we have the git config file?

That way we can always be sure that we're setting the `HOME` variable properly on windows and linux-ey flavors too

PR #972, opened 2018-12-07
Aerial doc 2

arschles:

The Athens docs are not super well organized and I think part of the problem is the current layout of the docs hides that.

**How is the fix applied?**

I haven't improved the organization in here, but I've changed the [theme](https://github.com/thingsym/hugo-theme-techdoc) to something that encourages a more end-to-end journey as @marwan-at-work explained in #924. It also un-collapses the docs section as in https://github.com/gomods/athens/issues/924#issuecomment-444997361

This might be a good start to improving our docs. If folks like it, I'll continue with docs organization.

cc/ @marwan-at-work 

Note: I'm replacing https://github.com/gomods/athens/pull/968 with this PR because I accidentally added a submodule in that one and I couldn't figure out how to un-submodule my submodule and git defeated me. I have lost the battle but I won't lose the war.


michalpristas:
while you're at it.
what are your opinions on splitting configuring/installing sections to scenarios
for example scenario 1

- Running athens on docker
- Running athens in cloud
- RUnning athens on my host
- Running athens with GCP
- Running athens with Azure

and so on

michalpristas:
@arschles is this alive?

arschles:
@michalpristas I'm just picking this back up now

PR #973, opened 2018-12-07
Adding ability to run all tests inside docker containers

arschles:

Right now, you have to follow a few different steps to execute unit tests on your local machine, and we provide no instructions on how to run end to end tests on your local machine. So the result is that it's hard to test everything locally that CI will test after you submit your pull request.

**How is the fix applied?**

This PR adds Makefile targets and docker compose services for running unit and e2e tests completely inside docker containers, so that folks don't have to do any setup or configuration on their local machines (if they don't want). It also adds documentation for how to run the tests.

Ref #262


DEVELOPMENT.md
```diff
@@ -62,19 +62,62 @@ Note that `make dev` only runs the minimum amount of dependencies needed for thi
...
+```console
+make dev-teardown
+```
+
+# Run End to End Tests
+
+End to end tests ensure that the Athens server behaves as expected from the `go` CLI tool. These tests run exclusively inside Docker containers. To run them, execute this command:
+
+```console
+make test-e2e-docker
```
robjloranger: should this be `make test-e2e` as the instructions for docker based end to end is above now?

robjloranger: this whole sub section would need to be corrected as well.

docker-compose.yml
```diff
@@ -1,5 +1,24 @@
 version: '3'
 services:
+  testunit:
+    build:
+      context: .
+      dockerfile: Dockerfile.test
```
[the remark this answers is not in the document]
arschles: Yup, it was accidentally git-ignored, so I added the exception for it in https://github.com/gomods/athens/pull/973/commits/138a5f3cd9d7688289cf775af152ec5c776041cc

DEVELOPMENT.md
arschles: good catch. I will collapse this section with the "completely in containers" section, so that we'll have only instructions on running e2e tests inside docker

arschles: @robjloranger after I made the previous comment I realized that the above instructions are for unit tests

arschles:
@manugupt1 @robjloranger I think I addressed your comments here. Can you let me know what you think?

PR #976, opened 2018-12-07
Switching to document the use of stable tags

arschles:

Various documents have instructions that use the `:latest` tag for our docker image, but that has code from the head of master, which hasn't been officially released.

**How is the fix applied?**

I've changed the docs to use the `:stable` tag instead

cc/ @robjloranger @marpio @paganotoni


robjloranger:
I want to note as well, that the issue is we aren't getting any images tagged with `latest` or we need to push a patch version so the script will now push a `latest` tagged image, in which case the docs could stay the same.

In either case, as of today, a user running `docker pull gomods/athens:latest` receives an error of `Error response from daemon: manifest for gomods/athens:latest not found`.

carolynvs:
Sounds like the real problem is that latest isn't being pushed. Maybe it changed (I've been busy) but latest should be the latest tagged release, and canary is the tip of master.

I can take a look tomorrow and see what's wrong with the deployments?

robjloranger:
Please do @carolynvs, I tried but wasn't too familiar with it. The scripts look as though they should be correctly tagging but the most recent, non versioned, push was tagged with commit sha.

I found part of the problem, the version tag at `v0.2.0` was actually `release-v0.0.2`, which would fail the regex. This was back on `gomods/proxy` though.

ref: build log @docker push for v0.2 https://travis-ci.org/gomods/athens/builds/455582715#L1561

Again, I'm not really sure but should help get started.

ref: slack convo https://gophers.slack.com/archives/C9LRAQN8N/p1544217316305600

carolynvs:
Will do!

marpio:
@carolynvs @robjloranger @arschles 
So it seems that the problem is that, Carolyn's commit https://github.com/gomods/athens/commit/bee463053a358b0de4650af317996c9f57bf611f came too late. 
The v0.2.0 build used the old version of the `push-docker-images.sh` script where `:stable` tag was used.
For reference: https://travis-ci.org/gomods/athens/builds/454139475#L1591

arschles:
@marpio sorry for the long delay. Do you think we should close this then?

marpio:
@arschles yes!

carolynvs:
@arschles Can this be closed now?

marpio:
@arschles  I'm closing it now since we released v0.3.0

PR #979, opened 2018-12-08
Add support for TLS

leitzler:
Athens doesn't support serving via https, this PR adds support for that.


cmd/proxy/main.go
```diff
@@ -4,9 +4,12 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"net/http"
 	"os"
 	"path/filepath"
 
+	"github.com/gobuffalo/buffalo/servers"
+
```
michalpristas: rmeove newline please

pkg/config/config.go
```diff
@@ -48,6 +50,30 @@ func (c *Config) BasicAuth() (user, pass string, ok bool) {
...
+	if err != nil {
+		return "", "", fmt.Errorf("Could not access TLSCertFile: %v", err)
+	}
+
+	keyFile, err := os.Stat(c.TLSKeyFile)
+	if err != nil {
+		return "", "", fmt.Errorf("Could not access TLSKeyFile: %v", err)
+	}
+
+	if keyFile.Mode()&077 != 0 {
```
michalpristas: this will not work on windows, please add GOOS check and skip check for windows.

pkg/config/config.go
```diff
@@ -67,7 +67,7 @@ func (c *Config) TLSCertFiles() (cert, key string, err error) {
 		return "", "", fmt.Errorf("Could not access TLSKeyFile: %v", err)
 	}
 
-	if keyFile.Mode()&077 != 0 {
+	if keyFile.Mode()&077 != 0 && runtime.GOOS != "windows" {
```
michalpristas: :+1:

PR #989, opened 2018-12-10
pushing the latest tag on master builds

### LINKED ISSUE #980: Vote: Should we use : latest tag on docker hub
At the moment the script that's pushing to docker hub sets :latest tag on every release:
 https://github.com/gomods/athens/blob/5ab820de96fa69c1a9e69119d9ca9e171b33667f/scripts/push-docker-images.sh#L25
There are pros and cons of using :latest and we should decide if we want to use it.

arschles:

We aren't pushing the a docker image with the `latest` tag on it when we commit to the master branch, and that's been causing confusion with folks who are used to doing a `docker pull gomods/athens`, without the tag. That's because the `docker` CLI defaults to `latest`.

**How is the fix applied?**

I changed the push docker images script around to push `latest` on master pushes, and `stable` on tag pushes (so that `stable` always holds the last tag release)

Also, this PR is "plane quality" - I haven't tested anything out yet because I'm on wifi on the plane and this branch took about 45 seconds just to push to GH 😆 . I wanted to get this started for now.

cc/ @carolynvs @robjloranger @marpio @leitzler because I know all of you have commented on this topic a few times.

Fixes #980 if we the vote for "latest" wins.


manugupt1:
I think the convention is 
canary -> latest master
latest -> latest release
vX.X.X-> image:vX.X.X

@carolynvs might be able to correct me on this one!

marpio:
@manugupt1 Yeah, that is the convention we're following now. 
This PR would change this to:
canary -> latest master
latest -> latest master
vX.X.X -> image:vX.X.X
vX.X.X -> stable

manugupt1:
@marpio I looked at hub.docker.com
latest usually points to the latest release (which is latest vX.X.X)

marpio:
Makes sense @manugupt1
@arschles should we leave the push script unchanged and follow the convention :latest-> last release vX.X.X as we do now?

manugupt1:
@marpio That would be my thought and then we can update docker hub to reflect what we mean so that if someone asks what should we do, we can point them to the documentation that we have over there?

marpio:
:+1: it's safer that way imo. Even if someone does docker pull gomods/athens in production she/he will get a stable version.
Also I'm with you @manugupt1 that we should do a patch release.

arschles:
Repeating what we discussed in slack: the next release should push the latest tag, so we are going to close this and let the `latest` tag issue be pushed then, so I am closing this

PR #997, opened 2018-12-13
Allow mgo.ParseURL to set [/database]

### LINKED ISSUE #906: Ability to set MongoDB database name

I deployed athens to heroku. If I wanted to use the sandbox mLab offering of mongoDB for example I'd have to take their automatically generated database name (something like `heroku_aowd89adw` which wouldn't be `athens`. `athens` is currently required by this storage driver implementation.

At the moment we hardcode the database name as [`athens`](https://github.com/gomods/athens/blob/f776bc66671f5d58e6346ed7d540b989c84d9709/pkg/storage/mongo/mongo.go#L59) and there's a `TODO` to make a better solution.
 
Either creating environment variables for the database inline with our current environment variable naming or my preffered solution is reading the `[/database]` from `mgo.ParseURL` 

Leave it as it'd be and force users to have `athens` as their database name.

**Additional context**

Some discussion around the subject https://gophers.slack.com/archives/C9LRAQN8N/p1542026809792000

chriscoffee:

Note: I've marked this as WIP for a short time as I forgot to put the PR in ages ago and now it's fallen out of date, I'll fix up the code asap.

As per Michael's solution we need to be able to set the database from
either the parsed connection string, or fallback to athens if not set.

**How is the fix applied?**

Connection string can either be overridden by `ATHENS_MONGO_DEFAULT_DATABASE` or in the connection string. 

Fixes #906


pkg/storage/mongo/mongo.go
```diff
@@ -79,6 +78,12 @@ func (m *ModuleStore) newSession(timeout time.Duration, insecure bool) (*mgo.Ses
 
 	dialInfo.Timeout = timeout
 
+	if dialInfo.Database != "" {
+		m.d = dialInfo.Database
+	} else {
+		m.d = conf.DefaultDBName
```
arschles: @chriscoffee I don't think we need to set it manually b/c envconfig will do it automatically, just an FYI. But this doesn't hurt anything so nothing to change here 😄

arschles:
@chriscoffee if this isn't WIP anymore, can you take `WIP` off of the title? the WIP bot will "sign off" on it then 😄

PR #1005, opened 2018-12-20
Refactoring code to improve HTTP status code

nafisfaysal:
     *   In our codebase, there have so many palaces we use HTTP status code sometimes we write the [Value](https://github.com/gomods/athens/blob/master/pkg/log/buffalo_formatter.go#L22-#L24) and sometimes we use [Value+Description](https://github.com/gomods/athens/blob/master/pkg/middleware/filter.go#L32).

**Describe the issue you have been trying to solve.**
    * Here I update and maintained HTTP status code with Value+Description also Improving the Readability of our Code.


nafisfaysal:
@michalpristas :smile:

PR #1009, opened 2018-12-22
Modified S3 storage to check for AWS default credentials.

### LINKED ISSUE #882: Add support for aws credentials file
Some users when running Athens with S3 storage want to use aws default configuration aka `.aws/credentials`. Unfortunately we can't omit validation of our storage parameters from configuration file (or could we in that case?)

1. We can drop `required` flag on every parameter of `S3config` and check config inside `New` method in `storage/s3`. 
2. We can create alternative s3 storage which will don't require any configuration(what name should it have?), also we will add seperate `New` method inside S3 storage for that.

We could also move this credentials as we do with .netrc and .hgrc files https://github.com/gomods/athens/issues/716

**Additional context**
https://gophers.slack.com/archives/C9LRAQN8N/p1541454892427000

markopotamus:

Users that want to use S3 storage must currently specify their access key ID and secret access key to authenticate with the AWS S3 API.  Some users would like to be able to read their credentials from the file installed by the AWS CLI.

[AWS CLI config file documentation](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html)

**How is the fix applied?**

I made the access key ID and secret access key fields on the S3 config optional.  If present in the config, they will be used to construct new static credentials for S3.  Otherwise, I build the path to the file in the user's home directory and use it to construct new shared credentials for S3.  For now, only the credentials for the default profile are returned.

Fixes #882


michalpristas:
Looks good, could you please update [these docs](https://docs.gomods.io/configuration/storage/#aws-s3) as well so it is clear when you use it that config file has priority

docs/content/configuration/storage.md
```diff
@@ -101,7 +101,7 @@ how to create `bucket` inside it.
 
 This driver is using the [AWS S3](https://aws.amazon.com/s3/) and assumes that you already have `account` and `bucket` created in it.
 If you never used Amazon Web Services there is [quick guide](https://docs.aws.amazon.com/AmazonS3/latest/gsg/GetStartedWithS3.html) how to create `bucket` inside it.
-After this you can pass you credentials inside `config.toml` file
+After this you can pass you credentials inside `config.toml` file.  If the access key ID and secret access key are not specified in `config.toml`, the driver will attempt to load credentials for the default profile from the [AWS CLI configuration file](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html) created during installation.
```
michalpristas: could you please fix the typo laying there: your credentials

markopotamus: Will do.  Good catch :)

PR #1010, opened 2018-12-22
Remove Buffalo

marwan-at-work:
1. Replaces Buffalo with Standard Library
2. Removes Jaeger as default option so that we don't ping it
3. Remove Buffalo hacks for tracing/stats. 

Follow up PRs: 
1. Update DEVELOPMENT.md to remove buffalo mentions 
2. Remove observ.StartSpan and just use trace.StartSpan directly.


cmd/proxy/actions/auth.go
```diff
@@ -4,26 +4,13 @@ import (
...
 	"strings"
 
-	"github.com/gobuffalo/buffalo"
-	"github.com/markbates/goth"
-	"github.com/markbates/goth/gothic"
-	"github.com/markbates/goth/providers/github"
 	"github.com/mitchellh/go-homedir"
 )
 
-func initializeAuth(app *buffalo.App) {
```
manugupt1: This can be removed in a separate PR.

cmd/proxy/actions/home.go
```diff
@@ -1,10 +1,9 @@
...
 
 import (
-	"github.com/gobuffalo/buffalo"
 	"net/http"
 )
 
-func proxyHomeHandler(c buffalo.Context) error {
-	return c.Render(http.StatusOK, proxy.JSON("Welcome to The Athens Proxy"))
+func proxyHomeHandler(w http.ResponseWriter, r *http.Request) {
+	w.Write([]byte(`"Welcome to The Athens Proxy"`))
```
manugupt1: We can have one default helper function to make sure that everything is always a json response.

config.dev.toml
```diff
@@ -154,7 +147,7 @@ StatsExporter = "prometheus"
     [Storage.Disk]
         # RootPath is the Athens Disk Root folder
         # Env override: ATHENS_DISK_STORAGE_ROOT
-        RootPath = "/path/on/disk"
+        RootPath = "/Users/208581/Desktop/ans"
```
manugupt1: Let us keep it empty rather than a folder that does not exist.

pkg/download/catalog.go
```diff
@@ -21,29 +21,34 @@ type catalogRes struct {
 }
 
 // CatalogHandler implements GET baseURL/catalog
-func CatalogHandler(dp Protocol, lggr log.Entry, eng *render.Engine) buffalo.Handler {
+func CatalogHandler(dp Protocol, lggr log.Entry) http.Handler {
```
manugupt1: Content-type json should be in the header.

pkg/download/list.go
```diff
@@ -15,21 +14,24 @@ import (
...
+		mod, err := paths.GetModule(r)
 		if err != nil {
 			lggr.SystemErr(errors.E(op, err))
-			return c.Render(http.StatusInternalServerError, nil)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
 		}
 
-		versions, err := dp.List(c, mod)
+		versions, err := dp.List(r.Context(), mod)
```
manugupt1: This is sweet!

pkg/log/log_context.go
```diff
@@ -1,25 +1,25 @@
...
 // If there is no value for the key or the type assertion fails, it returns a new
 // entry from the provided logger
-func EntryFromContext(ctx buffalo.Context) Entry {
-	d := ctx.Data()
-	e, ok := d[logEntryKey].(Entry)
-	if e == nil || !ok {
+func EntryFromContext(ctx context.Context) Entry {
+	e, ok := ctx.Value(logEntryKey).(Entry)
+	if !ok || e == nil {
 		return NoOpLogger()
```
manugupt1: If we are doing a NoOp() then we should remove it altogether from the source code first and then move to the removal of buffalo imo

cmd/proxy/actions/home.go
marwan-at-work: Yes, but let's do it in a future pr

config.dev.toml
marwan-at-work: This was a slip up while testing things, will put it back to what it was can be changed in a future pr. trying to keep it as small as possible haha.

pkg/log/log_context.go
marwan-at-work: @manugupt1 we've always done NoOpLogger, this is not new. Or did I misunderstand you?

pkg/download/catalog.go
marwan-at-work: @manugupt1 done

pkg/log/log_context.go
manugupt1: No, you did not I made a mistake!

pkg/middleware/filter.go
```diff
@@ -5,42 +5,43 @@ import (
...
 				// Include: please handle this module in a usual way
-				return next(c)
+				h.ServeHTTP(w, r)
+				return
 			case module.Direct:
 				// Direct: do not store modules locally, use upstream proxy
-				newURL := redirectToUpstreamURL(upstreamEndpoint, c.Request().URL)
-				return c.Redirect(http.StatusSeeOther, newURL)
+				newURL := redirectToUpstreamURL(upstreamEndpoint, r.URL)
+				http.Redirect(w, r, newURL, http.StatusSeeOther)
```
manugupt1: Can we simply do return http.Redirect and same for the above?

marwan-at-work: @manugupt1 we cannot return `http.Redirect` because it does not return anything, nor does the http.HandlerFunc has a return type :)

manugupt1: Interesting, I was not aware that about with go. Some languages return nil and I do not use go daily.

.gitignore
```diff
@@ -18,7 +18,6 @@ tmp/*
 vgp
 bin/*
 .envrc
-athens
 yarn-error.log
```
manugupt1: Let us remove it as well, since we have removed buffalo altogether, there is no reason that we should have yarn-log as well.

cmd/proxy/actions/app.go
```diff
@@ -104,52 +86,44 @@ func App(conf *config.Config) (*buffalo.App, error) {
...
+	flushStats, err := observ.RegisterStatsExporter(r, conf.StatsExporter, Service)
 	if err != nil {
 		lggr.Infof("%s", err)
 	} else {
 		defer flushStats()
-		app.Use(observ.StatsMiddleware())
-	}
-
-	// Automatically redirect to SSL
-	app.Use(forcessl.Middleware(secure.Options{
```
manugupt1: Does this break any functionality?

marwan-at-work: @manugupt1 I'm not aware if this was used at all. I'll bring it back just in case :)

marwan-at-work:
@manugupt1 thanks for the second look, all comments addressed.

PR #1019, opened 2018-12-27
docs: remove commit hash from known limitations

marwan-at-work:
once https://github.com/gomods/athens/pull/1015 is merged, this is no longer a limitation.


marpio:
@marwan-at-work should we wait with it until go1.12 ?

marwan-at-work:
@marpio it doesn't hurt to merge #1015 before 1.12, but yeah for this PR it might be good to mention that hashes and branches only work with 1.12+

PR #1022, opened 2019-01-03
Use default conf values if file not found

### LINKED ISSUE #1008: Add defaults for Athens config values

If you don't pass Athens a config file, it won't have all the configuration values that it needs

I'd like all of the default configurations that we see in the sample config file to be hard-coded into the binary, so that you only need to use the config file, or set environment variables, to override the defaults if you need to

None

**Additional context**

We discussed this during tghe Dec 20, 2018 dev meeting

marpio:
Use default values if config file is not present.

Fixes #1008


pkg/config/config.go
```diff
@@ -41,6 +42,37 @@ type Config struct {
 	Storage          *StorageConfig
 }
 
+// Load loads the config from a file.
+// If file is not present returns default config
+func Load(configFile string) (*Config, error) {
```
marwan-at-work: I think it's better just to check if `configFile == ""`. In other words, if we provide a `-config` flag, then we *must* have a valid config file, or we should fail. But if no flag is provided, then use the default stuff in the code base. What do you think?

manugupt1: I agree with @marwan-at-work  as well! I think it is better that way.

I think we should only add defaults for Dev and test. For production, using memory as a storage is a bad idea. Further, maybe in a newer commit, we just look at the environment and build defaults. Until then, I think we should have defaults only for dev. What do you think?

pkg/config/config.go
```diff
@@ -41,6 +42,37 @@ type Config struct {
...
+		GoEnv:            "development",
+		GoGetWorkers:     30,
+		ProtocolWorkers:  30,
+		LogLevel:         "debug",
+		CloudRuntime:     "none",
+		TimeoutConf:      TimeoutConf{Timeout: 300},
+		StorageType:      "memory",
+		Port:             ":3000",
+		GlobalEndpoint:   "http://localhost:3001",
+		TraceExporterURL: "http://localhost:14268",
```
manugupt1: Would you not want to specify type as jaeger if you have an export url?

manugupt1: Yep! Actually, I am thinking that we should remove all parameters related to observability now from dev.

pkg/config/config.go
marpio: @marwan-at-work 
I was thinking about this as well. The reason I've made it this way, was that I din't want to specify the `config_file` flag while developing all the time. I can customize the config file - i.e. I typically use FS storage instead of mem - and then just do `go run .`
Here we set the default value for the config file path:
https://github.com/gomods/athens/blob/23b285d49d2760e6b627b3ec074a45740a860568/cmd/proxy/main.go#L20
I guess typing `go run . -config_file ../../config.dev.toml` instead of just `go run .` is a small inconvenience so maybe you're right... I agree it's safer and more explicit that way.
Maybe I'll just add the `config_file` flag to the `run` target in our Makefile?

@manugupt1 :+1: for the env check

pkg/config/config.go
```diff
@@ -41,6 +42,37 @@ type Config struct {
 	Storage          *StorageConfig
 }
 
+// Load loads the config from a file.
+// If file is not present returns default config
+func Load(configFile string) (*Config, error) {
+	_, err := os.Stat(configFile)
+	if err == nil {
+		conf, err := ParseConfigFile(configFile)
```
robjloranger: if you skip this first if/else you could just return the call to `ParseConfigFile` on 56.

not a deal breaker, I just try to avoid nesting the 'happy path' in if statements.

robjloranger: I _think_ `ParseConfigFile` would already return an error if the file is not found when trying to decode.

pkg/config/config.go
marwan-at-work: @marpio good point, i'm always on the side of convenience 😄 -- you can either put `config_file` in the Makefile or just create your own alias such as `alias goathens='go run . -config_file=<path>`  which is something i often do as well

manugupt1: @marpio In that case, should we not have default settings close to production deployment rather than a dev deployment?

marpio: @manugupt1 it is not easy to tell what are the sensible defaults for production, in my opinion. i.e. which storage is the right one? maybe FS but then what should be the default directory?

manugupt1: @marpio Sorry for the late reply, Personally, I am a huge fan of doing magic with configuration variables within the code. It is easy for any human to miss it and make an error and affect systems that were not intended to.

As to Warning messages are useful only when people read logs, logs are read only when something fails unfortunately in the real world. The same goes for docs. 

I had rather have fail early, fail fast in critical path than introduce personal setups. Of course, when things fail, I would like to give them very helpful messages. 

It is just me TBF, so I would definitely like a second opinion.

marpio: @manugupt1 are you suggesting we should fail if the config file is not provided?

manugupt1: @marpio Well, if the right configuration are not there to run it minimally, then I would suggest to fail rather than suggesting our own internally through code.

`config.dev.toml` actually makes clear that it is to be used in dev environment, so I am not bothered by that.

What do you think?

marpio: @manugupt1 not sure I understand what you're proposing. Are you saying we should just fail if the config file isn't there?

michalpristas: agree with @manugupt1, defaults may make you feel that everything is ok after you deploy to your prod cluster but the opposite is true. i would rather fail fast

marpio: @michalpristas then we should reconsider #1008 b/c providing defaults the whole point of this issue.

pkg/config/config.go
```diff
@@ -41,6 +42,32 @@ type Config struct {
...
+func Load(configFile string) (*Config, error) {
+	if configFile == "" {
+		log.Print("config file not provided - using default settings")
+		return createDefault(), nil
+	}
+	return ParseConfigFile(configFile)
+}
+
+func createDefault() *Config {
+	log.Print("def conf")
```
arschles: @marpio can you remove this log line?

marpio: @arschles done!

Issue #1023, opened 2019-01-04
Minio saver: wrong path for module  .info


Zyqsempai:
**Describe the bug**
Minio saver sets extra version for a module `.info`, module path already contains version
Module path: `github.com/gomods/testCatalogModule/v1.2.0014`
Extra version: `v1.2.0014`
Result: `github.com/gomods/testCatalogModule/v1.2.0009/v1.2.0009.info`
https://github.com/gomods/athens/blob/master/pkg/storage/minio/saver.go#L20

I can solve it in the scope of https://github.com/gomods/athens/issues/996

Zyqsempai:
@fedepaol Can you check this one?

manugupt1:
@Zyqsempai I looked into it a bit. Seems like it can be done in its own PR. I would prefer to do it that way unless absolutely necessary.

Let me know what do you think.

fedepaol:
Agree with that, unless it blocks the minio catalog.
In any case it could be fixed & merged before the other

Zyqsempai:
Agree, opened PR to fix it. Better if we will merge it asap, cause catalog method implementation, related to this issue.

arschles:
@Zyqsempai I wanted to let you know on this that we're going to hold off on it until after we release [v0.3.0](https://github.com/gomods/athens/projects/2), because it has a potential breaking change. We haven't forgotten about it though!

Zyqsempai:
@arschles Ok, thanks!

marpio:
@Zyqsempai I'm closing this since the change would require a breaking change.

PR #1024, opened 2019-01-04
Minio saver, lister and getter fix module path format

Zyqsempai:

Minio saver, lister and getter fix module path format from `moduleName/version/version.info`
 to `moduleName/version.info`

Fixes https://github.com/gomods/athens/issues/1023


Zyqsempai:
cc @fedepaol @manugupt1

manugupt1:
LGTM so far, can you add tests please to make sure this works as intended?

Zyqsempai:
@manugupt1 Mmm, it's a little bit hard to do, and also test is already there in some way.
Its hard to do because we don't have method which will return for us full module path, to check it with original one. And it's kind of implemented implicitly, here https://github.com/gomods/athens/blob/master/pkg/storage/compliance/tests.go#L97 cause it will not found this module if saved path was incorrect or not equal to our pattern, and we will get error and test fail then.

Zyqsempai:
Yep, we can try to do it only for minio package, but still, we don't provide any api or method to fetch full module path.

robjloranger:
@Zyqsempai are you talking about something like `PackageVersionName` from https://github.com/gomods/athens/blob/0d94af25034eae47f4408a6830a5cccfb78c74b1/pkg/config/module.go#L9-L13

pkg/storage/minio/saver.go
```diff
@@ -17,7 +17,7 @@ func (s *storageImpl) Save(ctx context.Context, module, vsn string, mod []byte,
 	dir := s.versionLocation(module, vsn)
 	modFileName := dir + "/" + "go.mod"
 	zipFileName := dir + "/" + "source.zip"
-	infoFileName := dir + "/" + vsn + ".info"
+	infoFileName := dir + ".info"
```
marwan-at-work: This is incorrect because it's missing a `/` which will end up not nesting the info bytes inside the `dir` -- it should at least be like the above such as `dir + "/" + "source.info"` -- Which brings up the question, what's wrong with `dir + "/" + vsn + ".info"`? Besides consistency that is.

marwan-at-work: This is especially incorrect, because in the near future `storage.Checker` will need to check the length of an object to equal 3

arschles:
@Zyqsempai would you be able to come back and work on this again? Now that the `v0.3.1` release is out, we can get going with this again 😁

Zyqsempai:
Yep, i am still here, sorry for delay;)

arschles:
Not a problem @Zyqsempai - thanks for sticking with this PR 😁 

Can you resolve the conflicts in here, and look at @marwan-at-work's comments at https://github.com/gomods/athens/pull/1024#discussion_r245858755?

manugupt1:
Hi @Zyqsempai It would be a breaking change for athens minio users and as of now compliance tests are passing, would you be fine closing this PR.

Zyqsempai:
@manugupt1 Sure, no problem.

PR #1025, opened 2019-01-04
mongo: use explicit URL for testing

marwan-at-work:
This seems like reverting a regression because our storage tests are set up such that they "skip by default, unless explicitly set by variable". 

Let me know if I'm missing something.

### DIFF
--- a/pkg/storage/mongo/mongo_test.go
+++ b/pkg/storage/mongo/mongo_test.go
@@ -1,16 +1,12 @@
 package mongo
 
 import (
-	"path/filepath"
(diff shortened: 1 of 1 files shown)


robjloranger:
So the idea here is to have mongo tests skipped by default, unless the developer sets the variable and runs a mongo instance?

marwan-at-work:
@robjloranger yes, and if you look at all the other storages that require a spun-up server, that's how they work as well :)

robjloranger:
just double checking

Issue #996, opened 2018-12-13
Add catalog method to minio storage 


fedepaol:
**Is your feature request related to a problem? Please describe.**
With #955 a new catalog endpoint has been introduced.
#961 was opened to add missing storages and mongo is on its way to be merged.

This issue is for implementing the cataloger interface in minio.
It should be pretty straightforward since s3 is already implemented.

**Describe the solution you'd like**
Implement the cataloger in minio.

Zyqsempai:
I can take it;)

fedepaol:
Great. You could give a look at the S3 implementation to take inspiration
from.


Il giorno sab 29 dic 2018, 19:21 Boris Popovschi <notifications@github.com>
ha scritto:

Zyqsempai:
@fedepaol I faced few difficulties on my way to implement Catalog method:
First, one I mentioned above in new issue, the second is that fact that minioClient.ListObjectsV2 and ListObjects, don't want to continue list from specific key(module version), so we easily can fetch 1001 item but we can't continue from last module and fetch for example next 5.

fedepaol:
Not sure I understood.
Can you provide a failing test?

fedepaol:
Sorry for the late reply and for the dismissive reply I gave before.
I gave a look at the minio client. Even if the docs describes only the Client struct with its method, looking at the sources there is also this Core struct that exposes lower level methods such one similar to the one we are using against S3. I'd suggest you to try to use [Core.ListObjectsV2](https://github.com/minio/minio-go/blob/master/core.go#L54)

Zyqsempai:
Thanks, I will try it and get with feedback.

Zyqsempai:
@fedepaol So, i opened PR, but it still has one big problem, performance.
Honestly, have no clue why every minio function works so slow.
274.693178ms
368.370131ms
346.225552ms

Here is a few examples, this is minioCore.ListObjectsV2 function, but minioClient.ListObjectsV2 doesn't show the better results.

manugupt1:
Does ListObjectV2 download entire object, if so that can explain the slowness?
Which storage did you use against minio to take these benchmarks?

Zyqsempai:
@manugupt1 Sorry, i didn't make it clear, those benchmarks above are from simple unit test.
That's why i am so frustrated.
Maybe i am missing something.

Zyqsempai:
Yeah, but other storages handle 1000+ without any trouble, fast enough.

PR #1027, opened 2019-01-04
catalog: clean up unused interfaces

marwan-at-work:
1. Removed `Catalog` method from the download.Protocol interface (and its implementations) 
2. Removed `Catalog` from `storage.Backend` to make it optional
3. Moved `RegisterHandler` from the `storage` package to be a simple endpoint in the `actions` package. This is because our `storage` package should not have to deal with anything http related.
4. Cleaned up tests.


cmd/proxy/actions/app_proxy.go
```diff
@@ -65,7 +66,5 @@ func addProxyRoutes(
 	handlerOpts := &download.HandlerOpts{Protocol: dp, Logger: l}
 	download.RegisterHandlers(r, handlerOpts)
 
-	storage.RegisterHandlers(r, s)
```
manugupt1: This is inline with download protocol, would not the same apply download protocol as well.

I would still namespace into storage actions as an empty struct{} type.

cmd/proxy/actions/catalog.go
```diff
@@ -18,16 +19,15 @@ type catalogRes struct {
...
-func RegisterHandlers(r *mux.Router, s Backend) {
-	r.Handle("/catalog", CatalogHandler(s)).Methods(http.MethodGet)
-}
-
-// CatalogHandler implements GET baseURL/catalog
-func CatalogHandler(s Backend) http.Handler {
+// catalogHandler implements GET baseURL/catalog
+func catalogHandler(s storage.Backend) http.HandlerFunc {
 	const op errors.Op = "download.CatalogHandler"
+	cs, isCataloger := s.(storage.Cataloger)
```
manugupt1: I am not a huge fan of type assertion and finding out if a type has been implemented or not. When we had the conversation, I thought it was already implemented it this way.

Is this how go recommends doing it? If so, can you show me, I would like to know more.

pkg/storage/compliance/tests.go
```diff
@@ -140,19 +142,13 @@ func testCatalog(t *testing.T, b storage.Backend) {
 		ver := fmt.Sprintf("v1.2.%04d", i)
 		b.Save(ctx, modname, ver, mock.Mod, bytes.NewReader(zipBts), mock.Info)
 	}
-	defer func() {
```
manugupt1: Is there a reason to delete this function?

cmd/proxy/actions/catalog.go
marwan-at-work: @manugupt1 interface-assertion is actually a good pattern if used correctly, as an example look at https://golang.org/pkg/net/http/#Flusher (in other words, the main http.ResponseWriter interface *does* have the Flush method, but only optionally, so you use interface-assertion to find out)

pkg/storage/compliance/tests.go
marwan-at-work: @manugupt1 yes, because we delete everything above. Look at `clearFunc()`

cmd/proxy/actions/catalog.go
manugupt1: yeah! the issue that I have right now is that I am not sure which way is better, both seem very equivalent to me. 

Also, if you think about it, we have already implemented these functions in the storage: 
https://github.com/fedepaol/athens/commit/0258e17d8905fa851c78f8e8ed2e73358135f816#diff-b646c51b29c85b8406f5792043c69de7R12

and isCataloger will never be false (with the code that we have written so far); which makes it functionally equivalent for the user but still uses the underlying interfaces.

Another thing, is error types, right now we have ServerError and NotImplemented. I think there are storages that won't be able to support catalog which should have a NotSupported error. See here: https://github.com/gomods/athens/issues/996

With this implementation, in my mind, we are increasing the probablity that a contributor is not able to identify the implementation is missing and misses opportunity to contribute to  our repo.

marwan-at-work: @manugupt1 

>Also, if you think about it, we have already implemented these functions in the storage:
I have removed those from this PR, therefore they'll come back as false. That's the whole point, we don't want to provide non-implemented implementations :) 

>I think there are storages that won't be able to support catalog
is exactly the point of this PR. 

>With this implementation, in my mind, we are increasing the probability that a contributor is not able to identify the implementation is missing and misses opportunity to contribute to our repo.

We want people to have an easier time to contribute, so the smaller the interface the quicker they can PR something. The Catalog method is complex due to pagination and backend implementation, and so it should definitely be optional. 

Lastly `NotImplemented` and `NotSupported` are too similar, no need to be that verbose, we can revisit this later on.

manugupt1: azureblob.Cataloger?

cmd/proxy/actions/app_proxy.go
manugupt1:
```
type storage struct{}
func (s * storage) RegisterHandlers(r, s)
```

Which makes it clear that the download protocol is not even involved. but this is a minor change that I won't worry about it as much.

cmd/proxy/actions/catalog.go
```diff
@@ -18,16 +19,15 @@ type catalogRes struct {
...
-// all the storage routes.
-func RegisterHandlers(r *mux.Router, s Backend) {
-	r.Handle("/catalog", CatalogHandler(s)).Methods(http.MethodGet)
-}
-
-// CatalogHandler implements GET baseURL/catalog
-func CatalogHandler(s Backend) http.Handler {
+// catalogHandler implements GET baseURL/catalog
+func catalogHandler(s storage.Backend) http.HandlerFunc {
 	const op errors.Op = "download.CatalogHandler"
```
manugupt1: Let us also rename the op while we are at it.

cmd/proxy/actions/catalog.go
manugupt1: I missed azureblob, LGTM

manugupt1:
Can you please rename the op, that is one change I think we should absolutely do. Other than that, I will approve it and let you merge when ready.

PR #1033, opened 2019-01-06
WIP: Fix exists

### LINKED ISSUE #1032: Exists checks only for .mod
**Describe the bug**

marpio:
This is trying to solve #1032 by checking all mod, info and zip parts. If one or more of them are missing it reports false.
Tests are missing and also every storage needs to use it but please review this if you get to it.

will fix part of #1032

### DIFF
--- a/pkg/storage/module/check.go
+++ b/pkg/storage/module/check.go
@@ -1,0 +1,72 @@
+package module
+
+import (
+	"context"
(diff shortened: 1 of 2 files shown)


marpio:
Closing. This can be fixed by simply checking the count of mod@ver elements (go.mod, .info, .zip).

PR #1036, opened 2019-01-08
cmd/proxy/actions: Add basic unit-tests for some proxy routes.

PascalDierich:
There are no tests for proxy routes yet (#138).

This PR adds unit-tests for:
* `checkAuth` in basicauth.go
* `healthHandler` in health.go
* `proxyHomeHandler` in home.go

The tests are pretty basic, but maybe it's a start to get more test-coverage :)


cmd/proxy/actions/health_test.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestHealthHandler(t *testing.T) {
```
manugupt1: I am not sure what benefit does this test gives us? It is a unit test that does not check if the system is up or not. I would not add this one

cmd/proxy/actions/home_test.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestProxyHomeHandler(t *testing.T) {
```
manugupt1: The same for this test also, the benefit from this test is not a lot imo. If you can convince me otherwise, that will be sweet.

cmd/proxy/actions/basicauth_test.go
```diff
@@ -0,0 +1,57 @@
...
+		name:     "password incorrect",
+		user:     "athens",
+		pass:     "1234",
+		httpUser: "athens",
+		httpPass: "not 1234",
+		want:     false,
+	},
+}
+
+func TestCheckAuth(t *testing.T) {
```
manugupt1: I like authentication tests regardless of how simple they are. I would like another maintainer to have a look at it though.

cmd/proxy/actions/home_test.go
marwan-at-work: Agree with @manugupt1 can we remove this test and keep the basic auth one. I'll merge afterwards. Thanks @PascalDierich

PascalDierich: Removed the handler tests. @marwan-at-work @manugupt1 thanks for your input 👍

PR #1041, opened 2019-01-13
Use ambient aws credential provider for s3

joshuarubin:

Athens does not pick up aws "ambient" role credentials when run in ec2, ecs or fargate containers.

**How is the fix applied?**

A new variable `UseAmbientCredentials` (or `$AWS_USE_AMBIENT_CREDENTIALS`) is added. If `false`, and `Key` and `Secret` are defined then the aws config is built as it was previously. If `true` then just a default `aws.Config` is used with Region provided. This will use ec2/ecs credentials (in addition to using the `~/.aws/credentials` that were previously used).

The reason the new variable is required is because ec2/ecs systems will also define `$AWS_ACCESS_KEY_ID` and `$AWS_SECRET_ACCESS_KEY`.

### DIFF
--- a/pkg/config/s3.go
+++ b/pkg/config/s3.go
@@ -2,10 +2,10 @@ package config
 
 // S3Config specifies the properties required to use S3 as the storage backend
 type S3Config struct {
-	Region                string `validate:"required" envconfig:"AWS_REGION"`
-	Key                   string `envconfig:"AWS_ACCESS_KEY_ID"`
-	Secret                string `envconfig:"AWS_SECRET_ACCESS_KEY"`
-	Token                 string `envconfig:"AWS_SESSION_TOKEN"`
-	Bucket                string `validate:"required" envconfig:"ATHENS_S3_BUCKET_NAME"`
-	UseAmbientCredentials bool   `envconfig:"AWS_USE_AMBIENT_CREDENTIALS"`
+	Region                  string `validate:"required" envconfig:"AWS_REGION"`
+	Key                     string `envconfig:"AWS_ACCESS_KEY_ID"`
+	Secret                  string `envconfig:"AWS_SECRET_ACCESS_KEY"`
+	Token                   string `envconfig:"AWS_SESSION_TOKEN"`
+	Bucket                  string `validate:"required" envconfig:"ATHENS_S3_BUCKET_NAME"`
+	UseDefaultConfiguration bool   `envconfig:"AWS_USE_DEFAULT_CONFIGURATION"`
 }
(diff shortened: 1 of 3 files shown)


pkg/config/s3.go
```diff
@@ -2,9 +2,10 @@ package config
...
-	Key    string `envconfig:"AWS_ACCESS_KEY_ID"`
-	Secret string `envconfig:"AWS_SECRET_ACCESS_KEY"`
-	Token  string `envconfig:"AWS_SESSION_TOKEN"`
-	Bucket string `validate:"required" envconfig:"ATHENS_S3_BUCKET_NAME"`
+	Region                string `validate:"required" envconfig:"AWS_REGION"`
+	Key                   string `envconfig:"AWS_ACCESS_KEY_ID"`
+	Secret                string `envconfig:"AWS_SECRET_ACCESS_KEY"`
+	Token                 string `envconfig:"AWS_SESSION_TOKEN"`
+	Bucket                string `validate:"required" envconfig:"ATHENS_S3_BUCKET_NAME"`
+	UseAmbientCredentials bool   `envconfig:"AWS_USE_AMBIENT_CREDENTIALS"`
```
manugupt1: Can we use a better name?
Also, can we add it to config.dev.toml so that users know that this one exists?
Let us make sure that we add documentation for that. Right from the docs, I think this was the best  description, I got: 
```
session Provides initial default configuration, and load
      configuration from external sources such as environment and shared
      credentials file.
```
but feel free to improvize

pkg/storage/s3/s3.go
```diff
@@ -73,18 +68,10 @@ func New(s3Conf *config.S3Config, timeout time.Duration, options ...func(*aws.Co
 
 // buildAWSCredentials builds the credentials required to create a new AWS session.  It will prefer the access key ID and
```
manugupt1: Can we update the documentation here as well?

pkg/config/s3.go
joshuarubin: Yes, I can update the naming. What would you call it?

joshuarubin:
There's better documentation about how things work here: https://godoc.org/github.com/aws/aws-sdk-go/aws/session

pkg/config/s3.go
joshuarubin: `AWS_USE_DEFAULT_CONFIGURATION`?

manugupt1:
I like this naming better. Thank you!

Cheers
Manu

________________________________

joshuarubin:
OK @manugupt1 I think I've addressed your concerns.

Issue #998, opened 2018-12-13
Official MongoDB Go Driver


chriscoffee:
The official MongoDB Go driver just hit beta. Looking forward we could look to replace the current globalsign/mgo package we use.

https://github.com/mongodb/mongo-go-driver/releases/tag/v0.1.0

arpitgogia:
Hi @chriscoffee I would like to take up on the `good-first-issue` badge and volunteer for this task 😅

marpio:
Awesome @arpitgogia ! It's yours!

robstrong:
Can this issue be closed now that #1042 is merged?

marpio:
Yes. Thank for the reminder.

PR #1042, opened 2019-01-14
Mongo go driver beta integration

arpitgogia:

Replacing the mgo Mongo Driver with the official beta release 

**How is the fix applied?**

Changed imports and method calls to correspond to the new API

Addresses #998


arpitgogia:
I have changed quite a few parts. Had two doubts while doing this:
* The new driver has both client and session paradigms. Currently, I've based everything on the client paradigm [pkg/storage/mongo/mongo.go](https://github.com/gomods/athens/pull/1042/files#diff-511dd4cece028a268d387fa7b7c36c98R57)
* I'm yet to integrate the GridFS API.

pkg/storage/mongo/cataloger.go
```diff
@@ -20,22 +20,46 @@ func (s *ModuleStore) Catalog(ctx context.Context, token string, pageSize int) (
 
 	fields := bson.M{"module": 1, "version": 1}
 
-	c := s.s.DB(s.d).C(s.c)
+	compositeQ := bson.D{q, fields}
+
+	c := s.s.Database(s.d).Collection(s.c)
+
+	documentCount, erro := c.Count(compositeQ)
```
manugupt1: I am not sure if this required or not. If you have a more than 0 count, then you end up making two calls to the MongoDB thus degrading performance.

pkg/storage/mongo/mongo.go
```diff
@@ -54,59 +54,62 @@ func (m *ModuleStore) connect() error {
...
+
+	// 	tlsConfig.ClientCAs = roots
+
+	// 	dialInfo.DialServer = func(addr *mgo.ServerAddr) (net.Conn, error) {
+	// 		return tls.Dial("tcp", addr.String(), tlsConfig)
+	// 	}
+	// }
+
+	// return mgo.DialWithInfo(dialInfo)
+	return &Client.NewClient(m.url)
```
manugupt1: This has changes related to AutoCert and TLS? How does your changes reflect that as we do not have tests related to that.

manugupt1:
Nice work! Left a few comments, but it is definitely good progress.

michalpristas:
Nice work in this PR. But I would postpone merging until stable release is out.

pkg/storage/mongo/mongo.go
```diff
@@ -9,14 +9,14 @@ import (
...
-	"github.com/globalsign/mgo"
 	"github.com/gomods/athens/pkg/config"
 	"github.com/gomods/athens/pkg/errors"
+	"github.com/mongodb/mongo-go-driver/mongo"
 )
 
 // ModuleStore represents a mongo backed storage backend.
 type ModuleStore struct {
-	s        *mgo.Session
+	s        *mongo.NewClient
```
arschles: @arpitgogia would you mind renaming this struct field name to something more explicit?

arschles:
@arpitgogia I am going to mark this with `do not merge` until the mongo driver is out of beta and this PR has been updated with it

arschles:
@arpitgogia we're still watching the mongo driver to see when it goes out of beta. we haven't forgotten about this :)

arpitgogia:
No worries @arschles 😃

marpio:
@arschles @michalpristas @arpitgogia 
The `v1.0.0-rc2` has been just released https://github.com/mongodb/mongo-go-driver/releases/tag/v1.0.0-rc2
Do you think we can finish this PR?

arschles:
@marpio yes, I think we can merge this! @arpitgogia would you mind resolving the conflicts so we can have CI tests run again and then we can merge?

arpitgogia:
Will do @arschles!

pkg/storage/mongo/cataloger.go
arpitgogia: Yeah I wasn't sure myself too. Somehow I thought it wouldn't matter much if the fields are indexed, do an early exit. Will use the length of `modules` array only.

pkg/storage/mongo/mongo.go
arpitgogia: At the time I wrote this the driver didn't have any authentication mechanism. 
Will check if there's something new in RC2

arschles:
@arpitgogia sounds good. just `@` me in here or `@` on the gophers slack when ready 😁

arpitgogia:
@arschles I have adapted to the RC2 driver. 
* One error that I can't seem to get rid of is 
```
../../pkg/storage/mongo/mongo.go:70:31: unknown field 'coll' in struct literal of type mongo.IndexView
``` 
`IndexView` does have `coll` as a field according to the [source code](https://github.com/mongodb/mongo-go-driver/blob/v1.0.0-rc2/mongo/index_view.go#L36)

* `make test-unit` also fails but I couldn't understand the exact reason it is failing.

pkg/storage/mongo/cataloger.go
```diff
@@ -3,34 +3,39 @@ package mongo
...
-		All(&modules)
+	cursor, err := c.Find(context.Background(), q, &options.FindOptions{Projection: projection})
 
 	if err != nil {
 		return nil, "", errors.E(op, err)
 	}
+	for cursor.Next(context.Background()) {
+		var module storage.Module
+		elem := &bson.D{}
+		if err := cursor.Decode(module); err == nil {
```
marpio: I don't think we should ignore errors. It's better to fail than to return incomplete data.
Could you use `multierror` like here https://github.com/gomods/athens/blob/d29efb7d61c34b988fe4428772fd8089bc6e1f88/pkg/storage/minio/saver.go#L104 ?

pkg/storage/mongo/cataloger.go
```diff
@@ -3,34 +3,39 @@ package mongo
...
+	c := s.client.Database(s.db).Collection(s.coll)
 
-	c := s.s.DB(s.d).C(s.c)
 	modules := make([]storage.Module, 0)
-	err := c.Find(q).
-		Select(fields).
-		Sort("_id").
-		Limit(pageSize).
-		All(&modules)
+	cursor, err := c.Find(context.Background(), q, &options.FindOptions{Projection: projection})
```
marpio: Can we use `context.WithTimeout` here? 
Like 
```
tctx, cancel := context.WithTimeout(ctx, s.timeout) 
cursor, err := c.Find(tctx, q, &options.FindOptions{Projection: projection})
```
Can you also check other places where context.Background() is used?

pkg/storage/mongo/deleter.go
```diff
@@ -21,13 +23,17 @@ func (s *ModuleStore) Delete(ctx context.Context, module, version string) error
...
+	bucket, err := gridfs.NewBucket(db, &options.BucketOptions{})
 	if err != nil {
 		return errors.E(op, err, errors.M(module), errors.V(version))
 	}
-	err = c.Remove(bson.M{"module": module, "version": version})
+	err = bucket.Delete(s.gridFileName(module, version))
+	if err != nil {
+		return errors.E(op, err, errors.M(module), errors.V(version))
+	}
+	result, err := c.DeleteOne(context.Background(), bson.M{"module": module, "version": version})
```
marpio: Can we use context.WithTimeout here instead?

pkg/storage/mongo/getter.go
```diff
@@ -3,22 +3,25 @@ package mongo
...
 	ctx, span := observ.StartSpan(ctx, op.String())
 	defer span.End()
-	c := s.s.DB(s.d).C(s.c)
+	c := s.client.Database(s.db).Collection(s.coll)
 	result := &storage.Module{}
-	err := c.Find(bson.M{"module": module, "version": vsn}).One(result)
+	err := c.FindOne(context.Background(), bson.M{"module": module, "version": vsn}).Decode(result)
 	if err != nil {
 		kind := errors.KindUnexpected
 		if err == mgo.ErrNotFound {
```
marpio: Will this work? We should not use `github.com/globalsign/mgo` at all, no?
Does mongo-drive define it's own errors? It would be good to know if it was found or not.

pkg/storage/mongo/getter.go
```diff
@@ -35,9 +38,9 @@ func (s *ModuleStore) GoMod(ctx context.Context, module, vsn string) ([]byte, er
...
```
marpio: As above - we probably should not use github.com/globalsign/mgo

pkg/storage/mongo/getter.go
```diff
@@ -56,15 +59,17 @@ func (s *ModuleStore) Zip(ctx context.Context, module, vsn string) (io.ReadClose
...
-		kind := errors.KindUnexpected
-		if err == mgo.ErrNotFound {
-			kind = errors.KindNotFound
-		}
-		return nil, errors.E(op, err, kind, errors.M(module), errors.V(vsn))
+		return nil, errors.E(op, err, errors.M(module), errors.V(vsn))
+	}
+
+	dStream, err := bucket.OpenDownloadStreamByName(zipName, options.GridFSName())
+	defer dStream.Close()
```
marpio: We should not close the reader here otherwise we would try to read from a closed reader later on.

pkg/storage/mongo/getter.go
```diff
@@ -56,15 +59,17 @@ func (s *ModuleStore) Zip(ctx context.Context, module, vsn string) (io.ReadClose
...
+	if err != nil {
```
marpio: Can we somehow say if the mod@ver wasn't found - check for ErrNotFound?

pkg/storage/mongo/getter.go
```diff
@@ -56,15 +59,17 @@ func (s *ModuleStore) Zip(ctx context.Context, module, vsn string) (io.ReadClose
...
+		return nil, errors.E(op, err, errors.M(module), errors.V(vsn))
 	}
 
-	return f, nil
+	return ioutil.NopCloser(dStream), nil
```
marpio: Why do we need `ioutil.NopCloser`?

pkg/storage/mongo/lister.go
```diff
@@ -4,28 +4,37 @@ import (
...
 		if err == mgo.ErrNotFound {
 			kind = errors.KindNotFound
 		}
 		return nil, errors.E(op, kind, errors.M(module), err)
 	}
+	result := make([]storage.Module, 0)
+	for cursor.Next(context.Background()) {
+		var module storage.Module
+		err := cursor.Decode(&module)
+		if err == nil {
```
marpio: As above - we should not ignore errors. If we do List could provide different results at t1 and t2.

pkg/storage/mongo/mongo.go
```diff
@@ -27,69 +28,70 @@ type ModuleStore struct {
...
 	return ms, nil
 }
 
-func (m *ModuleStore) connect(conf *config.MongoConfig) error {
+func (m *ModuleStore) connect(conf *config.MongoConfig) (*mongo.Collection, error) {
 	const op errors.Op = "mongo.connect"
 
 	var err error
-	m.s, err = m.newSession(m.timeout, m.insecure, conf)
+	err = m.client.Connect(context.Background())
```
marpio: Can we use context with timeout here as well?

pkg/storage/mongo/saver.go
```diff
@@ -25,14 +28,24 @@ func (s *ModuleStore) Save(ctx context.Context, module, version string, mod []by
...
-	defer f.Close()
 
-	numBytesWritten, err := io.Copy(f, zip)
+	uStream, err := bucket.OpenUploadStream(zipName, options.GridFSUpload())
+	if err != nil {
+		return errors.E(op, err, errors.M(module), errors.V(version))
+	}
+	defer uStream.Close()
+
+	zipBytes, err := ioutil.ReadAll(zip)
```
marpio: Zip can be big so we don't want to load everything into memory. Can you use `io.Copy`?

marpio:
@arpitgogia 
Also regarding the error you mentioned `unknown field 'coll' in struct literal of type mongo.`
It seems that coll is not exported:
https://github.com/mongodb/mongo-go-driver/blob/82905e3077cb5ce7f194748a47b55bf7fff65ccf/mongo/index_view.go#L37

pkg/storage/mongo/saver.go
```diff
@@ -25,14 +28,24 @@ func (s *ModuleStore) Save(ctx context.Context, module, version string, mod []by
...
+	if err != nil {
+		return errors.E(op, err, errors.M(module), errors.V(version))
+	}
+	numBytesWritten, err := uStream.Write(zipBytes)
```
marpio: Can you use io.Copy?

pkg/storage/mongo/cataloger.go
arpitgogia: The only two places where `context.Background()` remains are lines [78](https://github.com/arpitgogia/athens/blob/mongo-go-beta-integration/pkg/storage/mongo/mongo.go#L78) and [57](https://github.com/arpitgogia/athens/blob/mongo-go-beta-integration/pkg/storage/mongo/mongo.go#L57) in mongo.go

arpitgogia:
@marpio thanks for the comments
I've made some changes based on your review

marpio:
@arpitgogia Thanks for the changes!
Could you check the build error?

arpitgogia:
@marpio is there some issue with `check_gofmt.sh`?

pkg/storage/mongo/lister.go
```diff
@@ -3,28 +3,43 @@ package mongo
...
+
+	cursor, err := c.Find(tctx, query, &options.FindOptions{Projection: projection})
 	if err != nil {
-		kind := errors.KindUnexpected
-		if err == mgo.ErrNotFound {
-			kind = errors.KindNotFound
+		return nil, errors.E(op, err, errors.M(module))
+	}
+	result := make([]storage.Module, 0)
+	var errs error;
```
marpio: no need for `;`

pkg/storage/mongo/lister.go
```diff
@@ -3,28 +3,43 @@ package mongo
...
+	for cursor.Next(ctx) {
+		var module storage.Module
+		if err := cursor.Decode(module); err != nil {
+			kind := errors.KindUnexpected
+			if err == mongo.ErrNoDocuments {
+				kind = errors.KindNotFound
+			}
+			errs = multierror.Append(errs, errors.E(op, kind))
```
marpio: Please add the err -> `errs = multierror.Append(errs, errors.E(op, err, kind))`

marpio:
> @marpio is there some issue with `check_gofmt.sh`?

don't think so... Could you run it on your machine?

arpitgogia:
@marpio I got the script to work. Turns out the script throws an error if any of the files being inspected don't match gofmt's expected formatting. For now I've changed it to show the diff between current and expected states, hope that's more verbose 😅

arschles:
@arpitgogia @chriscoffee said that https://github.com/mongodb/mongo-go-driver is now [stable](https://github.com/mongodb/mongo-go-driver/releases/tag/v1.0.0). We're gonna review it once more and aim to merge it now 🎉 🚀 :shipit:

pkg/storage/mongo/mongo.go
```diff
@@ -27,69 +28,70 @@ type ModuleStore struct {
...
-	index := mgo.Index{
-		Key:        []string{"base_url", "module", "version"},
-		Unique:     true,
-		Background: true,
-		Sparse:     true,
-	}
-	c := m.s.DB(m.d).C(m.c)
-	return c.EnsureIndex(index)
+func (m *ModuleStore) initDatabase() *mongo.Collection {
+	// TODO: database and collection as env vars, or params to New()? together with user/mongo
```
arschles: I think it's ok to leave this TODO out. We've hard-coded the db and collection names since the first version and nobody has asked to change them so far, so we can wait and see if someone does...

pkg/storage/mongo/getter.go
```diff
@@ -4,27 +4,27 @@ import (
...
 	ctx, span := observ.StartSpan(ctx, op.String())
 	defer span.End()
-	c := s.s.DB(s.d).C(s.c)
+	c := s.client.Database(s.db).Collection(s.coll)
 	result := &storage.Module{}
-	err := c.Find(bson.M{"module": module, "version": vsn}).One(result)
+	tctx, cancel := context.WithTimeout(ctx, s.timeout)
+	defer cancel()
+
+	err := c.FindOne(tctx, bson.M{"module": module, "version": vsn}).Decode(result)
```
marpio: @arpitgogia 
Can you split the calls in two and check for error in each?
Also, please check for `ErrNoDocuments` and if present, return error with `errors.KindUnexpected`
This should also fix the CI error.

marpio:
@arpitgogia thanks for the changes! We're ready to go when the conflicts are resolved.

arpitgogia:
@marpio I think the CI is failing because of [some error with Mongo](https://cloud.drone.io/gomods/athens/277/1/2)

marpio:
@arpitgogia could you return KindNotFound on the ErrNoDocuments?
The ci is failing b/c of that I guess.

arpitgogia:
@marpio sorry, I might have missed it earlier 😅. 
I've changed it but I think it could be because of this [error Mongo is throwing](https://cloud.drone.io/gomods/athens/277/1/2) because the return code in the test is 500

arschles:
@arpitgogia I think there are two options on the face of it:

1. Configure the test mongo with an SSL cert
1. Connect to Mongo in tests without SSL

If it's possible in the code, (2) sounds much easier to me. Let me know what you think 😄

arpitgogia:
@arschles I feel both with and without SSL connections should be tested, because both of the scenarios are possible in real life. 
What I think I can do is, I'll add a flag (in `MongoConfig`) that allows connecting without SSL and use that flag in test for now. I've finally got the test running on my machine now, so I'll be pushing some minor fixes as well as soon as the tests pass.
In a follow up pull request, I'll setup tests for mongo connections with SSL. 

What do you think? 😅

arschles:
@arpitgogia answers below

>I feel both with and without SSL connections should be tested, because both of the scenarios are possible in real life.

I'd like to really discourage non-SSL connections in production

> What I think I can do is, I'll add a flag (in MongoConfig) that allows connecting without SSL and use that flag in test for now. I've finally got the test running on my machine now, so I'll be pushing some minor fixes as well as soon as the tests pass.

I saw you added the insecure flag to the config and added the insecure parameter. What do you think about removing the flag, but leaving the parameter as-is?

>In a follow up pull request, I'll setup tests for mongo connections with SSL.

Wonderful! I think at that point, it's safe to remove the tests for the non-SSL connections

pkg/storage/mongo/cataloger.go
```diff
@@ -3,35 +3,46 @@ package mongo
...
+	cursor, err := c.Find(tctx, q, &options.FindOptions{Projection: projection})
 
 	if err != nil {
 		return nil, "", errors.E(op, err)
 	}
 
+	var errs error
+	for cursor.Next(ctx) {
+		var module storage.Module
+		if err := cursor.Decode(module); err != nil {
```
marpio: Decode takes a pointer

pkg/storage/mongo/lister.go
```diff
@@ -3,28 +3,43 @@ package mongo
...
-		kind := errors.KindUnexpected
-		if err == mgo.ErrNotFound {
-			kind = errors.KindNotFound
+		return nil, errors.E(op, err, errors.M(module))
+	}
+	result := make([]storage.Module, 0)
+	var errs error
+	for cursor.Next(ctx) {
+		var module storage.Module
+		if err := cursor.Decode(module); err != nil {
```
marpio: Decode takes a pointer

pkg/storage/mongo/lister.go
```diff
@@ -3,28 +3,43 @@ package mongo
...
 func (s *ModuleStore) List(ctx context.Context, module string) ([]string, error) {
 	const op errors.Op = "mongo.List"
 	ctx, span := observ.StartSpan(ctx, op.String())
 	defer span.End()
-	c := s.s.DB(s.d).C(s.c)
-	fields := bson.M{"version": 1}
-	result := make([]storage.Module, 0)
-	err := c.Find(bson.M{"module": module}).Select(fields).All(&result)
+	c := s.client.Database(s.db).Collection(s.coll)
+	projection := bson.M{"version": 1}
```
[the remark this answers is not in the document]
arpitgogia: It's there in the current implementation too.
https://github.com/gomods/athens/blob/master/pkg/storage/mongo/lister.go#L21 so I thought to not change it

pkg/config/config.go
```diff
@@ -44,6 +44,7 @@ type Config struct {
 	SingleFlightType string `envconfig:"ATHENS_SINGLE_FLIGHT_TYPE"`
 	SingleFlight     *SingleFlight
 	Storage          *StorageConfig
+	Insecure         bool
```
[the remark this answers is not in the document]
arpitgogia: Removed and now using the one already defined in `MongoConfig`

arpitgogia:
@arschles @marpio the CI seems to be happy now 😃

pkg/storage/mongo/cataloger.go
```diff
@@ -37,6 +43,7 @@ func (s *ModuleStore) Catalog(ctx context.Context, token string, pageSize int) (
 	for cursor.Next(ctx) {
 		var module storage.Module
 		if err := cursor.Decode(&module); err != nil {
+			fmt.Println(module)
```
marpio: One last thing - can you remove the Print here?

arpitgogia: Sorry, stray statement from debugging 😅

marpio: wow you're fast :smile:

PR #1043, opened 2019-01-15
actions: improve basic auth tests

marwan-at-work:
Improved our basic auth middleware tests to make sure that on unauthorized /healthz requests, we're logging a warning to the user.


cmd/proxy/actions/basicauth_test.go
```diff
@@ -1,57 +1,80 @@
...
-		pass:     "1234",
-		httpUser: "athens",
-		httpPass: "1234",
-		want:     true,
+		name:   "happy_path",
+		user:   "correctUser",
+		pass:   "correctPass",
+		path:   "/",
+		logs:   "",
+		status: 404,
```
manugupt1: Let us change it to use the same endpoint and get a 200, for example `/healthz`. That is far better than a 404. Correct me if I am reading something wrong here @marwan-at-work

cmd/proxy/actions/basicauth_test.go
```diff
@@ -1,57 +1,80 @@
...
+		logs:   "",
+		status: 401,
+	},
+	{
+		name:   "log_on_healthz",
+		user:   "wrongUser",
+		pass:   "wrongPassword",
+		path:   "/healthz",
+		logs:   healthWarning,
+		status: 401,
```
michalpristas: can we name it expectedStatus or something similar which tells that we are checking against it and not actually passing in

PR #1052, opened 2019-02-09
pkg/config: Update GoGetWorkers default/docs

marwan-at-work:
Update the GoGetWorkers documentation to be clear about what it does under the hood, I also lowered the number to a more sane default...doing 30 git clones at the same time will more likely run you out of resources.


config.dev.toml
```diff
@@ -15,8 +15,11 @@ GoEnv = "development"
 # GoGetWorkers specifies how many times you can concurrently
 # go mod download, this is so that low performance instances
 # can manage go get more sanely and not run out of disk or memory.
+# One way to think about is: the number of GoGetWorkers corresponds
```
michalpristas: 👍

PR #1058, opened 2019-02-14
[WIP] Adding exists checks for all module files

arschles:

When storages check if a module exists, they only check the `.mod` file

**How is the fix applied?**

I made each storage check all 3 files.

Storage drivers to fix:

- [x] Azure blob
- [x] fs
- [x] gcp
- [x] memory
- [x] minio
- [x] mongo
- [x] s3

Other TODOs:

- [ ] `pkg/storage/module/check_test.go`

Ref https://github.com/gomods/athens/issues/1032, but does not fix because this PR doesn't handle fixing the missing case


arschles:
@marpio I was thinking of going in and adding checks for the rest of the storages before we merge this. Is that ok with you, or would you rather I do each storage in a separate PR?

marpio:
Oh, sure. I thought you want to add them one by one but I'm completely ok with doing all at once.

arschles:
@marpio I have made some progress and roughly implemented the checks in all the drivers. I still have a test to write and I think a test is failing still, but this is the general shape I'm proposing. can you take a quick look and let me know what you think?

marpio:
Thanks @arschle ! Looks good to me!

marpio:
@marwan-at-work @arschles 

_1. Accept this PR as is, and just expect users to have beefy machines._

I think we should strive to make athens work for everyone, on weaker machines as well so if this poses a problem then we should look into others solutions. Out of curiosity, what's your setup?

>     2. Close the issue as "won't fix" and just be optimistic that when a module is saved, all of its pieces are saved. Therefore, if Check returns `true` (.mod piece exists) but `.info` returns `NotFound`, the user will notice the error and probably fix it easily by deleting something manually from storage.

This is probably a viable solution if we assume that athens users have direct access to the storage which may not always be the case. Asking someone to delete the pieces can be cumbersome.

>  1. Instead of making 3 network requests to check if a module exists, we make 1 that checks for the length of a storage space. Hopefully all storages provide that option as it would be a big win. This can also mean a breaking change for how mods are stored so we should def think more about this.

This would be the cleanest solution to the problem IMO but, as you mentioned, it would be a breaking change since we're saving the versions like this: `module/@v/{version.info, version.zip, gomod}` so a simple `ListObjects` with prefix 'module/version` would not work. As you said this would need some more thinking and I don't know what is our stand on making a breaking change.

>  2. We can maybe just put the burden on the `Save` function to ensure that all modules are saved, and if one errors just make sure the `.mod` file is deleted so that checker returns false, and if it can't be deleted, then Athens did its best and the user should manually fix whatever is wrong with their storage.

If we don't want to make a breaking change then this would be my preferred solution.

>     3\. We can potentially remove the `Check` method from the interface all together and just rely on two things: `errors.KindNotFound` to indicate something was not found, and two: the order in which module pieces are fetched. For example, go always pings `info` first, then `mod`, then `zip`. Therefore, we only stash+serve on `info` but on `mod` and `zip` we would just stash and return an error (to avoid mismatched pieces). This solution is potentially the best, but also potentially the buggiest so it would require some looking into.

I have mixed feelings about this b/c it would make athens depend on `go cmd` implementation detail.

marpio:
@arschles Can we close this PR?

arschles:
@marpio yup! I'm closing now, sorry for the loooong delay

PR #1061, opened 2019-02-15
Update SHOULDERS.md

stanislas-m:

Ensure the SHOULDERS.md list is up to date with the current deps.

**How is the fix applied?**

This is the result of the following command: 
```
$ shoulders -w -n Athens
```


marpio:
Thanks @stanislas-m !

PR #1063, opened 2019-02-16
Use Go 1.12rc for Docker Image

marwan-at-work:
Now that the Release Candidate is out for 1.12, it's good that Canary users can test it out and so that they can catch any backwards-incompatible changes in Go Modules early on (i.e. the checksum mismatch issue)


cmd/proxy/Dockerfile
```diff
@@ -2,7 +2,7 @@
 # https://github.com/docker-library/buildpack-deps/blob/1845b3f918f69b4c97912b0d4d68a5658458e84f/stretch/scm/Dockerfile
 # https://github.com/golang/go/blob/f082dbfd4f23b0c95ee1de5c2b091dad2ff6d930/src/cmd/go/internal/get/vcs.go#L90
 
-FROM golang:1.11-alpine AS builder
+FROM golang:1.12rc1-alpine AS builder
```
manugupt1: Can you use either `rc` or `1.12-rc` as the tag so that we do not have to make multiple changes as the images get updated?

marwan-at-work:
@manugupt1 good suggestion, updated the pr.

PR #1069, opened 2019-02-18
Dockerfile: remove redundant call to apk update

fsouza:
- ``apk update`` fetches the index and saves it locally
- ``apk add --update`` fetches the index and installs the requested
packages

### DIFF
--- a/cmd/proxy/Dockerfile
+++ b/cmd/proxy/Dockerfile
@@ -20,8 +20,7 @@ COPY --from=builder /bin/athens-proxy /bin/athens-proxy
 COPY --from=builder /go/src/github.com/gomods/athens/config.dev.toml /config/config.toml
 COPY --from=builder /usr/local/go/bin/go /bin/go
 
-RUN apk update &&  \
-	apk add --no-cache bzr git mercurial openssh-client subversion procps fossil && \
+RUN apk add --update bzr git mercurial openssh-client subversion procps fossil && \
 	mkdir -p /usr/local/go
 
 ENV GO_ENV=production


cmd/proxy/Dockerfile
```diff
@@ -20,8 +20,7 @@ COPY --from=builder /bin/athens-proxy /bin/athens-proxy
 COPY --from=builder /go/src/github.com/gomods/athens/config.dev.toml /config/config.toml
 COPY --from=builder /usr/local/go/bin/go /bin/go
 
-RUN apk update &&  \
-	apk add --no-cache bzr git mercurial openssh-client subversion procps fossil && \
+RUN apk add --no-cache bzr git mercurial openssh-client subversion procps fossil && \
```
manugupt1: Thanks for this! I would rather `apk update` and use `apk add`. This will easily allow us to add future packages as well without getting indexes all the time for example
```
RUN apk add
RUN apk add 
```
and so on..
My suggestion would be to have 
`RUN apk update && apk add bzr git ....`

Let me know what you think?

fsouza: We can do ``apk add --update`` then. Though the reason why people use ``apk add --no-cache`` is to get smaller Docker images. For example:

- with ``apk add --no-cache``, the image ends up with 161MB
- with ``apk add --update``, the image ends up with 163MB

The difference is very very small, so I'm not sure if you care. Let me know and I can make the change either way! :)

manugupt1: Yep! I do not :)

fsouza: Cool, updated it to use ``--update``! Thanks for the feedback! :D :D

PR #1070, opened 2019-02-18
pkg/stash: add Etcd as a SingleFlight backend

### LINKED ISSUE #760: Implement Distributed Deduplication Mechanisms

marwan-at-work:
This is a more polished version of my previous PR (https://github.com/gomods/athens/pull/933)

It fixes the review comments, merge conflicts, and is also no longer WIP so it's ready to merge.
Apologies if I missed any of the previous comments, feel free to call them out. 

Fixes #760


docker-compose.yml
```diff
@@ -52,3 +52,74 @@ services:
...
+      - http://etcd2:2380
+      - -listen-peer-urls
+      - http://0.0.0.0:2380
+      - -initial-cluster
+      - etcd0=http://etcd0:2380,etcd1=http://etcd1:2380,etcd2=http://etcd2:2380
+
+volumes:
+  etcd0:
+  etcd1:
+  etcd2:
```
arschles: @marwan-at-work nit: can you add a newline at the end of the file?

pkg/config/singleflight.go
```diff
@@ -0,0 +1,15 @@
...
+// lock or single flight mechanism.
+type SingleFlight struct {
+	Etcd *Etcd
+}
+
+// Etcd holds client side configuration
+// that helps Athens connect to the
+// Etcd backends.
+type Etcd struct {
+	Endpoints string
```
arschles: Can you explicitly name this and the above struct field with tags? I think explicit is better than implicit here

pkg/stash/with_etcd_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	var eg errgroup.Group
+	for i := 0; i < 5; i++ {
+		eg.Go(func() error {
+			_, err := s.Stash(context.Background(), "mod", "ver")
+			return err
+		})
+	}
+
+	err = eg.Wait()
```
arschles: is there a way to do this with a timeout? it'd be nice to avoid having integration tests hang forever if there's something wrong with an etcd server running in docker or something else. maybe this can be done by creating `eg` with a context that times out? I don't remember if that's how it works...

pkg/config/singleflight.go
marwan-at-work: @arschles +1 on this field, missed the env override. But the above struct field does not need it just like the `StorageConfig`

pkg/stash/with_etcd_test.go
marwan-at-work: @arschles nice suggestion, I'll add a timeout on the context of the Stash function, should work fine.

pkg/config/config.go
```diff
@@ -39,6 +39,8 @@ type Config struct {
 	HGRCPath         string `envconfig:"ATHENS_HGRC_PATH"`
 	TLSCertFile      string `envconfig:"ATHENS_TLSCERT_FILE"`
 	TLSKeyFile       string `envconfig:"ATHENS_TLSKEY_FILE"`
+	SingleFlightType string `envconfig:"ATHENS_SINGLE_FLIGHT_TYPE"`
```
arschles: 👍

PR #1073, opened 2019-02-18
Check code defaults and config.toml defaults are the same

marpio:

We should keep code defaults and config.toml defaults the same.
I've added a test so that we know if they're out of sync.

### DIFF
--- a/config.dev.toml
+++ b/config.dev.toml
@@ -132,13 +132,13 @@ TraceExporter = ""
 # TraceExporterURL is the URL to which Athens populates distributed tracing
 # information such as Jaeger. In Stackdriver, use this as the GCP ProjectID.
 # Env override: ATHENS_TRACE_EXPORTER_URL
-TraceExporterURL = ""
+TraceExporterURL = "http://localhost:14268"
 
 # Stats is not a requirement for Athens. If the value is not set,
 # Athens will keep on running and stats won't be collected.
 # Possible values are: prometheus.
 # Env overide: ATHENS_STATS_EXPORTER
-StatsExporter = ""
+StatsExporter = "prometheus"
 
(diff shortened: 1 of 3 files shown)


config.dev.toml
```diff
@@ -132,13 +132,13 @@ TraceExporter = ""
 # TraceExporterURL is the URL to which Athens populates distributed tracing
 # information such as Jaeger. In Stackdriver, use this as the GCP ProjectID.
 # Env override: ATHENS_TRACE_EXPORTER_URL
-TraceExporterURL = "http://localhost:14268"
+TraceExporterURL = ""
```
marwan-at-work: Why did we remove this? It's a sane default since our trace backend is Jaeger by default. However, Jaeger is off but once it's on we don't need to specify the URL

config.dev.toml
```diff
@@ -132,13 +132,13 @@ TraceExporter = ""
...
 
 # Stats is not a requirement for Athens. If the value is not set,
 # Athens will keep on running and stats won't be collected.
 # Possible values are: prometheus.
 # Env overide: ATHENS_STATS_EXPORTER
-StatsExporter = "prometheus"
+StatsExporter = ""
```
marwan-at-work: @marpio I believe the original default was "prometheus" as I just recently updated in my PR in the config.go file and not the other way around. This is because we don't actually ping prometheus, but prometheus pings us so it's okay to prepare the handler for it whether there is a scraper or not. 

cc: @manugupt1 as I believe he helped with this default

t-tomalak: @marwan-at-work I was working on that, and yes you are right that, we setup is as default because we don't need external service to actually be runing and scraping that stats.

PR #1074, opened 2019-02-19
pkg/config: make sure envconfig doesn't override config values for port

### LINKED ISSUE #1037: envconfig incorrectly overrides config values from file
envconfig will check if appropriate environment variable variable is set, and it not will set the corresponding config value to default, if default tag is present.

However, by doing so it will rewrite the values set via the config file, because it is run after the config file was parsed (consider: https://github.com/gomods/athens/blob/fd10bed60949bb14d612222b3bf600f07808d65d/pkg/config/config.go#L87-L99)

This makes it impossible to set a `Port` value (the only one with default tag right now) via the config file, as envconfig will revert it to default unconditionally.

fsouza:
Fixes #1037.

envconfig can't actually support that use case, and because athens was
using the `default:` tag only for the port field, handling this manually
is the easiest and functional solution.

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -64,6 +64,7 @@ func TestPortDefaultsCorrectly(t *testing.T) {
 }
 
 func TestEnvOverrides(t *testing.T) {
+	os.Clearenv()
 	expConf := &Config{
 		GoEnv:           "production",
 		GoGetWorkers:    10,
@@ -88,10 +89,7 @@ func TestEnvOverrides(t *testing.T) {
 	}
 
 	envVars := getEnvMap(expConf)
-	envVarBackup := map[string]string{}
 	for k, v := range envVars {
-		oldVal := os.Getenv(k)
-		envVarBackup[k] = oldVal
 		os.Setenv(k, v)
 	}
 	conf := &Config{}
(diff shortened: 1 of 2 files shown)


pkg/config/config_test.go
```diff
@@ -88,10 +89,7 @@ func TestEnvOverrides(t *testing.T) {
 	}
 
 	envVars := getEnvMap(expConf)
-	envVarBackup := map[string]string{}
 	for k, v := range envVars {
-		oldVal := os.Getenv(k)
-		envVarBackup[k] = oldVal
 		os.Setenv(k, v)
 	}
 	conf := &Config{}
```
manugupt1: I am having a bit of trouble understanding this test. Expected Port is set as `:7000`. If you override an empty `PORT` with your change, it should be `:3000` and then the two config, should be equal which they are not.

Can you help me understand it please?

fsouza: Hey, thanks for the feedback. Do you mean this test or the new test?

This test was already here, but what it does is:  ``getEnvMap(expConf)`` returns a ``map[string]string`` with the values for environment variables, then ``os.Setenv`` is called for each entry in the map. By the time ``envOverride(conf)`` is called, the environment variable ``ATHENS_PORT`` is defined as ``":7000"``, therefore the test pass as ``expConf`` expects it to be ``":7000"``. The test continues to pass because nothing changed in the behavior.

In the test below (``TestEnvOverridesPreservingPort``), ``envOverride`` receives a Config object that has a Port defined (and that port is not the default one). That test fails without the change in config.go (because of the issue described in #1037.

Should I add some comments to the code or make changes to clarify the intent?

manugupt1: Ah.. yep! I missed that part! That makes sense!

Can you add a test that sets `:3000` when neither env var is set nor is it defined in config file? I think that is what stumped me. Another thing is that the default ATHENS_PORT can probably be a global variable so that you can test against that value directly in the pkg. I do not see a problem with constants being in a global.

Issue #1051, opened 2019-02-04
Add support for `~`, `^` and `<` in filtering 


michalpristas:
**Is your feature request related to a problem? Please describe.**
After #1050 is in we have a way how to enable specific versions of the module(those verified) 
With repo being more active filtering might become a bit noisy which we can address by supporting `~` and `^`

**Describe the solution you'd like**
With this issue fixed we will have 5 ways how to enable module:
- `+ repo`: enables all versions of the repo
- `+ repo 1.2,1.3`: enables just the listed versions of the repo
- `+ repo ~1.2.3`: enables `1.2.x`  which are at least `1.2.3`
- `+ repo ^1.2.3`: enables `1.x.x` which are at least `1.2.3` 
- `+ repo <1.2.3`: enables everything lower than `1.2.3` includes `1.2.2` and `0.58.9` as well

Kunde21:
The other extension I would like to have on the version filtering is handling multiple rule entries.  We would need to define precedence rules, but I can see an approved list that has multiple approval points and one shared instance.  
```
- 
# entries from team A
+ github.com/gomods/athens v1.1,v1.3
# many other entries

# entries from team B
+ github.com/gomods/athens v1.2.1
```

Another option might be a block list that is append-only:
```
+
- github.com/buggy/mcbuggerson v2.27.301
- github.com/buggy/mcbuggerson v2.28.110
# and so on
```

fedepaol:
Can I grab this to get back on track?

Kunde21:
Yeah, go ahead @fedepaol.  Thank you!

marwan-at-work:
Looking at @michalpristas 's examples in the description above, any reason the versions don't have a `v` prefixed in them? I feel like we should stay consist with how the Go community is starting to expect SemVer tags to look like. 

Also, if 

> `+ repo ~1.2.3`: enables `1.2.x`

Why not just make it `+ repo 1.2.x`? where you can do `2.x.x` and so on. 

Also, I'd like us to step and think more thoroughly about this: 

Are these features worth the complexity? Is it widely used and widely requested? 

I'm not against, but personally I'll probably never use it which doesn't mean no one will, but would love to see some sort of up votes before features are accepted.

fedepaol:
@marwan-at-work I'll hold my keyboard, and I agree to not overcomplicate the filters.
We should think about general scenarios / use cases. 
[...]

PR #1075, opened 2019-02-19
Filter

fedepaol:
Enhance the currently supported filter rules as described in https://github.com/gomods/athens/issues/1051
The filter lines are enhanced with new match patterns:

* `+ repo ~1.2.3`: enables `1.2.x`  which are at least `1.2.3`
* `+ repo ^1.2.3`: enables `1.x.x` which are at least `1.2.3`
* `+ repo <1.2.3`: enables everything lower than `1.2.3` includes `1.2.2` and `0.58.9` as well

**How is the fix applied?**

Modified the filtering process

Fixes  https://github.com/gomods/athens/issues/1051


docs/content/configuration/filter.md
```diff
@@ -68,3 +68,19 @@ D enterprise.github.com/company
...
+
+<pre>
+-
+# external dependency approved list
++ github.com/gomods/athens <1.2.3
+</pre>
+
+The currently supported modifiers are 
+
+- ~1.2.3: enables 1.2.x  which are at least 1.2.3
```
marwan-at-work: Out of curiosity: why aren't we using the `v` prefix for semver tags just like go users expect them to be?

fedepaol: That's reasonable (and now I remember you said the same thing on the issue). Will change that to ~v1.2.3 and such.

fedepaol: Changed

docs/content/configuration/filter.md
```diff
@@ -68,3 +68,19 @@ D enterprise.github.com/company
...
+-
+# external dependency approved list
++ github.com/gomods/athens <v1.2.3
+</pre>
+
+The currently supported modifiers are 
+
+- ~v1.2.3: enables 1.2.x  which are at least 1.2.3
+- ^v1.2.3: enables 1.x.x which are at least 1.2.3
+- <v1.2.3: enables everything lower than 1.2.3 includes 1.2.2 and 0.58.9 as well
```
arschles: I have some wording suggestions to make this more human readable and more formally specific

* `~1.2.3` will enable all patch versions from 1.2.3 and above (e.g. 1.2.3, 1.2.4, 1.2.5)
    * Formally, `1.2.x` where `x >= 3`
* `^1.2.3` will enable all patch and minor versions from 1.2.3 and above (e.g. 1.2.4, 1.3.0 and 1.4.5)      
    * Formally, `1.x.y` where `x >= 2` and `y >= 3`
* `<1.2.3` will enable all versions lower than 1.2.3 (e.g. 1.2.2, 1.0.0 and 0.58.9)
    * Formally, `x.y.z` where `x <= 1`, `y < = 2` and `z < 3`

fedepaol: Done!

pkg/module/filter.go
```diff
@@ -189,13 +190,94 @@ func initFromConfig(filePath string) (*Filter, error) {
...
+// - plain versions
+// - v1.2.3 enables v1.2.3
+// - ~1.2.3: enables 1.2.x  which are at least 1.2.3
+// - ^1.2.3: enables 1.x.x which are at least 1.2.3
+// - <1.2.3: enables everything lower than 1.2.3 includes 1.2.2 and 0.58.9 as well
+func matches(version, qualifier string) bool {
+	prefix := qualifier[0]
+	first := qualifier[1]
+
+	// v1.2.3 means we accept every version starting with v.1.2.3
```
marwan-at-work: s/v.1.2.3/v1.2.3

pkg/module/filter.go
```diff
@@ -189,13 +190,94 @@ func initFromConfig(filePath string) (*Filter, error) {
...
+	// handle this special case first, then go for ~v1.2.3 and similar
+	if prefix == 'v' && first >= '0' && first <= '9' { // a number
+		return strings.HasPrefix(version, qualifier)
+	}
+
+	v, err := getVersionSegments(version[1:])
```
marwan-at-work: will this always be guaranteed to have a minimum length of 2? Otherwise it might panic.

pkg/module/filter.go
```diff
@@ -189,13 +190,94 @@ func initFromConfig(filePath string) (*Filter, error) {
...
+	if err != nil {
+		return false
+	}
+
+	q, err := getVersionSegments(qualifier[2:])
```
marwan-at-work: will this always be guaranteed to have a minimum length of 3? Otherwise it might panic.

pkg/module/filter.go
fedepaol: That's true.There's no guarantee in the wild internet. 
Will make the function more robust.

pkg/module/filter.go
fedepaol: See above

docs/content/configuration/filter.md
```diff
@@ -68,3 +68,24 @@ D enterprise.github.com/company
...
+
+Athens provides advanced filter modifiers to cover cases such as api compatibility or when a given dependency changes its license from a given versions. The modifiers are intended to be used in the pattern list of the filter file.
+
+<pre>
+-
+# external dependency approved list
++ github.com/gomods/athens <v1.2.3
+</pre>
+
+The currently supported modifiers are
```
michalpristas: please add a Warning that in filter file you need always specify version using semver with 3 parts and (`1.2` will not work)

pkg/module/filter.go
```diff
@@ -189,13 +190,94 @@ func initFromConfig(filePath string) (*Filter, error) {
...
+	first := qualifier[1]
+
+	// v1.2.3 means we accept every version starting with v.1.2.3
+	// handle this special case first, then go for ~v1.2.3 and similar
+	if prefix == 'v' && first >= '0' && first <= '9' { // a number
+		return strings.HasPrefix(version, qualifier)
+	}
+
+	v, err := getVersionSegments(version[1:])
+	if err != nil {
```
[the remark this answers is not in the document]
fedepaol: Passing the logger down there would involve messing up with a lot of layers. 
Not sure if it's worth the effort. That function might fail only if an element of the semver is not a string which would not mach any filter by definition.
On the other hand, I was thinking that we could validate the filter file after loading it. It would prevent having invalid qualifiers.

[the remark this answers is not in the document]
fedepaol: Obviuosuly when I say "the effort" I mean the effort in adding that kind of complexity to the codebase, not that I don't want to write the code :-)

pkg/module/filter.go
```diff
@@ -172,30 +173,115 @@ func initFromConfig(filePath string) (*Filter, error) {
...
 
+// matches checks if the given version matches the given qualifier.
+// Qualifiers can be:
+// - plain versions
+// - v1.2.3 enables v1.2.3
+// - ~1.2.3: enables 1.2.x  which are at least 1.2.3
+// - ^1.2.3: enables 1.x.x which are at least 1.2.3
+// - <1.2.3: enables everything lower than 1.2.3 includes 1.2.2 and 0.58.9 as well
+func matches(version, qualifier string) bool {
+	if len(qualifier) < 2 || len(qualifier) < 2 {
```
michalpristas: `len(qualifier) < 2` two times

michalpristas:
After update still LGTM. besides this one condition

pkg/module/filter.go
fedepaol: Done

michalpristas:
@marwan-at-work can you check once you have some time, if all your comments are addressed or anything else pops up

fedepaol:
@arschles they are already covered, sorry for being late in replying

marpio:
@arschles merging since everything has been addressed.

arschles:
@fedepaol got it! thanks

PR #1086, opened 2019-02-25
Helm chart: Fix ingress setup

zerok:
The ingress part of the Helm chart contained references to
not-existing templates and values. Additionally, the deployment used
some no longer available health-check URLs which prevented it from
being marked as ready.

The ingress configuration inside the Helm chart was faulty which prevented the chart to be applied if `ingress.enabled` was set to `true`.

**How is the fix applied?**

I've updated the ingress part of the Helm chart to reference existing templates and also the appropriate values.


charts/proxy/templates/deployment.yaml
```diff
@@ -23,11 +23,11 @@ spec:
         imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
         livenessProbe:
           httpGet:
-            path: /healthz
```
michalpristas: this should be accessible 
https://github.com/gomods/athens/blob/12ce2a6499bd9fe3b0486968008497f0fb03f9ee/cmd/proxy/actions/app_proxy.go#L25

zerok: @michalpristas My bad. The helm chart currently references v0.2.0 in its default configuration where the health checks were only bound to `/`. Would setting the default value to `canary` be an option here?

arschles: @zerok good catch. would it be possible to add an `if` on the version in here, so that either way (`v0.2.0` and `canary` - no need to look for other versions than that) this health check can reference the right path?

zerok: @arschles I've put the paths used for these checks behind a template 🙂

charts/proxy/templates/_helpers.tpl
```diff
@@ -2,3 +2,9 @@
 {{- $name := default .Chart.Name .Values.nameOverride -}}
 {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
+{{- define "livenessPath" -}}
```
michalpristas: Great 👍

michalpristas:
@zerok please take care of these nasty conflicts

zerok:
@michalpristas done 🙂

arschles:
@zerok looks awesome, thank you again for this!

Looking forward to your next PR 😁

zerok:
@arschles Thanks 😄

Issue #1087, opened 2019-02-25
Storage/GCP: Allow authentication for GCS outside of GCP


marwan-at-work:
If Athens is running outside of GCP, there needs to be a way to authenticate it to access Google Cloud Storage. This is a useful case for us if we are say running a CI/CD build that can be anywhere (github-actions, droneCI, circleCI, etc) but have our modules stored in GCS.

PR #1084, opened 2019-02-24
Adding a CircleCI config file

arschles:
**What is the problem I am trying to address?**

Travis-CI got acquired and we don't know if it's gonna be stable. See https://github.com/gomods/athens/issues/1077 for details

**How is the fix applied?**

This adds a [CircleCI config file](https://circleci.com/docs/2.0/configuration-reference/#section=configuration) to test out how CircleCI runs our tests. 

[here](https://circleci.com/gh/gomods/athens/tree/pull%2F1084) are the builds for this PR

**Mention the issue number it fixes or add the details of the changes if it doesn't have a specific issue.**

This is a proposed fix for #1077. I've tagged with do not merge until we decide what to do

PR #1088, opened 2019-02-25
storage/gcp: allow json key to be used for GCS access

marwan-at-work:
Fixes https://github.com/gomods/athens/issues/1087


arschles:
Ignore the circle CI build failure. it's a work in progress on #1084

PR #1092, opened 2019-02-26
build: switch to go1.12

marwan-at-work:
Once go1.12 is up on Docker Hub, this can be merged so that we can release v0.3.0 ✌️

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ services:
 language: go
 install: false
 go:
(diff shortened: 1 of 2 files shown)


chriscoffee:
1.12 is out! https://hub.docker.com/_/golang?tab=tags

marwan-at-work:
@chriscoffee thanks for the note, restarted the build

PR #1096, opened 2019-03-01
Upstream goproxy

yittg:

Support configure one upstream proxy, like mirror of another. Often use in private develop environment.

**How is the fix applied?**

If one go proxy configured, all handler for `GOPROXY` will try to fetch from the upstream proxy configured if local miss, or do the original fetch if failed.

proxy repo code edited from [go modfetch package](https://github.com/golang/go/tree/master/src/cmd/go/internal/modfetch).

### DIFF
--- a/cmd/proxy/actions/app_proxy.go
+++ b/cmd/proxy/actions/app_proxy.go
@@ -57,6 +57,11 @@ func addProxyRoutes(
 
 	lister := download.NewVCSLister(c.GoBinary, fs)
 
+	if c.Proxy != "" {
+		mf = module.WithProxy(c.Proxy, mf)
+		lister = download.WithProxy(c.Proxy, lister)
+	}
+
 	withSingleFlight, err := getSingleFlight(c, s)
 	if err != nil {
 		return err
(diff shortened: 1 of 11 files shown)


marwan-at-work:
@yittg Athens already has a way to redirect requests to another proxy. See the `GlobalEndpoint` field in the config file. Please see if that solves your use case, otherwise feel free to open an issue issue about what you need and we'll discuss design/implementation there first. 

Furthermore, when it comes to redirecting to other proxies, we are most likely not going to make any major decisions until the Go team resolves this issue: https://github.com/golang/go/issues/26334

I will close this PR for now but feel free to open an issue and if it makes sense I'll re-open this PR. If you also want some chat-style discussion, feel free to join and ask anything on the #athens channel on the Gopher Slack org. 

Thanks for contributing!

yittg:
oh, i missed the existed config, it's a better and more complete way. i write this patch when reading the athens and goproxy source code by the way, so i don't new an issue first 😊, it's all right. And thanks for the great job.

yittg:
I've have a closer look about this configuration and it's machanism. It doesn't match this PR exactly. It  redirects to proxy directly without local cache function, is it right?And cache is a important for private env. May we new an issue to discuss it?@marwan-at-work

marwan-at-work:
@yittg The entire multi proxy communication will be rethought for Athens once the Go issue above that I mentioned is resolved. It might be best to wait until then so we know what kind of solution we want to have. So I'd keep the conversation in the upstream and once that is resolved, we can discuss the how Athens should play out. If you see anything differently, you're always free to open an issue as you see fit :)

yittg:
@marwan-at-work That's all right, thanks for your reply. :+1:

PR #1097, opened 2019-03-01
Publish charts to azure storage

### LINKED ISSUE #566: Host charts
continuing the discussion from: https://github.com/gomods/athens/pull/547#discussion_r212679915

We should decide where to host our helm charts, maybe docker compose too?

@arschles suggested starting with an Azure storage container, which as @carolynvs mentioned we could then point a sub domain like `charts.gomods.io`.

carolynvs:

Publish our helm charts to azure storage so people can install it from a neutral location, such as https://charts.gomods.io. I need to work with @arschles to get the subdomain setup to point to the azure bucket but this will allow us to either round robin or host elsewhere later if we need to without causing the old links to go bad.

**How is the fix applied?**

* I had to move the charts/proxy directory to charts/azure-proxy to work with the helm standards, otherwise the tool refused to work
* The script packages (tars) up the chart and then updates the index file
* The tar files and index file and uploaded to azure using a tiny azure client
* I am using a docker build image to deal with the helm and azure client tool dependencies so that it's easy for someone else to run and test these scripts.
* The secret environment variable `AZURE_STORAGE_CONNECTION_STRING` is set directly in Travis. 

Fixes #566


carolynvs:
OK I've created https://athens.blob.core.windows.net/charts, put the key into Travis, and the container will be populated after this PR is merged.

carolynvs:
Note that going forward, when we make chart changes, we need to make sure to bump the chart version in charts/Chart.yaml.

carolynvs:
Since we can't find the right person to make a CNAME for us with charts.gomods.io, we decided to move forward with this PR and point a CNAME to the container later.

PR #1103, opened 2019-03-03
Config improvements - rename file & remove default

### LINKED ISSUE #914: Update default path for proxy config file
The current default path to the config file is `../../config.dev.toml` regardless from where you start the proxy.

I'd suggest to use a more generic name for the file, as `config.toml` and also either skip the path entirely or even better use (per OS) specific config path. E.g.:

 OS | Path |
 ---- | ---- |
 Linux | `$XDG_CONFIG_HOME` |
 macOS | `~/Library/Application Support/` ([details](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/MacOSXDirectories/MacOSXDirectories.html)) |
 Windows | `%LOCALAPPDATA%` |

leitzler:
Fixes #914.

The original problem in that issue have however changed since the default path have been removed already.

This PR solves the rest of the problems I had back then, and has two parts:
1) Rename the provided config to `config-example.toml`, a clarification for non developers.
2) Remove the unknown default config that is used if no file is specified. If a user accidentally doesn't specify the file when starting, there is no obvious way to find out what config is used. The default config was quite permissive and that is probably not what you want. Since the flag now have a default value, athens will still start if there is a `config.toml` in the working directory.


docs/content/install/install-on-kubernetes.md
```diff
@@ -91,7 +91,7 @@ This will deploy a single Athens instance in the `default` namespace with `disk`
 
 ### Give Athens access to private repositories via Github Token (Optional)
 1. Create a token at https://github.com/settings/tokens
-2. Provide the token to the Athens proxy either through the [config.toml](https://github.com/gomods/athens/blob/master/config.dev.toml#L111) file or by setting the `ATHENS_GITHUB_TOKEN` environment variable.
+2. Provide the token to the Athens proxy either through the [config.toml](https://github.com/gomods/athens/blob/master/config-example.toml#L111) file or by setting the `ATHENS_GITHUB_TOKEN` environment variable.
```
manugupt1: This line number is too specific. I would rephrase it in this or another PR a that is not the focus of the PR

manugupt1:
Seems like CI is failing because of config file missing.

docs/content/install/install-on-kubernetes.md
leitzler: I agree that line numbers shouldn't be included in the docs. Did however not change it here since it isn't the scope for this PR.

marpio:
@leitzler
The hardcoded defaults are in-sync with the config file defaults (it's ensured by a unit test)
Do you think writing those to the std out and adding docs would make the experience better? 
Having the proxy be easy to tryout without having to think about our rather big configuration is a nice feature IMO.

leitzler:
Yes, my main concern is that there is no obvious way for the user to know what configuration they are running if it is hardcoded inside the binary.

@marpio yeah, writing them to stdout would make much more sense, good idea!
In my experience programs either:
- have a configuration file with sane defaults to start with
- writes a new config file if not found and uses it (to home directory for example)
- have a flag that prints out the default configuration to stdout so that the user can create a new file if they lost the provided

There is also the way Hashicorp Vault does it, where you can start with `vault server -dev` and get a developer friendly instance up & running with (well documented) sane defaults.

The risk with built in hard coded defaults is also that they might be used suddenly without the user knowing, if the user starts the binary and forgets to point out the config file. And if the defaults happen to be more permissive than the intended config file there might be data leaks.

I think all three (and a half) suggestions above are nice and wouldn't mind any of them. I just picked one for this PR since I think it makes sense to change the current behaviour :).

marwan-at-work:
@leitzler I like the idea of writing a defaults file when none exists. This might mean, however, that we still have to hardcode a file in the codebase to make sure it's writeable on other users machines. Or, embed our config.toml file into the binary. 

Maybe your 3rd option is best: just be informative to the user. Something like "no config file found, using defaults" might be enough? 

Btw, other parts of this PR are definitely good such as getting rid of the `../../configfile` to something more sane such as $HOME or just `./configfile`. But maybe those two things are worth separating into two PRs.

leitzler:
Hardcoded embedded file that is written if missing is fine, then the user can see the default values. The cons is that we still need to decide where to write that file (not obvious, see commends in #914).

And about being informative to the user, that's kind of how it works today (since there is a log entry). But that log entry doesn't contain anything about how the default looks like.

Imagine a user editing the config to enable TLS and basic auth. The at some point starts the proxy and forgets to specify that config, then the proxy will be exposed without TLS or password by default. I think that an insecure option should be a active choice :)

The best solution is probably to write-defaults-if-not-exist, and then use that file as default if no file is specified. That needs a decision on where the default location should be to work cross platform `($HOME/.athens/config.toml)`?

docs/content/intro/why.md
```diff
@@ -21,7 +21,7 @@ The Go community has had lots of problems with libraries disappearing or changin
...
+The fact that the Go command line can now ping _your own_ server to download dependencies, that means you can program whatever logic you want around providing such dependencies. Things like Access Control (discussed below), adding custom versions, custom forks, and custom packages. For example, Athens provides a [Validation Hook](https://github.com/gomods/athens/blob/master/config-example.toml#L87) that will get called for every module download to determine whether a module should be downloaded or not. Therefore, you can extend Athens with your own logic such as scanning a module path or code for red flags etc.
```
arschles: nit for a follow-up PR: I think we should reword that first sentence:

>Since the Go command line can now ping _your own_ server to download dependencies, you can implement whatever logic you want around providing such dependencies.

leitzler:
> @leitzler I think I'm 👍 with this but I want to confirm - if you don't pass a config file at all, do you still get the defaults that are hard-coded?

In the current state the default behaviour, when not passing a config file, is to look for `config.toml` in the current directory. I can change that to one of the suggestions above, if we just decide how it should behave.

arschles:
@leitzler I think that's fine for now 😄

arschles:
And, sorry for the long delay in my response!

arschles:
@leitzler can you resolve conflicts and then we can merge?

marwan-at-work:
@arschles there are still unresolved questions here, this should not be merged yet. 

I'm still hesitant about writing a default file somewhere in the FS because you can't guarantee what people's environments look like and if they can write to the FS. At the same time, I do think that looking for a file in `../..` is super weird. My opinion is that the config should be the following, *with precedence*: 

1. -config_file comes first if it was present. 
2. If above present either look in $HOME/.athens.config.toml or look in $CWD/.athens.config.toml
3. If above not present, log to the user "no config file found, using defaults" and use defaults. We already do this logging. 

For option 2, wehther we do $HOME or $CWD is up to everyone. Either way is fine with me. 

Thanks

leitzler:
I agree with @marwan-at-work, there are still some questions to solve.

1. `-config_file` should have highest prio, no doubt. :+1:
2. Reading from an existing file that isn't in the folder you start the binary from leads to problems. There is no env var named `$CWD` by default in Linux or macOS. Did you mean the dir from where the binary is started? Would prefer that before `$HOME`, do that env var exist on all supported OS:es?

3. I think that it is kind of weird to start a server application like this with unknown defaults. If I, as a end user, download the binary to try it out and starts it, I wouldn't even know what port it listens to, or that it listens on all interfaces and not just localhost. Servers, in my experience, tend to bind to loopback interface (only) when they want to provide a quick up-and-running for trying out purposes). Without me knowing, it also connects to other services that happens to run on my machine (like etcd & jaeger), if I understand the default configuration.

That's my opinion at least, I see that it isn't shared so I'll remove those commits from this PR and keep first one that renames the config file and removes `../../`. That commit was the initial reason for this PR anyway. And if you prefer to keep it the way it is now, just close this PR, no harm done!

marwan-at-work:
@leitzler for no. 2, I'm happy with either of the following: 

1. Look for the CWD of where the binary started. 
2. Use github.com/mitchellh/go-homedir to get the home dir and look for something there. If an error occurred, log it and fall to no. 3 

If that's ok with you, I'd say the PR can lean towards those steps and I'm happy to help if you don't have time. Much appreciated!

marwan-at-work:
As for your comment regarding number 3, the big argument towards having "unknown defaults" is that users can start 'trying out Athens' without having to do absolutely anything. All they have to do is just `go install github.com/gomods/athens && athens` 

They'll be able to read that the code is running on port 3000, and just use it immediately. One thing we can do, is have a more unique port number.

arschles:
ok, won't merge yet (obviously). I'll say that I agree with @marwan-at-work that we should make the `go install github.com/gomods/athens && athens` experience work well.

If the config is to use in-memory or filesystem storage (in the current working directory) and serve on a specific port that we log, we can write clear logs and documentation in our "getting started" to say what's going to happen and what the user should do to clean up after they're done

leitzler:
So, for 2), would we want cwd, home or both? Did push a changed that does both.

Makefile
```diff
@@ -15,7 +15,7 @@ build-image:
 
 .PHONY: run
 run:
-	cd ./cmd/proxy && go run . -config_file ../../config.dev.toml
+	go run ./cmd/proxy -config_file config-example.toml
```
marwan-at-work: Let's keep this `config.dev.toml` so that we don't break people who already have links saved for reference as well as scripts that use the default one. It also says `.dev` so it should be clear that it's an "example" and not production ready. We can even add a comment at the top of the file to say `this is not for production` if that's your motivation behind the breaking change : )

pkg/config/config.go
```diff
@@ -47,14 +47,31 @@ type Config struct {
...
-	if configFile == "" {
-		log.Print("config file not provided - using default settings")
-		return createDefault(), nil
+	// User explicitly specified a config file
+	if configFile != "" {
+		return ParseConfigFile(configFile)
 	}
-	return ParseConfigFile(configFile)
+
+	defaultConfigFile := ".athens.toml"
```
marwan-at-work: nit: let's pull this into a package level `const` so that it's easier for users to spot.

config.dev.toml
```diff
@@ -2,6 +2,9 @@
 # Most properties can be overridden with environment variables specified in this file
 # Most properties also have defaults (mentioned in this file) if they are not set in either the config file or the corresponding environment variable
 
+# If you put this file where you start Athens from (or in your home directory) as ".athens.toml", athens will use it when starting.
```
marwan-at-work: this says `.athens.toml` but the `defaultConfigFile` says `athens.toml`. I think I like `.athens.toml` better but either way is cool with me 👍 

It also mentions the home directory but I see you removed it from the code. Also either way is cool with me 😄

docs/content/install/install-on-kubernetes.md
arschles: @manugupt1 I'll open up an issue (and a PR for it) to remove the line number, so we can go forward with this PR

manugupt1:
Cool! Makes sense

Cheers
Manu

________________________________

docs/content/install/install-on-kubernetes.md
arschles: https://github.com/gomods/athens/issues/1157

PR #1108, opened 2019-03-06
Delete the maintainer resources file

arschles:
We were trying to track all the resources we use so that maintainers can get access to them, but this is harder to do in Markdown than in a google sheet, so we're going to stick with that

**How is the fix applied?**

Fixes #


chriscoffee:
LGTM 👍

PR #1109, opened 2019-03-07
config.toml: add cautionary note around basic auth

marwan-at-work:
Added a helpful documentation to let people know that the Basic Auth option is only temporary and not a great production solution :)


chriscoffee:
LGTM, I like the fact it's all caps :shipit:

PR #1114, opened 2019-03-09
Filter out pseudo versions from storage on list endpoint

### LINKED ISSUE #1000: can't get new commit, when no version tag
use a no version tag, example: `github.com/kataras/pio`

run this when no `github.com/kataras/pio` in disk

```shell
# download commit: v0.0.0-20180502220154-1d218932f864
curl localhost:3000/github.com/kataras/pio/@v/v0.0.0-20180502220154-1d218932f864.info

# looking version, and will list v0.0.0-20180502220154-1d218932f864
curl localhost:3000/github.com/kataras/pio/@v/list

# actually, it has a new v0.0.0-20180511174041-a9733b5b6b83,
# so, when I do tidy on module(GO111MODULE=on go mod tidy), it can't pull this
```

marpio:

This PR prevents the scenario described here: https://github.com/gomods/athens/issues/1000#issuecomment-449658598

**How is the fix applied?**
The pseudo versions coming from storage will be filtered out.

Fixes #1000


pkg/download/protocol.go
```diff
@@ -100,8 +102,30 @@ func (p *protocol) List(ctx context.Context, mod string) ([]string, error) {
...
+	// from now on every time user runs go get github.com/my/mod she/he will get pseudo version x1 even if a newer version x2 exists
+	// this is because /list returns non-empty list of versions (x1) and so /latest wont get hit
+	return union(goList, removePseudoVersions(strList)), nil
+}
+
+var pseudoVersionRE = regexp.MustCompile(`^v[0-9]+\.(0\.0-|\d+\.\d+-([^+]*\.)?0\.)\d{14}-[A-Za-z0-9]+(\+incompatible)?$`)
+
+func removePseudoVersions(allVersions []string) []string {
+	// copied from go cmd https://github.com/golang/go/blob/master/src/cmd/go/internal/modfetch/pseudo.go#L93
+	isPseudoVersion := func(v string) bool {
```
marwan-at-work: nit: it might be a tiny memory saver to define it once in this package?

marpio: Wouldn't the compiler optimize that? Anyway, I've removed the func entirely.

marpio:
Thanks for pointing this out @marwan-at-work ! Totally forgot about that scenario.

Issue #1115, opened 2019-03-10
Stasher: Implement Redis SingleFlight


marwan-at-work:
Most cloud platforms provide a redis-compliant  in-memory store and so it's worth trying having a Stasher implementation that locks stashing modules. This way, users wouldn't have to maintain their own Etcd clusters.

chriscoffee:
This is ace @marwan-at-work! We could also mention a few of the fun redis stuff you can do too, or atleast point people to the docs that show it can do `append-only` and persist to disk in our docs

marwan-at-work:
@chriscoffee not too familiar with Redis but interested to know how any of those features you mentioned could help Athens (and the Stasher in particular)

PR #1116, opened 2019-03-10
pkg/stash: implement a Redis lock for stashing

marwan-at-work:
Creates a Redis implementation for locking a module stash. Currently supports only one endpoint as input and once proven stable, we can add cluster management. 

Fixes https://github.com/gomods/athens/issues/1115


pkg/stash/with_redis.go
```diff
@@ -0,0 +1,73 @@
...
+	lock, err := lock.Obtain(s.client, mv, &lock.Options{
+		LockTimeout: time.Minute * 5,
+		RetryCount:  60 * 5,
+		RetryDelay:  time.Second,
+	})
+	if err != nil {
+		return ver, errors.E(op, err)
+	}
+	defer func() {
+		const op errors.Op = "etcd.Unlock"
```
chriscoffee: Do you mean `etcd.Unlock` here or `redis.Unlock`

PR #1118, opened 2019-03-11
storage/gcp: Optimize Exists check from Getter

marwan-at-work:
There's no need to call storage.Checker because the Object function can return a NotFound error which will give us the same answer. This saves us from calling GCP twice on every request. 

Furthermore, the current implementation of Checker could lie about whether a piece of a module exists or not. For example, say the module.info piece is missing but the module.mod piece exists; this means that the checker would pass but the Object function would return a NotFound which is a 500 before this PR. 

In any case, there's a future design-detail to be discussed once we figure out how the checker should actually work which might mean we'd put this back in, but it wouldn't be a difficult task. 

cc @marpio and @chriscoffee as I think they're familiar with GCP (feel free to ignore otherwise)


pkg/storage/gcp/getter.go
```diff
@@ -6,6 +6,8 @@ import (
 	"io"
 	"io/ioutil"
 
+	"cloud.google.com/go/storage"
+
```
marpio: Please remove empty line

PR #1121, opened 2019-03-11
Remove dummy mod dir

marpio:
From go 1.12 release notes:

> When GO111MODULE is set to on, the go command now supports module-aware operations outside of a module directory, provided that those operations do not need to resolve import paths relative to the current directory or explicitly edit the go.mod file. Commands such as go get, go list, and go mod download behave as if in a module with initially-empty requirements. In this mode, go env GOMOD reports the system's null device (/dev/null or NUL). 

This means we can remove the dummy mod directory required before by` go list` and `go mod download `


marwan-at-work:
Wow, windows failed but linux didn't :) -- could be the CI/CD set up is not making a clean GOPATH/pkg

arschles:
Adding `do not merge` until https://github.com/golang/go/issues/29522 is resolved

arschles:
Now that Athens uses 1.12 to do the downloads, I think we can merge this now, no?

@marpio @marwan-at-work what do you both think?

marwan-at-work:
@arschles this would completely break what this PR promises https://github.com/gomods/athens/pull/1201 because nothing would work :) 

So I suggest we say that Athens is strictly 1.12+ and merge this PR as long as @marpio is cool w/ it

arschles:
Good point RE #1201, I didn't put 2+2 together...

I'm cool with saying 1.12 across the board. @marpio can you fix the azure pipelines build? Looks like there is a version mismatch...

marpio:
@arschles @marwan-at-work 
I think we can merge this if you approve

marwan-at-work:
@marpio all good on my end as long as we are cool with being 1.12+ and not 1.11+ (go1.13 is soon anyway)

marpio:
+1 for 1.12
Are we referring to 1.11 somewhere now that #1201 is in?

Issue #1111, opened 2019-03-07
Disable Azure Pipelines on PRs But Allow On Merge/Tag


marwan-at-work:
Azure Pipelines are super slow and since Go is cross-platform, we can almost confidently say if it passes on Linux it can pass on Windows. 

### Proposal 
We should disable Azure Pipelines **only** in PRs then enable them for merge to master and releases.

PR #1122, opened 2019-03-11
ci/cd: remove bottlenecks from azure pipelines

marwan-at-work:
Since Drone is now in place and tests almost all of our storages, there's no need for Azure Pipelines to run the same tests. All we need it for is that it's Windows compatible, and the biggest difference between Windows/Linux is the FileSystem so we should insure the regular tests pass on Windows and I'd say that's a pretty confident sign that things are working properly. 

Fixes https://github.com/gomods/athens/issues/1111

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -22,38 +22,6 @@ steps:
   displayName: 'set up the Go workspace'
 
 - powershell: |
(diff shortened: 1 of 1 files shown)


marwan-at-work:
@arschles gonna merge this but please feel free to review after and I'm happy to make further PRs.

arschles:
all good with me, thanks for pinging me @marwan-at-work !

Issue #1123, opened 2019-03-11
[Proposal] GCS Distributed Locking


marwan-at-work:
This is a proposal to alter the GCS backend storage implementation so it can support distributed locking. Related read: https://cloud.google.com/storage/docs/consistency

There's a pro and a con to this. I believe the pro outweighs the con significantly. 

### Pro
You can run multiple stateless Athens containers and hook them up to a GCS storage and never have to maintain any state at all. No etcd cluster, no redis server and so on. 

### Con 
We need to make sure that the storage.Saver implementation is synchronous and not upload all 3 parts in parallel. This is so that two concurrent requests don't upload parts of each other. 

I have this working locally so I'll be submitting a PR shortly, but I'd love to have this issue open for any discussions or concerns. 

**Important note**: some backend storages don't provide a way to synchronize concurrently. For example, AWS S3 does not provide strong consistency as I was told so if you're running against S3 at scale, you'll need a distributed lock.

marpio:
Nice idea @marwan-at-work ! This should be also possible with Azure Storage.
One question - in a scenario where the upload i.e. succeeded for `mod` and `info` but failed for `zip` would we be able to somehow overwrite the `mod` and ìnfo` the next time we're saving the mod@ver?

Alternatively we could use GCS separately to implement the distributed lock.
Don't know if it's better or not but just wanted to throw in another idea.

This could work as follows:
- To acquire a Lock try to create a file `module@version.lock` (or similar). As long as we get 412 we know the module@version is currently stashed and so we wait a little and try again if we get 200 we can proceed.
- To unlock we try to delete the file until we get 204.

Of course we would need to think about setting some deadline.

Pros of this solution:
- we could leave the uploads parallel
- this implementation could be used independently with any other storage (admittedly it would not happen a lot)
Cons:
- it's not as straightforward as this solution
- also we could have a problem if athens crashes before it unlocks
-

marwan-at-work:
@marpio 
> would we be able to somehow overwrite the mod and ìnfo the next time we're saving the mod@ver?

Yes, the `storage.Delete` method should still work and therefore if something goes wrong, we should be able to clean up with Delete then the next a save happens, we're alright again. 

>also we could have a problem if athens crashes before it unlocks. The minimum life span is 1 day so the module would be locked for the whole day. This might be a no-go.

I didn't know about this limitation, that would leave some unhappy users. 

[...]

PR #1124, opened 2019-03-12
pkg/stash: Add GCS implementation

marwan-at-work:
1. Added a new helper to the errors package
2. Made the GCP saver be synchronous in favor of using it as a distributed stash locker. 

Fixes https://github.com/gomods/athens/issues/1123


pkg/stash/with_gcs.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"context"
+
+	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/observ"
+)
+
+// WithGCSLock returns a distributed singleflight
+// using an redis cluster. If it cannot connect, it will return an error.
```
marpio: s/using an redis cluster/using GCS

marwan-at-work:
@marpio update the pr, thanks for taking a look!

pkg/stash/with_gcs.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func (s *gcsLock) Stash(ctx context.Context, mod, ver string) (newVer string, err error) {
+	const op errors.Op = "gcslock.Stash"
+	ctx, span := observ.StartSpan(ctx, op.String())
+	defer span.End()
+	newVer, err = s.stasher.Stash(ctx, mod, ver)
+	if err != nil {
+		// already been saved before, move on.
+		if errors.Is(err, errors.KindAlreadyExists) {
```
marpio: @marwan-at-work
Can you walk me through a following scenario?
Let's say we have 2 Athens instances - a1 and a2.
a1 wants to stash mod@ver, it doesn't exist so it starts uploding it to the gcs. Now a2, 10ms later wants to stash mod@ver as well. When will a2 geht 412? Will it wait until a1 upload completed? Or will it return immediately?

marwan-at-work: @marpio 
A2 will not wait for A1 to be finished. A2 will keep writing, and the moment you call `Close` it will check if there were previous writes. Keep in mind when two different clients write to the same Object path, they are not overriding each other, they both get a different "version" to write to. However, with the `ExactMatch` condition, the `Close` fails if there was a previous version already finished and closed. 

Therefore, if A2 finished before A1, then A2 is the *winner* and A1 will get a 412. Which is totally okay because from our perspective we don't care who finished first, we just care that a module is immutable and does not have "mixed" content from two different client.

It's basically a race to whoever finishes first. The `with_gcp_test.go` file ensures that because we launch 5 goroutines at the same tiem that all write to the same bucket with no guarantee to who finishes first. Therefore, the test check is that the content across all 3 files are consistent and that's what matters. 

I'm more than happy to listen to more scenarios or hear any concerns cuz this is def a very important piece of the Athens logic. Thanks!

marpio: > It's basically a race to whoever finishes first.

Makes totaly sense now. Thanks!

arschles:
@marpio I merged this because your comment was resolved, I hope that was ok!

PR #1125, opened 2019-03-12
docs: update references to Go 1.11

johanbrandhorst:
Go 1.12 is out now, which might confuse some readers.

Clarify the Go versions that can be used.

**How is the fix applied?**

By rewording some of the docs.


docs/content/try-out.md
```diff
@@ -8,7 +8,7 @@ menu: shortcuts
 
 To quickly see Athens in action, follow these steps:
 
-First, make sure you have [Go 1.11 installed](https://gophersource.com/setup/),
+First, make sure you have [Go 1.11 or later installed](https://gophersource.com/setup/),
```
marwan-at-work: Actually we should say `Go 1.12` here because we only guarantee support from 1.12 on due to the checksum mismatches and the ability to do `go get` outside of a module repository.

johanbrandhorst: OK!

johanbrandhorst:
Always happy to help, and I'm grateful for your work on this @marwan-at-work :).

PR #1126, opened 2019-03-12
Use docker on azure pipelines

marpio:
It would be good to use docker on azure pipelines. This way we would have access to go1.12

Update:
So this is working **but**
- it takes 6:30 min just to run unit tests compared to <1min on drone
- it has hiccups from time to time with the error `Windows named pipe error: The system cannot find the file specified. (code: 2)` i.e. https://dev.azure.com/gomods/athens/_build/results?buildId=747
- I wasn't able to use Nanoserver, which would prob spped this up, b/c I got `The operating system of the container does not match the operating system of the host.` error


arschles:
@marpio it looks like most builds are taking ~4:30 now. Maybe this was shortened after #1122 ?

marwan-at-work:
I'm ok if removed Azure Pipelines until it can support go1.12 because that's the minimum version we support. Another option is make azure pipelines only work for merges, and not PRs. But since this is an integration, and not a webhook, I'm not sure how to enable that.

marpio:
@marwan-at-work
This PR introduces go1.12 
It's just that it's still rather slow - 6:30min

arschles:
@marwan-at-work I'd rather not disable our windows builds altogether, but I'd be ok with doing it only for merges to master. I believe there's a filter in the yaml we can turn on to prevent pipelines from doing any PR builds.

On another note, @marpio do you think we could speed up these builds if we pre-build a docker image for our tests, and use that image to run our tests inside (instead of building the test image each time)?

marpio:
@arschles 
I tried nanoserver b/c the image is much smaller. The build failed b/c the image doesn't contain git but pulling the image alone takes ~4min
Not sure what we can do to reduce that.
Ref:
https://dev.azure.com/gomods/athens/_build/results?buildId=760

arschles:
@marpio makes sense. Let me know what you think about this:

- We create a new repo (`gomods/windows-testing-docker` for example) to hold the `Dockerfile` for our tests
- We write that Dockerfile from `go:1.12-nanoserver` as the base, and install `git` on top
- We hook up CI/CD to deploy that new docker image to `gomods/windows-testing` _and_ Azure container registry (ACR)
    - I think we should push to `gomods/windows-testing` so people can see what we're doing, but the ACR part is key because we can speed up image pull times from Azure pipelines
- We use the docker image from ACR to run our tests
    - Instead of using a `COPY` directive in the `Dockerfile.win.test`, we can mount a volume

I'm suggesting this because:

- We can pre-build a smaller image that has the minimal set of tools we need
- Speed up the image pulls for it
- Still use Docker for all the things in Azure Pipelines, which is what I ❤️ so much about this PR in the first place

marpio:
@arschles sounds like a good plan! Let me create issues for each of the steps.

arschles:
@marpio great! in the meantime, IMO would you be ok with holding off on this PR until the other image is created?

marpio:
@arschles
Yes, of course.

arschles:
@marpio I've created https://github.com/gomods/athens/issues/1160 for now. Feel free to split it up, close it, or whatever else you'd like to do, if you think it's appropriate 😀

arschles:
Note: the work for https://github.com/gomods/athens/issues/1160 is in https://github.com/gomods/windows-testing-docker

arschles:
@marpio gomods/windows-testing-docker#4 is now merged, so `gomods.azurecr.io/athens-win-test:1038` is now available. Can you try that out in here?

Going forward, we'll need to decide how to version the images, but this is a good start!

arschles:
@marpio our windows testing docker image works now! https://dev.azure.com/gomods/windows-testing-docker/_build/results?buildId=1038 is available. I think that the image will be available if you `az login` like [here](https://github.com/gomods/windows-testing-docker/blob/f1d968d34f56097a895afce912f80c81df1f9f71/azure-pipelines.yml#L35) and `az acr login` like [here](https://github.com/gomods/windows-testing-docker/blob/f1d968d34f56097a895afce912f80c81df1f9f71/azure-pipelines.yml#L37)

But if you don't want to do that, I can DM you the username & password so you can `docker login` and `docker pull`. Let me know which you prefer!

arschles:
@marpio what would you like to do with this PR?

marpio:
@arschles 
I'll close it for now.
We can reopen it once the performance of Azure Pipelines improves. It would be nice to be able to use docker and the win1803 image would be perfect for that.

Issue #1130, opened 2019-03-14
Implement storage.Backend for Artifactory storage


tylerchr:
**Is your feature request related to a problem? Please describe.**

At Qualtrics we use Artifactory as an artifact repository, which is not currently supported by Athens. We want to deploy Athens internally. but strong corporate incentives exist to use our internal repository for storage instead of Athens' S3 support.

**Describe the solution you'd like**

I'd like Athens to support a storage backend which I can use to integrate with our Artifactory instance. Notably, this doesn't necessarily mean a bespoke `artifactory` storage backend—Artifactory also supports using raw HTTP requests with Basic Authentication.

From my perspective, the better solution would be to implement an `http` storage backend that exploits Artifactory's direct HTTP API. This has the added upside of looking remarkably similar to the Go module proxy HTTP API itself, though some extensions would be necessary to support the write operations.

**Describe alternatives you've considered**

Another option is to use [Artifactory's REST API](https://www.jfrog.com/confluence/display/RTF/Artifactory+REST+API) and implement a storage backend more tailored to Artifactory's unique API endpoints.

To me this is an inferior option. While it may provide tighter integration with Artifactory, it means that changes to the Artifactory API over time must be mirrored by updated support in Athens.

**Additional context**

I got curious and wrote up an `http` storage backend to demonstrate what this would look like. I'll submit that as a PR shortly, for a starting point for discussions on what we might be able to achieve here.

**Update**: Submitted #1131.

Issue #1110, opened 2019-03-07
Plugin architecture for storages


marpio:
**Is your feature request related to a problem? Please describe.**
There are still interesting storage options which we do not support i.e. B2. We cannot possibly support all storages in our code base but we can make it possible for the user to provide one i.e. through https://golang.org/pkg/plugin/

**Describe alternatives you've considered**
Implementing them like we do now but it will make the maintenance difficult.

cc: @marwan-at-work since you mentioned that on slack

marwan-at-work:
Another alternative would be is to provide an HTTP interface to storages, almost an extension of the download protocol. And also we can look at how K8s does plugin extensions

tylerchr:
I've implemented an `http` storage type in an effort to support an integration with a new storage backend (Artifactory) in a more generic way than simply integrating with that backend (this came out of a Slack conversation with @arschles). It works essentially as @marwan-at-work described, in that it maps the storage operations to generic HTTP GET/PUT/DELETE calls with BasicAuth, and looks pretty much just like the download protocol with some extensions. Perhaps that would be applicable or useful here?

My intention is to open an proposal issue and then a PR, though I've been held up somewhat (internal review by my employer). However that's nearly resolved—barring any surprises I expect to do that this week.

arschles:
@tylerchr so awesome! Looking forward to this

I'm +1 on this over the Go plugin interface, so that folks can build storage interfaces and have flexibility to run them either on `localhost` or over a network

tylerchr:
I opened #1130 (issue describing my needs) and #1131 (prototype implementation of an HTTP storage type).

Would be interested to hear your thoughts (@marpio, @marwan-at-work) especially since you're coming from the "storage plugins" angle which I hadn't initially considered.

Kunde21:
Missed this when it was first discussed, but there's an option to use https://github.com/hashicorp/go-plugin and define a gRPC service as the interface.  Then, storage plugins can be created and maintained in a separate repo, so long as they fulfill the service definition.

marwan-at-work:
@Kunde21 the readme says: 

>While the plugin system is over RPC, it is currently only designed to work over a local [reliable] network. Plugins over a real network are not supported and will lead to unexpected behavior.

Is that outdated?

Kunde21:
@marwan-at-work the readme kind of buries the gRPC option:
[...]

PR #1131, opened 2019-03-14
Prototype: Implement new storage backend `http`

tylerchr:
Per my needs in #1130 I started hacking on a generic `http` backend. It started as a prototype but turned out not to be all that complex.

The specific decisions I made in this implementation may or may not be ideal. The whole premise of a generic `http` backend is just an idea. I'm looking for feedback on that and assume that some serious iteration would be necessary before considering a merge.

At Qualtrics we use an artifact repository that isn't already supported by Athens. We want to deploy Athens internally. but strong corporate incentives exist to use our internal repository for storage (rather than, say, Athens' S3 support).

I could have just used its API and implemented another platform-specific storage backend, but after working through our requirements and talking with @arschles I saw an opportunity to achieve our goals with a more open-ended solution.

**How is the fix applied?**

What resulted was a reasonably generic `http` storage type. In general, it maps the methods of `storage.Backend` to HTTP calls in the following manner (I'm using the `golang.org/x/net` module as an example):

- `Lister`
    - `List` requests `GET /golang.org/x/net/@v/` and scrapes versions (actually, links to `*.mod` files) from the directory listing
- `Getter`
    - `Info` requests `GET /golang.org/x/net/@v/v0.0.0-20180724234803-3673e40ba225.info` and returns the contents
    - `GoMod` requests `GET /golang.org/x/net/@v/v0.0.0-20180724234803-3673e40ba225.mod` and returns the contents
    - `Zip` requests `GET /golang.org/x/net/@v/v0.0.0-20180724234803-3673e40ba225.zip` and returns the contents
- `Checker`
    - `Exists` requests `GET /golang.org/x/net/@v/v0.0.0-20180724234803-3673e40ba225.mod` and returns `false` if it receives a `404 Not Found` response

The above calls are largely identical to the usual API defined in `go help goproxy`.

The big exception is the handling of `List` due the different semantics used by Athens' storage interface. I went through a couple ideas here:

- Perhaps Athens itself could manage a real `/list` file, but that's a relatively larger change and also raises concurrency questions—I opted to just steer clear of that. Admittedly this decision does have the unfortunate consequence that the HTTP backend is not itself a compliant Go module proxy.
- Since raw file servers traditionally show full directory listings when no index file exists, I thought perhaps this directory could be parsed to collect all the available versions. It seemed a little silly at first but in practice it turned out to be pretty concise and straightforward, so I went for this. I found that directory listing support is pretty universal among file servers (and [easy to implement](https://golang.org/pkg/net/http/#FileServer) too).

The following extensions to the standard protocol support write operations expected by the Storage interface:

- `Saver`
    - `Save` uses `module.Upload` to do the following things, potentially using BasicAuth credentials from the Athens config file:
        - `PUT /golang.org/x/net/@v/v0.0.0-20180724234803-3673e40ba225.info`
        - `PUT /golang.org/x/net/@v/v0.0.0-20180724234803-3673e40ba225.mod`
        - `PUT /golang.org/x/net/@v/v0.0.0-20180724234803-3673e40ba225.zip`
- `Deleter`
    - `Delete` uses `module.Delete` to do the following things, again potentially using BasicAuth credentials:
        - `DELETE /golang.org/x/net/@v/v0.0.0-20180724234803-3673e40ba225.info`
        - `DELETE /golang.org/x/net/@v/v0.0.0-20180724234803-3673e40ba225.mod`
        - `DELETE /golang.org/x/net/@v/v0.0.0-20180724234803-3673e40ba225.zip`

Addresses #1130


arschles:
Hey @tylerchr I saw this is still a draft PR. Are you ready for reviews on it?

arschles:
Also @tylerchr would this close both #1110 and #1130?

tylerchr:
Yeah—I didn't really intend that. Clearly I don't know how Draft PRs work.

It'd obviously close #1130. I suppose whether it addresses #1110 is a question for @marpio, but it seems to me that it would. I see no reason why arbitrary storage backends couldn't be pretty simply implemented as little HTTP translation layers.

I haven't put a whole lot of thought into this, but deploying a couple of containers (Athens + B2 storage layer, for example) does seem a little more cloud-friendly than attempting something with package `plugin`, what with [pods](https://kubernetes.io/docs/concepts/workloads/pods/pod/) and [task groups](https://www.nomadproject.io/docs/job-specification/group.html) and the like. Plus, as I think you pointed out elsewhere, there's the extra flexibility of being able to run them from separate places on the network.

marpio:
@tylerchr I like the `http` backend idea very much. It definitely has advantages to the plugin package. so yeah that would close #1110

arschles:
Sounds like we agree that this will close #1110, that's great to hear!

@tylerchr I'll review this now 😁

config.dev.toml
```diff
@@ -306,3 +307,16 @@ SingleFlightType = "memory"
         # Name of container in the blob storage
         # Env override: ATHENS_AZURE_CONTAINER_NAME
         ContainerName = "MY_AZURE_BLOB_CONTAINER_NAME"
+
+    [Storage.HTTP]
+        # The root URL into which modules will be uploaded
```
arschles: nit: will they be downloaded from this URL as well? If so, can you add that here?

docs/content/configuration/storage.md
```diff
@@ -48,6 +48,27 @@ Disk storage allows modules to be stored on a file system. The location on disk
 
 where `/path/on/disk` is your desired location. Also it can be set using `ATHENS_DISK_STORAGE_ROOT` env
 
+## HTTP
+
+This driver stores files to an HTTP server via standard GET and PUT requests. The files are laid out in a manner identical to the proxy URL used to access them, and the requests are optionally (but hopefully!) authenticated using Basic Auth.
+
+The HTTP storage driver can be used to integrate with systems like Artifactory that offer blob storage over HTTP.
```
arschles: nit: can you mention that this is different from other blob storage systems like S3, Azure blob storage, or Google Cloud Storage that also implement (proprietary) HTTP interfaces?

pkg/storage/http/checker.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/gomods/athens/pkg/observ"
+)
+
+// Exists checks for a specific version of a module
+func (s *ModuleStore) Exists(ctx context.Context, module, vsn string) (bool, error) {
+	var op errors.Op = "http.Exists"
+	ctx, span := observ.StartSpan(ctx, op.String())
+	defer span.End()
+
+	req, _ := http.NewRequest(http.MethodHead, s.moduleRoot(module)+vsn+".mod", nil)
```
arschles: would you mind making a function that does all the appending for you? so you could write something like `s.modulePath(module, version, "mod")`

pkg/storage/http/http.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil, errors.E(op, err)
+	}
+
+	return ms, nil
+}
+
+func (s *ModuleStore) connect() error {
+	const op errors.Op = "http.connect"
+
+	// I guess just GET the base URL and see if it 401's?
```
arschles: Sounds good to me! Maybe we should require that it returns a 204. What do you think?

pkg/storage/http/lister.go
```diff
@@ -0,0 +1,113 @@
...
+// collectLinks traverses the contents of r, which is assumed to be HTML, and gathers
+// all <a> tag references that pass filter.
+func collectLinks(r io.Reader, filter func(string) bool) ([]string, error) {
+
+	t := html.NewTokenizer(r)
+
+	var links []string
+	var err error
+
+floop:
```
arschles: Personally I feel like this label makes reading and reasoning about this code confusing. Do you think there's a way to remove it so that the code is more understandable and not way longer?

pkg/storage/http/lister.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+		return nil, errors.E(op, kind, errors.M(module))
+	}
+
+	// Find all the links to ".mod" files within the current module
+	//
+	// Ok, so admittedly this scheme for listing versions is a little harebrained.
+	// But that said it _is_ pretty standard for directory indexes to be formatted
+	// very closely to what we expect here. I'm not settled on this but it does
+	// work surprisingly well.
```
arschles: Doesn't seem harebrained to me! If this is what you want to go with, let's make sure to add it to our docs (on the hugo site) and explain some options for what you can do if your storage system doesn't spit this format out

pkg/storage/http/http.go
tylerchr: Given the following facts:

- I use the base URL for this check
- The base URL will necessarily refer to a directory
- Elsewhere, `GET`s on directories are expected to yield a directory listing

...it seems a little inconsistent to hope for `No Content` here. For example, my backend of interest (Artifactory) responds with `200 OK` and a directory listing.

You do make a good point though: right now it fails on status codes `>= 400` but I could tighten that to `>= 200 && < 300` to more clearly reflect the expectation of this check. Thoughts?

pkg/storage/http/lister.go
tylerchr: I hadn't initially approached this from the "storage plugin" angle, but I see now that it may be inconvenient/impossible to render a directory listing for certain backends.

Perhaps `List` could support two strategies for obtaining a listing, and attempt them one at a time in order:

1. Fetch a listing in the documented format from `GET /<module>/@v/list`. Athens itself won't update this file or otherwise manage it, but this endpoint may be simpler/cleaner to implement for some storage plugins than a directory listing.
2. Scrape a version list from the directory listing at `GET /<module>/@v/` as currently implemented. This option caters to storage servers based on filesystem semantics like Artifactory or `http.FileServer`.

This approach would make implementing custom storage plugins a bit nicer at the cost of slowing down `List` for cases like mine, where the server only supports the second strategy.

pkg/storage/http/lister.go
tylerchr: Good suggestion. I addressed this with an early return, which did simplify things slightly.

tylerchr:
@arschles I addressed most of your inline comments, some with follow-up questions. I haven't tackled your top-level comments yet.

tylerchr:
I tried using goquery to do the link parsing and personally didn't find it all that compelling. It essentially trades

```
mods, err := collectLinks(resp.Body, func(ref string) bool {
	return strings.HasPrefix(absolute(baseURL, href), baseURL) && strings.HasSuffix(href, ".mod")
})

// convert mod file references to versions
versions := make([]string, len(mods))
for i, m := range mods {
	versions[i] = strings.TrimSuffix(path.Base(m), ".mod")
}
```

and the `collectLinks` and `getAttrs` helpers (47 lines) for

```
var versions []string
doc.Find("a").Each(func(i int, s *goquery.Selection) {
	href, ok := s.Attr("href")
	if !ok {
		return
	}
	if !strings.HasPrefix(absolute(baseURL, href), baseURL) || !strings.HasSuffix(href, ".mod") {
		return
	}
	versions = append(versions, strings.TrimSuffix(path.Base(href), ".mod"))
})
```

and the entirety of the `goquery` library, which is a dependency of nontrivial size. It's not the sort of tradeoff I'd accept in my own code, but I'm happy to do that here if it's what you'd prefer.

pkg/storage/http/http.go
arschles: Got it. Then I have a more generic question - do you intend to build a generic HTTP storage driver in this PR, or one that can use Artifactory as a storage driver?

The two goals overlap, but I think there will need to be a spec for the driver (it looks like you started on that in here 😁) if you want to do the former. With the latter, I think it is fine to make everything in this PR work with Artifactory, without worrying about other possible backends.

Personally, I think it would be best to go with the latter (Artifactory-specific) since that achieves your immediate needs. Later on, we could of course always build a more generic HTTP backend with a spec and all...

arschles:
@tylerchr regarding goquery, all good with me to go with what you have! Thanks for explaining your logic 😁

arschles:
@tylerchr did you want to pick this up again? I'm happy to work with you to finish this up if you're up for it. I think it's really cool and would love to see it go in.

tylerchr:
@arschles Yes, in fact I was just getting back up to speed over the weekend. I’m thinking the best way forward is for me to implement the more generic form of HTTP support here (e.g., with the 204 you proposed earlier, etc.) with an eye toward the pluggable HTTP Storage API we’ve discussed in #1110.

In the intervening weeks I’ve become more comfortable with implementing my Artifactory support externally via that new API rather than take that leaning here—having two similar-but-incompatible HTTP APIs in Athens (Artifactory HTTP + Extensible HTTP) itself doesn’t seem like the best thing for the project.

With that said, what guidance do you have in the design of said HTTP plugin API? I suspect we’re fairly close here already but if it’s going to be used as an extension point I’d think you’ll want to help me get it right. Is there any discussion about it other than #1110, and if not, what would you like to be changed in this PR to achieve that?

arschles:
@tylerchr 

>Yes, in fact I was just getting back up to speed over the weekend.

Awesome! Welcome back 😀 

>having two similar-but-incompatible HTTP APIs in Athens (Artifactory HTTP + Extensible HTTP) itself doesn’t seem like the best thing for the project.

I agree. It'll be nice to have Athens storages be extensible and have Artifactory as one of the extensions

>I suspect we’re fairly close here already

Absolutely! 😀 

>what guidance do you have in the design of said HTTP plugin API?

Looking back at the history of this PR, I agree with @marwan-at-work said in https://github.com/gomods/athens/issues/1110#issuecomment-471779673 that we should use prior art if possible.

What do you think about using the standard download HTTP API, plus an extension for catalogers to use?

arschles:
@tylerchr just checking in, would you be interested in continuing with this PR? If not, absolutely no worries - I'll continue on with it.

jpreese:
@tylerchr similarly, I'm curious about what functionality you felt Artifactory lacked that you also wanted to deploy Athens? We use Artifactory here, and I always saw Athens as the open and free approach.

arschles:
@jpreese @tylerchr are either of you interested in continuing this conversation? We have had conversations about HTTP backends in https://github.com/gomods/athens/issues/1488.

PR #1138, opened 2019-03-14
try to build athens with athens

marwan-at-work:
Just trying out to see if we can build Athens with Athens without hick ups. 
My opinion is that we should probably continue to use vendor until 1.13 but if people like this I'm happy to merge. 

Also, we should look into pricing and not make Aaron go bankrupt. 

cc: @arschles @marpio


marwan-at-work:
@arschles they weren't at first, but after the first CI run now they are. 
Happy to pair on azure pipelines, feel free to ping me on slack

arschles:
@marwan-at-work regarding:

>Also, we should look into pricing and not make Aaron go bankrupt.

We can put the code onto Azure blob storage and not break the bank. Cool with you?

marwan-at-work:
@arschles the problem is that azure backend does not support distributed locking like gcp does now. This is a problem when multiple users submit PRs (or update them) at overlapping times.

arschles:
oh you're right! Ok RE my wallet, GCS isn't gonna break the bank (🤞 !!!!)

Let's plan to merge this and then when https://github.com/gomods/athens/issues/1140 is done, we can switch over to Azure.

Also, removing our vendor directory is a _huge_ deal for us. I'm really proud, this will be so awesome when we merge it!

.drone.yml
```diff
@@ -3,19 +3,28 @@ kind: pipeline
 name: default
 
 steps:
+- name: download-dependencies
```
arschles: @marwan-at-work out of curiousity, why download the dependencies in a separate step from the `go build`? Is it because you also run the `go mod vendor`?

azure-pipelines.yml
```diff
@@ -22,6 +22,6 @@ steps:
   displayName: 'set up the Go workspace'
 
 - powershell: |
-    go test -mod=vendor -race ./...
+    go test -race ./...
```
arschles: @marwan-at-work this one is a bit problematic because now we'll be building Athens from VCS as the upstream. Maybe we should leave in the vendor directory for now so az pipelines still works?

What do you think?

arschles: Another option is we can run an Athens in pipelines after #1126 is done

.drone.yml
marwan-at-work: @arschles it's for security reasons, we don't want to expose the GCS credentials within the build step. Which leads me to another pull request that would only test GCP on merges (or maybe a nicer option where we only run _that_ test manually)

azure-pipelines.yml
marwan-at-work: @arschles yes the azure fix in this PR is not intended to be merged :)

.drone.yml
arschles: ok, cool. makes sense!

azure-pipelines.yml
arschles: ok, thanks not what I meant, sorry for being unclear @marwan-at-work 

I basically meant we could merge this now and then move the dependencies from GCS over to Azure after #1126 is done

Cool with that?

marwan-at-work: @arschles hmm sorry not sure I understand. What would moving the deps from GCS to Azure fix? I imagine Azure pipelines is a containerized environment so whether the Deps are in GCS or Azure Blobs, it'll need credentials to access them. But maybe im missing something. Lemme know!

arschles: Sorry about that @marwan-at-work. I'll try to explain more explicitly, inline:

>What would moving the deps from GCS to Azure fix?

Just the billing issue. I don't think we 100% definitely need to do this, but since we can get Azure resources for free at the moment, I figure it should be on the table. At the same time, though, I don't want that to slow _this_ PR down.

>I imagine Azure pipelines is a containerized environment so whether the Deps are in GCS or Azure Blobs, it'll need credentials to access them

Yup, you're right. But regardless of whether we do GCS or Azure blob storage, could we set permissions to read-only by the public?

marwan-at-work: >could we set permissions to read-only by the public? Or do we then need to figure out a different way to get dependencies into storage?

@arschles I don't think so, because when we add a new dependency we need to be able to persist to the storage. We can potentially do something weird or manual to avoid it but I don't think that's necessary. My opinion is that we should give read/write permissions and ensure that they don't get exposed during CI. But I'm happy to explore alternatives of course. 

As for the cost issue, that sounds good to me. Has it been expensive? I'm happy to contribute to the cost. But also if moving to Azure Blobs removes that headache, then sounds great. Maintainers, would need some access to Azure blobs to be able to debug issues when they arise.

However, one big blocker to this is the fact that we don't have a distributed SingleFlight for Azure. But with GCP it's _both_ a Backend Storage and a Distributed SingleFlight. 

If you make an Athens Azure account, I'm happy to look more into that side of things :)

marpio: @marwan-at-work 
> However, one big blocker to this is the fact that we don't have a distributed SingleFlight for Azure.

There's an issue for that #1140 
I'm working on it and should have something in the next few days.

arschles: @marwan-at-work the creds thing makes sense. I was hoping there would be a way to make the bucket read only to the public, but no dice!

Cost is not a big deal at the moment. I know we haven't been building Athens with Athens yet and so storage and traffic in GCS hasn't been as it will be, but this month's bill is $0.01 so I think we should see how it goes 😄

arschles:
@marwan-at-work all good with merging this?

marwan-at-work:
@arschles go for it if everyone's happy.

arschles:
Update: we talked about this PR in the developer meeting on March 28, 2019, and decided to wait for https://github.com/gomods/athens/pull/1152 to go in, so that we fix that bug before merging something so large.

arschles:
@marwan-at-work what do you think about this? Now that we've fixed #1152 and done the v0.4.0 release, I'm totally cool merging this is you are!

marwan-at-work:
@arschles I'm going to open a new PR for this :)

PR #1141, opened 2019-03-16
rm buffalo question

marpio:
We do not use Buffalo anymore so we shouldn't ask about it in our bug tmpl

**How is the fix applied?**

Removed the question


arschles:
@marpio I noticed that there are some other places where we still have buffalo commands (`init.ps1` for example). Would you be able to open an issue to remove those?

arschles:
From https://github.com/gomods/athens/pull/1141#pullrequestreview-215847210:

>Going to leave it up for one more review to make sure I'm not crazy

I think I'm not crazy, so I'm going to merge 😄

PR #1143, opened 2019-03-17
fs: Use no. of objects for Exists check

### LINKED ISSUE #1032: Exists checks only for .mod

marpio:

**How is the fix applied?**

I'm checking if the no. of objects equals 3 (zip, mod, info)

Partially fixes #1032


pkg/storage/fs/checker.go
```diff
@@ -14,6 +14,14 @@ func (v *storageImpl) Exists(ctx context.Context, module, version string) (bool,
 	defer span.End()
 	versionedPath := v.versionLocation(module, version)
 
+	ok, err := afero.Exists(v.filesystem, versionedPath)
```
marwan-at-work: is this needed? Can we not just return false if the `afero.ReadDir` below returned a "NotExists" error?

michalpristas:
how about checking explicitly for mod/zip/info. what if somebody tempered with storage and we have 3 cat gifs there now. even more probable with FileSystem.

marwan-at-work:
@michalpristas very interesting idea. Probably worth opening a separate issue for? 
Not sure what kinda efficient solution would achieve this. Either a separate auditing endpoint, or a background job etc. We can also just do nothing since the storage is fully owned by the Athens user.

michalpristas:
what I meant was to check whether .mod .zip and info files exists instead of blindly checking for a number of files inside a directory.

marwan-at-work:
@michalpristas ah. Well since Athens owns the storage saving part and never lets the client save anything, we might be ok just checking the length. The only two possibilities for something getting mishandled are: 
1. If the DB was directly manipulated
2. If the module in VCS itself was incorrect 


For 1, I wanna think more about that...I keep thinking we might not even need the storage.Checker but I could be wrong. This could also be were auditing from the Athens side might make sense but im not sure yet. 

For 2, I think that's not an issue Athens has to worry about (maybe im wrong), but basically whoever owns a module in VCS the first time asks for it, they win. That's kinda by design.

marpio:
@michalpristas I'm cool with checking for each part explicitly but don't we assume that the storage is only in athens control? What if I name that cat picture v1.0.0.info?

Maybe users could prevent tempering with the files by using immutable storage? AFAIK azure, aws and gcp offer that. We will probably need to delete modules at some point so not sure how this would play out though. :thinking:

michalpristas:
i'm ok with followup, something from my experience is: do not assume anything. somewhere sometimes will break.

marpio:
@michalpristas 

> do not assume anything. somewhere sometimes will break.

I'm with you on that :smile: It's just that if we do not trust the storage then there are many more scenarios where something can go wrong i.e. what if someone modifies the files? or deletes them and athens will fetch force-pushed version?
I think it's a broader problem and checking just the names won't solve it.

arschles:
>I think it's a broader problem and checking just the names won't solve it.

@marpio I agree that it won't solve a broader problem, but it may help an operator detect one

marwan-at-work:
@arschles at the moment, it won't help an operator detect one because when Exists returns false, we just fetch from VCS and override whatever was in there.

arschles:
Sorry "detect" was a bad choice of words. What I mean is that we can log when all 3 files are not present. This would be slightly similar to a disk or database integrity check. Athens can still work even if this log pops up, and it tries to recover by going to the VCS (an imperfect solution but the best that the server can do in this situation) as you said @marwan-at-work.

arschles:
If we're ok with this logging, I can go in and add it in follow-ups

PR #1144, opened 2019-03-17
mongo: Use no. of objects for Exists check

### LINKED ISSUE #1032: Exists checks only for .mod

marpio:

**How is the fix applied?**

I'm checking if the no. of objects equals 3 (zip, mod, info)

Partially fixes #1032


marpio:
CLosing as this is not needed for mongo.

PR #1148, opened 2019-03-19
helm chart: configure ssh-secured git servers

minhdanh:

This PR adds support for working with private ssh-secured git servers in the helm chart. The idea is taken from here: https://docs.gomods.io/configuration/authentication/#altassian-bitbucket-and-ssh-secured-git-vcs-s

**How is the fix applied?**

One or many git servers are added to `sshGitServers`, and the corresponding config files (git config and ssh config) and ssh keys will be created. Athens then will use these configs and keys to download the source from the git servers.

```
sshGitServers:
  ## hostname of the git server
  - host: git.example.com
    ## ssh username
    user: git
    ## ssh private key for the user
    privateKey: |
      -----BEGIN RSA PRIVATE KEY-----
      -----END RSA PRIVATE KEY-----
    ## ssh port
    port: 22
```

### DIFF
--- a/charts/athens-proxy/Chart.yaml
+++ b/charts/athens-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: athens-proxy
-version: 0.2.0
+version: 0.2.1
 appVersion: 0.2.0
 description: The proxy server for Go modules
 icon: https://bitnami.com/assets/stacks/wordpress/img/wordpress-stack-220x234.png
--- a/charts/athens-proxy/Chart.yaml
+++ b/charts/athens-proxy/Chart.yaml
@@ -1,6 +1,6 @@
 name: athens-proxy
 version: 0.2.1
-appVersion: 0.3.0
+appVersion: v0.3.1
 description: The proxy server for Go modules
 icon: https://bitnami.com/assets/stacks/wordpress/img/wordpress-stack-220x234.png
 keywords:
--- a/charts/athens-proxy/Chart.yaml
+++ b/charts/athens-proxy/Chart.yaml
@@ -1,6 +1,6 @@
 name: athens-proxy
 version: 0.2.1
-appVersion: 0.2.0
+appVersion: 0.3.0
 description: The proxy server for Go modules
 icon: https://bitnami.com/assets/stacks/wordpress/img/wordpress-stack-220x234.png
 keywords:
--- a/charts/athens-proxy/values.yaml
+++ b/charts/athens-proxy/values.yaml
@@ -58,3 +58,16 @@ jaeger:
   # you must set this on the command line when you run 'helm install'
   # for example, you need to run 'helm install --set jaeger.url=myurl ...'
   url: "SET THIS ON THE COMMAND LINE"
+
+# private git servers over ssh
+sshGitServers:
+  ## hostname of the git server
+  - host: git.example.com
+    ## ssh username
+    user: git
+    ## ssh private key for the user
+    privateKey: |
+      -----BEGIN RSA PRIVATE KEY-----
+      -----END RSA PRIVATE KEY-----
+    ## ssh port
+    port: 22
--- a/charts/athens-proxy/values.yaml
+++ b/charts/athens-proxy/values.yaml
@@ -2,7 +2,7 @@ replicaCount: 1
 image:
   registry: docker.io
   repository: gomods/athens
-  tag: v0.3.0
+  tag: "{{ .Chart.AppVersion }}"
   ## Specify a imagePullPolicy
   ## Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
   ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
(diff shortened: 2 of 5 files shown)


rimusz:
could you please also bump chart and app version in https://github.com/gomods/athens/blob/master/charts/athens-proxy/Chart.yaml#L2-L3?

charts/athens-proxy/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: athens-proxy
-version: 0.2.0
+version: 0.2.1
 appVersion: 0.2.0
```
minhdanh: @rimusz I bumped `version` to `0.2.1`. But for `appVersion`, shouldn't it be the version of athens image: https://github.com/gomods/athens/blob/master/charts/athens-proxy/values.yaml#L5?

rimusz: yes, app version has to be the same as the image tag

minhdanh: @rimusz Done

rimusz:
I see the last version is `v0.3.1` you should update to it

rimusz:
Ideally to fetch the `appVersion` in the chart you could update:
https://github.com/gomods/athens/blob/master/charts/athens-proxy/values.yaml#L5 to `tag: "{{ .Chart.AppVersion }}"`

and https://github.com/gomods/athens/blob/master/charts/athens-proxy/templates/deployment.yaml#L22 to `image: "{{ .Values.image.registry }}/{{ .Values.image.repository }}:{{ tpl .Values.image.tag }}"`

Then only https://github.com/gomods/athens/blob/master/charts/athens-proxy/Chart.yaml#L3 needs to be updated with the new athens version

minhdanh:
@rimusz Done. Please check

rimusz:
LGTM
@arschles you can merge the PR :)

charts/athens-proxy/values.yaml
```diff
@@ -2,7 +2,7 @@ replicaCount: 1
 image:
   registry: docker.io
   repository: gomods/athens
-  tag: v0.3.0
+  tag: "{{ .Chart.AppVersion }}"
```
andreasolund: I think we should avoid using templates in values files, since it makes it harder for users of this chart. Keeping `image.tag` and `appVersion` in sync manually is better for readability.

rimusz: why you think so?
it makes only one place to change the tag. it is used widely in helm charts

rimusz: add a comment e.g. `# Note that by default we use appVersion to set image tag` above the tag would make things easier to understand.
even the current chart version is not in sync between `appVersion` and `tag`

andreasolund: Yes, I understand why you want to reuse `appVersion`. I'm just pointing out that this change makes it harder to reuse the helm chart in an enterprise environment where you have to use your own images and where you usually override `image.registry`, `image.repository` and `image.tag`. :)

rimusz: right, I see your point :-), let's use a bit different solution then changing files:
values.yaml
```
...
image:
  registry: docker.io
  repository: gomods/athens
  # Note that by default we use appVersion to get image tag
  # But you can set the tag if there are needs to overwrite it
  # tag:
....
```

deployment.yaml
```
...
image: "{{ .Values.image.registry }}/{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
...
```

But default tag will be taken from `appVersion` and can be overwritten in the `values.yaml` as before

rimusz: @minhdanh can you make those ^^ changes, please? thanks

minhdanh: It's ok for me, but there's another thing: https://github.com/gomods/athens/blob/master/charts/athens-proxy/templates/_helpers.tpl#L6

The image tag is used to decide the endpoint of livenessProbes and readinessProbes.
`v0.3.0` is hardcoded. Looks like the endpoints won't be rendered with image tag/appVersion  `v0.3.1`.

rimusz: remove https://github.com/gomods/athens/blob/master/charts/athens-proxy/templates/_helpers.tpl#L5-L10` if it is supported in v0.3.0 and onwards?

or move that part to the `values.yaml` with instructions to setup for all versions?

rimusz:
also @minhdanh you need to make install oh ssh keys optional, right now if I set `sshGitServers: {}`, initContainer still runs, `ssh-keys` volume gets mounted, and same for the ssh configmap

rimusz:
I have created #1154 which also addresses this PR

minhdanh:
Ok, so anything else needs doing on my side?

rimusz:
we can close this one, as for me was easier to make changes by myself in the #1154 when explaining all needed changes :)

minhdanh:
Closing.

arschles:
Hey @minhdanh, thanks for submitting this PR though! I really appreciate it. I hope you're still interested in doing one. Send me a message on gophers slack `@arschles` any time if you'd like

PR #1150, opened 2019-03-19
Azure Pipelines: use go1.12

marpio:
Azure pipelines uses g1.11 but Athens requires 1.12
Using windows containers on Azure Pipelines is slow - it takes around ~12min (8min just to pull the image) ref: #1126 
This PR uses chocolatey to install go 1.12 on the VM


marpio:
@arschles yes, let's merge that as a temporary solution and work on #1126

PR #1152, opened 2019-03-19
Fixing GCP Flakes

### LINKED ISSUE #1151: Flake in GCP tests
**Describe the bug**

Check out line 224-245 in this build: https://cloud.drone.io/gomods/athens/138/1/5

I've seen this across a few different PRs. Without looking in depth at the GCP code, I wonder if we're hitting a race condition?

**Error Message**

```
 expected: []string{"v1.1.0", "v1.2.0", "v1.3.0"} 
 actual : []string{"v1.3.0"} 
```

**To Reproduce**

It's tough to reproduce this because it's a flakey test. I suspect that it's due to a race condition.

**Expected behavior**

The tests should always pass if the code is correct

 - OS: Mac OS X 64 bit
 - Go version : 1.12
 - Buffalo Version : N/A
 - Proxy version : canary built from c59aa69848b3909f31f3a4ec796e167bf56075b3
 - Storage (fs/mongodb/s3 etc.) : GCP

**Additional context**

arschles:

There have been a few PRs failing intermittently on GCP tests. This smells like a race condition since every GCP test was using the same `athens_drone_bucket` bucket in Google Cloud Storage

**How is the fix applied?**

I've used [moniker](https://github.com/technosophos/moniker) (written by the super awesome @technosophos - thanks!) to generate human-readable, unique bucket names on each test run.

TODOs:

- [ ] Refactor code so that tests can manually create the bucket

Fixes #1151


marwan-at-work:
@arschles one more comment, this PR can be to disable GCP tests and only allow them on merges to master. This way, _I think_, there will be no race conditions and we don't expose GCP credentials. Afterwards, we can work on making a gcp bot to run "privileged tests" when we know the PR is not malicious.

marwan-at-work:
So feel free to pick one of the solutions I mentioned above or suggest anything else otherwise 👍 💯

arschles:
@marwan-at-work good idea on moving bucket creation to `getStorage`. I'll see if I can make that work! 😁 

Also, it feels like running the GCP tests only in master would be removing test functionality from the PR tests. But I think the problem you're talking about is that right now, we're exposing GCP credentials to PRs?

marwan-at-work:
@arschles yes we are. Anyone who opens a PR could log out the credentials. *Ideally*, we create a PR bot that runs that one test on demand. So we have 3 options: 

1. Create buckets during tests (keeps risk of exposed creds) 
2. Only allow merge to master (no risk of exposed credentials, but can't test GCP in PRs) 
3. Develop a testbot (best option, requires some work)

I think we can do _either_ 1 or 2, then eventually do 3.

arschles:
@marwan-at-work are you cool with (1) - since it's not worse than what is currently in master with respect to leaking credentials) - and then we can do a follow-up to do the bot in (3)?

Also, I don't know much about bots, so if you can send me a link to docs I'll do the follow up

marwan-at-work:
@arschles yes, 1 then 3 sounds great. Let me know when the PR is good to go and I'll review. Thanks!

arschles:
@marwan-at-work RE this part:

>I think a nicer solution would be to create the bucket itself directly inside getStorage since it is part of the gcp package and has access to the unexported GCS client

I refactored the `New(...)` function into a `newClient(...)` function, which `New(...)` calls under the covers. That way, test code can call `newClient(...)`, create the bucket, and then call `New(...)`

pkg/storage/gcp/gcp_test.go
```diff
@@ -48,8 +53,24 @@ func getStorage(t testing.TB) *Storage {
...
+	// Create the GCP bucket, so that the below call to New doesn't fail.
+	//
+	// The 'New' function calls bucket.Attrs(), so the bucket has to exist
+	bkt := cl.Bucket(cfg.Bucket)
+	if err := bkt.Create(ctx, cfg.ProjectID, nil); err != nil {
+		t.Fatal(err)
+	}
+
+	// create a new Storage implementation for GCP
+	s, err := New(ctx, cfg, config.GetTimeoutDuration(30))
```
marwan-at-work: New will call `newClient` one more time, and we'll end up with two clients.

I think you can avoid all of that and just replace this line with the following: 

```go
return &Storage{bucket: bkt, timeout: config.GetTimeoutDuration(30)}
```

This is because we just created a new bucket, and so we know it exists so no need to call `New` and have it instantiate a newClient then check if the bucket exists.

pkg/storage/gcp/gcp_test.go
```diff
@@ -39,7 +41,10 @@ func (s *Storage) clear() error {
 			return err
 		}
 	}
-	return nil
+	// since this is a unique bucket name (see getTestConfig), we need to
+	// not only delete all the objects in the bucket (above), we need to then
+	// delete the bucket itself
+	return s.bucket.Delete(ctx)
```
marwan-at-work: Actually, I don't know if this will work...clean is going to be called multiple times by the compliance tests. You can put this functionality inside `testing.M` if you want, or you'd have to have to different "clean" functions that the compliance test must accept, one that will clean contents, and one that will do a "final clean up"

marwan-at-work:
I'll be taking this over so I'll close this PR and open a new one :v:

arschles:
Replaced by https://github.com/gomods/athens/pull/1176

PR #1154, opened 2019-03-21
Chart - gocenter and configured ssh-secured git servers support

rimusz:
- Optional support for using [gocenter.io](https://gocenter.io) as upstream proxy 
- Optional support for configured ssh-secured git servers #1148 (thanks @minhdanh)
- Chart's readme


rimusz:
Still needs to be addressed:
https://github.com/gomods/athens/blob/master/charts/athens-proxy/templates/_helpers.tpl#L6

The image tag is used to decide the endpoint of livenessProbes and readinessProbes.
v0.3.0 is hardcoded. Looks like the endpoints won't be rendered with image tag/appVersion v0.3.1.

rimusz:
`appVersion` is not used for setting tag, as that breaks `livenessPath` and `readinessPath` setting in deployment

Ready to be reviewed!

charts/athens-proxy/templates/deployment.yaml
```diff
@@ -8,18 +8,43 @@ metadata:
...
+          {{- range $server := .Values.sshGitServers }}
+          - name: ssh-git-servers-secret
+            mountPath: /root/.ssh/id_rsa-{{ $server.host }}
+            subPath: id_rsa-{{ $server.host }}
+          {{- end }}
+      {{- end }}
       containers:
       - name: {{ template "fullname" . }}
-        image: "{{ .Values.image.registry }}/{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+        image: "{{ .Values.image.registry }}/{{ .Values.image.repository }}:{{ tpl .Values.image.tag . }}"
```
minhdanh: I see you're setting image tag to v0.3.1 here: https://github.com/gomods/athens/pull/1154/files#diff-71a5c45e593365379cc2572e15c7db38R5
So `tpl` won't work.

charts/athens-proxy/templates/_helpers.tpl
```diff
@@ -3,8 +3,8 @@
 {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 {{- define "livenessPath" -}}
-{{- if eq .Values.image.tag "v0.3.0" -}}/{{- else -}}/healthz{{- end -}}
+{{- if contains "v0.2" .Values.image.tag -}}/healthz{{- else -}}/{{- end -}}
```
minhdanh: Maybe it's `v0.3` instead of `v0.2`?

rimusz: by my understanding v0.3.x` has different readiness/liveness check endpoints just `/`,
so the new logic is fine :)

charts/athens-proxy/templates/deployment.yaml
rimusz: it works, but I removed it anyway now

charts/athens-proxy/README.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+### GoCenter as upstream proxy server
+
+[GoCenter](https://gocenter.jfrog.com) is a public central repository of Go modules which provides repeatable and fast module resolution. GoCenter automatically processes and validates Go projects directly from source code repositories into immutable, versioned modules, making them freely available to the Go developer community. In this way, it empowers the Go community to reliably generate reproducible builds of their Go applications.
+
+```yaml
+goCenter:
+  # Set https://gocenter.io as upstream proxy
+  enabled: enable
```
ankushchadha: should boolean be used instead?

charts/athens-proxy/templates/_helpers.tpl
```diff
@@ -3,8 +3,8 @@
 {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 {{- define "livenessPath" -}}
-{{- if eq .Values.image.tag "v0.3.0" -}}/{{- else -}}/healthz{{- end -}}
+{{- if contains "v0.2.0" .Values.image.tag -}}/healthz{{- else -}}/{{- end -}}
 {{- end -}}
 {{- define "readinessPath" -}}
-{{- if eq .Values.image.tag "v0.3.0" -}}/{{- else -}}/readyz{{- end -}}
+{{- if contains "v0.2.0" .Values.image.tag -}}/readyz{{- else -}}/{{- end -}}
```
marpio: @rimusz `v0.2.0` did not have the `/readyz` endpoint (but /healthz was there)
https://github.com/gomods/athens/blob/c6644f82f9f9cf583a12566d844d9f132d06daa0/cmd/proxy/actions/app_proxy.go#L22-L23
`v0.3.0` and up does have /healthz and /readyz enpoints
https://github.com/gomods/athens/blob/a4450c114235eb18093328f1755c5258b771787b/cmd/proxy/actions/app_proxy.go#L25-L26

rimusz: so the old one:
```
{{- if eq .Values.image.tag "v0.3.0" -}}/{{- else -}}/healthz{{- end -}}
{{- end -}}
{{- define "readinessPath" -}}
{{- if eq .Values.image.tag "v0.3.0" -}}/{{- else -}}/readyz{{- end -}}
{{- end -}}
```
is incorrect

rimusz: ideally the new chart version should only support `v0.3.x`

rimusz: updated per @marpio comments in the commit `hardcode /healthz to deployment.yaml`

charts/athens-proxy/templates/deployment.yaml
```diff
@@ -8,22 +8,47 @@ metadata:
...
+      release: "{{ .Release.Name }}"
   template:
     metadata:
       labels:
         app: {{ template "fullname" . }}
         chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
         release: "{{ .Release.Name }}"
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/config-gocenter.yaml") . | sha256sum }}
+        checksum/config: {{ include (print $.Template.BasePath "/config-ssh-git-servers.yaml") . | sha256sum }}
```
minhdanh: This maybe will only work for `config-ssh-git-servers.yaml`, not for `config-gocenter.yaml` as the two annotation names are the same.

rimusz: it will work for both, as files listed there have different names

rimusz: `helm install athens-proxy/ -n athens --dry-run --debug`
```
...
  template:
    metadata:
      labels:
        app: athens-athens-proxy
        chart: "athens-proxy-0.2.1"
        release: "athens"
      annotations:
        checksum/config: c1c40d75065587c49691755b457d72d8f188c8762ef24b61d45c476393a7d90c
        checksum/config: 95356df6ef3f8b1b87e99b54f0d42bc918c839a0681ab705c79125d3abfbcfdd
        checksum/secret: 51c479514eb71530bfe8f6e4777fd957ad89413993b0804256d20b7521cc1739
...
```
`gocenter` and `sshGitServers` enabled there, and all checksum are different

charts/athens-proxy/README.md
```diff
@@ -0,0 +1,150 @@
+# Athens Proxy Helm Chart
+
+## What is Athens?
+
+[Athens](https://docs.gomods.io) is a Server for Your Go Packages.
+
+Athens provides a server for [Go Modules](https://github.com/golang/go/wiki/Modules) that you can run. It serves public code and your private code for you, so you don't have to pull directly from a version control system (VCS) like GitHub or GitLab.
+
+## Prerequisites Details
```
arschles: @rimusz can you remove the "Details" from the title here?

charts/athens-proxy/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+* Kubernetes 1.10+
+
+## Requirements
+
+- A running Kubernetes cluster
+- [Kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) installed and setup to use the cluster
+- [Helm](https://helm.sh/) installed and setup to use the cluster (helm init) or [Tilerless Helm](https://github.com/rimusz/helm-tiller)
```
arschles: Two requests here:

- Can you replace `Tilerless` with `Tillerless`?
- Can you include links to documentation on how to install helm? Your tillerless helm documentation looks great fwiw 😁 💯

charts/athens-proxy/README.md
```diff
@@ -0,0 +1,150 @@
...
+    ## ssh username
+    user: git
+    ## ssh private key for the user
+    privateKey: |
+      -----BEGIN RSA PRIVATE KEY-----
+      ...
+      -----END RSA PRIVATE KEY-----
+    ## ssh port
+    port: 22
+```
```
arschles: @rimusz a lot of this file starting with "Deploy Athens" is also in https://docs.gomods.io/install/install-on-kubernetes/. The added parts look like the "GoCenter as upstream proxy server" and "Private git servers over ssh support" sections. Could you remove everything from "Deploy Athens" and below, add the new parts to https://docs.gomods.io/install/install-on-kubernetes/, and link to that install documentation in here?

charts/athens-proxy/values.yaml
```diff
@@ -47,6 +48,12 @@ storage:
 netrc:
   # if enabled, it expects to find the content of a valid .netrc file imported as a secret named netrcsecret
   enabled: false
+  existingSecret: netrcsecret
+
+goCenter:
```
arschles: @rimusz I think this section is more general, since you can set any URL you'd like. So could we rename the section name to "upstream"?

charts/athens-proxy/values.yaml
```diff
@@ -47,6 +48,12 @@ storage:
...
+  # Set https://gocenter.io as upstream proxy
+  enabled: false
+  url: "https://gocenter.io"
```
arschles: can we add a comment here to indicate possible values? for example:

```yaml
# This is where you can set the URL for the upstream module repository.
# If 'enabled' is set to true, Athens will try to download modules from the upstream when it doesn't find them in its own storage.
# You can use 'https://gocenter.io' to use JFrog's GoCenter as an upstream here, or you can also use another Athens server as well
```

rimusz:
Right, ok I'm going to make a new PR which only does the upstream proxy with Gocenter as one of examples, readme, and other fixes/changes:
```
Updating the docker image tag
Adding a README for the helm chart
Generalizing the readiness path template to use contains
Adding checksum for the upstream proxy secret
Adding support for setting the upstream URL in the helm chart values file
Adding selectors to lots of the manifests that didn't have them already
```
All these above should come as one PR.

The `configured ssh-secured git servers` then can be done afterwards in a new PR, @minhdanh can continue there and I will review :)

rimusz:
new PR #1161 for the upstream support

arschles:
@rimusz what is left of this PR that wasn't done in #1161? I think the SSH part at least, right?

rimusz:
yes, only ssh-secured git servers support is left

arschles:
@rimusz great! if you resolve conflicts here, seems like only the SSH changes will be left in this PR? Otherwise, if you open a new PR, let's close this one. Let me know what you decide to do 😄

rimusz:
new PR makes more sense :)

arschles:
@rimusz sounds good. I am gonna close this for now then. Ping me when you're ready with the new PR!

Issue #1157, opened 2019-03-26
Remove the line number from the config link in the Kubernetes docs


arschles:
**Describe the bug**

The link to the `config.toml` file in https://github.com/gomods/athens/blob/master/docs/content/install/install-on-kubernetes.md#give-athens-access-to-private-repositories-via-github-token-optional goes to a specific line number. Making it do that might be a problem in the future because we would need to keep our docs up to date with code as it changes.

We should change the link to remove the line number, and if it's appropriate update the documentation to describe approximately where in the config file to look.

**Error Message**

N/A

**To Reproduce**

N/A

**Expected behavior**

N/A

**Environment (please complete the following information):**
N/A

**Additional context**

This is a follow-up to https://github.com/gomods/athens/pull/1103#discussion_r269309877

PR #1158, opened 2019-03-26
Removing the line number from the config link

arschles:

There's a link to the config file in the Kubernetes install docs 
(https://docs.gomods.io/install/install-on-kubernetes/), and the link has a line
number in it. The code in the config file has changed and now the link points
to the wrong line

**How is the fix applied?**

I removed the line number in the link (so it points to just the file) and
instead just describes where the config field is.

Fixes https://github.com/gomods/athens/issues/1157


manugupt1:
thanks @arschles for this!

PR #1162, opened 2019-03-27
pkg/stash: Add Azure Blob implementation

### LINKED ISSUE #1140: Azure Storage Distributed Locking
Azure storage supports [Lease Blob operation](https://docs.microsoft.com/en-us/rest/api/storageservices/lease-blob) which means we could use it similarly to GCS (issue #1123) to support `SingleFlight` on Azure storage.

marpio:
Implements distributed SingleFlight using Azure Blob Storage

Fixes #1140


pkg/stash/with_azureblob_test.go
```diff
@@ -0,0 +1,133 @@
...
+		mod = "stashmod"
+		ver = "v1.0.0"
+	)
+	cfg := getAzureTestConfig()
+	strg := getAzureStorage(t)
+	strg.Delete(ctx, mod, ver)
+	defer strg.Delete(ctx, mod, ver)
+
+	// sanity check
+	_, err := strg.GoMod(ctx, mod, ver)
```
manugupt1: Is this check really necessary?

pkg/stash/with_azureblob.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/gomods/athens/pkg/storage"
+	"github.com/google/uuid"
+)
+
+// WithAzureBlobLock returns a distributed singleflight
+// using a Azure Blob Storage backend. See the config.toml documentation for details.
+func WithAzureBlobLock(conf *config.AzureBlobConfig, timeout time.Duration, checker storage.Checker) (Wrapper, error) {
+	const op errors.Op = "stash.WithAzureBlobLock"
+
+	accountURL, err := url.Parse(fmt.Sprintf("https://%s.blob.core.windows.net", conf.AccountName))
```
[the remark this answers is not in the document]
marpio: We have the same problem in our storage. I would rather fix this in a separate PR

pkg/stash/with_azureblob.go
```diff
@@ -0,0 +1,156 @@
...
+	defer span.End()
+	_, err := blobURL.RenewLease(ctx, leaseID, azblob.ModifiedAccessConditions{})
+	return err
+}
+
+func (s *azblobLock) acquireLease(ctx context.Context, blobURL azblob.BlockBlobURL) (string, error) {
+	const op errors.Op = "azblobLock.acquireLease"
+	ctx, span := observ.StartSpan(ctx, op.String())
+	defer span.End()
+	tctx, cancel := context.WithTimeout(ctx, 300*time.Second)
```
marwan-at-work: nit: `5*time.Minute` is a little more readable than `300*time.Second`

pkg/stash/with_azureblob.go
```diff
@@ -0,0 +1,156 @@
...
+	defer cancel()
+
+	// first we need to create a blob which can be then leased
+	_, err := blobURL.Upload(tctx, bytes.NewReader([]byte{1}), azblob.BlobHTTPHeaders{}, nil, azblob.BlobAccessConditions{})
```
marwan-at-work: So it seems that the lease acquiring process has a timeout of 5 minutes, but the lease itself doesn't have a timeout. In other words, if the stashing process takes too long, the lease will keep getting refreshed every 10 seconds forever right? 

I know that our _main_ stasher has its own timeout of 10 minutes so just double checking that I'm following the logic here.

marpio: @marwan-at-work 
Good question! I figured I will rely on the inner stasher timeout. Probably not a good idea if for some reason the stasher doesn't have a timeout and hangs forever.
For some reason I was convinced that the stasher timeout is 5 and not 10 min.
Which leads me to another question - if the stasher has timeout = 10 min shouldn't the lease acquiring process also have the same timeout? It will need to wait for the first request so max 10 min. I'm asking b/c the redis lock uses 5 min.

marwan-at-work: @marpio yeah that's a good point. We should probably make the timeouts more consistent. I don't know if 5 is better than 10 but either way is cool with me. The whole point is that we have some sane timeout on the stashing process...I still think we don't need to make it configurable (just yet), so maybe the longer the better.

PR #1163, opened 2019-03-27
Include path in the proxy command example

shrajfr12:

Minor typo in command example

**How is the fix applied?**

Replaced "/proxy" with "<path_to_athens>/proxy"

### DIFF
--- a/docs/content/configuration/upstream.md
+++ b/docs/content/configuration/upstream.md
@@ -23,6 +23,6 @@ Additionally in the current or new config file, set the following parameters as 
 1. Restart Athens specifying the updated current or new config file.
 
     ```
-    /proxy  -config_file <path-to updated  current or new configfile>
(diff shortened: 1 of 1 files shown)


docs/content/configuration/upstream.md
```diff
@@ -23,6 +23,6 @@ Additionally in the current or new config file, set the following parameters as
 1. Restart Athens specifying the updated current or new config file.
 
     ```
-    /proxy  -config_file <path-to updated  current or new configfile>
+     ./proxy  -config_file <path-to updated  current or new configfile>
```
manugupt1: Instead of `./proxy`, I would probably change it to `<path_to_athens>/proxy`

shrajfr12: Makes sense. Will update soon.

PR #1172, opened 2019-03-31
storage/gcp: fix lister prefix

marwan-at-work:
I noticed that our `/@v/list` endpoint was adding versions that didn't exist for `github.com/gobuffalo/buffalo` which turns out to be a bug in our GCP lister because it was considering all modules that had the previous prefx so it took version from other mods like `github.com/gobuffalo/buffalo-plugin` 

Most of our backend implementations have this correct, but I believe Azure/S3 might have this incorrect as well. We should hopefully add those implementations in our CI soon

### DIFF
--- a/pkg/storage/gcp/lister.go
+++ b/pkg/storage/gcp/lister.go
@@ -2,7 +2,6 @@ package gcp
 
 import (
 	"context"
-	"path/filepath"
 	"strings"
 
 	"cloud.google.com/go/storage"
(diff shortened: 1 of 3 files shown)


pkg/storage/gcp/lister.go
```diff
@@ -17,7 +18,8 @@ func (s *Storage) List(ctx context.Context, module string) ([]string, error) {
 	ctx, span := observ.StartSpan(ctx, op.String())
 	defer span.End()
 
-	it := s.bucket.Objects(ctx, &storage.Query{Prefix: module})
+	modulePrefix := filepath.Join(module, "/@v")
```
marpio: Isn't filepath.Join Os dependent? On Windows it will use `\` instead of `/`.

marwan-at-work: @marpio great catch, I live in my own world of `/` so it didn't occur to me. Honestly, it's a good point for windows testing in CI :) -- anywho I updated the code. Let me know if there's anything else, thanks!

PR #1174, opened 2019-04-01
Metrics

### LINKED ISSUE #969: Support for Metrics using DataDog and StackDriver
After #816 will be merged we should add support for DataDog and StackDriver. 

`RegisterStatsExporter` should support `datadog` and `stackdriver` and export stats to them.

**Additional context**
https://opencensus.io/exporters/supported-exporters/go/datadog/#0
https://opencensus.io/exporters/supported-exporters/go/stackdriver/#0

fedepaol:
Issue #969: Adding support to metrics for datadog and stackdriver.

**How is the fix applied?**

Added the aforementioned exporters for metrics.

Fixes #969

### DIFF
--- a/pkg/observ/stats.go
+++ b/pkg/observ/stats.go
@@ -18,17 +18,19 @@ import (
 // Currently it supports: prometheus
 func RegisterStatsExporter(r *mux.Router, statsExporter, service string) (func(), error) {
 	const op errors.Op = "observ.RegisterStatsExporter"
+	var stop = func() {}
+	var err error
 	switch statsExporter {
 	case "prometheus":
 		if err := registerPrometheusExporter(r, service); err != nil {
 			return nil, errors.E(op, err)
 		}
 	case "stackdriver":
-		if err := registerStatsStackDriverExporter(r, service); err != nil {
+		if stop, err = registerStatsStackDriverExporter(r, service); err != nil {
 			return nil, errors.E(op, err)
 		}
 	case "datadog":
-		if err := registerStatsDataDogExporter(r, service); err != nil {
+		if stop, err = registerStatsDataDogExporter(r, service); err != nil {
 			return nil, errors.E(op, err)
 		}
 	case "":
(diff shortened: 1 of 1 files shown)


pkg/observ/stats.go
```diff
@@ -51,6 +62,33 @@ func registerPrometheusExporter(r *mux.Router, service string) error {
 	return nil
 }
 
+func registerStatsDataDogExporter(r *mux.Router, service string) (func(), error) {
```
t-tomalak: Do we need to pass `mux.Router` here? I think it was needed only for prometheus because it need to register Handler. In DataDog is not needed anymore because it not create any handler for that.

pkg/observ/stats.go
```diff
@@ -51,6 +62,33 @@ func registerPrometheusExporter(r *mux.Router, service string) error {
...
+
+	dd := datadog.NewExporter(datadog.Options{Service: service})
+	if dd == nil {
+		return nil, errors.E(op, "Failed to initialize data dog exporter")
+	}
+	view.RegisterExporter(dd)
+	return dd.Stop, nil
+}
+
+func registerStatsStackDriverExporter(r *mux.Router, projectID string) (func(), error) {
```
t-tomalak: Same as for DataDog. We probably don't need `mux.Router` here :slightly_smiling_face:

pkg/observ/stats.go
fedepaol: That's right. I blatantly copied the prometheus exporter. Will fix.

pkg/observ/stats.go
```diff
@@ -15,11 +18,21 @@ import (
 // Currently it supports: prometheus
 func RegisterStatsExporter(r *mux.Router, statsExporter, service string) (func(), error) {
 	const op errors.Op = "observ.RegisterStatsExporter"
+	var stop = func() {}
+	var err error
```
arschles: @fedepaol can you define the error inside of the `case` statements below? I think it would be a bit easier to follow that way

arschles:
@fedepaol I have only nits on this, and it looks like @t-tomalak does too. 

If you’d like, we can merge it and fix the rest in follow-ups. Let me know!

pkg/observ/stats.go
arschles: Nevermind, that ends up looking messier. Ignore me!

Issue #893, opened 2018-11-10
Proposal: cache the /list endpoint for a short period of time?


arschles:
**Is your feature request related to a problem? Please describe.**

The `/list` endpoint contacts the upstream VCS to get version numbers, so that corresponds to one `git` (or `hg` etc...) operation per dependency. For a public deployment of athens, I believe that's untenable.

**Describe the solution you'd like**

I think it'd be a good idea to add an option to the proxy to cache the `/list` endpoint. That would mean that it fetches the version list directly from storage until a TTL expires. At that point it would do the list request to the upstream once, refresh storage, and reset the TTL. If there are multiple proxy servers running in a cluster, they would use #760 to ensure only one refresh is done after the TTL.

If we did this, we would need to tell package authors that if they push a new commit or release a new tag, it won't show up in Athens until the TTL expires. We could let them trigger a refresh for their module with #104, but I'm not sure if that's a reasonable solution.

**Describe alternatives you've considered**

I haven't thought of anything else that would reduce upstream requests for each `/list` call.

**Additional context**

I'm researching and preparing for #772 and found that some way to tell whether a response was cached or not would be really helpful for us.

robjloranger:
I think some level of caching would be a big bonus here. Could we have manual invalidation on new version pushes?

arschles:
@robjloranger yea, I think we could definitely have that. Maybe some variation on the webhook that got proposed in #104 could do the invalidation - what do you think?

robjloranger:
Exactly 😁 some endpoint to say some module was "updated" and we can even check to see if it does have a new version, and if so update the list cache.

arschles:
Update: https://github.com/gomods/athens/issues/1068 is the endpoint that @robjloranger talks about in https://github.com/gomods/athens/issues/893#issuecomment-443395041

I am going to start prototyping this now

arschles:
We've discussed this a few times and decided this caching logic shouldn't live inside of Athens. We are discussing an "offline mode" in #1506, which will treat the list endpoint as immutable. If Athens gets an offline mode, then caching will become unnecessary if you run it in that configuration. If you were to still run it in online mode configuration and still want caching, the best thing to do would be to use a caching proxy. There are several widely used open source ones and several cloud based ones. I have also built a "toy" one [here](https://github.com/arschles/lathens) for reference.

I am going to close this.

Issue #1068, opened 2019-02-16
Webhooks for notifying Athens of a new module/version


arschles:
**Is your feature request related to a problem? Please describe.**

It would be nice for module authors to tell an Athens instance when a new version of their module is published, without relying on someone to do a `go get` for that new version.

**Describe the solution you'd like**

A simple webhook that can be called which:

- Notifies Athens of a new module@version
- Prompts Athens to go fetch that module@version and store it in its storage
    - Just as if someone ran a `go get` against Athens

**Describe alternatives you've considered**

None

**Additional context**

I'm reviving this idea from the following related issues:

- https://github.com/gomods/athens/issues/106
- https://github.com/gomods/athens/issues/10
- https://github.com/gomods/athens/issues/104

cc/ @marpio

phaizel:
I am keen to work on this.

arschles:
@phaizel great! DM me if you like and I'll help get you started

arschles:
@phaizel are you still interested in doing this one?

arschles:
Please seee #1341 for a more detailed idea of how this could be implemented

PR #1175, opened 2019-04-03
[WIP] Implementing a cache for the /list endpoint

arschles:

The `/list` and `/latest` endpoints (I'm going to call these the "mutable endpoints") by default make requests to the upstream VCS that corresponds 
to the given module. That means that every `go get`corresponds to at least one VCS operation.

For many deployments of Athens, that can be annoying because the server will try to contact the given module's backing VCS at least once for every `go get` operation. It also tries to gracefully recover if the VCS is unavailable at the time. In either case, `go get` operations might be slower than they need to be.

Many teams don't need version lists to be absolutely up to date to the minute, because most build from an existing `go.sum` much more than they add new dependencies to their codebase. We should allow folks to optimize Athens for the former if they want.

**How is the fix applied?**

I've added a cache in front of the to the `List` call in the protocol implementation, so that a deployment of Athens can choose to only fetch the version lists from the upstream VCS after a specified TTL.

This PR would complement https://github.com/gomods/athens/issues/1068 nicely, because an Athens  admin could set a very long TTL, and notify Athens of new versions via the webhook mechanism in #1068.

Fixes https://github.com/gomods/athens/issues/893

This doesn't add caching for the `/latest` endpoint. I'll add a new issue for that after this is merged.


arschles:
@marwan-at-work let's talk in here so folks don't have to scroll up in slack.

Totally agree RE the in-memory cacher. I built is as a proof of concept in here, similar to how we started with the in-memory single flight implementation, then built implementations for production & horizontally scalable deployments in follow-ups.

When someone does a `go build` and there's a `go.sum` present, there will be no traffic going to the `/latest` endpoint, but `/list` will still be hit, is that correct? It's been months since I looked at the client code, so apologies if it's not.

In either case, I think (b) could be a nice solution instead of what I did here if we also implemented https://github.com/gomods/athens/issues/1068.

Those two things combined, plus some config changes in the mod/zip/info handlers to turn off the stasher, could be used to turn Athens into a "dumb" proxy that never looks upstream, and forces a user to hit a webhook if they want to add a module to the proxy. The result would be a proxy that's 100% isolated from the VCS in the critical path, which I think could be pretty handy.

What do you think?

marwan-at-work:
@arschles 

> When someone does a go build and there's a go.sum present, there will be no traffic going to the /latest endpoint

`go.sum` is definitely irrelevant so I still believe you're talking about `go.mod`

> but /list will still be hit, is that correct?
No, I don't believe so. There's no need to hit /list when go.mod already says exactly what version is needed so it will immediately proceed to stating that version with the `.info ` endpoint

>The result would be a proxy that's 100% isolated from the VCS in the critical path, which I think could be pretty handy.

That is definitely very interesting and I assume important in some scenarios. But maybe it's worth brainstorming how exactly is it helpful to isolate an athens from vcs? 

Furthermore, we should definitely brainstorm some designs and architectures on how to approach the solution, outside of just "caching list/latest" and we might come up with more elegant solutions 

Thanks for opening this discussion up :)

arschles:
Ok, maybe it's been too long and I just forgot how `go get` and `go build` work 😆. I'll go educate myself before I keep talking about these behaviors!

We can also talk about this in a developer meeting, but I'll write some stuff in here about pros/cons of an isolated Athens and how it might work.

In terms of immutability and isolation here's how Athens stacks up, if I'm not mistaken (please correct me if I'm wrong somewhere here):

1. The mod, info and zip endpoints are completely isolated and immutable after a given version for a given module path is written to storage
1. The list endpoint is immutable but not isolated. The lack of isolation can be a problem in the below scenario. This probably won't happen but is possible:
    - `/my/module/@v/list` returns a v2.1.2 from the VCS
    - The VCS has downtime
    - `/my/module/@v/v2.1.2.mod` does `go mod download my/module@v2.1.2` and fails
1. The latest endpoint is not immutable or isolated

Eventually I'd like to fix the problems in both 2 and 3. I also think another value of an isolated Athens is if you want to pre-store modules and simply use Athens to serve modules directly from storage. Specifically, here are three use cases for this approach that I've been thinking of

- Use the webhook I talked about to tell Athens which module paths & versions to store, and treat all others are excluded
    - Gocenter works similarly to this, except a web form submission stands in for the webhook
- Load modules from another source and treat all others as excluded (you could also do this one with a CDN fwiw)
- Load modules from a notary log, according to a filter

Finally, Athens definitely works great for lots of use cases, so I'm not suggesting that we change anything fundamental here. I think anything we do should be opt-in via configuration.

arschles:
I'm going to close this in favor of moving this caching functionality into a separate process. I prototyped it in github.com/arschles/lathens, and we can bring it in here (as another `package main` under `cmd/`) if there's any interest in it.

Issue #1165, opened 2019-03-28
Github Bot To Run Sensitive Tests


marwan-at-work:
Right now if a malicious user can open a PR and print our gcp credentials for CI/CD tests. We'd 100% be able to see it and invalidate those credentials but we should have a better mechanism. I suggest we create a github bot that we can ask it to run those sensitive tests after we see that the PR has no malicious code.

arschles:
@marwan-at-work can you go into some more detail on the bot? How would it "hide" the credentials for the user?

marwan-at-work:
The bot wouldn't hide the credentials. The idea is that you have a step in the build that will run the GCP tests, but a maintainer has to look at that PR before you run those tests to ensure that that PR is not malicious. For exaple, if the PR has a "fmt.Println(credentials)" then we wouldn't tell the bot to run the tests. Only a maintainer can run the bot obviously.

arschles:
@marwan-at-work I totally dropped this one, but I'm running through all the issues now and saw this. I think that generally, a bot would be a really good idea.

CircleCI has a feature like this that requires maintainers to manually sign off before PR tests run (if you turn it on). It comes with a problem, though, that it requires a maintainer to approve every single PR's tests. So, for example, if someone does a docs change, we would need to manually approve it if we were using circle.

Maybe we could start with a bot that replicates the circleCI functionality and in a follow-up, make it smarter about what it blocks (i.e. only block on changes that include `*_test.go` files)

marwan-at-work:
@arschles that sounds good. We can start as a super simple bot to just trigger a build as long as it's run by one of the maintainers. As long as DroneCI has an API to say "trigger this build"

arschles:
@marwan-at-work what if someone submits a PR that turns off the bot and also prints out a secret in part of the standard unit test code? Are you thinking that the bot runs somewhere else outside of drone and injects secrets?

PR #1176, opened 2019-04-04
storage/gcp: create a bucket on every test run

### LINKED ISSUE #1151: Flake in GCP tests

marwan-at-work:
Creates GCS buckets on the fly, deletes the bucket after tests finish 

Fixes #1151


pkg/storage/gcp/gcp_test.go
```diff
@@ -2,14 +2,27 @@ package gcp
...
 )
 
+var bucketName = randomBucketName(os.Getenv("DRONE_PULL_REQUEST"))
+var strg *Storage
+
+func TestMain(m *testing.M) {
+	exitCode := m.Run()
+	if strg != nil {
+		strg.bucket.Delete(context.Background())
+	}
```
arschles: @marwan-at-work would you be able to create and delete a bucket inside of each the individual test and benchmark? I think that would be a little easier to follow than the globals.

marwan-at-work: @arschles that would require a refactoring of all the tests, we can do that as a follow up in my opinion if we deem it necessary. It would also be more costly because you'd be creating/destroying buckets 6x more per PR ;)

arschles: Oh, I didn't see all 6 places. Where are the others beside `TestBackend` and `BenchmarkBackend`?

marwan-at-work: @arschles to create/remove bucket for every single test suite, you'd have to have a new "setup" function before every test suite in this file: https://github.com/gomods/athens/blob/master/pkg/storage/compliance/tests.go

Which means every single storage we have tests for should update its signatures. 

It's a totally good solution and would allow our tests to be more flexible but it's a big lift for just fixing the ci/cd part :)

arschles: @marwan-at-work I think I'm missing something because you already call `getStorage` at the beginning of `TestBackend` and `BenchmarkBackend` in the GCP tests. It seems like the only difference is that you need the global `strg` variable so that `TestMain` works properly, but `TestMain` doesn't need to exist in these GCP tests, does it?

marwan-at-work: @arschles you're so right! I forgot that we actually call `compliance.RunTests` from with the test function itself, so we can actually defer cleaning up. 

Thanks for the call out :) I updated the code

marwan-at-work: feel free to comment more, cuz i mighta missed a couple things here and there

arschles:
@marwan-at-work this looks great! :shipit: 💥 🚀 

We have https://github.com/gomods/athens/issues/1165 to follow-up on the leaked google credentials thing. I'm gonna merge this now

PR #1178, opened 2019-04-05
Refactoring the developer meetings page

arschles:

The developer meetings page was a little bit unclear, and it didn't show the rotating times for the meetings properly. Folks also had trouble converting UTC to their own time zone.

**How is the fix applied?**

I tried to make the developer meetings page clearer and put in some JS code to guess the time zone. I have no idea what I'm doing though...

N/A


docs/static/js/dev-meetings.js
```diff
@@ -0,0 +1,55 @@
...
+    nextThurs.year(),
+    nextThurs.month(),
+    nextThurs.date(),
+    20
+));
+
+
+// get next thursday in moment format for each time
+// (so that we can more easily print it out)
+const time20 = moment().
```
manugupt1: I can't see this being used anywhere.

docs/static/js/dev-meetings.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+const time22ID = document.getElementById("time22");
+const time22Str = nextThurs22.getHours() + ":00";
+if (time22.day() == 4) {
+    time22ID.innerHTML = time22Str;
+} else {
+    time22ID.innerHTML = time22Str + " (Friday)";
+}
+// const tz = jstz.guess();
```
manugupt1: Can you clean up comments as it is confusing?

arschles:
Thanks for the reviews @chriscoffee and @manugupt1 !

@manugupt1 I addressed your feedback. Can you take another look when you have time?

PR #1184, opened 2019-04-08
pkg/module: include SSH_AUTH_SOCK in go get env

dfinkel:
When running locally, it's convenient to use a user's ssh-agent. For
that to work, we need to propagate SSH_AUTH_SOCK.

Not all SSH keys are stored as files without passwords in `~/.ssh`. It is quite convenient to use `ssh-agent` (or `gpg-agent` in `ssh-agent` mode) to provide limited access to keys without having to enter the SSH key every time. (most notably when running on a workstation/laptop)

This is most useful when running `athens` as a docker container in the background.

**How is the fix applied?**

`openssh` generally handles the `SSH_AUTH_SOCK` environment variable automatically, so if it is present in the environment of `go mod download`, ssh-agent-based public key auth should just work.

I verified this locally on my machine with my company's github-enterprise setup.

### DIFF
--- a/pkg/module/go_get_fetcher.go
+++ b/pkg/module/go_get_fetcher.go
@@ -185,9 +185,13 @@ func PrepareEnv(gopath string) []string {
 		gitSSHCmd,
 	}
 
-	if sshAuthSockVal, hasSSHAuthSock := os.Getenv("SSH_AUTH_SOCK"); hasSSHAuthSock {
-		sshAuthSock := fmt.Sprintf("SSH_AUTH_SOCK=%s", sshAuthSockVal)
-		cmdEnv = append(cmdEnv, sshAuthSock)
+	if sshAuthSockVal, hasSSHAuthSock := os.LookupEnv("SSH_AUTH_SOCK"); hasSSHAuthSock {
+		// Verify that the ssh agent unix socket exists and is a unix socket.
+		st, err := os.Stat(sshAuthSockVal)
+		if err == nil && st.Mode()&os.ModeSocket != 0 {
+			sshAuthSock := fmt.Sprintf("SSH_AUTH_SOCK=%s", sshAuthSockVal)
+			cmdEnv = append(cmdEnv, sshAuthSock)
+		}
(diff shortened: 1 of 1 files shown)


pkg/module/go_get_fetcher.go
```diff
@@ -169,6 +169,7 @@ func PrepareEnv(gopath string) []string {
 	cacheEnv := fmt.Sprintf("GOCACHE=%s", filepath.Join(gopath, "cache"))
 	gitSSH := fmt.Sprintf("GIT_SSH=%s", os.Getenv("GIT_SSH"))
 	gitSSHCmd := fmt.Sprintf("GIT_SSH_COMMAND=%s", os.Getenv("GIT_SSH_COMMAND"))
+	sshAuthSock := fmt.Sprintf("SSH_AUTH_SOCK=%s", os.Getenv("SSH_AUTH_SOCK"))
```
arschles: @dfinkel can you check for the existence of `SSH_AUTH_SOCK` before adding this to the `cmdEnv` slice?

Also, are there any situations you can think of where someone would explicitly not want to use `SSH_AUTH_SOCK`?

dfinkel: I've switched it to verify existence of the environment variable (and also added a check to verify that the file it points at both exists and is a unix socket (taking into account symlinks))

I don't know of a case where someone would have a valid `SSH_AUTH_SOCK` and wouldn't want to use it. It strictly provides better control, and openssh is generally smart enough to fall back to public key files and password auth if that doesn't work.

One can always unset the variable from the environment if they don't want it to propagate.

arschles: @dfinkel ok, that sounds good. I am going to merge this, but will you be able to add a small section to the [configuring authentication documentation](https://docs.gomods.io/configuration/authentication/) about this env var?

dfinkel: Sure, I'll send a PR in the next couple days to extend the documentation.

PR #1188, opened 2019-04-15
Adding parameterized ATHENS_GO_GET_WORKERS

arschles:

It's impossible to set the value of the `ATHENS_GO_GET_WORKERS` environment variable from the helm chart

**How is the fix applied?**

I added it as a parameter 😄 


manugupt1:
@arschles I am trusting you on Helm charts. LGTM!

arschles:
@manugupt1 can you merge if you're all good with this?

PR #1194, opened 2019-04-22
Add gitconfig support for helm chart

hypnoglow:
This commit adds support for gitconfig in the athens helm chart.

Reflects https://docs.gomods.io/configuration/authentication/#altassian-bitbucket-and-ssh-secured-git-vcs-s

We deploy athens to Kubernetes using helm chart, and we want it to be able to access our private repositories with modules.

**How is the fix applied?**

This patch allows injecting any custom git config to the athens container. This is kind of similar to netrc support, but in our case gitconfig support would provide more flexible configuration options.

### DIFF
--- a/docs/content/install/install-on-kubernetes.md
+++ b/docs/content/install/install-on-kubernetes.md
@@ -257,3 +257,34 @@ In order to instruct athens to fetch and use the secret, `netrc.enabled` flag mu
 ```console
 helm install gomods/athens-proxy -n athens --namespace athens --set netrc.enabled=true
 ```
+
+### gitconfig support
+
+A [gitconfig](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration) file can be shared as a secret to allow
+the access to modules in private git repositories. For example, you can configure access to private repositories via HTTPS
+using personal access tokens on GitHub, GitLab and other git services.
+
+First of all, prepare your gitconfig file:
+
+```console
+cat << EOF > /tmp/gitconfig
+[url "https://user:token@git.example.com/"]
+    insteadOf = ssh://git@git.example.com/
+    insteadOf = https://git.example.com/
+EOF
+```
+
+Next, create the secret using the file created above:
+
+```console
+kubectl create secret generic athens-proxy-gitconfig --from-file=gitconfig=/tmp/gitconfig
+```
+
+In order to instruct athens to use the secret, set appropriate flags (or parameters in `values.yaml`):
+
+```console
+helm install gomods/athens-proxy --name athens --namespace athens \
+    --set gitconfig.enabled=true \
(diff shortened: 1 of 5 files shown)


docs/content/install/install-on-kubernetes.md
```diff
@@ -223,3 +223,34 @@ In order to instruct athens to fetch and use the secret, `netrc.enabled` flag mu
 ```console
 helm install gomods/athens-proxy -n athens --namespace athens --set netrc.enabled=true
 ```
+
+### gitconfig support
+
+A [gitconfig](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration) file can be shared as a secret to allow
+the access to modules in private git repositories. For example, you can configure access to private repositories via HTTPS
+using personal access tokens on GitHub, GitLab and other git services.
```
arschles: @hypnoglow would you mind mentioning here that this approach uses basic authentication?

hypnoglow: Strictly speaking, if there would be a possibility to inject ssh keys into a container,  then this approach would be not only about basic auth - we could use git config to rewrite https:// urls to ssh:// urls to force cloning via SSH using keys.

But, as long as there is currently no option to pass SSH keys to a container yet, I agree that this is mostly about basic auth (or token auth on GitHub/GitLab, that is implemented on top of basic auth protocol).

hypnoglow: Added sentence: `Basically, this approach uses basic authentication.`

marpio:
@arschles Can you re-review?
@hypnoglow Can you resolve the conflicts?

hypnoglow:
Rebased. @arschles please take a look.

rimusz:
Please rebase from master, as the chart got many changes since you opened PR, thanks

hypnoglow:
Rebased a few more times, please take a look.

hypnoglow:
Rebased again.

arschles:
@hypnoglow continuing from slack - it looks like the conflict resolution will be fairly straightforward. The conflict is this: 

```yaml
<<<<<<< add-gitconfig-support
        {{- if .Values.gitconfig.enabled }}
        - name: gitconfig
          mountPath: "/etc/gitconfig"
          subPath: "gitconfig"
        {{- end }}
=======
      {{- with .Values.resources }}
        resources:
{{ toYaml . | indent 10 }}
      {{- end }}
>>>>>>> master
```

And it should be ok to take both changes:

```yaml
        {{- if .Values.gitconfig.enabled }}
        - name: gitconfig
          mountPath: "/etc/gitconfig"
          subPath: "gitconfig"
        {{- end }}
      {{- with .Values.resources }}
        resources:
{{ toYaml . | indent 10 }}
      {{- end }}
```

Let me know if you'd like me to resolve it. I can do it from the GitHub UI

PR #1195, opened 2019-04-22
Adding a skeleton proposal for webhooks

### LINKED ISSUE #1068: Webhooks for notifying Athens of a new module/version

arschles:

There's currently no way to do two things with Athens:

- Tell it to download and store module versions in the background
- Tell it that a vanity domain is backed by a VCS path (which solves part of https://github.com/golang/go/issues/31458)

**How is the fix applied?**

I've added configuration, docs, and empty HTTP handlers for these webhooks

Fixes #1068

cc/ @marwan-at-work


manugupt1:
@arschles two cents. I really like the idea of async and sync handlers. 

We should have think of naming the web-hooks. So, When the user names a web-hook we just call the name and write it to a list of config file, so that a user can enable / disable a web-hook without 
A) hopefully restarting
B) or just restarting the server

but definitely not re-compiling the code.

config.dev.toml
```diff
@@ -85,6 +85,17 @@ Port = ":3000"
...
+# Enables webhooks that allow you to two things outside of the standard 
+# module download protocol:
+#
+#   - Trigger Athens to download a module asynchronously, without downloading
+#       any module code or metadata yourself
+#   - Tell Athens that a vanity module path is backed by a given VCS path.
+#       When you do this, Athens will download at the version in the /@latest
+#       endpoint for the given VCS path, store it, and register the given VCS
+#       path with the vanity domain forever
+EnableWebhooks = false
```
marwan-at-work: This wouldn't work unless we're supporting GitHub webhooks (and not GitLab, BitBucket etc) 
So we might want to just make an enum where "empty" means off.

pkg/download/handler.go
```diff
@@ -51,3 +51,36 @@ func RegisterHandlers(r *mux.Router, opts *HandlerOpts) {
 	r.Handle(PathVersionModule, LogEntryHandler(ModuleHandler, opts)).Methods(http.MethodGet)
 	r.Handle(PathVersionZip, LogEntryHandler(ZipHandler, opts)).Methods(http.MethodGet)
 }
+
+func RegisterWebhookHandlers(r *mux.Router, opts *HandlerOpts) {
```
marwan-at-work: I think the `download` package should be strictly for the download protocol...so this function and its handlers can probably live in their own package called `webhooks` either in the `actions` directory or in the `pkg` directory. Or you might want to flatten it out and just put it in the `actions` dir directly.

config.dev.toml
arschles: @marwan-at-work I don't understand. If you turned on webhooks, couldn't you htt them with any tool? They're essentially `curl` commands

pkg/download/handler.go
arschles: Good point. I'll push a change that moves this to another package!

config.dev.toml
marwan-at-work: @arschles yes, but we can't control what the json shape of the body request will look like. We need to unmarshal the body to know what the module/version/event is...unless we make the url smart enough to figure all of that out which might be a weird devex

arschles: Oh, well we need to specify what the JSON needs to be. I don't think we should necessarily use the GH webhook format, but we could. I didn't specify it in here yet, so here's an idea:

"Async fetch" webhook:

```console
GET /download/async/github.com/my/module
```

"Synchronous alias" webhook:

```console
POST /alias/sync/alias.com

{"origin": "github.com/my/module"}
```

Thoughts?

marwan-at-work: If we don't use the GH webhook format, how will we know what version just got released from that webhook call?

Kunde21: We have the option to handle all vcs webhook formats (github/gitlab/bitbucket/etc) or we could start with a library and cmd for integration into other tooling like `goreleaser`.  I think the latter would be a better starting point, then we can consider a webhook-detection routine or separate endpoints for handling the webhooks from each vcs site.

arschles: @marwan-at-work @Kunde21 I think the version should be included in the async fetch webhook - probably as `.../@v/v1.0.1` at the end of the path above - but we can make it optional in the sync alias webhook.

Overall, I'm not looking for any support for any GH webhook formats in here (or other VCS's). IO want to define a simple format that someone can `curl` against to tell Athens either about the existence of a new module path & version, or about a new alias.

arschles:
@manugupt1 

>but definitely not re-compiling the code.

Absolutely agree. I'm defining whether the webhooks are on or off via config flags. If [this](https://github.com/gomods/athens/pull/1195/files#diff-422a9f62303968af1694a4097c7fb5b2R97) what you were thinking?

manugupt1:
@arschles Yes and no! I was thinking that we could allow users can write their own plugins via single interface. As long as they satisfy the interface, it is a web-hook that athens will support.  This can be compiled in let's say `webhooks`

Then, athens should be able to generate list of available webhooks at runtime, all turned off by default pointing to a file which is in config. So, for example
WEBHOOKS_CONFIG=~/.athens/webhooks

Now users can turn on / off webhooks by default.

arschles:
@manugupt1 that sounds interesting, but I feel like it's overkill for the problem here because at the core, we're trying to allow folks to register aliases and trigger module downloads in the background (aside: I expect that the most common use case for the latter is to heat up the storage). What do you think about sticking to those two features and then looking into custom webhooks if we need to support more use-cases

Also, @marwan-at-work I think that the async-fetch webhook (i.e. `GET /download/async/github.com/my/module`) could be solved in option (2) of your proposal at #1211. What do you think?

manugupt1:
Makes sense! I was thinking more in terms of a general web hook.


Cheers,

Manu Gupta

marwan-at-work:
@arschles how will Athens know which version to grab?

config.dev.toml
marwan-at-work: @arschles that sitll doesn't work as a webhook, because the version is a moving target. We can't just have one version to be overridden

arschles:
@marwan-at-work good question. I'm assuming you mean for the alias webhook? 

In that case, I'm not sure if that endpoint makes sense anymore, because `go mod download` may not work for `go.mod` files that have the vanity domain in their `module` section.

that aside, if we decide to sidestep that problem in some way, I think requiring the user to put the version in the post body like the below is a reasonable solution:

```
POST /alias/sync/alias.com

{"origin": "github.com/my/module, "version": "v1.x.x"}
```

After this request returns, the user can count on an `alias.com` module path being on the server with that specific version. The user can also do another `POST` with another `v0.x.x` or `v1.x.x` and Athens would download that version as well. A POST with a version that was already downloaded could return a `409 Conflict`.

`"version": "v2.x.x"` would correspond to an `alias.com/v2` module path, at that specific version, and so on.

I also think we should disallow pseudoversions for now.

arschles:
Closing this until we can have more detailed discussion in #1068

PR #1197, opened 2019-04-26
Clusterrole

moeryomenko:

Add optional ServiceAccount and ClusterRoleBinding templates


rimusz:
Athens do not access Kubernetes API, no needs to have RBAC rules set there.

PR #1198, opened 2019-04-26
Trying out appveyor

marpio:

Azure pipelines VMs does not have go 1.12 and using docker is very slow (12 min).
Appveyor VM images contain go 1.12 and the build takes 1:30min.

**How is the fix applied?**
We can try out appveyor and see if it works for us.


arschles:
Hm, weird that this build took only 4 min for Azure Pipelines. Looks like the setup step was significantly faster than usual. Anyway, merging for now!

PR #1199, opened 2019-04-29
Fix wg.Done() usage

moeryomenko:

`defer` should be used when multiple return points.

### DIFF
--- a/pkg/download/protocol.go
+++ b/pkg/download/protocol.go
@@ -73,13 +73,13 @@ func (p *protocol) List(ctx context.Context, mod string) ([]string, error) {
 	wg.Add(2)
 
 	go func() {
-		defer wg.Done()
 		strList, sErr = p.storage.List(ctx, mod)
+		wg.Done()
 	}()
 
 	go func() {
-		defer wg.Done()
 		_, goList, goErr = p.lister.List(ctx, mod)
+		wg.Done()
 	}()
 
 	wg.Wait()


moeryomenko:
> Couple of points:
> 
>     1. In protocol.go, if `p.storage.List` panics, then Defer never gets called.

No one of the Lister implementations cause panic 

>    2. Why does this CL also change our vendor directory? It definitely shouldn't do that because we don't want different behavior between running go build -mod=vendor and running go build

ok, agree it's not correct.

marwan-at-work:
@moeryomenko 

> No one of the Lister implementations cause panic
We can't guarantee that for the future. Whoever changes that behavior in the future must also know to make sure to defer here. 

Another question, are these `defers` causing any performance problems? Why do they need to be removed? 

Thanks for the PR in any case :)

moeryomenko:
> @moeryomenko
> 
> > No one of the Lister implementations cause panic
> > We can't guarantee that for the future. Whoever changes that behavior in the future must also know to make sure to defer here.
> 
> Another question, are these `defers` causing any performance problems? Why do they need to be removed?
> 
> Thanks for the PR in any case :)

It seems abusive to use deferred work at any time to work with WaitGroup.Done ()

marwan-at-work:
@moeryomenko 

> It seems abusive to use deferred work at any time to work with WaitGroup.Done()

I'm aware that `defer` incurs a performance cost, but we shouldn't introduce potential code instability for an unmeasured performance gain. If we want to enhance performance, it'd be wise to first measure where our bottlenecks are first and fix those first. 

I'll leave this PR open for a little longer to see if anyone else has potential comments 

Thanks

marpio:
I'm with @marwan-at-work on that we shouldn't trade  unmeasured performance gains for code stability.
But out of curiosity - wouldn't a panic in the list func crash the whole process anyway?

moeryomenko:
> But out of curiosity - wouldn't a panic in the list func crash the whole process anyway?

No, read about the recovery mechanism.

And little simple example https://goplay.space/#IOnbHpgCwc0

manugupt1:
I am with @marwan-at-work as well.
Another reason to have defer is that. If you have a panic, and if there is a panic recoverer as a middleware, then you would never the done() part of the function.

marwan-at-work:
@marpio yes the program might crash, but the `defer` still gets run before the program crashes, and in this case it will clean up the `disk` and remove the temporary GOPATH where things got downloaded. 

I'm closing this PR but feel free to continue the discussion here or in an issue. 

Thanks!

PR #1200, opened 2019-04-29
cgo enabled

moeryomenko:

cgo enabled, because final stage contain c standard library. 
Also changed source path on stage builder

### DIFF
--- a/cmd/proxy/Dockerfile
+++ b/cmd/proxy/Dockerfile
@@ -4,13 +4,14 @@
 
 FROM golang:1.12-alpine AS builder
 
-WORKDIR $GOPATH/src/github.com/gomods/athens
+WORKDIR /src
 
 COPY . .
 
 ARG VERSION="unset"
 
-RUN DATE="$(date -u +%Y-%m-%d-%H:%M:%S-%Z)" && GO111MODULE=on CGO_ENABLED=0 go build -mod=vendor -ldflags "-X github.com/gomods/athens/pkg/build.version=$VERSION -X github.com/gomods/athens/pkg/build.buildDate=$DATE" -o /bin/athens-proxy ./cmd/proxy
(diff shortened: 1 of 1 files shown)


moeryomenko:
> More importantly, can we guarantee to have _all_ C libraries in alpine that Go needs?

```sh
$ ldd /bin/athens-proxy
      /lib/ld-musl-x86_64.so.1 (0x7f76e1123000)
      libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7f76e1123000)
```

alpine always contain libc(musl), go uses libc only

> Out of curiosity: what's the problem with building a static binary, _even_ if `alpine` has C libraries?

Don't have any problems, but don't have sense. This is reasonable when scratch or busybox is used as the final stage

marwan-at-work:
Note that it's not just Go that uses C libraries. Go calls other binaries such as `git`, `bzr` and potentially others. I'm not sure if that affects anything, but doing a static build usually guarantees no need for external C dependencies. 

That said, if other maintainers are comfortable that this change won't have any regressions then sounds good. 

cc: @marpio @arschles

manugupt1:
two cents: Athens does not require any CGO dependency, so we should be fine without CGO. Along with that athens spins up vgo processes and they do not require CGO as well, so we are okay with not having it.  We should only enable cgo when it is actually required by any of the code that we are using.

moeryomenko:
> two cents: Athens does not require any CGO dependency, so we should be fine without CGO. Along with that athens spins up vgo processes and they do not require CGO as well, so we are okay with not having it. We should only enable cgo when it is actually required by any of the code that we are using.

meant that there is no need to build a completely static binary.

manugupt1:
@moeryomenko yep! We should not build everything unless required. If we need it, we can always add it later.

PR #1201, opened 2019-04-30
Refer to a consistent minimum Go version

scottilee:

> The docs refer to two different versions of Go and there should only be reference to one for consistency.

**How is the fix applied?**

> I searched the Athens repo for all mentions of Go 1.11 and Go 1.12 and changed them to refer to Go 1.11+. This was discussed with Aaron Schlesinger on Slack as Go 1.11 is the minimum version supported.

Fixes #

> There is no associated issue.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -6,7 +6,16 @@ Athens uses [Go Modules](https://golang.org/cmd/go/#hdr-Modules__module_versions
 
 See our [Contributing Guide](CONTRIBUTING.md) for tips on how to submit a pull request when you are ready.
 
+**All the instructions in this document assume that you have checked out the code to your local machine.**
+
+If you haven't done that, please do with the below command before you proceed:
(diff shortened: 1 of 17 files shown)


scottilee:
Let me know if I should just change the "1.11"s to "1.12" now and make a note somewhere about most things working with 1.11.

arschles:
nah, I think it's ok for now

arschles:
@scottilee I didn't put it together that https://github.com/gomods/athens/pull/1121 will mean we require Go 1.12. Can you update these documentation to `1.12+` after all? Sorry for confusing everything here...

scottilee:
No problem, I just pushed the change. I left the "+" so that it's easier for future releases but if that's not the case or it looks weird since 1.12 is the latest release let me know and I can remove the "+" so that it's just "v1.12".

I don't know why those chart YAML files got pulled in...

PR #1204, opened 2019-05-02
[chart] add support to configure ssh-secured git servers

rimusz:


charts/athens-proxy/values.yaml
```diff
@@ -54,8 +54,8 @@ upstreamProxy:
   # This is where you can set the URL for the upstream module repository.
   # If 'enabled' is set to true, Athens will try to download modules from the upstream when it doesn't find them in its own storage.
   # You can use 'https://gocenter.io' to use JFrog's GoCenter as an upstream here, or you can also use another Athens server as well.
-  enabled: false
-  url: "https://upstreamproxy_url"
+  enabled: true
```
arschles: @rimusz can we default upstreams to off (can keep the URL to gocenter) - the majority of people use the VCS as their upstream, and this would break them

rimusz: sure

PR #1209, opened 2019-05-07
chart, fix workers

rimusz:


charts/athens-proxy/templates/NOTES.txt
```diff
@@ -1,6 +1,6 @@
 {{- if .Values.ingress.enabled }}
 {{- range .Values.ingress.hosts }}
-  http://{{ . }}
+  https://{{ . }}
```
arschles: @rimusz can this be configurable and defaulted to `https`?

arschles: @rimusz and I discussed offline, we'll make this in a follow up, since it's only in `NOTES.txt`

arschles:
For context, this PR is related to the fact that Athens is now up on the [helm hub](https://hub.helm.sh/charts/gomods/athens-proxy) 🎉

PR #1215, opened 2019-05-09
Pointing to meetup.com for meeting times

arschles:

The meeting times section of the developer meetings doc page didn't show the right times always (for example, after a month has an odd number of weeks in it).

**How is the fix applied?**

I removed that section and just pointed to the [meetup page](https://www.meetup.com/The-Athens-Developer-Meeting)


arschles:
This PR is a docs change that I'd like to be in before this week's developer meeting (last week the site conflicted with the meetup page), so I'm going to self-merge this

PR #1225, opened 2019-05-14
use circleci for chart testing

rimusz:

PR #1226, opened 2019-05-14
Adding basic circleci config

arschles:
I think this will help @rimusz in https://github.com/gomods/athens/pull/1225/files


arschles:
I'm merging my own PR so that we can get this CircleCI config into master and start testing https://github.com/gomods/athens/pull/1225/files

PR #1231, opened 2019-05-18
build: remove vendor folder

marwan-at-work:
This PR removes the vendor and builds Athens in two ways: 

1. On DroneCI: it builds Athens with the latest master version of Athens
2. On AppVeyor: it builds Athens with proxy.golang.org :)


arschles:
This was huuuuuuge and I'm so happy we made it to this point :tada: 💚  :shipit: 🚀

PR #1240, opened 2019-05-23
auth404

storyicon:
Here is the reason: https://github.com/storyicon/gos/blob/master/docs/on-goproxy.md

### DIFF
--- a/pkg/errors/gocmd.go
+++ b/pkg/errors/gocmd.go
@@ -9,3 +9,9 @@ import (
 func IsRepoNotFoundErr(err error) bool {
 	return strings.Contains(err.Error(), "remote: Repository not found")
 }
+
+// IsRepoNoAuthErr returns true if the Go command line
+// hints at a repository no auth.
+func IsRepoNoAuthErr(err error) bool {
+	return strings.Contains(err.Error(), "fatal: could not read Username")
+}
--- a/pkg/download/protocol.go
+++ b/pkg/download/protocol.go
@@ -92,12 +92,13 @@ func (p *protocol) List(ctx context.Context, mod string) ([]string, error) {
 
 	// if i.e. github is unavailable we should fail as well so that the behavior of the proxy is stable.
 	// otherwise we will get different results the next time because i.e. GH is up again
-	isUnexpGoErr := goErr != nil && !errors.IsRepoNotFoundErr(goErr)
+	isUnexpGoErr := goErr != nil && !errors.IsRepoNotFoundErr(goErr) && !errors.IsRepoNoAuthErr(goErr)
 	if isUnexpGoErr {
 		return nil, errors.E(op, goErr)
 	}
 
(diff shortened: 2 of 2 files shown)


pkg/errors/gocmd.go
```diff
@@ -9,3 +9,9 @@ import (
 func IsRepoNotFoundErr(err error) bool {
 	return strings.Contains(err.Error(), "remote: Repository not found")
 }
+
+// IsRepoNoAuthErr returns true if the Go command line
+// hints at a repository no auth.
+func IsRepoNoAuthErr(err error) bool {
+	return strings.Contains(err.Error(), "fatal: could not read Username")
```
marwan-at-work: is this string *always* the case for unauthorized requests? Or is it just when the repo is `git`?

storyicon: When pull a private repo, print goErr [here](https://github.com/gomods/athens/blob/master/pkg/download/protocol.go#L92), you will get the following error message.

```
exit status 1: go list -m  xxxx.com/private_repo: git ls-remote -q origin in /tmp/athens815520410/pkg/mod/cache/vcs/8848b6dab38d29dde08703be3e1a0218487530e7342a4e7bddb5731f05a98873: exit status 128:
	fatal: could not read Username for 'https://xxxx.com': terminal prompts disabled
If this is a private repository, see https://golang.org/doc/faq#git_https for additional information.
```

In fact, most developers use `athens`, they do not want to build an `athens` in their own local, but directly set the `GOPROXY` environment variable to `https://athens.azurefd.net/`
So setting `Github Token` or a `.netrc` file doesn't work for them.

Issue is open: https://github.com/gomods/athens/issues/1241

marwan-at-work: @storyicon the developers using `GOPROXY` will not set a `Github Token`, the Athens maintainers who deployed athens behind i.e. `athens.azurefd.net` are the ones that will set up a `Github Token` for them.

storyicon: @marwan-at-work  Sure, I know, but will you give your GitHub/Gitlab token to `athens`?
I now think that for a `public proxy provider`, such as `athens.azurefd.net`, the normal working mode should be: 
1. Use `athens` when pulling public repositories.
2. Returns a specific status code when pulling a `private repository` to allow developers to use local `git`.

Developers don't need to give any `token` to any `proxy provider`.
Maybe there are some mistakes in my understanding. If so, I apologize.

arschles:
@storyicon what do you think of https://github.com/gomods/athens/pull/1240#pullrequestreview-241157285?

matt0x6F:
Closing this request for now, feel free to resubmit. Athens responds with a 500 because if a user is using the proxy and intends to access a private repository without additional configuration on the client (eg: `direct`) then Athens should not respond the same as if it could not _find_ the repository.

PR #1244, opened 2019-05-27
clarifying the readme

arschles:
we don't have a public stable proxy and don't have concrete plans to release one, so let's not say we do for now.


scottilee:
It would be nice if you could link to the experimental public proxy.

As a side note, have you considered making a list of companies using Athens in production to display somewhere?

arschles:
@marpio I made the changes you suggested - thanks for that. 

@scottilee - I also added a link to the experimental proxy. If I get some time I'll put a domain in front of it. I have it on my list to ask folks to let us know if they're using Athens at their company, but haven't gotten to it yet. Would you be interested in helping?

scottilee:
@arschles Yeah, would be happy to. I've been busy but starting next week I should have more time. It looks like @marwan-at-work already started something at https://github.com/gomods/athens/issues/1323. Let me know what else I need to do.

arschles:
@scottilee all good! I tweeted about #1323 from the Athens account. A retweet and any other way you can spread the word would be much appreciated! I really appreciate your help 😄

PR #1247, opened 2019-05-27
fix:  '/catalog' endpoint, which didn't allow fetch URL query parameters

### LINKED ISSUE #1236: token is not passed in /catalog from URL
**Describe the bug**
token/pagesize is not passed to /catalog from URL.

**Error Message**
token == "" instead of passed in value

image

image

**To Reproduce**
just pass any variable

**Expected behavior**
variables extracted from mux.Vars(r) is what's passed in URL

 - OS: [e.g. Linux 64bit]
 - Go version :1.12.4
 - Proxy version :master
 - Storage (fs/mongodb/s3 etc.) :

**Additional context**
https://stackoverflow.com/questions/31371111/mux-vars-not-working
Looks like shouldn't use `mux.Vars`

5onic:
Fixed `/catalog` endpoint which didn't retrieve correctly `token `and `pagesize` parameters.

**How is the fix applied?**

In order to correctly fetch `pagesize` and `token` parameters, I had replaced to `r.Query()` values instead of `mux.Vars()`.

Fixes #1236

### DIFF
--- a/cmd/proxy/actions/catalog.go
+++ b/cmd/proxy/actions/catalog.go
@@ -9,7 +9,6 @@ import (
 	"github.com/gomods/athens/pkg/log"
 	"github.com/gomods/athens/pkg/paths"
 	"github.com/gomods/athens/pkg/storage"
-	"github.com/gorilla/mux"
 )
 
 const defaultPageSize = 1000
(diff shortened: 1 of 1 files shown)


cmd/proxy/actions/catalog.go
```diff
@@ -28,10 +27,12 @@ func catalogHandler(s storage.Backend) http.HandlerFunc {
 			w.WriteHeader(errors.KindNotImplemented)
 			return
 		}
+
 		lggr := log.EntryFromContext(r.Context())
-		vars := mux.Vars(r)
-		token := vars["token"]
-		pageSize, err := getLimitFromParam(vars["pagesize"])
+		urlQuery := r.URL.Query()
+		token := urlQuery.Get("token")
```
marwan-at-work: you can remove the line above and just use `r.FormValue("token")` directly.

cmd/proxy/actions/catalog.go
```diff
@@ -28,10 +27,12 @@ func catalogHandler(s storage.Backend) http.HandlerFunc {
...
+
+		pageSize, err := getLimitFromParam(urlQuery.Get("pagesize"))
```
marwan-at-work: same here: `r.FormValue("pagesize")`

5onic:
@marwan-at-work No problems, fixed.

Thanks 👍

Issue #1237, opened 2019-05-22
Set up a dev environment with sail?


arschles:
**Is your feature request related to a problem? Please describe.**

Not really, this would be a pretty cool addition to our dev experience

**Describe the solution you'd like**

We already have a pretty good documentation page for how to set up your local development environment. I think it would be nice & convenience to provide another option for anyone looking to develop on Athens using [sail](https://github.com/cdr/sail) to preconfigure their dev environment.

**Describe alternatives you've considered**

We could of course not do this

**Additional context**

https://sail.dev/

brainflake:
Has anybody taken a crack at this yet @arschles? I'd be happy to give it a shot

robjloranger:
@brainflake it looks up for grabs still. Please do. 🤓

brainflake:
Sounds good @robjloranger. Sail seems actually really easy to setup - is there anything in particular I should ensure is included in the dev environment (aside from enabling GO111MODULE)? The current go version is 1.12.4 in their base golang ubtuntu image, which I'm assuming works based on the Dockerfile currently in `cmd/proxy`.

Currently just seeing if i can get the unit tests to work from the console in the sail window...

robjloranger:
Awesome, sorry for the delay @brainflake. How is it coming? Did you get tests working?

I'm not yet familiar with sail myself, I am very interested in giving it a shot though. It should really help with on boarding new contributors who may be unfamiliar with Go development.

I think if you can build and run Athens then you've got a great start for the sail environment set up.

arschles:
+1 to what @robjloranger said. Any progress with Sail I think is a great win. @brainflake if you have something working for local dev and/or unit tests, we can PR and merge that and then improve from there in follow-ups!

brainflake:
Sounds good @robjloranger @arschles 

I'll cut a PR with what I have. It's just a Dockerfile w/ a few lines at the moment but seems to work for building + running.

arschles:
I think that #1248 fixed this, so I'm gonna close. @brainflake if I'm wrong, let me know please!

brainflake:
Sounds good 🚀

PR #1248, opened 2019-05-27
Add support for Sail development environments

brainflake:
This will make it easy for new contributors to build and run the proxy locally using [Sail](https://sail.dev). The open issue for this is #1237 

## Overview

Sail aims to provide a turn key, immutable development environment for projects using it. To use it you need to install the CLI available at [sail.dev](https://sail.dev) and run `sail run gomods/athens`.

## Configuration

Configuration is done through a Dockerfile at `.sail/Dockerfile`. The command will use this to spin up a container the is preconfigured for a development environment. The base image includes [code-server](https://github.com/cdr/code-server) to enable a 'remote' development environment for VS Code on the container, accessible in your browser.

## Current status

`make` targets can be run from the VS Code console in the browser. Currently most of the non-docker targets work: `make build`, `make run`, `make test-unit`, etc. The unit tests are currently failing for me (it seems related to minio?).


arschles:
@brainflake looks really cool. Do you know if there is a way to tell the sail CLI to use a different docker socket? I couldn't find one on the sail docs site.

arschles:
And by the way, thanks for doing this!

brainflake:
Hmm, I'm not sure to tell you the truth @arschles . Are you referring to the `EXPOSE` statement? I added that for the proxy, so when someone does a `make run` in the terminal inside sail they can access it from their localhost. Although now that I'm looking at it it doesn't seem to be working...

arschles:
@brainflake sorry for the delay. I was referring to having sail talk to a different docker host on my machine, because I was trying to test with Docker for Mac and iirc it doesn't use the domain socket that's standard on most linux systems. did the same thing not work for you too?

brainflake:
Ah, I see @arschles. So I was testing w/ Docker Desktop for Mac, which does use the standard domain socket at `/var/run/docker.sock` (although I think earlier versions for Mac use docker-machine, which uses a TCP socket).

As far as I can tell, sail will use the same configuration that the docker CLI uses, so it may be possible to point it to elsewhere using `DOCKER_HOST`, `DOCKER_MACHINE_NAME`, etc.

brainflake:
Yeah after poking around a bit, looks like this is the entrypoint https://github.com/cdr/sail/blob/master/hat_builder.go#L35, which eventually calls https://github.com/moby/moby/blob/master/client/options.go#L19

So I think the those standard env vars should be respected

arschles:
@brainflake There's a tiny conflict in `DEVELOPMENT.md`. Can you let me know if you'd like me to fix it (I can do it from the GitHub UI, but I don't want to commit directly to your `master` branch w/o your permission)?

brainflake:
@arschles I just went ahead and merged the latest master into my branch so I think it should be good, but definitely don't hesitate if need be

PR #1251, opened 2019-05-28
Azureblob tests + checker fix

### LINKED ISSUE #1243: Fix Azure storage checker
**Describe the bug**
We check modules in v0.4.0 based on prefixes not based on explicit paths, so when two Versions have the same prefix, we get into trouble.
More on that here: https://github.com/gomods/athens/issues/1219#issuecomment-494973122

There's already a fix for s3 (thanks @marwan-at-work) but we need to do the same for Azure storage.

**To Reproduce**
Described here https://github.com/gomods/athens/issues/1219#issuecomment-494900010

 - OS: [e.g. Linux 64bit]
 - Go version :
 - Proxy version : v0.4.0
 - Storage (fs/mongodb/s3 etc.) :

**Additional context**
https://github.com/gomods/athens/issues/1219

marpio:

Described in #1243 
We were also missing tests for azure blob storage.

Fixes #1243


marwan-at-work:
@arschles I'm not too savvy with Azure Blobs, but can you take a look and hit the merge button if all good? Thanks!

arschles:
@marwan-at-work @marpio I think we should definitely hook up the tests to Azure. I'll set this up now

arschles:
Ok, I've added `ATHENS_AZURE_ACCOUNT_KEY` and `ATHENS_AZURE_ACCOUNT_NAME` to the drone env (turned on for PRs 😱!). I believe future builds should be running the azure tests. Did I miss anything @marpio @marwan-at-work ?

marwan-at-work:
@arschles I'll take a look at the next PR build, and if the tests don't run it shouldn't be a biggie to fix as long as we have the credentials, thanks for adding them!!

PR #1254, opened 2019-05-29
FIXED type-o in docs

EwanValentine:
Type-o in docs.

**How is the fix applied?**
Fixed the type-o :D

N/a

### DIFF
--- a/docs/content/design/from-vcs2user.md
+++ b/docs/content/design/from-vcs2user.md
@@ -48,7 +48,7 @@ Then the `Fetcher`:
 - invokes `go mod download -json {module}`
 
 This command downloads the module into the storage. Once the download is completed, 
-- the `Fetch` function reads the module bits from stroage and returns them to the caller. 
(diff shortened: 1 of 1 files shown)


docs/content/design/from-vcs2user.md
```diff
@@ -130,4 +130,4 @@ In _Download protocol_ it is used to check whether or not the module is already
 
 Otherwise, _Download protocol_ uses `Stasher` to download module, store it into a `storage` and then it serves it back to the user.
 
-You can also see `addons.WithPool` in a code snippet above. This addon is something similar to `Stash wrapper - Pool`. It controls the number of concurrent requests proxy can handle.
\ No newline at end of file
+You can also see `addons.WithPool` in a code snippet above. This addon is something similar to `Stash wrapper - Pool`. It controls the number of concurrent requests proxy can handle.
```
EwanValentine: Not sure why this is being picked up as a change?

marwan-at-work: invisible characters 😄

Issue #1227, opened 2019-05-16
Publish Athens server docker image for armv6


atomi:
Since Athens is using Drone which also offers a plugin to publish manifest files can multi arches  be supported? 

http://plugins.drone.io/drone-plugins/drone-manifest/   

I was able to build 955929d just fine on a Raspberry Pi 3 with only a tiny change in the Dockerfile   

``` patch
diff --git a/cmd/proxy/Dockerfile b/cmd/proxy/Dockerfile
index fe2a691..88506ad 100644
--- a/cmd/proxy/Dockerfile
+++ b/cmd/proxy/Dockerfile
@@ -14,7 +14,7 @@ RUN DATE="$(date -u +%Y-%m-%d-%H:%M:%S-%Z)" && GO111MODULE=on CGO_ENABLED=0 go b
 
 # Add tini, see https://github.com/gomods/athens/issues/1155 for details.
 ENV TINI_VERSION v0.18.0
-ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-static /tini
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-static-armhf /tini
 RUN chmod +x /tini
 
 FROM alpine
```

Edit: Obviously this is not critical but it would be a nice to have for further updates. Thanks.

atomi:
So for what it's worth, I've been playing around with Athens on the Pi and it does indeed work but it is pretty slow.    

So I'm not sure there will be many takers. 🤔

Kunde21:
Is there a reason the Dockerfile doesn't use the alpine `tini` package directly?

atomi:
> Is there a reason the Dockerfile doesn't use the alpine tini package directly?

I wasn't aware tini was available in alpine but probably not. 🤷    
It looks like it's the same version available as well.

marpio:
I think it's interesting b/c it would potentially enable Athens to run on [AWS EC2 A1](https://aws.amazon.com/ec2/instance-types/a1/) or Scaleway ARM servers https://www.scaleway.com/en/arm-instances/

Kunde21:
I've built athens on `arm`, `arm64`, and `amd64` with a multiarch manifest here:  hub.docker.com/r/kunde21/athens-arm

I'll be testing on my local hardware (rpi3 and odroid XU4), but it would be good to get others to test, too.  Main focus for testing in performance on different hardware, so we can give some general recommendations when it's released on the official gomods dockerhub.

[...]

PR #1260, opened 2019-05-30
Use alpine tini

marpio:

Fallowing the conversation in https://github.com/gomods/athens/issues/1227
It seems that `tini` is available on alpine.
This way we wouldn't need to adjust the Dockerfile for different archs (like ARM in case of #1227).

cc: @Kunde21


cmd/proxy/Dockerfile
```diff
@@ -26,13 +21,14 @@ COPY --from=builder /go/src/github.com/gomods/athens/config.dev.toml /config/con
 COPY --from=builder /usr/local/go/bin/go /bin/go
 COPY --from=builder /tini /bin/tini
```
Kunde21: Also remove the copy, since that `tini` doesn't exist.

Issue #1253, opened 2019-05-29
No access log since 0.4.0


kumy:
**Describe the bug**
Since docker image v0.4.0 the access logs are not present on stdout.

**Error Message**
none

**To Reproduce**
* Launch Athens using docker image tag `v0.4.0`
* Download some module `go run .`

**Expected behavior**
I expect to see access logs like in version `0.3.1`

```
handler: GET /github.com/athens-artifacts/samplelib/@v/v1.0.0.info [200]
handler: GET /github.com/athens-artifacts/samplelib/@v/v1.0.0.mod [200]
handler: GET /github.com/athens-artifacts/samplelib/@v/v1.0.0.zip [200]
```

**Environment (please complete the following information):**
 - OS: Linux 64bit
 - Go version : go version go1.12.5 linux/amd64
 - Proxy version : 0.4.0
 - Storage (fs/mongodb/s3 etc.) : fs

**Additional context**
BTW, running `athens-proxy -version` returns `unset` for `v0.4.0` while it worked on `0.3.1`

```
/config # athens-proxy 
2019-05-29 16:02:51.584248 I | config file not provided - using default settings
INFO[0000] Exporter not specified. Traces won't be exported 
2019-05-29 16:02:51.614218 I | Starting application at port :3000
handler: GET /github.com/athens-artifacts/samplelib/@v/v1.0.0.info [200]
handler: GET /github.com/athens-artifacts/samplelib/@v/v1.0.0.mod [200]
handler: GET /github.com/athens-artifacts/samplelib/@v/v1.0.0.zip [200]
/config # athens-proxy -version
Build Details:
	Version:	v0.3.1
	Date:		2019-03-09-22:58:46-UTC

# […]

# Using v0.4.0
/ # athens-proxy -version
Build Details:
	Version:	unset
	Date:		2019-05-09-20:44:12-UTC
/ # athens-proxy -config_file /config/config.toml 
INFO[0000] Exporter not specified. Traces won't be exported 
2019-05-29 16:05:59.621565 I | Starting application at port :3000
^C/ # 

```

marwan-at-work:
@kumy thank you for reporting this, there are two separate issues here: 
1. logs in the container
2. build tag in the container 

I'll look into both today. However, the default behavior of the container is `ENV=production` which means only error-level logs would show up. Let me look into why v0.3.1 was showing those.

marwan-at-work:
[...]

PR #1261, opened 2019-05-30
Set build-args - VERSION

marpio:

#1253 mentions that the `athens-proxy -version` returns `unset`
This is bc we are not setting the VERSION build-arg in our CI

**How is the fix applied?**

http://plugins.drone.io/drone-plugins/drone-docker/ allows using build_arg


marpio:
@marwan-at-work 
good point!
I've added it to `docker-latest`

PR #1269, opened 2019-06-04
Fix #1268: Launch tini as subreaper

### LINKED ISSUE #1268: Tini not launched as subreaper
**Describe the bug**
As used, tini is not registered as subreaper. So theorically, it doesn't works.

**Error Message**
```console
 [WARN  tini (6)] Tini is not running as PID 1 and isn't registered as a child subreaper.
 Zombie processes will not be re-parented to Tini, so zombie reaping won't work.
 To fix the problem, use the -s option or set the environment variable TINI_SUBREAPER to register Tini as a child subreaper, or run Tini as PID 1.
 time="2019-06-04T10:36:25+02:00" level=info msg="Exporter not specified. Traces won't be exported"
 2019-06-04 10:36:25.862059 I | Starting application at port :3000
```

**To Reproduce**
Nothing special, it's written in the startup logs. `docker run --rm -p 3000:3000 gomods/athens:v0.4.0`

**Expected behavior**
Tini should by default be launched as subreaper

 - OS: Linux 64bit
 - Go version : N/A
 - Proxy version : v4.0.0

**Additional context**
Adding `-s` to the tini options should be sufficient. It can also be configured using environment variable, but it's not default.

kumy:
Added the `-s` parameter to launch `tini` as subreaper by default.

Fixes #1268

### DIFF
--- a/cmd/proxy/Dockerfile
+++ b/cmd/proxy/Dockerfile
@@ -28,6 +28,6 @@ ENV GO_ENV=production
 
 EXPOSE 3000
 
-ENTRYPOINT [ "/sbin/tini", "--" ]
+ENTRYPOINT [ "/sbin/tini", "-s", "--" ]
 
 CMD ["athens-proxy", "-config_file=/config/config.toml"]


marpio:
Closing as we decided to just add documentation instead.

PR #1273, opened 2019-06-05
Fixes helm chart volumeMounts if only netrc is enabled

briancharous:
Fixes a couple issues with the Helm chart regarding the netrc file.
* If only `netrc` is enabled but disk storage is disabled, previously the `volumeMounts` key of the deployment would not render causing the netrc mount to show up under the `ports` key.
* The documentation describes creating a Kubernetes secret with a key of `.netrc` but the chart was mounting the secret at `/etc/netrc/netrc` which is inconsistent with the docs.

### DIFF
--- a/charts/athens-proxy/templates/deployment.yaml
+++ b/charts/athens-proxy/templates/deployment.yaml
@@ -75,7 +75,7 @@ spec:
         {{- end }}
         {{- if .Values.netrc.enabled }}
         - name: ATHENS_NETRC_PATH
-          value: "/etc/netrc/netrc"
+          value: "/etc/netrc/.netrc"
         {{- end }}
         {{- if .Values.upstreamProxy.enabled }}
         - name: ATHENS_FILTER_FILE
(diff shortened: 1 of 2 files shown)


charts/athens-proxy/templates/deployment.yaml
```diff
@@ -91,8 +91,10 @@ spec:
         {{- end }}
         ports:
         - containerPort: 3000
-        {{- if eq .Values.storage.type "disk" }}
+        {{- if or (eq .Values.storage.type "disk") .Values.upstreamProxy.enabled .Values.netrc.enabled }}
```
marpio: Just a question - why is `.Values.upstreamProxy.enabled` included?

briancharous: I believe that's because if the upstream proxy is enabled, the filter file configmap gets mounted into `/usr/local/lib` on line 102.

rimusz:
LGTM

arschles:
What @rimusz said 🤣

PR #1276, opened 2019-06-07
minio : Use no. of objects for Exists check

marpio:
What is the problem I am trying to address?

How is the fix applied?

I'm checking if the len of objects equals 3 (zip, mod, info)

### DIFF
--- a/pkg/storage/minio/checker.go
+++ b/pkg/storage/minio/checker.go
@@ -19,15 +19,25 @@ func (v *storageImpl) Exists(ctx context.Context, module, version string) (bool,
 	defer span.End()
 	versionedPath := v.versionLocation(module, version)
 	modPath := fmt.Sprintf("%s/go.mod", versionedPath)
-	_, err := v.minioClient.StatObject(v.bucketName, modPath, minio.StatObjectOptions{})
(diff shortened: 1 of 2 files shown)


pkg/storage/minio/checker.go
```diff
@@ -19,15 +18,25 @@ func (v *storageImpl) Exists(ctx context.Context, module, version string) (bool,
...
 	modPath := fmt.Sprintf("%s/go.mod", versionedPath)
-	_, err := v.minioClient.StatObject(v.bucketName, modPath, minio.StatObjectOptions{})
+	infoPath := fmt.Sprintf("%s/%s.info", versionedPath, version)
+	zipPath := fmt.Sprintf("%s/source.zip", versionedPath)
 
-	if minio.ToErrorResponse(err).Code == minioErrorCodeNoSuchKey {
-		return false, nil
-	}
+	var count int
+	objectCh, _ := v.minioCore.ListObjectsV2(v.bucketName, versionedPath, "", false, "", 0, "")
```
marwan-at-work: any reason for not handling the error here?

marpio: no, it should be handled.

PR #1280, opened 2019-06-09
add warning logs for situations where files are missing from a packag…

### LINKED ISSUE #1171: Add logs to warn when not all 3 files (mod, info and zip) are present

The file system storage currently checks for all three files (`.mod`, `.zip`, `.info`) exist and fails any one endpoint if they don't _all_ exist. This can be problematic if, for some reason, 1/3 or 2/3 of the files exist. In this case, Athens will "fall back" to downloading the module from the VCS and _overwriting_ it in its storage.

That's not an optimal solution, but without a larger refactoring, it's the best we can do

We should clearly log when we run into a situation when a module doesn't have all 3 files present. The log should be a "warning" with detailed information on why this happens and what Athens is doing to fix it (i.e. falling back to the VCS), and what that means (the module might get overwritten)

A larger refactor of how files are stored on disk

**Additional context**

This is a follow-up from https://github.com/gomods/athens/pull/1143#issuecomment-474602212.

cByst:
…e in athens storage

The file system storage currently checks for all three files (.mod, .zip, .info) exist and fails any one endpoint if they don't all exist. This results in the entire module being refreshed from vcs overwriting what ever is currently in Athens storage for this version of this module.

**How is the fix applied?**

The solution is to log a warning about what is happening when this one of these situations occurs.

Fixes #1171

### DIFF
--- a/pkg/download/addons/with_pool.go
+++ b/pkg/download/addons/with_pool.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/gomods/athens/pkg/download"
 	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/log"
 	"github.com/gomods/athens/pkg/storage"
 )
 
@@ -64,13 +65,13 @@ func (p *withpool) List(ctx context.Context, mod string) ([]string, error) {
 	return vers, nil
 }
 
-func (p *withpool) Info(ctx context.Context, mod, ver string) ([]byte, error) {
+func (p *withpool) Info(ctx context.Context, mod, ver string, lggr log.Entry) ([]byte, error) {
 	const op errors.Op = "pool.Info"
 	var info []byte
 	var err error
 	done := make(chan struct{}, 1)
 	p.jobCh <- func() {
-		info, err = p.dp.Info(ctx, mod, ver)
+		info, err = p.dp.Info(ctx, mod, ver, lggr)
 		close(done)
 	}
 	<-done
@@ -96,13 +97,13 @@ func (p *withpool) Latest(ctx context.Context, mod string) (*storage.RevInfo, er
 	return info, nil
 }
 
-func (p *withpool) GoMod(ctx context.Context, mod, ver string) ([]byte, error) {
+func (p *withpool) GoMod(ctx context.Context, mod, ver string, lggr log.Entry) ([]byte, error) {
 	const op errors.Op = "pool.GoMod"
 	var goMod []byte
 	var err error
 	done := make(chan struct{}, 1)
 	p.jobCh <- func() {
-		goMod, err = p.dp.GoMod(ctx, mod, ver)
+		goMod, err = p.dp.GoMod(ctx, mod, ver, lggr)
 		close(done)
 	}
 	<-done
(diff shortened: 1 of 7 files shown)


pkg/download/addons/with_pool.go
```diff
@@ -64,13 +65,13 @@ func (p *withpool) List(ctx context.Context, mod string) ([]string, error) {
 	return vers, nil
 }
 
-func (p *withpool) Info(ctx context.Context, mod, ver string) ([]byte, error) {
+func (p *withpool) Info(ctx context.Context, mod, ver string, lggr log.Entry) ([]byte, error) {
```
marwan-at-work: We don't need to update the function signatures all over, because a `log.Entry` is already existent in the Context map. All you have to do is call `log.EntryFromContext(ctx)` and you will get the Entry. If the Entry is not there, you'll get a NoOp logger so that we never panic.

pkg/download/protocol.go
```diff
@@ -146,13 +147,14 @@ func (p *protocol) Latest(ctx context.Context, mod string) (*storage.RevInfo, er
...
 
-func (p *protocol) Info(ctx context.Context, mod, ver string) ([]byte, error) {
+func (p *protocol) Info(ctx context.Context, mod, ver string, lggr log.Entry) ([]byte, error) {
 	const op errors.Op = "protocol.Info"
 	ctx, span := observ.StartSpan(ctx, op.String())
 	defer span.End()
 	info, err := p.storage.Info(ctx, mod, ver)
 	var newVer string
 	if errors.IsNotFoundErr(err) {
+		lggr.Warnf("missing info file for module: %s restoring entire module from vcs this will overwrite anything existing for version: %s of module: %s in storage", mod, ver, mod)
```
marwan-at-work: This is probably the wrong place to put the `lggr.Warnf` because this means that for **every** non existent Module, we'll see this warning.

We should probably only log this Entry when we actually *do* have .GoMod and .Zip but not .Info (and so on). 

In other words, we should log this in the `fs.Exists` method because it's the only one that checks for *all* three files at the same time. 

For now, I think logging in fs.Exists is enough but in the future we should probably return a specific Error Kind such as `errors.IncompleteModule` and have the Download Protocol log a Warning only if the error was of that kind.

cByst: @marwan-at-work thanks for the feedback. I was originally looking at fs.Exists to do the logging but I'm having some trouble understanding what fs.Exists is really used for. From what I can tell the only time fs.Exists is called is after the module has been fetched from VCS and only if semver doesn't equal the requested version.

cByst: Also just for reference I added a log message to the top of fs.Exists and then replicated the situation we are talking about running into here and I never see fs.Exists being called. Any help understanding this a bit more would be greatly appreciated.

marwan-at-work: Thanks for looking more into this. I missed that the `fs` package does not directly use it in its `GoMod/Info/Zip` methods. Other storage implementations use it. Here's some background:

`Exists` currently exists for some backend storages to use it before serving an object (info/mod/zip) such as the `azure` implementation. Or they use to make sure a module is never overwritten (such as the `mongo` implementation). 

More importantly, our `stash` package uses it for implementing a Single Flight and for when `go mod download` converts dynamic versions such as `master` and `latest` to semver versions. See https://github.com/gomods/athens/blob/master/pkg/stash/stasher.go#L53:L64

Long story short, we should definitely do one of the following

1. Do the warning on the Save method, when 1-or-2 out of 3 files are saved. 
2. Do the warning on the `Stash` method when Save returns a new error Kind such as `KindSaveIncomplete`  
3. Have the `download` package _always_ call `Exists` on the `Info` method/endpoint. This is because the `.info` endpoint always comes first before Go calls for `.mod` and `.zip`. This way, we don't waste too many redundant calls. 

I think no.1 is the easiest and least error-prone to do. However, it will only warn once per Save. It might be more favorable in the future to implement 2 and/or 3. 

Also, this is definitely opening up some eyes into how the `Exists` method is a bit confusing because the codebase does not clearly state its intention and how/where it should be used. 

I think no.2 and no.3 would help in that regard and we should definitely make all storages do the same behavior (all should use .Exists in the same exact places/logic) 

If any of this is still too foreign, please feel free to reach out to me on #athens gopher slack channel and I'm more than happy to guide you more.

Thanks!

arschles:
@cByst are you still up for continuing with this PR? I think there are only a few things left before we can merge it.

Absolutely no pressure either way. Just let us know and we can pick it up if you want.

cByst:
@arschles my bad I think I lost track of this one. I will try to take a look at it this week and try to clean it up.

arschles:
@cByst not a problem! Just let us know when you're ready, we'll be here 😄

PR #1284, opened 2019-06-12
[debug] allow to expose pprof via config

### LINKED ISSUE #1177: Enable pprof/expvar/goroutine dumps
I would like to debug my current running athens proxy.  It would be great if [pprof](https://golang.org/pkg/net/http/pprof/) or [expvar](https://golang.org/pkg/expvar/) were exposed (at least optionally in the config).  This would allow me to collect goroutine dumps or inspect environment variables so I can debug the proxy.

benjaminch:

This CL allows to activate `pprof` when needed to investigate any potential performances issues and / or proceed to profilings.

**How is the fix applied?**

Introducing two new configuration params
allowing to activate pprof.
- `ATHENS_ACTIVATE_DEBUG_HANDLER`
- `ATHENS_DEBUG_HANDLER_PORT`

If `ATHENS_ACTIVATE_DEBUG_HANDLER` is set to `true` then `pprof` is started on the `ATHENS_DEBUG_HANDLER_PORT` port which is ideally different from the application port for security reasons as stated [here](https://mmcloughlin.com/posts/your-pprof-is-showing).

`pprof` won't be exposed by default.

Fixes #1177

### DIFF
--- a/config.dev.toml
+++ b/config.dev.toml
@@ -43,6 +43,14 @@ LogLevel = "debug"
 # Env override: ATHENS_CLOUD_RUNTIME
 CloudRuntime = "none"
 
+# EnablePprof specifies if the pprof endpoints should be exposed.
+# Note that this option is not meant to be activated forever on a server
+# and should be desactivated once not needed.
+EnablePprof = false
+
+# PprofPort specifies the port on which pprof will be exposed if activated.
+PprofPort = ":3001"
+
 # The filename for the include exclude filter.
 # Env override: ATHENS_FILTER_FILE
 #
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -25,6 +25,8 @@ type Config struct {
 	ProtocolWorkers  int       `validate:"required" envconfig:"ATHENS_PROTOCOL_WORKERS"`
 	LogLevel         string    `validate:"required" envconfig:"ATHENS_LOG_LEVEL"`
 	CloudRuntime     string    `validate:"required" envconfig:"ATHENS_CLOUD_RUNTIME"`
+	EnablePprof      bool      `envconfig:"ATHENS_ENABLE_PPROF"`
+	PprofPort        string    `envconfig:"ATHENS_PPROF_PORT"`
 	FilterFile       string    `envconfig:"ATHENS_FILTER_FILE"`
 	TraceExporterURL string    `envconfig:"ATHENS_TRACE_EXPORTER_URL"`
 	TraceExporter    string    `envconfig:"ATHENS_TRACE_EXPORTER"`
@@ -78,6 +80,8 @@ func defaultConfig() *Config {
 		ProtocolWorkers:  30,
 		LogLevel:         "debug",
 		CloudRuntime:     "none",
+		EnablePprof:      false,
+		PprofPort:        ":3001",
 		StatsExporter:    "prometheus",
 		TimeoutConf:      TimeoutConf{Timeout: 300},
 		StorageType:      "memory",
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -10,6 +10,8 @@ import (
 	"os/signal"
 	"time"
 
+	_ "net/http/pprof"
+
 	"github.com/gomods/athens/cmd/proxy/actions"
 	"github.com/gomods/athens/pkg/build"
 	"github.com/gomods/athens/pkg/config"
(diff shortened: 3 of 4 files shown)


benjaminch:
Hello guys ! 

This is my first PR here :)
The first version is pretty naive but I would like to get your thoughts about it.

I am feeling like it would be good to add some tests but I am not quite sure where to put those. IMO those are e2e tests since we would like to verify that:

1. `pprof` is not activated by default.
2. when `ATHENS_ACTIVATE_DEBUG_HANDLER=true` then `pprof` is exposed on port `ATHENS_DEBUG_HANDLER_PORT`

Also, maybe good to document this feature somewhere (maybe in a developper / configuration page)?

Let me know what you guys think.

Thanks ! :)

config.dev.toml
```diff
@@ -43,6 +43,14 @@ LogLevel = "debug"
 # Env override: ATHENS_CLOUD_RUNTIME
 CloudRuntime = "none"
 
+# ActivateDebugHandler specifies if the pprof endpoints should be exposed.
+# Note that this option is not meant to be activated forever on a server
+# and should be desactivated once not needed.
+ActivateDebugHandler = false
```
marwan-at-work: What do you think about calling this `EnablePprof` ? I think most Go developers would immediately know what this means since `pprof` is part of the standard tooling. Debug, on the other hand could mean a lot of things, such as verbose debug logging.

config.dev.toml
```diff
@@ -43,6 +43,14 @@ LogLevel = "debug"
...
+
+# DebugHandlerPort specifies the port on which pprof will be exposed if activated.
+DebugHandlerPort = ":8081"
```
marwan-at-work: same as above, we can just call it `PprofPort`.

Also since we run at port 3000 by default, we should make this port 3001. Unles 8081 is a convention port for pprof? Not sure :)

pkg/config/config.go
```diff
@@ -19,36 +19,38 @@ const defaultConfigFile = "athens.toml"
...
-	SingleFlightType string    `envconfig:"ATHENS_SINGLE_FLIGHT_TYPE"`
-	SingleFlight     *SingleFlight
-	Storage          *StorageConfig
+	GoEnv                string    `validate:"required" envconfig:"GO_ENV"`
+	GoBinary             string    `validate:"required" envconfig:"GO_BINARY_PATH"`
+	GoGetWorkers         int       `validate:"required" envconfig:"ATHENS_GOGET_WORKERS"`
+	ProtocolWorkers      int       `validate:"required" envconfig:"ATHENS_PROTOCOL_WORKERS"`
+	LogLevel             string    `validate:"required" envconfig:"ATHENS_LOG_LEVEL"`
+	CloudRuntime         string    `validate:"required" envconfig:"ATHENS_CLOUD_RUNTIME"`
+	ActivateDebugHandler bool      `envconfig:"ATHENS_ACTIVATE_DEBUG_HANDLER"`
```
marwan-at-work: same here, `ATHENS_ENABLE_PPROF`

config.dev.toml
```diff
@@ -43,6 +43,14 @@ LogLevel = "debug"
 # Env override: ATHENS_CLOUD_RUNTIME
 CloudRuntime = "none"
 
+# ActivateDebugHandler specifies if the pprof endpoints should be exposed.
+# Note that this option is not meant to be activated forever on a server
+# and should be desactivated once not needed.
```
marwan-at-work: typo: r/desactivated/deactivated

Also, AFAIK it's totally okay to keep it activated forever because it does not affect performance or the program unless the user actually pings the `/pprof` endpoint. Right?

config.dev.toml
benjaminch: Yes, you are right, it will be clearer :) 
I was thinking that it might be relavant to activate other tools as well but if needed later, will change the config name.

config.dev.toml
benjaminch: No problem, no it's not a convention.

pkg/config/config.go
benjaminch: Sure

config.dev.toml
benjaminch: Yes !
Regarding having it enabled forever, I don't think it's a good idea. I can be source of DoS attacks  since it loads a lot of stuff on call. I think it's better to state it so people are aware of it and enable it when needed or in particular envs.

benjaminch:
Thanks for the review @marwan-at-work !
I was wondering, don't you think we should add tests on pprof activation? Do you think the doc is enough for this feature?

Cheers !

config.dev.toml
marwan-at-work: Sounds good :)

marwan-at-work:
@benjaminch A test would certainly not be bad, but since we're just enabling/proxying a tool and not writing any core business logic, I'm okay if don't write tests for this particular PR. So up to you (and others' opinions who review this PR as well). Thanks!

benjaminch:
> @benjaminch A test would certainly not be bad, but since we're just enabling/proxying a tool and not writing any core business logic, I'm okay if don't write tests for this particular PR. So up to you (and others' opinions who review this PR as well). Thanks!

Alright !
If others thinks tests are needed and can guide me to where those should be added, I will be happy to add those.

Cheers !

benjaminch:
Thanks @marwan-at-work :)

PR #1297, opened 2019-07-04
Updates to mongo getter to reduce duplicate code

bou1der:

In the mongo/getter implementation there is duplicated code in the Info and GoMod functions. The purpose of this PR is to reduce the duplicated code without any changes to the operation of either of the functions. 

There was also a difference between how the Info and GoMod functions passed the reference for the query result to the Decode function. The Info passed a pointer to the reference while the GoMod function just passed the reference. Mongo handles both so there was no error during run time but this refactor ensures that Info and GoMod call the mongo.Decode function the same way.

**How is the fix applied?**

The common/duplicated code from Info and GoMod functions has been moved into a new getter.Query function.

Fixes # (No Existing Issue)

### DIFF
--- a/pkg/storage/mongo/getter.go
+++ b/pkg/storage/mongo/getter.go
@@ -16,26 +16,11 @@ import (
 // Info implements storage.Getter
 func (s *ModuleStore) Info(ctx context.Context, module, vsn string) ([]byte, error) {
 	const op errors.Op = "mongo.Info"
-	ctx, span := observ.StartSpan(ctx, op.String())
-	defer span.End()
-	c := s.client.Database(s.db).Collection(s.coll)
 
-	result := &storage.Module{}
+	result, err := Query(ctx, s, module, vsn, op)
 
-	tctx, cancel := context.WithTimeout(ctx, s.timeout)
-	defer cancel()
-
-	queryResult := c.FindOne(tctx, bson.M{"module": module, "version": vsn})
-	if queryErr := queryResult.Err(); queryErr != nil {
-		return nil, errors.E(op, queryErr, errors.M(module), errors.V(vsn))
-	}
-
-	if err := queryResult.Decode(&result); err != nil {
-		kind := errors.KindUnexpected
-		if err == mongo.ErrNoDocuments {
-			kind = errors.KindNotFound
-		}
-		return nil, errors.E(op, err, kind, errors.M(module), errors.V(vsn))
+	if err != nil {
+		return nil, errors.E(op, err)
 	}
 
 	return result.Info, nil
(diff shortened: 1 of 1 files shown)


pkg/storage/mongo/getter.go
```diff
@@ -16,26 +16,11 @@ import (
 // Info implements storage.Getter
 func (s *ModuleStore) Info(ctx context.Context, module, vsn string) ([]byte, error) {
 	const op errors.Op = "mongo.Info"
-	ctx, span := observ.StartSpan(ctx, op.String())
```
manugupt1: Can you keep this span here to collect traces?

pkg/storage/mongo/getter.go
```diff
@@ -44,24 +29,11 @@ func (s *ModuleStore) Info(ctx context.Context, module, vsn string) ([]byte, err
 // GoMod implements storage.Getter
 func (s *ModuleStore) GoMod(ctx context.Context, module, vsn string) ([]byte, error) {
 	const op errors.Op = "mongo.GoMod"
-	ctx, span := observ.StartSpan(ctx, op.String())
```
manugupt1: Please leave it as it is to collect traces?

pkg/storage/mongo/getter.go
```diff
@@ -91,3 +63,30 @@ func (s *ModuleStore) Zip(ctx context.Context, module, vsn string) (io.ReadClose
 
 	return dStream, nil
 }
+
+// Query connects and queries storage module
+func Query(ctx context.Context, s *ModuleStore, module, vsn string, op errors.Op) (*storage.Module, error) {
```
manugupt1: Can you have it as `func (s* moduleStore) query(ctx, module, .....)`  this way query will be private to the package?

bou1der: Definitely, that makes a lot more sense.

pkg/storage/mongo/getter.go
bou1der: No problem. Thanks for taking the time to review these changes!

pkg/storage/mongo/getter.go
bou1der: No problem - Should I also start a trace for the new function (query) or will the trace on Info capture the desired metrics?

manugupt1: yes! that is super important imo as that is where we will do majority of compute time. By having traces here, we can see which function the query was a subset of

```
----Info----- ------GoMod------
 --query---     ----query---
```

arschles:
@manugupt1 looks like @bou1der made your requested changes, so I'm going to click the "dismiss review" button so we can get this merged

PR #1298, opened 2019-07-04
Add mod version decoding to fix branches with capital letters

### LINKED ISSUE #1292: Can't "go get" branches with capital letters in their name
**Describe the bug**
Attempting to `go get` a branch with capital letters in the name currently results in a HTTP 500. It appears that this is because Athens isn't decoding the branch name before using it. This is outlined in [this issue](https://github.com/golang/go/issues/26456) in the golang/go repository.

**Error Message**
If applicable add error message to help explain your problem or console output enclose in triple back-ticks eg,

```
$ go get -u bitbucket.org/icelolly/<...>@ICE-163-<...>
go: finding bitbucket.org/icelolly/<...> ICE-163-<...>
go: finding bitbucket.org/icelolly ICE-163-<...>
go: finding bitbucket.org ICE-163-<...>
go get bitbucket.org/icelolly/<...>@ICE-163-<...>: unexpected status (http://athens.<...>/bitbucket.org/icelolly/<...>/@v/ICE-163-<...>.info): 500 Internal Server Error
```

(Side note here, this is the same thing that happens if the branch doesn't exist, the error could probably be more useful / obvious, like returning a 404 if the branch wasn't found. In this case, the branch _definitely_ exists).

**To Reproduce**

1. Create a branch on Go Modules based project that includes capital letters.
2. Attempt to download that branch via `go get`.

**Expected behavior**

It to successfully download the branch.

 - OS: `Linux <...> 5.1.15-arch1-1-ARCH #1 SMP PREEMPT Tue Jun 25 04:49:39 UTC 2019 x86_64 GNU/Linux`
 - Go version: `go version go1.12.6 linux/amd64`
 - Proxy version: `0.4.0` (I think, I'm using the Helm chart, version 0.2.11, but if you go to the version endpoint of Athens, that just reports "unset" as the version).
 - Storage: `fs`

t-tomalak:

To properly handle module version (branches) with capital letters. 

**How is the fix applied?**

Added path decode for module version which caused issue for branches with capital letters. Used same function which is used to decode module names.

Fixes #1292

### DIFF
--- a/pkg/paths/path.go
+++ b/pkg/paths/path.go
@@ -25,7 +25,7 @@ func GetVersion(r *http.Request) (string, error) {
 	if version == "" {
 		return "", errors.E(op, "missing version parameter")
 	}
-	return version, nil
+	return DecodePath(version)
 }
 
 // AllPathParams holds the module and version in the path of a ?go-get=1


t-tomalak:
I'm thinking that we could probably add repository with branch which would have capital letter to athens-artifacts to handle such cases in our e2e tests.

PR #1299, opened 2019-07-06
pkg/download: return 404 on list/latest

marwan-at-work:
In 1.13, the Go command deduces a "module root" from an "import path" concurrently which means all paths are tried for `github.com/pkg/errors` such as `github.com/pkg/errors`, `github.com/pkg` and `github.com`. 

However, if the proxy returns a non 404/410 on any of the above calls, then the Go command will log to the user the returned error which could confuse the user thinking the `go get` command failed even though it didn't. 

For now, all attempts for `go mod` commands should return 404/410 until the mentioned issue in this PR gets resolved. Will follow up with another PR for .info/.mod/.zip

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,7 @@
 module github.com/gomods/athens
 
+go 1.12
+
 require (
 	cloud.google.com/go v0.26.0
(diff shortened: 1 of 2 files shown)


arschles:
Ref https://github.com/golang/go/issues/30134

PR #1302, opened 2019-07-09
Fixed Minio NewStorage to return the correct error

### LINKED ISSUE #1295: Minio MakeBucket errors not being returned.
**Describe the bug**
[See code for minio MakeBucket](https://github.com/gomods/athens/blob/a5eba1812bc48a4df1d9e7868288ed8feb31cdfd/pkg/storage/minio/minio.go#L39-L47)
If MakeBucket fails, the BucketExists call will overwrite the real error.  The result is a generic error being logged.

**Error Message**

```console
error getting storage configuration (Internal Server Error)
```
**To Reproduce**
Set the minio bucket name to an invalid name like "bucket_test".  The underscore is not an acceptable character.

```console
$ make dev
$ ATHENS_STORAGE_TYPE=minio ATHENS_MINIO_BUCKET_NAME=bucket_test make run
```

**Expected behavior**
Return the "MakeBucket" error.  For this specific error, the message should be:

```console
error getting storage configuration (Bucket name contains invalid characters)
```

 - OS: Linux 64bit
 - Go version : go version go1.12.6 linux/amd64
 - Proxy version : N/A for this test case
 - Storage (fs/mongodb/s3 etc.) : Minio

I'll submit a pull request to address the issue in the next few days.  Will be away till Friday.

jamesnaftel:
The incorrect error was being bubbled up when using an "_" in the in Minio bucket name.  It was returning the BucketExists error which uses a less strict form of error checking.

**How is the fix applied?**
The fix simply ignores the BucketExists error (the execution path is already inside an error condition) when there is an error.  If the bucket does exist, it correctly returns a storage backend.   I added test cases locally to minio_test.go since they are specific to minio and not the overall compliance tests.  Let me know if this isn't ok but I felt it made more sense.

Fixes #1295

### DIFF
--- a/pkg/storage/minio/minio.go
+++ b/pkg/storage/minio/minio.go
@@ -39,8 +39,9 @@ func NewStorage(conf *config.MinioConfig, timeout time.Duration) (storage.Backen
 	err = minioClient.MakeBucket(bucketName, region)
 	if err != nil {
 		// Check to see if we already own this bucket
-		exists, err := minioClient.BucketExists(bucketName)
-		if err == nil && !exists {
+		exists, _ := minioClient.BucketExists(bucketName)
+		if !exists {
+			// MakeBucket Error takes priority
 			return nil, errors.E(op, err)
 		}
 	}
--- a/pkg/storage/minio/minio_test.go
+++ b/pkg/storage/minio/minio_test.go
@@ -13,6 +13,66 @@ func TestBackend(t *testing.T) {
 	compliance.RunTests(t, backend, backend.clear)
 }
 
+// TestNewStorageExists tests the logic around MakeBucket and BucketExists
+func TestNewStorageExists(t *testing.T) {
+	url := os.Getenv("ATHENS_MINIO_ENDPOINT")
+	if url == "" {
+		t.SkipNow()
+	}
+
+	tests := []struct {
+		name         string
+		deleteBucket bool
+	}{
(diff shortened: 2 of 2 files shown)


pkg/storage/minio/minio.go
```diff
@@ -39,8 +39,9 @@ func NewStorage(conf *config.MinioConfig, timeout time.Duration) (storage.Backen
 	err = minioClient.MakeBucket(bucketName, region)
 	if err != nil {
 		// Check to see if we already own this bucket
-		exists, err := minioClient.BucketExists(bucketName)
-		if err == nil && !exists {
+		exists, _ := minioClient.BucketExists(bucketName)
+		if !exists {
```
arschles: @jamesnaftel should we check `err` after this check, in case there was something else? Or will an error always be `nil` if `exists` is true?

jamesnaftel: In the current minio-go client, true only gets return with a nil.  We could chain the BuckExists error in the failure case with the MakeBucket error but it looks like we would usually get the same error.

[minio-go: BucketExists](https://github.com/minio/minio-go/blob/26dec1a6ef2e943908b058233dbd091b411696d5/api-stat.go#L31)

arschles: Got it, that makes sense. Thanks for explaining @jamesnaftel !

PR #1304, opened 2019-07-12
support fetch from goproxy

### LINKED ISSUE #1294: support fetch from proxy when download url provided
1. now download url is provided for redirect, when async redirect mode specified, and the fetcher still download module from original repo only. If the original repo is not reachable, the async fetch will always fail.
2. download rules is only applied to info/mod/zip apis, list is leave unconfigurable.

provide a method to both tell `go mod download`/`go list` fetch from proxy(i.e. set GOPROXY=${download url})

yittg:

support fetch from goproxy instead of from the original repo

**How is the fix applied?**

provide a config/env for GOPROXY to Go mod/list Command

Fixes #1294

### DIFF
--- a/config.dev.toml
+++ b/config.dev.toml
@@ -15,6 +15,13 @@ GoBinary = "go"
 # Env override: GO_ENV
 GoEnv = "development"
 
+# GoProxy specifies GOPROXY env for go list or mod download inside athens
+# which can be configured totally same with GOPROXY of Go Command.
+# Notes that the comma-separated GOPROXY (e.g. <proxy1>,<proxy2>,direct) is only available in Go 1.13 or higher,
+# otherwise only single proxy URL can be set.
+# Env override: GOPROXY
+GoProxy = "direct"
+
 # GoGetWorkers specifies how many times you can concurrently
 # go mod download, this is so that low performance instances
 # can manage go get more sanely and not run out of disk or memory.
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -21,6 +21,7 @@ type Config struct {
 	TimeoutConf
 	GoEnv            string    `validate:"required" envconfig:"GO_ENV"`
 	GoBinary         string    `validate:"required" envconfig:"GO_BINARY_PATH"`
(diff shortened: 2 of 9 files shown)


yittg:
cc @marwan-at-work

pkg/config/config.go
```diff
@@ -48,6 +48,7 @@ type Config struct {
 	NoSumPatterns    []string  `envconfig:"ATHENS_GONOSUM_PATTERNS"`
 	DownloadMode     mode.Mode `envconfig:"ATHENS_DOWNLOAD_MODE"`
 	DownloadURL      string    `envconfig:"ATHENS_DOWNLOAD_URL"`
+	DownloadProxy    bool      `envconfig:"ATHENS_DOWNLOAD_PROXY"`
```
marwan-at-work: Let's call this `GoProxy` and the env itself will also be `GOPROXY`

pkg/contexts/context.go
```diff
@@ -0,0 +1,50 @@
+package contexts
```
marwan-at-work: no need for this package, you can just pass something directly if need be

yittg:
@marwan-at-work thanks for your suggestions, it also meet my scenario :), I'll push a simpler solution.

yittg:
@marwan-at-work a pure GOPROXY version is pushed, please have a look.

config.dev.toml
```diff
@@ -15,6 +15,11 @@ GoBinary = "go"
 # Env override: GO_ENV
 GoEnv = "development"
 
+# GoProxy specifies GOPROXY env for list/download command
+# which can be configured totally same with GOPROXY of Go Command
+# e.g. <proxy1>,<proxy2>,direct
```
marwan-at-work: Forgot to mentionn: we should note that the comma-separated feature are only available in Go 1.13 and higher since the Dockerfile still uses Go 1.12

yittg: fixed it.

yittg:
Almost done, please trigger it to merge if no more problems, :)

PR #1309, opened 2019-07-19
ref: upstream lister and it's impl to module pkg

yittg:
it's a refactor about upstream lister which should in module package.

cc @arschles

### DIFF
--- a/pkg/module/go_vcs_lister.go
+++ b/pkg/module/go_vcs_lister.go
@@ -1,4 +1,4 @@
-package download
+package module
 
 import (
 	"bytes"
@@ -10,17 +10,10 @@ import (
 
 	"github.com/gomods/athens/pkg/config"
 	"github.com/gomods/athens/pkg/errors"
-	"github.com/gomods/athens/pkg/module"
 	"github.com/gomods/athens/pkg/observ"
 	"github.com/gomods/athens/pkg/storage"
 	"github.com/spf13/afero"
 )
-
-// UpstreamLister retrieves a list of available module versions from upstream
-// i.e. VCS, and a Storage backend.
-type UpstreamLister interface {
-	List(ctx context.Context, mod string) (*storage.RevInfo, []string, error)
-}
 
 type listResp struct {
 	Path     string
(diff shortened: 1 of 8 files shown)


pkg/module/go_runtime.go
```diff
@@ -0,0 +1,67 @@
+package module
```
marwan-at-work: let's call this file according to the only function it's in: `prepare_env.go` instead of `go_runtime` which can be confusing as to what it means.

pkg/module/go_runtime.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+)
+
+// PrepareEnv will return all the appropriate
+// environment variables for a Go Command to run
+// successfully (such as GOPATH, GOCACHE, PATH etc)
+func PrepareEnv(gopath, goProxy string) []string {
```
marwan-at-work: let's make this function private: `preparEnv` now that no one is using it outside this package AFAIK.

pkg/module/go_vcs_lister.go
```diff
@@ -61,8 +54,8 @@ func (l *vcsLister) List(ctx context.Context, mod string) (*storage.RevInfo, []s
 	if err != nil {
 		return nil, nil, errors.E(op, err)
 	}
-	defer module.ClearFiles(l.fs, gopath)
-	cmd.Env = module.PrepareEnv(gopath, l.goProxy)
+	defer ClearFiles(l.fs, gopath)
```
marwan-at-work: I think we can make `ClearFiles` also private

pkg/module/go_runtime.go
yittg: i want to provide unified functions relative to go command execute in this file next.
by now, it also ok to name it as `prepare_env.go`, :)

yittg:
@marwan-at-work comments fixed, PTAL

pkg/module/prepare_env.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+)
+
+// prepareEnv will return all the appropriate
+// environment variables for a Go Command to run
+// successfully (such as GOPATH, GOCACHE, PATH etc)
+func prepareEnv(gopath, goProxy string) []string {
```
marpio: I'm not strictly against it but does moving this function to a separate file improves anything? Having to jump to another file to read it seems cumbersome to me when the function is used only in one place.

yittg: it is used by `go_get_fetcher` and `go_vcs_lister` both. 
May we merge lister and fetcher, then all the three could be put in one file.
But for now, they implement different interface, i think it would be better to keep them independent to make the refactor clear.

arschles:
@marwan-at-work are you ok with this to merge?

PR #1310, opened 2019-07-19
[WIP] ref: go cmd runtime

yittg:
following part of #1309, for providing unified go command runtime.
mark as WIP before #1309 merged.
cc @arschles

### DIFF
--- a/pkg/module/go_get_fetcher.go
+++ b/pkg/module/go_get_fetcher.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"runtime"
 	"strings"
 
 	"github.com/gomods/athens/pkg/errors"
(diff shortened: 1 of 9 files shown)


pkg/module/go_get_fetcher.go
```diff
@@ -42,9 +39,11 @@ func NewGoGetFetcher(goBinaryName string, goProxy string, fs afero.Fs) (Fetcher,
 		return nil, errors.E(op, err)
 	}
 	return &goGetFetcher{
-		fs:           fs,
-		goBinaryName: goBinaryName,
-		goProxy:      goProxy,
+		fs: fs,
+		goTool: goTool{
```
marwan-at-work: Nesting arguments inside a private struct feels like an overkill, the strut name is `goGetFetcher` which already implies the fact that it uses a "goTool" i don't think it needs another layer of abstraction.

pkg/module/go_vcs_lister.go
```diff
@@ -30,43 +21,25 @@ type listResp struct {
...
-		config.FmtModVer(mod, "latest"),
-	)
-	cmd.Dir = tmpDir
-	stdout := &bytes.Buffer{}
-	stderr := &bytes.Buffer{}
-	cmd.Stdout = stdout
-	cmd.Stderr = stderr
-
-	gopath, err := afero.TempDir(l.fs, "", "athens")
+	runtime, err := prepareRuntime(l.fs, l.goTool, "")
```
marwan-at-work: `runtime` is a very confusing name. What runtime exactly? The Go runtime? The machine runtime?

pkg/module/go_vcs_lister.go
```diff
@@ -30,43 +21,25 @@ type listResp struct {
...
 	if err != nil {
 		return nil, nil, errors.E(op, err)
 	}
-	defer module.ClearFiles(l.fs, gopath)
-	cmd.Env = module.PrepareEnv(gopath, l.goProxy)
+	defer runtime.clean()
 
-	err = cmd.Run()
+	err = runtime.run("list", "-m", "-versions", "-json", config.FmtModVer(mod, "latest"))
```
marwan-at-work: this could be called `gocmd.run(...)` but i still prefer us to flat out the logic instead of abstract it.

pkg/module/go_get_fetcher.go
```diff
@@ -55,23 +54,10 @@ func (g *goGetFetcher) Fetch(ctx context.Context, mod, ver string) (*storage.Ver
 	ctx, span := observ.StartSpan(ctx, op.String())
 	defer span.End()
 
-	// setup the GOPATH
-	goPathRoot, err := afero.TempDir(g.fs, "", "athens")
+	m, cleanFunc, err := downloadModule(g.goTool, g.fs, mod, ver)
```
marwan-at-work: the function `Fetch` implements the contract of `module.Fetcher`, and i think it's okay to flat out the implementation inside of it instead of abstract it, so I feel that we should keep it the way it is.

yittg:
@marwan-at-work  Sincerely thank you for your suggestions prompting me to rethink more. My intention is to abstract a unified command build and execution. Indeed, the implementation may be a bit overkill. hha, thanks again. 😄

marwan-at-work:
@yittg I think it might be a good idea to merge the `module.Fetcher` interface with the `module.UpstreamListerInterface` by making one interface that has the `Fetch` method as well as the `List` method since they both take the same dependencies and initialization. 

But for now, let's see how the other PR looks before making further refactors. 

Thanks!

yittg:
@marwan-at-work sounds good, keep as it is for now.

marpio:
I'm with Marwan on that - I think the current code is easy to read and nesting it doesn't provide much value.

PR #1312, opened 2019-07-20
storage/fs,minio: fix list module with same prefix

### LINKED ISSUE #1311: There is a bug in the minior/lister.go that can cause get wrong versions.
if moudule = "github.com/labstack/echo",objectCh will contains the version of module "github.com/labstack/echo/v4" ,  because list all the file related to searchPrefix.

image

yittg:
check semver and go.mod existence to verify a module version

Fixes #1311

### DIFF
--- a/pkg/storage/compliance/tests.go
+++ b/pkg/storage/compliance/tests.go
@@ -63,40 +63,42 @@ func testNotFound(t *testing.T, b storage.Backend) {
 func testListSuffix(t *testing.T, b storage.Backend) {
 	ctx := context.Background()
 
-	otherMod := "github.com/one/two-other"
-	mock := getMockModule()
-	err := b.Save(
-		ctx,
-		otherMod,
-		"v0.9.0",
-		mock.Mod,
-		mock.Zip,
-		mock.Info,
-	)
-	require.NoError(t, err, "Save for storage failed")
-	modname := "github.com/one/two"
-	versions := []string{"v1.1.0", "v1.2.0", "v1.3.0"}
-	for _, version := range versions {
-		mock := getMockModule()
-		err := b.Save(
-			ctx,
-			modname,
-			version,
-			mock.Mod,
-			mock.Zip,
-			mock.Info,
-		)
-		require.NoError(t, err, "Save for storage failed")
+	modVers := map[string][]string{
+		"github.com/one/two":       {"v1.1.0", "v1.2.0", "v1.3.0"},
+		"github.com/one/two/v2":    {"v2.1.0"},
+		"github.com/one/two-other": {"v0.9.0"},
+		"github.com/one":           {}, // not a module but a valid query, so no versions
+	}
+	for modname, versions := range modVers {
+		for _, version := range versions {
+			mock := getMockModule()
+			err := b.Save(
+				ctx,
+				modname,
+				version,
+				mock.Mod,
+				mock.Zip,
+				mock.Info,
+			)
+			require.NoError(t, err, "Save for storage failed")
+		}
 	}
 	defer func() {
-		b.Delete(ctx, otherMod, "v0.9.0")
-		for _, ver := range versions {
-			b.Delete(ctx, modname, ver)
+		for modname, versions := range modVers {
+			for _, version := range versions {
+				b.Delete(ctx, modname, version)
+			}
 		}
 	}()
-	retVersions, err := b.List(ctx, modname)
-	require.NoError(t, err)
-	require.Equal(t, versions, retVersions)
+	for modname, versions := range modVers {
+		retVersions, err := b.List(ctx, modname)
+		require.NoError(t, err)
+		if len(versions) == 0 {
+			require.Empty(t, retVersions)
+		} else {
+			require.Equal(t, versions, retVersions)
+		}
+	}
 }
 
 // testList tests that a storage Backend returns
--- a/pkg/storage/fs/lister.go
+++ b/pkg/storage/fs/lister.go
@@ -3,10 +3,12 @@ package fs
 import (
 	"context"
 	"os"
+	"strings"
 
 	"github.com/gomods/athens/pkg/errors"
 	"github.com/gomods/athens/pkg/observ"
 	"github.com/spf13/afero"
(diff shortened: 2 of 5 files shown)


yittg:
cc @marwan-at-work @marpio

pkg/storage/compliance/tests.go
```diff
@@ -63,40 +63,37 @@ func testNotFound(t *testing.T, b storage.Backend) {
...
-			version,
-			mock.Mod,
-			mock.Zip,
-			mock.Info,
-		)
-		require.NoError(t, err, "Save for storage failed")
+	modVers := map[string][]string{
+		"github.com/one/two":       {"v1.1.0", "v1.2.0", "v1.3.0"},
+		"github.com/one/two/v2":    {"v2.1.0"},
+		"github.com/one/two/three": {"v0.9.0"},
```
marwan-at-work: let's keep the old test of `github.com/one/two-other != github.com/one/two` to this new mix of scenarios.

pkg/storage/fs/lister.go
```diff
@@ -3,12 +3,35 @@ package fs
...
 
+func (l *storageImpl) isSemVerDir(module string, verFileInfo os.FileInfo) (bool, error) {
+	if !verFileInfo.IsDir() {
+		return false, nil
+	}
+	ver := verFileInfo.Name()
+	if ver[0] != 'v' {
+		return false, nil
+	}
+	if _, err := semver.NewVersion(ver[1:]); err != nil {
```
marwan-at-work: Go's interpretation of SemVer is not the same as the Spec which is what core-os I believe follows. Therefore, this could lead to some false negatives. You can check out godoc.org/golang.org/x/mod/semver for Go's interpretation.

pkg/storage/minio/lister.go
```diff
@@ -23,21 +36,15 @@ func (l *storageImpl) List(ctx context.Context, module string) ([]string, error)
...
+	ret := []string{}
 	for _, object := range objectCh.Contents {
 		if object.Err != nil {
 			return nil, errors.E(op, object.Err, errors.M(module))
 		}
-		parts := strings.Split(object.Key, "/")
-		ver := parts[len(parts)-2]
-		if _, ok := dict[ver]; !ok {
-			dict[ver] = struct{}{}
+		if ver, ok := l.getModuleSemVerFromKey(module, object.Key); ok {
```
marwan-at-work: same as above, no need for this function. You can check that the path is correct by just making sure the module paths are equal. Therefore, let's keep the old code and add those 3 new lines: 

```
modpath := strings.Join(parts[:len(parts)-2], "/")
if modpath != module {
	continue
}
```

yittg: basically agree, a bit different from the 3 new lines with checking go mod key only

pkg/storage/fs/lister.go
```diff
@@ -24,7 +44,9 @@ func (l *storageImpl) List(ctx context.Context, module string) ([]string, error)
 	}
 	ret := []string{}
 	for _, fileInfo := range fileInfos {
-		if fileInfo.IsDir() {
+		if ok, err := l.isSemVerDir(module, fileInfo); err != nil {
```
marwan-at-work: let's short-circuit this by removing the `isSemVerDir` and just doing this check: `fileInfo.IsDir() && len(fileInfo.Name()) > 2`

marwan-at-work: @marpio pointed out that checking length > 2 wouldn't work because you might have `v10` which is len == 3 and so on and so forth. 

Basically, I'm hoping to get around making extra system calls because the /list endpoint is already a slow operation. 

Two ways to get around this is by 
1. Just checking if `!strings.Contains(version, ".")` 
2. By checking that the directory name is *not* `vN` where N is a valid integer

yittg: @marwan-at-work @marpio thanks for suggestions about performance.

Only check `vN` is not enough, say we have a module `github.com/one/two` with some versions cached, a query for module `github.com/one` is valid which is expected to get nothing.

I committed another fix to only check semver without stats `go.mod` file, the only edge case left  uncovered i think is that a module contains a full semver part like `github.com/one/v1.1.1`, which maybe no need to deal with.

yittg:
@arschles @marwan-at-work  @marpio PTAL again.

pkg/storage/fs/lister.go
marwan-at-work: @yittg then let's go with the first option, I think it should be enough because no major version will ever have a `.` in it. 
In other words, there shouldn't be a need to use `semver` package at all, right?

yittg: I don't think so, dot in import path is valid besides in host part, maybe rarely though, not only in version. Full semver check is essential, i think.

marwan-at-work: @yittg right, but when you're inside a module folder (in fs storage), there are no sub packages, there are only versions (v0.3.0/...) and major versions such as (v2/v2.1.3/...).

marwan-at-work: would sub modules be included there by any chance?

marwan-at-work: It might be worth making an example in github.com/athens-artifacts

yittg: Say we have a repo `github.com/athens-artifacts/a.b.c`, athens will save a cache in fs by `go get github.com/athens-artifacts/a.b.c`, next `go get github.com/athens-artifacts` is a valid request may not correct will got a list contains `a.b.c` which is not a version though.

marwan-at-work: @yittg that makes sense, I actually created https://github.com/athens-artifacts/submod-conflict where we can add real tests to (in another PR) 

You can notice that `v1.9.9` is actually not accepted as a sub-module. So option 2 actually does work (parsing vN). But sub-modules will also be included. So checking that semver.Canonical should account for both cases. 

Thanks again for looking into this :+1:

yittg: yeah, some particular integrated tests for these senarios are required.

PR #1314, opened 2019-07-24
Write tests for the new Mongo query function

### LINKED ISSUE #1301: Write a test for the new Mongo query function

This is a follow-up to https://github.com/gomods/athens/pull/1297#pullrequestreview-259171326

We should have a test for the `query` function that was introduced in #1297. We probably will need to make it an integration test, since it will need to connect to a real MongoDB server

N/A

**Additional context**

This is a follow-up to #1297. I've talked to @bou1der about it, and he is going to try to do it soon

bou1der:

I am adding tests around the mongo query function. The query function contains common code that is used by both the Info and GoMod functions in getter.go. 

**How is the fix applied?**

I created two integration tests in mongo_test.go that test the function when there are modules in the database and when modules cannot be found. These tests are very similar to the existing tests in compliance.go that are used to test the Info and GoMod functions that in turn call query(). 

I am having trouble triggering unexpected type errors from within the mongo driver reliably. I would like to add tests for a non 'nil' queryResult error and KindUnexpected if possible (see snippet below).  

```
        if queryErr := queryResult.Err(); queryErr != nil {
		return nil, errors.E(op, queryErr, errors.M(module), errors.V(vsn))
	}

	if err := queryResult.Decode(result); err != nil {
		kind := errors.KindUnexpected
		if err == mongo.ErrNoDocuments {
			kind = errors.KindNotFound
		}
		return nil, errors.E(op, err, kind, errors.M(module), errors.V(vsn))
	}
```

I don't believe it is possible with a mock due to how the mongo driver is written and I am at a loss to a clean way to test it as part of an integration test. I really appreciate any comments, suggestions, or ideas!

Fixes #1301

### DIFF
--- a/pkg/storage/mongo/mongo_test.go
+++ b/pkg/storage/mongo/mongo_test.go
@@ -41,18 +41,7 @@ func getStorage(tb testing.TB) *ModuleStore {
 	return backend
 }
 
-func TestGetterQueryModuleNotFoundError(t *testing.T) {
-	modname, ver := "xxx", "yyy"
-
-	ctx := context.Background()
-	backend := getStorage(t)
-
-	_, err := query(ctx, backend, modname, ver)
-	require.Error(t, err)
-	require.Equal(t, errors.KindNotFound, errors.Kind(err))
-}
-
-func TestGetterQueryModuleExists(t *testing.T) {
+func TestQueryModuleVersionExists(t *testing.T) {
 	modname, ver := "getTestModule", "v1.2.3"
 	mock := &storage.Version{
 		Info: []byte("123"),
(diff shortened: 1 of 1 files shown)


bou1der:
Thanks, @arschles! I looked at adding the test for this PR but it will take more time to wrap my head around it than I expected. I wrote a new issue #1327 to capture the discussion for later.

marpio:
@bou1der since you created #1327, are you OK with merging that as it is?

bou1der:
@marpio Yes, that sounds good to me 😄

manugupt1:
@bou1der I am super sorry! but I am going to have to re-request a review. Can you resolve conflicts

bou1der:
@manugupt1 If it's okay with you, I'm going to wait until #1331 is merged so we don't have the same conflicts again. This change and #1331 add tests to the mongo_test.go file and those changes are conflicting with one another.

manugupt1:
I have merged it now!

PR #1321, opened 2019-08-05
Removing gofmt vendor ignores

arschles:
We don't have a vendor directory anymore :smile: :tada:

We have code in `./scripts/check_gofmt.sh` that ignores go code in the `vendor/` directory, but we don't have a `vendor` directory anymore

**How is the fix applied?**

I removed the check

### DIFF
--- a/scripts/check_gofmt.sh
+++ b/scripts/check_gofmt.sh
@@ -4,5 +4,5 @@
 # Fail if a .go file hasn't been formatted with gofmt
 set -euo pipefail
 
-GO_FILES=$(find . -iname '*.go' -type f | grep -v /vendor/)   # All the .go files, excluding vendor/
+GO_FILES=$(find . -iname '*.go' -type f)   # All the .go files
 test -z $(gofmt -s -d $GO_FILES | tee /dev/stderr)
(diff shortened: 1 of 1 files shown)


arschles:
drone says that formatting is broken. could be because I changed the code that does the formatting 😲 . gonna look at this as soon as I can

scripts/check_gofmt.sh
```diff
@@ -4,5 +4,5 @@
 # Fail if a .go file hasn't been formatted with gofmt
 set -euo pipefail
 
-GO_FILES=$(find . -iname '*.go' -type f | grep -v /vendor/)   # All the .go files, excluding vendor/
+GO_FILES=$(find . -iname '*.go' -type f)   # All the .go files
 test -z $(gofmt -s -d $GO_FILES | tee /dev/stderr)
```
chriscoffee: Wrap double quotes around variables and command substitution, so `"$GOFILES"` and it should stop complaining

arschles:
@chriscoffee thanks for the tip. Shellcheck is happy now. Can you re-review and merge if you're ok with this?

PR #1331, opened 2019-08-08
Fix Mongo initialization to use environment variables for db/coll names

bou1der:

The mongo database's initialization code uses hardcoded values for the database and collection names. An environment variable for the database name exists as ATHENS_MONGO_DEFAULT_DATABASE but is not used during the mongo initialization. There is no existing environment variable for the collection name. 

**How is the fix applied?**

Mongo will use the values set in the mongo storage configuration or will default to "athens" or "modules" during initialization. The environment variable, ATHENS_MONGO_DEFAULT_COLLECTION, is added to set the name of the collection. This will default to "modules" if not set. The value stored in the existing environment variable ATHENS_MONGO_DEFAULT_DATABASE is now used to set the mongo database name or will revert to the default of "athens" if not set. The configuration documentation was updated to reflect the environment variables. 

Fixes # N/A

### DIFF
--- a/docs/content/configuration/storage.md
+++ b/docs/content/configuration/storage.md
@@ -80,6 +80,10 @@ This driver uses a [Mongo](https://www.mongodb.com/) server as data storage. On 
             # Env override: ATHENS_MONGO_DEFAULT_DATABASE
             DefaultDBName = athens
 
+            # Not required parameter
+            # Allows for use of custom collection 
+            # Env override: ATHENS_MONGO_DEFAULT_COLLECTION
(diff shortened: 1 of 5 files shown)


manugupt1:
@bou1der I am super sorry! Can you please resolve conflicts?

bou1der:
@manugupt1 whoops 🤦‍♂ I tried to fix the conflicts in via Github and it looks like I missed some formatting/extra brackets. I'll wrap this up later today from my computer. Sorry about that.

PR #1332, opened 2019-08-09
fix proxy not work when lower case is unset.

psy-core:

When `HTTP_PROXY`( `HTTPS_RPOXY` or `NO_PROXY`) environment variable is set but `http_proxy`(`https_proxy` or `no_proxy` ) enviromnent variable is unset, The upper case environment variable not work.

**How is the fix applied?**

Set the lower case env variable only when it exist.

simplely add non-empty judgement before set the three lower case variable into `cmdEnv`.


jpreese:
I posted this in Slack, but for completeness, dumping it here as well.

Are you running windows @psy-core ? 

My guess is that because HTTP_PROXY is set, but http_proxy is not set, when Athens passes all of the environment variables, the http_proxy value is overwriting the HTTP_PROXY value as Windows doesn't care about casing.

psy-core:
@arschles @jpreese I'm running athens in docker. I used to set `HTTPS_PROXY` and `NO_PROXY` env variables at version 0.3.0, and it works well. Recently I upgraded version to 0.5.0, and found that my proxy isn't work well. Instead, environment variable `https_proxy` works perfectly. Upper case proxy environment variables not work in version 0.5.0. I think it should be that the upper case variables are overrided by the lower case, even if lower case variable value is empty.

I verified this commit on my local machine (mac Mojave, docker version 18.09.2),  and the upper case proxy variable works well when the lower case is unset.

marpio:
@arschles this is the PR that added support for the lower case env vars:
https://github.com/gomods/athens/pull/1193
The reasoning was, that go supports both upper and lower case so we should too.

The problem now is that we will override the upper case version with the lower case one, which will probably be just an empty string.
This PR will solve that issue. The only edge case to consider is when both the upper and lower case versions are set. I think the upper case should take precedence.
What do you think?

marpio:
Apparently the lower case version is used even if it's set to "".

jpreese:
That is correct. Windows will continue to overwrite the last value in the array because it is case insensitive. You can see the Env implementation here https://golang.org/src/os/exec/exec.go

jpreese:
I mentioned in slack as well, but worth mentioning here as well, but I think there's value in re-assessing if `prepare_env.go` needs to exist at all. I'm not familiar with its history, but all it seems to be doing is taking the environment variables from the environment and passing it to command.exec

If we don't have any strong opinions about the particulars of the variables, we can just let command.exec pull them from the environment natively, which would also fix this issue.

marwan-at-work:
@jpreese there were a couple of reasons: 
1. Security, so as to not leak any unintended env vars into sub-processes 
2. Being explicit which avoids any side-effects the underlying machine might have over Athens operations. Plus, we get to see exactly what env vars do to the underlying `go mod download` command and why the process cares about them.

jpreese:
Gotcha, makes sense to me! Thanks for the clarification.

psy-core:
This PR is blocked for about one month. Is it ok for merge or need further discussion?

psy-core:
@arschles  Thank you for acception.  😄

PR #1342, opened 2019-08-14
add support for a host in ensurePortFormat

jhawk28:
using a host enables the ability to bind to a specific interface

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -2,16 +2,16 @@ package config
 
 import (
 	"fmt"
-	"log"
-	"os"
-	"path/filepath"
-	"runtime"
-
 	"github.com/BurntSushi/toml"
 	"github.com/gomods/athens/pkg/download/mode"
 	"github.com/gomods/athens/pkg/errors"
 	"github.com/kelseyhightower/envconfig"
-	validator "gopkg.in/go-playground/validator.v9"
+	"gopkg.in/go-playground/validator.v9"
+	"log"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strconv"
 )
 
 const defaultConfigFile = "athens.toml"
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -187,14 +189,16 @@ func envOverride(config *Config) error {
 	return nil
 }
 
-func ensurePortFormat(s string) string {
-	if len(s) == 0 {
-		return ""
+func ensurePortFormat(hostport string) string {
+	host, port, err := net.SplitHostPort(hostport)
+	if err != nil && strings.Contains(err.Error(), "missing port in address") {
```
marwan-at-work: i'm not a big fan of string matching in logic, this is because there's no guarantee that the error message will always be the same. If the `net` library changed the string, we'll break. 

One way to do it is basically this: 
1. Check if the string is a number, if it is, then add a `:` to it and return. 
2. If it's not a number, then just return it as is. 

The function should be much simpler: 

```golang
func ensurePortFormat(s string) string {
  if _, err := strconv.Atoi(s); err == nil {
    return ":" + s
  }
  return s
}
```

I believe this should cover all 3 cases of `3000`, `:3000`, `<ip>:3000`

jhawk28: agree. The `net` package doesn't expose the error so that would be breakage across lots of projects.

jhawk28: I updated so that it uses the suggested ensurePortFormat

pkg/config/config.go
```diff
@@ -2,16 +2,16 @@ package config
...
-	"path/filepath"
-	"runtime"
-
 	"github.com/BurntSushi/toml"
 	"github.com/gomods/athens/pkg/download/mode"
 	"github.com/gomods/athens/pkg/errors"
 	"github.com/kelseyhightower/envconfig"
-	validator "gopkg.in/go-playground/validator.v9"
+	"gopkg.in/go-playground/validator.v9"
+	"log"
```
marwan-at-work: can you please run gofmt/goimports so that standard library imports are above and separated from third party dependencies?

PR #1345, opened 2019-08-18
Add GOROOT  to prepareEnv

vahid-sohrabloo:

If go need to define GOROOT get error

**How is the fix applied?**
Add GOROOT env to prepareEnv

### DIFF
--- a/pkg/module/prepare_env.go
+++ b/pkg/module/prepare_env.go
@@ -23,6 +23,7 @@ func prepareEnv(gopath, goProxy string) []string {
 	gopathEnv := fmt.Sprintf("GOPATH=%s", gopath)
 	goProxyEnv := fmt.Sprintf("GOPROXY=%s", goProxy)
 	cacheEnv := fmt.Sprintf("GOCACHE=%s", filepath.Join(gopath, "cache"))
+	goRoot := fmt.Sprintf("GOROOT=%s", os.Getenv("GOROOT"))
 	gitSSH := fmt.Sprintf("GIT_SSH=%s", os.Getenv("GIT_SSH"))
 	gitSSHCmd := fmt.Sprintf("GIT_SSH_COMMAND=%s", os.Getenv("GIT_SSH_COMMAND"))
 	disableCgo := "CGO_ENABLED=0"
@@ -31,6 +32,7 @@ func prepareEnv(gopath, goProxy string) []string {
 		pathEnv,
 		homeEnv,
 		gopathEnv,
+		goRoot,
 		goProxyEnv,
 		cacheEnv,
 		disableCgo,


marpio:
@valid-sohrabloo
Thanks for your contribution!
Can you describe the issue that is solved by this PR?
We didn't need this until now so I'm wondering why is it necessary to pass GOROOT.

vahid-sohrabloo:
The ZIP or tarball binary downloads from the golang.org website have a $GOROOT value of "/usr/local/go" and recommend you unpack them into that location. If you choose not to do this, then you must set $GOROOT to the location you chose.

marwan-at-work:
@vahid-sohrabloo out of curiosity, why should Athens care about the ZIP/tarball binary downloads from golang.org? 

Athens' job is to download "modules" using `go mod download` which usually just requires GOPATH and not GOROOT, unless im missing something. 

Thanks!

vahid-sohrabloo:
the default GOROOT  is "/usr/local/go" If I install in another directory I should set GOROOT. thus if I install go in another directory Athens will not work because go mod download will return error

marwan-at-work:
@vahid-sohrabloo for that you can use `GO_BINARY_PATH`. See [here](https://github.com/gomods/athens/blob/master/config.dev.toml#L11)

vahid-sohrabloo:
And if go bin and GOROOT in a different location?

marwan-at-work:
@vahid-sohrabloo AFAIK `go mod download` only needs the `go` binary path and not GOROOT (where the Go source and stdlib lives). Therefore, that should be enough. But if you know if a use case where that isn't, definitely let us know.

vahid-sohrabloo:
Install go in /usr/local/gonew location
move /usr/local/gonew/bin/go to /usr/local/bin/
set GoBinary = "/usr/local/bin/go" in athens config
then use athens proxy give me this error
```
ERRO[0006] exit status 2: go: cannot find GOROOT directory: /usr/local/go
```

marwan-at-work:
@vahid-sohrabloo I tried following those steps and it didn't reproduce for me. 

The reason is because every Go binary has the GOROOT set in the binary itself, and so even if you move the binary elsewhere, it'll still look for the original place where GOROOT was set. 

Here's my reproduction steps:  


```
Desktop me$ pwd
/Users/me/Desktop # the Go binary is out in Desktop


Desktop me$ ./go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/Users/me/marwan/go" # the GOROOT is somewhere totally different. The binary used to be in /Users/me/marwan/go/bin
...

# This worked for me:
Desktop me$ GOPATH=/Users/me/Desktop/temp GO111MODULE=on ./go mod download github.com/pkg/errors@v0.8.1
go: finding github.com/pkg/errors v0.8.1

```

I'll try to reproduce it with Athens to see if there's any difference but this poses another question: 

1. In what use case do you have your Go binary different from your GOROOT? Why not have your GO_BINARY_PATH be within the GOROOT as it usually is? In other words, you're still able to have multiple Go distributions on one machine and Athens is configurable to use any of them.

vahid-sohrabloo:
Did you test my use case? it makes the error.
For some complex reason, I can't put GO bin in the GOROOT.
I know it is an unusual UseCase. But I need to set GOROOT to another location

arschles:
@vahid-sohrabloo I don't think @marwan-at-work has been able to reproduce it since he commented, and I haven't been able to either. I am going to close this PR now, but please feel free to comment here if `GO_BINARY_PATH` still isn't going to work for you? We can continue the discussion from there then.

PR #1353, opened 2019-08-24
Add plugin structure for local and remote storage connectors

### LINKED ISSUE #1110: Plugin architecture for storages

Kunde21:
As discussed in #1110 there isn't a way to set up a new storage backend without committing the code to the main athens repo.  

This is a plugin architecture that allows for local and remote storage implementations using the same `storage.Backend` interface.  The connection to the server is done via a gRPC server/client setup with defaults in place.  

Fixes #1110


arschles:
@Kunde21 Can you let me know if you're still interested in doing this PR? If so, no rush at all. If not, I'll be happy to take it over.

Issue #1336, opened 2019-08-12
Confusing behavior with missing (required) configuration values


bluekeyes:
**Describe the bug**

_(This is arguably 3 distinct bugs, but they are related and each led to the next.)_

(1) The documentation suggested to me that configuration properties that are not set in a configuration file will use the documented defaults; this is the behavior if environment variables are used for configuration. Instead, any properties not explicitly defined in the configuration file use the zero values for their types.

(2) For some properties (I hit it on `DownloadMode`, but may apply to others), the zero value is invalid and the property is effectively required, but is not marked for validation.

(3) When the the `DownloadMode` property is missing (i.e. it is an empty string), the logged error does not clearly indicate the problem or what the cause was, printing a vague `Bad Request` error (even though no network request was made.)

**Error Message**

```console
2019-08-12 20:11:55.325066 I | error adding proxy routes (Bad Request)
```

**To Reproduce**

1. Create a configuration file that does not specify a `DownloadMode`
2. Start the server using the configuration file

**Expected behavior**

1. Values not explicitly set in a configuration file use the documented defaults
2. Failing that, all required properties fail validation with an error showing they are required
3. Failing that, logged errors include details about the operation/code that failed

**Environment:**
 - OS: Docker on Ubuntu 16.04
 - Go version: ?
 - Proxy version: v0.5.0
 - Storage (fs/mongodb/s3 etc.): s3

arschles:
@bluekeyes Do you think a log to indicate what's missing, and removing docs that talk about defaults be a decent start to solving these 3 problems?

bluekeyes:
I think logging some of the additional details attached to errors and making sure all required properties are marked with the `required` tag for validation would be a good start: that existing validation was pretty helpful when I first started the server with my minimal configuration file.

I'm not sure what the best way to improve the docs is, since the defaults are still relevant for people using environment variables. On a closer reading, I realized some parts say that you should copy the development configuration file and modify it as needed, so maybe that could be emphasized on the "Installing Athens" or "Configuring Athens" sections?

twexler:
I also just ran into this when handwriting my own config, I think the easy change is to make `DownloadMode` required, but I think the correct fix is to actually set defaults when they should be.

marwan-at-work:
[...]

PR #1365, opened 2019-08-30
Don't return an http error when a DownloadMode is invalid

twexler:

Right now `mode.NewFile()` returns an HTTP 400 error when `Config.DownloadMode` is not set. This doesn't make a lot of sense, because the consumer of the error is not an HTTP client, but an administrator attempting to configure Athens for the first time. 

**How is the fix applied?**

`mode.NewFile()` no longer returns HTTP errors, and instead returns either `mode.ErrDownloadMode` or a formatted error stating that the specified `DownloadMode` is not valid. This also adds some basic testing to ensure the function returns expected error values.

Refs #1336

### DIFF
--- a/pkg/download/mode/mode.go
+++ b/pkg/download/mode/mode.go
@@ -2,7 +2,6 @@ package mode
 
 import (
 	"encoding/base64"
-	e "errors"
 	"fmt"
 	"io/ioutil"
 	"path"
@@ -19,16 +18,14 @@ type Mode string
 
 // DownloadMode constants. For more information see config.dev.toml
 const (
-	Sync           Mode = "sync"
-	Async          Mode = "async"
-	Redirect       Mode = "redirect"
-	AsyncRedirect  Mode = "async_redirect"
-	None           Mode = "none"
-	invalidModeErr      = "unrecognized download mode: %s"
+	Sync            Mode = "sync"
+	Async           Mode = "async"
+	Redirect        Mode = "redirect"
+	AsyncRedirect   Mode = "async_redirect"
+	None            Mode = "none"
+	downloadModeErr      = "download mode is not set"
+	invalidModeErr       = "unrecognized download mode: %s"
 )
-
-// ErrNoDownloadMode is returned if the specified mode is empty string
-var ErrNoDownloadMode = e.New("download mode is not set")
 
 // DownloadFile represents a custom HCL format of
 // how to handle module@version requests that are
@@ -55,6 +52,11 @@ type DownloadPath struct {
 // directly.
 func NewFile(m Mode, downloadURL string) (*DownloadFile, error) {
 	const op errors.Op = "downloadMode.NewFile"
+
+	if m == "" {
+		return nil, errors.E(op, downloadModeErr)
+	}
+
 	if strings.HasPrefix(string(m), "file:") {
 		filePath := string(m[5:])
 		bts, err := ioutil.ReadFile(filePath)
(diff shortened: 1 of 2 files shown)


pkg/download/mode/mode.go
```diff
@@ -64,11 +69,16 @@ func NewFile(m Mode, downloadURL string) (*DownloadFile, error) {
 		}
 		return parseFile(bts)
 	}
+
+	if m == "" {
```
marwan-at-work: let's have this check at the top of the function right after the `op` declaration

pkg/download/mode/mode.go
```diff
@@ -64,11 +69,16 @@ func NewFile(m Mode, downloadURL string) (*DownloadFile, error) {
...
+		return nil, errors.E(op, ErrNoDownloadMode)
+	}
+
 	switch m {
 	case Sync, Async, Redirect, AsyncRedirect, None:
 		return &DownloadFile{Mode: m, DownloadURL: downloadURL}, nil
 	default:
-		return nil, errors.E(op, "unrecognized download mode: "+m, errors.KindBadRequest)
+		return nil, errors.E(op, fmt.Sprintf(invalidModeErr, m))
```
marwan-at-work: let's still pass a `errors.KindBadRequest` just in case the caller in the future needs to check if there was a parsing error, or a user error

pkg/download/mode/mode.go
```diff
@@ -18,13 +19,17 @@ type Mode string
...
+	Sync           Mode = "sync"
+	Async          Mode = "async"
+	Redirect       Mode = "redirect"
+	AsyncRedirect  Mode = "async_redirect"
+	None           Mode = "none"
+	invalidModeErr      = "unrecognized download mode: %s"
 )
 
+// ErrNoDownloadMode is returned if the specified mode is empty string
+var ErrNoDownloadMode = e.New("download mode is not set")
```
marwan-at-work: Given our `errors` package, this only needs to be a string, our errors package will actually call `errors.New` for you here: https://github.com/gomods/athens/blob/master/pkg/errors/errors.go#L96 so there's no need to import the errors package here. You can also put this under `invalidModeErr`

pkg/download/mode/mode.go
twexler: Okay, I think I'll pass it as an argument _before_ the human-friendly string, so that the log message makes sense in this case

marwan-at-work: @twexler the order doesn't matter, I think the reason the friendly message was not showing up is because of the fact that `"some string " + m` was actually a Mode type, and not a string that our errors package didn't understand. That's a very tiny bug for later :)

twexler: Order does matter, if you look at this:
https://github.com/gomods/athens/blob/master/pkg/errors/errors.go#L91-L106 the _last_ error will get set to `e.Err`, which is then used in `Error.Error()` as the error string to return.

marwan-at-work: The `errors.KindBadRequest` is an int, which will never get set as the *error*, but it will get set as the *Kind*. The problem with the confusing error message was happening here: https://github.com/gomods/athens/blob/master/pkg/errors/errors.go#L107:L109

So whether it comes first, or last shouldn't matter. 

That is because we never passed a `string` in the args, which a mistake in the codebase itself. I think a more accurate way to do it, is to log-out (or even panic) unrecognized types such as mode.Mode

twexler: You're absolutely right, I thought those actually implemented `error`, my bad.

I'm happy to switch these errors to full on panics, although they become harder to test for.

marwan-at-work: @twexler for now, let's leave the errors package and fix it in another PR. I think a `panic` should work, because it will call out bugs in our own codebase, even if it's reported by a user (unfortunately)

twexler: Sounds good, I just pulled in the latest changes.

twexler:
@marwan-at-work ready to merge when you are!

PR #1366, opened 2019-08-30
stasher/azure blob test: use - instead of _ for container name 

marpio:
_ is an invalid char in container name so let's use - instead.

### DIFF
--- a/pkg/stash/with_azureblob_test.go
+++ b/pkg/stash/with_azureblob_test.go
@@ -105,7 +105,7 @@ func randomContainerName(prefix string) string {
 	// see https://github.com/technosophos/moniker for more details
 	namer := moniker.New()
 	if prefix != "" {
-		return fmt.Sprintf("%s_%s", prefix, namer.NameSep(""))
+		return fmt.Sprintf("%s-%s", prefix, namer.NameSep(""))
 	}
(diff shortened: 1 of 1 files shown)


pkg/stash/with_azureblob_test.go
```diff
@@ -84,13 +85,15 @@ func (ms *mockAzureBlobStasher) Stash(ctx context.Context, mod, ver string) (str
...
 
-func getAzureTestConfig(containerName string) *config.AzureBlobConfig {
+func getAzureTestConfig(t *testing.T, containerName string) *config.AzureBlobConfig {
 	key := os.Getenv("ATHENS_AZURE_ACCOUNT_KEY")
 	if key == "" {
+		t.Log("ATHENS_AZURE_ACCOUNT_KEY not provided.")
 		return nil
 	}
-	name := os.Getenv("ATHENS_AZURE_ACCOUNT_NAME")
+	name := "athens" //os.Getenv("AZURE_ACCOUNT_NAME")
```
arschles: can you remove the commented `os.Getenv` call?

arschles:
@marpio Are you up for continuing with this PR? No pressure either way, just let me know and I can pick it up if you want.

Issue #1341, opened 2019-08-13
Webhook for proactively downloading modules into storage


arschles:
**Is your feature request related to a problem? Please describe.**

It would be useful to tell Athens to download all of a module's versions before they are needed.

**Describe the solution you'd like**

Essentially, I would like to have a webhook to tell Athens to do the equivalent of priming its cache. This webhook would do the following, possibly in the background:

- Determine all of the versions of a module (the same thing that `go list` does)
- Start downloading all those modules behind the scenes
- Return a HTTP 202 Accepted response

**Describe alternatives you've considered**

We could simply do a `curl athens.server/github.com/my/module/@v/list` command and then a `curl athens.server/github.com/my/module/@v/$VERSION.info` on the client side to do the same thing, but this has a few advantages I can see:

- If Athens does a `go list` and sees some versions already in its storage, it can skip those versions immediately (the alternative is accepting an HTTP request and returning a small amount of info metadata)
- Athens can do the downloads in the background, without taking up any protocol workers

**Additional context**

This issue narrows the scope that https://github.com/gomods/athens/pull/1195 originally had

cc/ @marwan-at-work @Kunde21 @manugupt1

arschles:
Duplicate of #1068, closing

PR #1369, opened 2019-09-03
Adding skeleton webhooks

arschles:

This isn't really a problem, but it would be nice to have a way to tell Athens to "bulk fetch" a bunch of modules all at once. That functionality would be nice to "warm up" an Athens before end-users start using it.

**How is the fix applied?**

I added a relatively simple webhook to allow this.

Fixes https://github.com/gomods/athens/issues/1341
Closes https://github.com/gomods/athens/pull/1195

(This is a simplified version of #1195, that doesn't touch any of the aliasing features because that is all up in the air)


arschles:
I started this PR during [office hours on September 3](https://teamup.com/event/show/id/8Mk2MjoxhgpJq9vDZHv4BY1PzhGoVP). Lots more work to do here, but I'd like to hear from folks what they think about this general idea.

arschles:
Closing this until we can have more discussion on this feature in https://github.com/gomods/athens/issues/1068

PR #1371, opened 2019-09-04
Updating the download config docs

arschles:

The [download mode](https://docs.gomods.io/configuration/download/) docs didn't say how to point Athens to the HCL file

**How is the fix applied?**

I added instructions on how to point Athens to the HCL file

N/A


jwentworth:
This looks clear to me! Thanks!

PR #1374, opened 2019-09-06
introduce go private configuration

### LINKED ISSUE #1373: introduce GOPRIVATE configuration for private modules
In #1294 #1304 , athens introduce a config to download from upstream proxy,
a approach to indicates module paths that are not publicly available and disable proxy is needed

The go env `GOPRIVATE` is what we want, we should introduce it as configuration also.

yittg:

provide an approach to disable `GOPROXY` for some modules

**How is the fix applied?**

introduce `GOPRIVATE` configuration

Fixes #1373

### DIFF
--- a/cmd/proxy/actions/app_proxy.go
+++ b/cmd/proxy/actions/app_proxy.go
@@ -73,12 +73,12 @@ func addProxyRoutes(
 	// 3. The stashpool manages limiting concurrent requests and passes them to stash.
 	// 4. The plain stash.New just takes a request from upstream and saves it into storage.
 	fs := afero.NewOsFs()
-	mf, err := module.NewGoGetFetcher(c.GoBinary, c.GoProxy, fs)
+	mf, err := module.NewGoGetFetcher(c.GoBinary, c.GoProxy, c.GoPrivate, fs)
 	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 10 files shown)


yittg:
it's good to support all of module aware go env, also to extend download files, if there's a plan for these, i'll vote for it.
it may be also good to supplement directly for GOPROXY, it's a server setup configuration, maybe we need not cover all configure path, just for the essentials.

yittg:
@marwan-at-work would we close this PR then, discuss in the original issue if needed?
cc @arschles .

arschles:
@yittg @marwan-at-work my understanding was that the use case for this was to selectively turn _off_ a `GOPROXY` for a module or glob of modules when the `downloadURL` is set. Did I get that wrong?

Anyway, if that was the goal here, I agree we should put it into the download mode file, just because that's where we're putting most of our module-specific configuration.

yittg:
@arschles Did you mean that we should extend the download file to support cache from upstream urls or only fetch from original repos. We discussed it in a previous Issue #1294 / PR #1304 before I mentioned in #1373. Maybe we should do more and deeper discuss about it.

arschles:
>Did you mean that we should extend the download file to support cache from upstream urls or only fetch from original repos

@yittg yup! Do you see a problem with adding the support in the download file? Happy to discuss more, feel free to `@` me on slack and we can discuss there or talk verbally

Issue #1390, opened 2019-09-20
v0.6.0 and Go1.13: unable to install new private repos


marwan-at-work:
In Go1.13, `go mod download` consults the sum db which means all private modules that are not in yet storage will fail. 

This means that all new modules that are private will fail in Athens v0.6.0 

Related: https://github.com/gomods/athens/issues/1386
Related: https://github.com/gomods/athens/issues/1383

PR #1391, opened 2019-09-20
cmd/proxy/actions: pass NoSumPatterns to GoGetFetcher

marwan-at-work:
This PR does two things: 

1. It adds a new config called `GoBinaryEnvVars` that will allow users to pass any env vars into `go mod download` and `go list -m` -- therefore it deprecates the recently added GoProxy config since that can just be passed in to the new config directly. 

2. This PR also makes it so that NoSumPatterns actually get turned into GONOSUMDB and added to the config above **if** the config is not present. 

~TODO: add unit tests~ done

Fixes https://github.com/gomods/athens/issues/1390


cmd/proxy/actions/app_proxy.go
```diff
@@ -74,12 +74,23 @@ func addProxyRoutes(
 	// 3. The stashpool manages limiting concurrent requests and passes them to stash.
 	// 4. The plain stash.New just takes a request from upstream and saves it into storage.
 	fs := afero.NewOsFs()
-	mf, err := module.NewGoGetFetcher(c.GoBinary, c.GoProxy, fs)
+
+	// TODO: remove before we release v0.7.0
+	if c.GoProxy != "direct" && c.GoProxy != "" {
+		fmt.Println("GoProxy is deprecated, please use GoBinaryEnvVars")
```
arschles: @marwan-at-work - use a logger here? We can do this in a follow-up PR if you like

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	cloud.google.com/go v0.26.0
 	contrib.go.opencensus.io/exporter/stackdriver v0.6.0
 	github.com/Azure/azure-storage-blob-go v0.7.0
+	github.com/Azure/go-autorest/autorest/adal v0.6.0 // indirect
```
arschles: I don't think you need to change anything because of this, but I wonder why this had to get added

cmd/proxy/actions/app_proxy.go
marwan-at-work: @arschles good point, i'll use it now

go.mod
marwan-at-work: probably due to using Go 1.13, it's added in the PR that fixes Thrift anyway :)

PR #1395, opened 2019-09-25
Add support for setting ATHENS_GONOSUM_PATTERNS in helm chart

rabadin:
When using go 1.13, `go get` will by default issue checksum requests to sum.golang.org. When using athens inside an organization and deploying it via the helm chart, it's important to be able to set ATHENS_GONOSUM_PATTERNS in order to avoid leaking internal information to the outside world.


rabadin:
On second thought, it's probably not that useful, the same thing can be done using `configEnvVars`. Closing.

PR #1397, opened 2019-09-26
Allow EnvList.Validate to have multiple = for GODEBUG var

### LINKED ISSUE #1396: EnvList.Validate doesn't allow GODEBUG with multiple = to be set
**Describe the bug**
Setting ATHENS_GO_BINARY_ENV_VARS=GODEBUG=netdns=cgo causes athens to throw an error.

**Error Message**
```console
incorrect env format: GODEBUG=netdns=cgo
```

**To Reproduce**
Run athens with ATHENS_GO_BINARY_ENV_VARS=GODEBUG=netdns=cgo

**Expected behavior**
I expected athens to run go commands with GODEBUG=netdns=cgo

Running the docker image
 - OS: Linux 64bit
 - Go version : 1.13
 - Proxy version : v0.6.1
 - Storage (fs/mongodb/s3 etc.) : fs

**Additional context**
Athens/Go is having issues with my companies internal bitbucket DNS causing athens API calls to fail more than half the time because it can't reach the host. Seems to be related to https://github.com/git-lfs/git-lfs/issues/3332 but the latest version of go did not fix it in my environment. Running athens with GODEBUG=netdns=cgo fixes my problem.

Mostly failing calls:
```
# GODEBUG="" GOPRIVATE=bitbucket.mysite.net go get bitbucket.mysite.net/cs/my-repo
go get bitbucket.mysite.net/cs/my-repo: git ls-remote -q https://bitbucket.mysite.net/scm/cs/my-repo.git in /root/go/pkg/mod/cache/vcs/a97a647e681b2d215498d6cad48a1fea66bbd44a6722ef8dd68169c96b637253: exit status 128:
    fatal: unable to access 'https://bitbucket.mysite.net/scm/cs/my-repo.git/': Could not resolve host: bitbucket.mysite.net
```

Successful calls:
```
# GODEBUG="netdns=cgo" GOPRIVATE=bitbucket.mysite.net go get bitbucket.mysite.net/cs/my-repo
go: downloading bitbucket.mysite.net/cs/my-repo v1.2.0
go: extracting bitbucket.mysite.net/cs/my-repo v1.2.0
...
```

kmhagan:
I would like to pass GODEBUG to athens that contains multiple "=" values.

Athens has issues connecting to my companies internal bitbucket when i don't use GODEBUG=netdns=cgo

**How is the fix applied?**
If the env variable is GODEBUG don't enforce only 1 "=" in the validation

Fixes #1396


kmhagan:
@arschles sure thing! Added tests

pkg/config/config.go
```diff
@@ -85,7 +85,8 @@ func (el *EnvList) Add(key, value string) {
 func (el EnvList) Validate() error {
 	const op errors.Op = "EnvList.Validate"
 	for _, env := range el {
-		if strings.Count(env, "=") != 1 {
+		// excluding GODEBUG because it can have multiple "=", such as GODEBUG=netdns=cgo
+		if strings.Count(env, "=") != 1 && !strings.HasPrefix(env, "GODEBUG=") {
```
marwan-at-work: maybe I was too strict to require the count to be 1, maybe we should just say `if strings.Count(env, "=") < 1` and therefore we don't need to make a special case for GODEBUG. This way, we can allow other env vars that might need multiple equal signs.

kmhagan: Makes sense to me! Updated

marwan-at-work:
@arschles gonna merge this to unblock, but definitely feel free to make comments still on this PR and I'm happy to make any follow up changes.

PR #1402, opened 2019-09-27
Add ability to annotate the proxy Service via Helm

RobAtticus:
It is often use to add custom annotations to a Service deployed
through Helm, so this PR adds the ability for a user to specify
annotations in their values file. For such use case we have seen
is when working in conjunction with external-dns, we want to assign
DNS properties of the Service, but currently cannot because this
Helm chart does not support it.

We are trying to apply custom annotations to the Athens k8s Service to make it work with our usage of external-dns.

**How is the fix applied?**

I implemented a pattern seen in several other helm charts.

(See above)

### DIFF
--- a/charts/athens-proxy/Chart.yaml
+++ b/charts/athens-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: athens-proxy
-version: 0.3.3
+version: 0.3.4
 appVersion: 0.6.0
 description: The proxy server for Go modules
 icon: https://raw.githubusercontent.com/gomods/athens/master/docs/static/banner.png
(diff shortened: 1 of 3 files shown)


RobAtticus:
The lint check is failing because I did not bump the version -- happy to make that change, but not sure if you want to handle that yourselves.

RobAtticus:
Just wanted to surface this as it is blocking us from using Athens in production.

RobAtticus:
No worries @arschles , thanks for the guidance. I've updated it and rebased on master, so it should be good to go. I'll find you on Slack after I return from lunch if its still waiting.

Thanks!

PR #1409, opened 2019-10-01
Use glob matching for package paths (Fix #1408)

### LINKED ISSUE #1408: Download directive paths do not match subpackages like GOPRIVATE
**Describe the bug**
When using a 'Download File' to configure download modes, it is not possible to match all packages with an import path at a domain level. For example, private repos behind a firewall.

**To Reproduce**
1. Create a `download.hcl` file and configure Athens to use it via `DownloadMode` config.

```hcl
downloadURL = "https://proxy.golang.org"

mode = "async_redirect"

download "*.example.com/*" {
    mode = "sync"
}
```

**Expected behavior**
All modules from _anything_ `example.com` should not redirect, but should download synchronously.

**Actual behavior**
All requests are redirected to proxy.golang.org (which is exactly what I **don't** want).

 - OS: All
 - Go version: 1.13
 - Proxy version: master
 - Storage: All

**Additional context**
The problem appears to stem from using `path.Match` for glob matching. For `path.Match`, `*` matches any sequence of **non-/** characters.

In order to support this as is, I need to define an indeterminate number of possible directory matches. For example:

```hcl
download "*.example.com/*" {
    mode = "sync"
}

download "*.example.com/*/*" {
    mode = "sync"
}

download "*.example.com/*/*/*" {
    mode = "sync"
}

download "*.example.com/*/*/*/*" {
    mode = "sync"
}

download "*.example.com/*/*/*/*/*" {
    mode = "sync"
}
```

The `GOPRIVATE`, `GONOPROXY`, and `GONOSUMDB` environment variable seem to handle this somehow.  I can define `GOPRIVATE="*.example.com"` and it works as expected for all packages that start with any `example.com` domain.

jorng:
This will match the behavior of GOPRIVATE/GONOPROXY/GONOSUMDB

This addresses an issue with download directives where subpackages do not match if there are any additional `/` characters (counter to how GOPRIVATE/GONOPROXY/GONOSUMDB all work)

**How is the fix applied?**
Matched the behavior used by `go mod`. Shared the new matching code with SumDB matcher as well.

Fixes #1408

### DIFF
--- a/pkg/download/mode/mode.go
+++ b/pkg/download/mode/mode.go
@@ -4,10 +4,10 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
-	"path"
 	"strings"
 
 	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/paths"
 	"github.com/hashicorp/hcl2/gohcl"
 	"github.com/hashicorp/hcl2/hclparse"
 )
@@ -118,7 +118,7 @@ func (d *DownloadFile) validate() error {
 // exist or match.
 func (d *DownloadFile) Match(mod string) Mode {
 	for _, p := range d.Paths {
-		if hasMatch, err := path.Match(p.Pattern, mod); hasMatch && err == nil {
+		if paths.MatchesPattern(p.Pattern, mod) {
 			return p.Mode
 		}
 	}
@@ -130,7 +130,7 @@ func (d *DownloadFile) Match(mod string) Mode {
 // the top level downloadURL is returned.
 func (d *DownloadFile) URL(mod string) string {
 	for _, p := range d.Paths {
-		if hasMatch, err := path.Match(p.Pattern, mod); hasMatch && err == nil {
+		if paths.MatchesPattern(p.Pattern, mod) {
 			if p.DownloadURL != "" {
 				return p.DownloadURL
 			}
--- a/pkg/paths/path.go
+++ b/pkg/paths/path.go
@@ -2,6 +2,8 @@ package paths
 
 import (
 	"net/http"
+	"path"
+	"strings"
 
 	"github.com/gomods/athens/pkg/errors"
 	"github.com/gorilla/mux"
@@ -49,4 +51,29 @@ func GetAllParams(r *http.Request) (*AllPathParams, error) {
 	}
 
 	return &AllPathParams{Module: mod, Version: version}, nil
+}
+
+// MatchesPattern reports whether the path prefix of target matches
+// pattern (as defined by path.Match)
+//
+// This tries to keep the same behavior as GOPRIVATE/GONOPROXY/GONOSUMDB,
+// and is adopted from:
+// https://github.com/golang/go/blob/a11644a26557ea436d456f005f39f4e01902bafe/src/cmd/go/internal/str/path.go#L58
+func MatchesPattern(pattern, target string) bool {
+	n := strings.Count(pattern, "/")
+	prefix := target
+	for i := 0; i < len(target); i++ {
+		if target[i] == '/' {
+			if n == 0 {
+				prefix = target[:i]
+				break
+			}
+			n--
+		}
+	}
+	if n > 0 {
+		return false
+	}
+	matched, _ := path.Match(pattern, prefix)
+	return matched
 }
--- a/cmd/proxy/actions/sumdb.go
+++ b/cmd/proxy/actions/sumdb.go
@@ -4,8 +4,9 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
-	"path"
 	"strings"
+
+	"github.com/gomods/athens/pkg/paths"
 )
 
 func sumdbPoxy(url *url.URL, nosumPatterns []string) http.Handler {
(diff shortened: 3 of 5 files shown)


pkg/paths/path.go
```diff
@@ -50,3 +52,28 @@ func GetAllParams(r *http.Request) (*AllPathParams, error) {
...
+func MatchGlobPattern(pattern, name string) (bool, error) {
+	const op errors.Op = "paths.MatchGlobPattern"
+	pCount := strings.Count(pattern, "/")
+	nCount := strings.Count(name, "/")
+
+	for nCount > pCount {
+		i := strings.LastIndex(name, "/")
+		if i == -1 {
+			break
+		}
```
marwan-at-work: Is this block ever possible to be true?

pkg/paths/path.go
```diff
@@ -50,3 +52,28 @@ func GetAllParams(r *http.Request) (*AllPathParams, error) {
...
+		name = name[:i]
+		nCount = strings.Count(name, "/")
```
marwan-at-work: why not keep the same implementation as `matchesPattern` above? That one (copied from the Go source code) has one call to strings.Count while this implementation has N+2 calls?

pkg/download/mode/mode.go
```diff
@@ -130,7 +130,7 @@ func (d *DownloadFile) Match(mod string) Mode {
 // the top level downloadURL is returned.
 func (d *DownloadFile) URL(mod string) string {
 	for _, p := range d.Paths {
-		if hasMatch, err := path.Match(p.Pattern, mod); hasMatch && err == nil {
+		if hasMatch, err := paths.MatchGlobPattern(p.Pattern, mod); hasMatch && err == nil {
```
marwan-at-work: We're only checking `if err == nil`. In other words, we're really not doing anything if `err != nil` so we might as well be explicit about ignoring it so we should just put `_` and check for `hasMatch`

pkg/paths/path.go
jorng: Good question. I actually wrote this several weeks ago; before the `matchesPattern` implementation (I just received permission for the postback).  It was written to be understandable, and not too much copy-paste from the Go source code.

After rebasing and seeing the new `matchesPattern` function, I benchmarked it. `MatchGlobPattern` runs ~4-7% faster than `matchesPattern`, so I left it with my new implementation.

```
name                        old time/op    new time/op    delta
PathMatch/MatchPattern/1-8     131ns ± 0%     125ns ± 0%  -4.58%
PathMatch/MatchPattern/2-8     148ns ± 0%     138ns ± 0%  -6.76%
PathMatch/MatchPattern/3-8     168ns ± 0%     160ns ± 0%  -4.76%
PathMatch/MatchPattern/4-8     200ns ± 0%     193ns ± 0%  -3.50%
PathMatch/MatchPattern/5-8     215ns ± 0%     207ns ± 0%  -3.72%
PathMatch/MatchPattern/6-8     236ns ± 0%     227ns ± 0%  -3.81%
PathMatch/MatchPattern/7-8     270ns ± 0%     249ns ± 0%  -7.78%
PathMatch/MatchPattern/8-8     290ns ± 0%     272ns ± 0%  -6.21%
PathMatch/MatchPattern/9-8     313ns ± 0%     292ns ± 0%  -6.71%
```

> The `/`+digit for each benchmark is the number of `/` in the pattern.

pkg/paths/path.go
jorng: As it stands, it shouldn't. I can remove it if needed. It's there as a backup, in case the code ever changes.

pkg/download/mode/mode.go
jorng: Agreed, I only wrote it this way to be a drop-in replacement, the `err == nil` was already there.

jorng: Fixed in d81c837

pkg/paths/path.go
marwan-at-work: Thanks for the stats. It definitely seems weird that more strings.Count calls is more performant.

I ran some benchmarks against the two functions with the following input

`MatchGlobPattern("github.com/one/two/*", "github.com/one/two/three/four/five/six/seven")`

and 

`matchesPattern("github.com/one/two/*", "github.com/one/two/three/four/five/six/seven")`

And the first one yielded `167 ns/op` while the second one yielded `115 ns/op` which makes a little more sense. 

I'm a bit curious as to how you got your stats, do you mind putting them in a gist? Just so I can see what the results are on my machine. 

Interestingly enough, `benchstat` returned this: 

```
name   old time/op    new time/op    delta
Match-8     167ns ± 0%     115ns ± 0%   ~     (p=1.000 n=1+1)
```

Where the `~` indicates that the difference is not significant enough :) 

That said, I think it's safe to keep the Go implementation because they have far more robust tests on their implementations and it's good to keep the code consistent.

jorng: Ahh, looks like I made a mistake on my benchmark. The number of slashes between the pattern and the target were always matching. 

I can switch out my function for `matchesPattern`.

jorng: So my original benchmark was testing increasing lengths of `pattern`, but it's much more likely that `pattern` will be short, and `target` long. So the `matchesPattern` implementation is much better, as it runs in constant time, even with increasing `target` length.

I've added a Benchmark so this can be monitored / compared in the future if needed.

jorng:
Should all be resolved now. Let me know if there's anything else @marwan-at-work.

PR #1422, opened 2019-10-17
Fix helm chart to the latest docker image tag

sergeyrolich:

Fix helm chart to the latest docker image tag

**How is the fix applied?**

Updated versions in Chart and values

Bump version for helm chart because release new version athens


arschles:
@sergeyrolich one more thing before we can merge, can you update the `version` key in the `Chart.yaml` to be `0.3.7`? Our CI checks to make sure that the `version` is updated every time the chart is updated.

PR #1370, opened 2019-09-04
Added check sshGitServer and gitConfig for creating volumeMounts

sah4ez:
Fixes #1367

PR #1423, opened 2019-10-17
[chart] fix deployment.yaml for enabled ssh git servers

### LINKED ISSUE #1367: Deployment to kubernets
In [this condition](https://github.com/gomods/athens/blob/master/charts/athens-proxy/templates/deployment.yaml#L146) missed the values for `sshGitServers` ([there](https://github.com/gomods/athens/blob/master/charts/athens-proxy/templates/deployment.yaml#L163))

```
{{- if or (eq .Values.storage.type "disk") .Values.upstreamProxy.enabled .Values.netrc.enabled }}
   volumeMounts:
{{- end }}
```
when I tried install to helm, I got an error:
```
$ helm install --namespace kube-public -f athens.yaml --set storage.mongo.url=mongodb://mongo.default.svc.cluster.local:27017 -n athens gomods/athens-proxy
Error: release athens failed: Deployment.apps "athens-athens-proxy" is invalid: [spec.template.spec.containers[0].ports[1].name: Invalid value: "ssh-git-servers-config": must be no more than 15 characters, spec.template.spec.containers[0].ports[1].containerPort: Required value, spec.template.spec.containers[0].ports[2].name: Invalid value: "ssh-git-servers-config": must be no more than 15 characters, spec.template.spec.containers[0].ports[2].containerPort: Required value, spec.template.spec.containers[0].ports[3].containerPort: Required value]
```
I disabled `upstreamProxy`, `netrc` and used `type: mongo` in storage.

sergeyrolich:

If storage type not equals "disk" also need check sshGitServers and gitconfig 

**How is the fix applied?**

Add check shGitServer and gitConfig for creating volumeMounts

Fixes #1367

### DIFF
--- a/charts/athens-proxy/Chart.yaml
+++ b/charts/athens-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: athens-proxy
-version: 0.3.5
+version: 0.3.6
 appVersion: 0.6.1
 description: The proxy server for Go modules
 icon: https://raw.githubusercontent.com/gomods/athens/master/docs/static/banner.png
--- a/charts/athens-proxy/templates/deployment.yaml
+++ b/charts/athens-proxy/templates/deployment.yaml
@@ -143,7 +143,7 @@ spec:
         {{- end }}
         ports:
         - containerPort: 3000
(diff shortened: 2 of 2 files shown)


sergeyrolich:
Duplicate  #1370, close

Issue #1334, opened 2019-08-11
Download module from  a specific subdirectory within the repository 


owenhaynes:
**Is your feature request related to a problem? Please describe.**
Be able to cache a module from a mono-repo which has a vanity URL in front of it.

E.G have a go module in some mixed language environment git repo.
/src/test.cpp
/src/b.cpp
/go/blah.go
/go/go.mod
/go/go.sum

make go.foo.io/foo  points to /go/

**Describe the solution you'd like**
Be able to configure Athens to know how to map a vanity URL to a repo sub folder, avoiding a call  to go.foo.io/foo to get the metadata.
 
**Describe alternatives you've considered**
1. Creating a custom mod proxy which does this then Athens caches the proxy. 
2. Publish to Athens, this means configuring a CI to create a publish job and would require a format to be defined.

**Additional context**
I had a look to see if I could hack in the solution using the current state but because Athens just calls go mod download we get blocked on the way go mod download works golang/go#33562.

All the other go module proxies I have found all seem to just call go mod download as well.

Go seems to be hiding all the module versioning code in internal packages, unless I have missed to look somewhere. This makes it difficult to reuse there pseudo versioning code and other bits to do with the coderepo, these could just copied but this seems bad. This makes it harder to make Athens not call go mod download and alternative 1.

Alternative 2 seems to much like JAVA and will just pollute the cache

arschles:
Hi @owenhaynes! discussions on aliasing and vanity URLs have been pretty popular recently, if you're interested.  Here are 3 for example: #1315, #1282, and the alias endpoint discussion in #1195.

For your case, just to be clear, does running `go get` or `go mod download` on the VCS subdirectory work for you? For example, does `GOPROXY=https://athens go get bitbucket.com/foo/foo/src/go` work? It looks like that won't solve the problem for you since you need a vanity URL, but I've never tried this myself with non Go code at the top level of the repository, so I want to see if you know before I go try it myself (in theory it should?)

Regarding the vanity URL - I agree with what you said about alternative 2. Allowing folks to publish directly to Athens would make it a point of origin for code and fundamentally push the project beyond being a proxy. I'd like to avoid that. There's room, however, to write code in Athens itself that fetches code and packages it into a valid zip file, without using `go mod download`.

owenhaynes:
It was the first thing I tried before seeing if a proxy would solve my
problem.

The metadata for go get does not support subdirectories golang/go#33562.
They don't seem to what to make any changes to the metadata fields.

[...]

Issue #1132, opened 2019-03-14
Add feature about replacing modules


sevennt:
**Is your feature request related to a problem? Please describe.**
Add feature about replacing modules, some modules like ``golang.org/x/tools`` are not easy to 
download because of CN bad network.

By replacing modules in athens, we can replace this modules' path with other reachable path (such as github.com), then we can ``go mod tidy`` smoothly. 

**Describe the solution you'd like**
User adds replacing rules to configuration toml, like [https://docs.gomods.io/configuration/filter/](https://docs.gomods.io/configuration/filter/).

**Describe alternatives you've considered**
none.

**Additional context**
none.

arschles:
@sevennt are you thinking of something like an alias?

sevennt:
@arschles yes

arschles:
Sorry for the long delay @sevennt. There's a similar discussion going on at https://github.com/gomods/athens/issues/1315. Do you have a use case for this?

One possible solution would be to use the [new download mode configuration file](https://docs.gomods.io/configuration/download/) (this was added since we talked last) and introduce a syntax like this:

```hcl
alias "golang.org/x/tools" {
    name "https://github.com/golang/tools"
}
```

Could this work for you?

daqingshu:
> Sorry for the long delay @sevennt. There's a similar discussion going on at #1315. Do you have a use case for this?
> 
> One possible solution would be to use the [new download mode configuration file](https://docs.gomods.io/configuration/download/) (this was added since we talked last) and introduce a syntax like this:
> 
> ```terraform
> alias "golang.org/x/tools" {
>     name "https://github.com/golang/tools"
> }
> ```
> 
> Could this work for you?

I think your idea is great, use alias  syntax is great

sevennt:
@arschles that works, thanks.

PR #1426, opened 2019-10-18
Feature mod aliases

### LINKED ISSUE #1315: could athens support mirror?
a simple way to replace in athens, not in go.mod
batch config in athens

Mirrors = [
"golang.org/x/(.*) => github.com/golang/$1",
"google.golang.org/appengine => github.com/golang/appengine",
"google.golang.org/genproto => github.com/google/go-genproto",
"google.golang.org/grpc => github.com/grpc/grpc-go",
"google.golang.org/api => github.com/googleapis/google-api-go-client",
"cloud.google.com/go => github.com/googleapis/google-cloud-go",
]

if impl..ed many project just simple write go.mod file

**Additional context**
NA

ferryvg:
We need set source repo url for module

**How is the fix applied?**
Added new block to `download.hcl` and method like `Mode.Match`

Fixes #1315, #1132, #1334


ferryvg:
Will be updated tomorrow, because not working for 2+ versions of module

ferryvg:
go.mod support `replace` directive
https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive

blueworrybear:
@ferryvg 
Hi, I'm looking forward to this **Alias** feature. Although, **Go** supports `replace`, it does not mean Alias is not useful.

For example, I cannot access some VCS repositories due to the firewall policy in my department. Although I can use either:

1. `go mod replace`
2. `git` `insteadof` configure
3. Other goproxy server (Unfortunately, sometimes it's not working)

It's more convenience to me if I could set some redirection via Athens. Especially I could help my colleagues to avoid the same painful environment setup when they want to use the same modules.

Therefore, please don't give up working on the feature 👍 

BTW, I would like to know what happened for 2+ version of module? As I know, `go mod` requests using `url/to/module/v2` to denote breaking update. However, it seems to me that alias those URL is not impossible. Could we just replace `v2` in `Alias` function?

letian0805:
This feature is very useful! Why not merge it?

PR #1428, opened 2019-10-21
Add GCS to helm chart

epk:
Opening in place of #1347 

cc @arschles


chriscoffee:
Hey @epk this is great! Thanks so much for taking this ticket 🙏

What do you think about quoting out `serviceAccount` to the `b64enc` function so users don't have to encode the value themselves, might make it a little easier for users. There's some docs on that here https://helm.sh/docs/chart_template_guide/#encoding?
https://github.com/gomods/athens/blob/c480801cab123aadf09c403b70e8c744d5f7ad2d/charts/athens-proxy/templates/deployment.yaml#L111

**Bonus task**:

And a little request that's more something that we've overlooked. If you get a chance would you be able to make the language we use for `serviceAccount` standard throughout the codebase. What I mean here is we have:
* JSON Key:
https://github.com/gomods/athens/blob/2792796172b63628718b0c57f60e53b206497f98/config.dev.toml#L299-L306
* ATHENS_STORAGE_GCP_JSON_KEY
* serviceAccount

I think serviceAccount or similar is probably the way to go, given GCP refers to it as Service Account rather than a JSON key (https://cloud.google.com/iam/docs/service-accounts). What do you think?

It would require a change https://github.com/gomods/athens/blob/a27ef13b711857ad4652afa9ab3f31f009c65d77/pkg/storage/gcp/gcp.go#L34 and https://github.com/gomods/athens/blob/a27ef13b711857ad4652afa9ab3f31f009c65d77/pkg/config/gcp.go#L7 and https://github.com/gomods/athens/blob/2792796172b63628718b0c57f60e53b206497f98/config.dev.toml#L306

epk:
Hey @chriscoffee done! 😄

epk:
@marwan-at-work Any comments on this?

epk:
Also, I _think_ changes in https://github.com/gomods/athens/pull/1428/commits/1d83810a2dd23b83c09f55c862160219b07e753e warrant a new patch release

charts/athens-proxy/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: athens-proxy
-version: 0.3.5
+version: 0.3.6
 appVersion: 0.6.1
```
epk:
```suggestion
appVersion: 0.7.1
```

If going ahead with a new patch release, is this the PR to bump this?

epk: Also needs a bump at https://github.com/gomods/athens/pull/1428/files#diff-71a5c45e593365379cc2572e15c7db38L5

epk:
> @epk oook, sorry for the delay. Regarding the bump to the `tag` in `values.yaml`, that's gonna be done in #1422. One more thing we need, though - can you change the `version` in `Chart.yaml` to `0.3.7`? Our CI requires that every time the chart is changed, the `version` needs to be increased.

Done

PR #1431, opened 2019-10-22
Replace GO_ENV in proxy Dockerfile with config-docker.toml

davidpfarrell:

Because the proxy Dockerfile adds `GO_ENV=production`, there is no way to override the setting via the config.

This prevents the user from being able to use the config to fully manage their installation.

It does occur to me that a hard-default to Production mode is what you were aiming for, but I think there's another option that is equally as useful (if not more-so).

**How is the fix applied?**

Removes the `ENV` statement from the proxy Docker file.

Adds a new, docker-specific config file, `config-docker.toml`, which defaults to `"production"` mode.

Relates to Ticket #1435

I purposefully decided on `-docker.toml` (vs say, `-prod.toml`) as I think a docker-specific config will prove to be useful in that we may decide on more docker-specific defaults over time (in fact I'm working on another PR that might apply).

*SIDE-NOTE:* You'll notice I added a `chmod` statement to the Dockerfile.  Working on this exposed the fact that the permission check of the config file is performed BEFORE the env replacement step, meaning that you can currently be in production mode with non-conforming permissions.   I'm working on another PR for that and will submit it shortly.

### DIFF
--- a/cmd/proxy/Dockerfile
+++ b/cmd/proxy/Dockerfile
@@ -17,14 +17,13 @@ FROM alpine
 ENV GO111MODULE=on
 
 COPY --from=builder /bin/athens-proxy /bin/athens-proxy
-COPY --from=builder /go/src/github.com/gomods/athens/config.dev.toml /config/config.toml
+COPY --from=builder /go/src/github.com/gomods/athens/config.docker.toml /config/config.toml
 COPY --from=builder /usr/local/go/bin/go /bin/go
 
 # Add tini, see https://github.com/gomods/athens/issues/1155 for details.
 RUN apk add --update bzr git mercurial openssh-client subversion procps fossil tini && \
-	mkdir -p /usr/local/go
-
-ENV GO_ENV=production
+	mkdir -p /usr/local/go && \
+	chmod 600 /config/config.toml
 
 EXPOSE 3000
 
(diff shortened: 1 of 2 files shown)


marwan-at-work:
Also, can you please open an issue and tag it here. Every PR should link to an issue so that we can track any pre and follow-up discussions. Thanks!

ferryvg:
You can pass `GO_ENV` with `-e` flag in `docker run` command:
`docker run -e GO_ENV=development gomods/athens:latest`
Or in docker-compose.yml, etc...

davidpfarrell:
> Also, can you please open an issue and tag it here. Every PR should link to an issue so that we can track any pre and follow-up discussions. Thanks!

Done !  Ticket #1435 created and referenced in the description.  Thanks for taking the time to reply !

davidpfarrell:
> You can pass `GO_ENV` with `-e` flag in `docker run` command:
> `docker run -e GO_ENV=development gomods/athens:latest`
> Or in docker-compose.yml, etc...

Indeed you can - My concern is that you *MUST* do so in order to engage dev mode, as having a hard-coded environment variable baked into the Dockerfile makes the configuration file property essentially useless.

Thanks for chiming in - I have created ticket #1435 to further discussion on matter - Please feel free to share your thoughts there.

davidpfarrell:
Closing this PR since #1443 (merged) solves in a better way.

PR #1436, opened 2019-10-23
Renamed redeploy annotations to fix duplicate key in yaml

damoon:

The chart creates a yaml that has duplicate keys. One key will be ignored by Kubernetes.

**How is the fix applied?**

Renamed the annotations to distinct names.

The yaml becomes valid.

### DIFF
--- a/charts/athens-proxy/Chart.yaml
+++ b/charts/athens-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: athens-proxy
-version: 0.3.8
+version: 0.3.9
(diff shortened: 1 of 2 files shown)


damoon:
the chart version is updated from 0.3.5 to 0.3.6.

arschles:
@damoon I'm super sorry about this - We have some other helm chart related PRs that got merged, so can you update the chart version _one more time_ to `0.3.9`? Also would you mind DMing me when you do on the gophers slack (at `arschles`) so I can merge it right away?

PR #1440, opened 2019-10-24
Use sed to set proxy docker image to production mode

### LINKED ISSUE #1435: Consider removing 'GO_ENV=production' from proxy Dockerfile

Because the proxy Dockerfile adds `GO_ENV=production`, there is no way to override the setting via the config.

This prevents the user from being able to use the config to fully manage their installation.

Since a permanent environment override (i.e baked into the Dockerfile) makes the related config file property essentially useless, we might consider avoiding them, leaving env overrides solely as a convenience to the user.

Alternate solutions could include:

* Remove the `GO_ENV` statement, leaving the default Go Environment settings as "development" as specified in `config-dev.toml`
 (See PR #1443)
* Create a docker-specific config file, i.e. `config-docker.toml` which contains settings deemed useful for the Dockerized use case.
 (See PR #1431)
* Add a step in the Dockerfile that changes the `GoEnv` property to `"production"` - A sed command should likely be able to pull it off easily.
 (See PR #1440)

**Additional context**

I do think there could be value in having opinionated settings for the Dockerized use case, but not sure if env overrides are better than configuration settings.

[edit: Adding links to PRs for each alternate solution mentioned - I think the `Let the default be "development"` option has the most traction per the various discussion threads]

davidpfarrell:

Because the proxy Dockerfile adds `GO_ENV=production`, there is no way to override the setting via the config.

This prevents the user from being able to use the config to fully manage their installation.

**How is the fix applied?**

Replaces the `ENV` directive within the proxy Dockerfile with a `sed` action to inject the `"production"` value into the config file.

Relates to #1435

I initially submitted PR #1431 which introduced a new proxy-docker-specific config file.  After some discussion, it was apparent that such a solution doesn't scale well.

I  mentioned in the referenced issue that another idea could be to use `sed` to set the config value to `"production"`

This PR is an example of what that would look like.

### DIFF
--- a/cmd/proxy/Dockerfile
+++ b/cmd/proxy/Dockerfile
@@ -22,9 +22,9 @@ COPY --from=builder /usr/local/go/bin/go /bin/go
 
 # Add tini, see https://github.com/gomods/athens/issues/1155 for details.
 RUN apk add --update bzr git mercurial openssh-client subversion procps fossil tini && \
-	mkdir -p /usr/local/go
-
-ENV GO_ENV=production
+	mkdir -p /usr/local/go && \
+	sed -i -r -e 's/(GoEnv\s*=\s*)"development"/\1"production"/' /config/config.toml && \
+	chmod 600 /config/config.toml
 
 EXPOSE 3000
 


cmd/proxy/Dockerfile
```diff
@@ -22,9 +22,9 @@ COPY --from=builder /usr/local/go/bin/go /bin/go
 
 # Add tini, see https://github.com/gomods/athens/issues/1155 for details.
 RUN apk add --update bzr git mercurial openssh-client subversion procps fossil tini && \
-	mkdir -p /usr/local/go
-
-ENV GO_ENV=production
+	mkdir -p /usr/local/go && \
+	sed -i -r -e 's/(GoEnv\s*=\s*)"development"/\1"production"/' /config/config.toml && \
```
marwan-at-work: That looks cray. Do you mind explaining what it does (i'm not a bash expert). 

But in any case, I think we can just remove it and have development be be the new default coming from the config. It's a soft breaking change but I can't imagine it actually breaking people. We're not 1.0 so that to me seems like an okay change.

davidpfarrell: Hey! Sures, a quick rundown:
* `-i` : Edit the file in place (don't create a backup nor send to stdou)
* `-r` : Enable extended regex features, in this case back-referencing (see `\1`)
* `-e` : Prelude to an expression, i.e the proceeding `"/s..."`
* `s/find/replace/`: Format for wrapping a substitution expression.
* `(GoEnv\s*=\s*)"development"` : Look for the `GoEnv = "development"` property assignment.
* `(Go...)`: Parens tell the script to remember what it found, as we might use it later in the replace.
* `\s*=\s*` : Allow whitespace to be optional before and after the `=` i.e. could match `GoEnv="development"` as well as `GoEnv = "development"`
* `\1` : Take the text we found in the earlier `()` and stick it here
* `/config.config.toml` : Target file to look in (course)
* `&& \` : After executing this replacement, continue to execute statements on the following line(s).

re: `I think we can just remove it and have development be be the new default`

I'm happy to submit another PR for that if you're sure its route you want to take - This PR (And the other) were just in case you were hold-fast on having production mode enable in the docker hub images.

-DF

davidpfarrell: If simply removing the statement is deemed viable, then I have created PR #1443 to accomplish that.

BTW: Thank you for reviewing my PRs !

davidpfarrell:
Closing this PR since #1443 (merged) solves in a better way.

PR #1430, opened 2019-10-21
Add GOLANG_VERSION arg support to proxy-docker make target + Dockerfile

davidpfarrell:
**What is the problem I am trying to address?**

I want to be able to utilize the existing makefile functionality for setting the golang version when building the proxy via `make docker`.

**How is the fix applied?**

Updated makefile and proxy Dockerfile.

**Mention the issue number it fixes or add the details of the changes if it doesn't have a specific issue.**

Fixes #1442

I modified the `proxy-docker` make target to pass  `GOLANG_VERSION` in the docker build command.

I also modified `cmd/proxy/Dockerfile` to utilize the argument, when present, defaulting to the existing version of `1.13`.

I verified the functionality by issuing several `make docker GOLANG_VERSION=1.X` commands, then logging into the resulting image and performing `go version` to confirm the expected version was installed.

PR #1441, opened 2019-10-24
Invoke envOverrides before checkFilePerms

### LINKED ISSUE #1438: FilePermissions not checked when config=dev + env=prod
**Describe the bug**
re: `pkg/config/config.go:ParseConfigFile()` :

The "production+filePermissions" check happens before the environment overrides are applied.
This can lead to the case where the proxy is running in production mode but that file permissions on `configFile`  and  `config.FilterFile` are out of compliance.

NOTE: This is in-fact the current case for the proxy docker image which sets GO_ENV to production but has non-complying permissions on the injected `config.toml` file.

**Error Message**
n/a - This issue is related to lack of error message.

**To Reproduce**
Start with a config file that has non-complying access rules
```console
$ chmod ugo+rw config.toml
```

Set `GoEnv = "production"` within the config file

Launch the application.

Notice permission check error:
```
config.toml should have at most rwx,-, - (bit mask 077) as permission
```

Now set `GoEnv = "development"` within the config file

Then set environment variable to production:
```console
$ export GO_ENV=production
```

Launch the application.

Notice lack of error message.

**Expected behavior**
When the GoEnv is deemed to be production, either through the config file, or the env variable, file permissions should be confirmed.

 - OS: MacOS 10.13.6 (High Sierra)
 - Go version : 1.13.3
 - Proxy version : 0.7.0

**Additional context**
The easy fix is just to move the "production+filePermissions" check after the env override step.   

However, you might consider if there's some form of security risk in being able to disable file permission checks (and maybe other features?) by setting an environment variable. (there may not be, I don't have a thorough enough understanding of the system, I'm just calling it out)

I'm happy to put the PR together for the change once we decide on a fix, if any.

Thank you for your time,

-DF

davidpfarrell:

re: `pkg/config/config.go:ParseConfigFile()` :

The "production+filePermissions" check happens before the environment overrides are applied.
This can lead to the case where the proxy is running in production mode but that file permissions on `configFile`  and  `config.FilterFile` are out of compliance.

NOTE: This is in-fact the current case for the proxy docker image which sets GO_ENV to production but has non-complying permissions on the injected `config.toml` file.

**How is the fix applied?**

Moves the `checkFilePerms` call to happen after the call to `envOverride`

Also adds a step in the proxy docker Dockerfile to set the permissions on the config file so that they will pass the permissions check.

Fixes #1438

#### re:  Other Dockerfiles Affected By The change
Looking through the code base, the proxy docker Dockerfile appears to be the only one that explicitly sets `GO_ENV` to production, so should be the only one affected by this change.

#### re: Testing
I used the `test-unit-docker` test to confirm that the changes (likely) don't break anything.

NOTE: The `test-unit-docker` tests don't pass (unrelated to my change) with with default docker version within the makefile (`v1.12`).  However, the tests pass fine when explicitly passing a docker version of `v.1.13`.  I actually have a change in a PR #1430 that updates the version within the Makefile.


davidpfarrell:
@marwan-at-work thanks for the review.

Please Note: If PR #1443 gets approved, then the Dockerfile step to correct the file permissions on the injected config file will not be necessary.

If that PR merges before this one, then I can modify this PR to remove the permissions fix.

If this PR merges before that one, I can simply modify that PR to remove the permissions fix.

-DF

davidpfarrell:
Greetings ! Just checking in to see if there's anything I can do help this PR along?  Please don't hesitate to let me know if you have any questions or suggestions.  Thanks!

marwan-at-work:
just waiting on one more review :) 
cc: @arschles

davidpfarrell:
Per my previous comment, I'm going to modify this PR to remove the `"Dockerfile step to correct the file permissions on the injected config file`" -- Will push update shortly.

davidpfarrell:
OK I've reverted my changes to the Docker since we no-longer need to change the config file permissions, as the default mode is now Development (instead of Production), per #1443.

Now the PR consists solely of moving the `envOverride` step to happen before the `checkFilePerms` step.

PR #1444, opened 2019-10-26
Add chinese translation of install section

wangkirin:
**How is the fix applied?**

Add chinese translation of install section


wangkirin:
> @wangkirin this is amazing. Thank you so much for adding these documentation 🎉 🚀 :shipit:

thank you for your reply 
I will try to keep contributing to Athens  : )

PR #1445, opened 2019-10-29
SystemD installation

### LINKED ISSUE #1354: Is there interest in SystemD installation?
It's generally quite easy to install Go services on Linux systems that support [SystemD](https://en.wikipedia.org/wiki/Systemd). For such servers, Docker is an unnecessary dependency and unwanted overhead. For example, I already run [Caddy](https://caddyserver.com/) this way in Ubuntu.

Is there much interest in running Athens this way? If so, I can help get it started.

rickb777:
This provides a SystemD installation script and instructions, targetted at Linux administrators.

See issue #1354 

Editor: I'm adding the following line to close out #1354 when this is merged:_
Closes #1354


rickb777:
`go.sum` is changed by the Go tools; I never edit it manually. Presumably, that's why it changed in this case and the changes are immaterial (unlike any changes to `go.mod`).

arschles:
@rickb777 I should have asked the question differently - what `go` commands did you run in your branch that changed the `go.sum` file? I ask because there are no changes to any go code in here.

arschles:
On hold until we fix https://github.com/gomods/athens/issues/1478. The build is failing but it isn't this PR's fault.

PR #1454, opened 2019-11-08
Adding a section on updating the helm chart

arschles:

A lot of folks including me forget to update the Helm chart version when they make changes to the chart. It's annoying to see CI failures when you did everything else right!

**How is the fix applied?**

I added some documentation explaining why this happens and how you can fix it.

N/A


CONTRIBUTING.md
```diff
@@ -30,6 +30,18 @@ you must run `make setup-dev-env` first, otherwise you will see errors like the
 error connecting to storage (no reachable servers)
 ```
 
+## Helm Chart
+
+This repository comes with a [Helm](https://helm.sh) [chart](https://github.com/gomods/athens/tree/master/charts/athens-proxy) to make it easier for anyone to deploy their own instance of Athens to their Kubernetes cluster.
+
+Our CI/CD system will look to ensure that the `version` field in [Chart.yaml](https://github.com/gomods/athens/blob/master/charts/athens-proxy/Chart.yaml) is updated if any part of the chart is changed. We do that to make sure that there is a new version for every change to the chart.
+
+_If you're planning to submit a pull request (PR) that updates any part of the Helm chart, please make sure you _increase_ the version number in that field._
```
geototti21: 💡 Does it worth to mention that it follows semver? so change the version accordingly

robjloranger: I think that's a good idea

arschles: agreed! just added

arschles:
@geototti21 @robjloranger I just added a mention of semver to the new docs. @robjloranger if both of you are ok with the change, can you merge?

PR #1455, opened 2019-11-09
Install git-lfs on proxy container

### LINKED ISSUE #1448: Git Large File Storage (LFS) support
Yes, Athens generates an invalid checksum for go packages that have files stored in [git-lfs](https://git-lfs.github.com) (as it's not downloading them and instead hashing the text file).

Would be great if we could install `git-lfs` in the docker container.

Well we could stop using git-lfs in go packages. Not really a good alternative though.

**Additional context**
This happened in a private repository that is cloned by Athens directly.

bahlo:

For repositories that store files in git-lfs, the checksum will
differ for clients with and without git-lfs. Clients without git-lfs
will use the [git-lfs pointer](https://github.com/git-lfs/git-lfs/blob/master/docs/spec.md#the-pointer) for calculation, clients with will use
the real files.

**How is the fix applied?**

git-lfs is installed with the other utilities in the `cmd/proxy/Dockerfile`.

Fixes #1448

### DIFF
--- a/cmd/proxy/Dockerfile
+++ b/cmd/proxy/Dockerfile
@@ -25,7 +25,7 @@ COPY --from=builder /go/src/github.com/gomods/athens/config.dev.toml /config/con
 COPY --from=builder /usr/local/go/bin/go /bin/go
 
 # Add tini, see https://github.com/gomods/athens/issues/1155 for details.
-RUN apk add --update bzr git mercurial openssh-client subversion procps fossil tini && \
+RUN apk add --update bzr git git-lfs mercurial openssh-client subversion procps fossil tini && \
 	mkdir -p /usr/local/go
 
 EXPOSE 3000


bahlo:
Sure, I'll check out how to test this 👍 

Not sure if this is the right place to talk about it, but I guess I'll need to add a module with some `git-lfsed` files to [github.com/athens-artifacts/happy-path](https://github.com/athens-artifacts/happy-path)? That module should probably also be on the `athens-artifacts` org, right?

arschles:
>Sure, I'll check out how to test this 👍

Thanks!

>Not sure if this is the right place to talk about it

Here is fine, or you can open a new issue. Your choice

>I guess I'll need to add a module with some git-lfsed files to github.com/athens-artifacts/happy-path? That module should probably also be on the athens-artifacts org, right?

Yea, git-lfsed files would fit well in that repository. I would create a new module in there so that we can isolate the tests a bit more

Issue #1453, opened 2019-11-08
Catalog request fails  for minio client with Dell EMC ECS S3 as storage


jabbrwcky:
**Describe the bug**
We are currently deploying athens with a Dell ECS system as S3 storage using the minio client.

Catalog requests fail with an array index out of bounds exception because ECS returns the key as URL-Encoded value:

```
&minio.ObjectInfo{ETag:"\"d5fa08a78739671cc99b8c9aa6dbeae4\"", Key:"cloud.google.com%2Fgo%2Fv0.26.0%2Fv0.26.0.info", LastModified:time.Time{wall:0xf053700, ext:63704404906, loc:(*time.Location)(nil)}, Size:51, ContentType:"", Expires:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}, Metadata:http.Header(nil), Owner:struct { DisplayName string "json:\"name\""; ID string "json:\"id\"" }{DisplayName:"", ID:""}, StorageClass:"STANDARD", Err:error(nil)}
```

A test with mino shows that the key is returned decoded already here:

```
&minio.ObjectInfo{ETag:"\"3614268b5afb93e21ed8f08ae273ca9f-1\"", Key:"github.com/minio/hdfs/v3/v3.0.0/v3.0.0.info", LastModified:time.Time{wall:0x2eca2640, ext:63708811839, loc:(*time.Location)(nil)}, Size:50, ContentType:"", Expires:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}, Metadata:http.Header(nil), Owner:struct { DisplayName string "json:\"name\""; ID string "json:\"id\"" }{DisplayName:"", ID:""}, StorageClass:"STANDARD", Err:error(nil)}
```

**Error Message**

```console
2019-11-08 12:48:29.614332 I | http: panic serving [::1]:53575: runtime error: index out of range [-1]
goroutine 130 [running]:
net/http.(*conn).serve.func1(0xc000678000)
        /usr/local/Cellar/go/1.13.4/libexec/src/net/http/server.go:1767 +0x139
panic(0x2065380, 0xc000536ba0)
        /usr/local/Cellar/go/1.13.4/libexec/src/runtime/panic.go:679 +0x1b2
github.com/gomods/athens/pkg/storage/minio.parseMinioKey(0xc000825ad0, 0x213a048, 0x5, 0x2f19a01, 0xed5141434, 0x0, 0x25)
[...]

PR #1456, opened 2019-11-11
Enable extended AWS S3 client configuration

jabbrwcky:

This change makes the AWS S3 client usable with other endpoints (e.g. minio or as in my installation a Dell EMC ECS system offering S3 API) by enabling to configure:

* Overriding the endpoint for S3
* Enable use of S3 path style
* Enable use of static credentials instead of credentials provider

**How is the fix applied?**

Changes the setup of the AWS client with additional optional settings. Default behaviour is unchanged.

This change make the AWS client usable with non-AWS hosts. It is also one possible solution for #1453


jabbrwcky:
Close in favor of #1457

PR #1458, opened 2019-11-11
WIP: Add support for IAM proxies for S3

### LINKED ISSUE #1452: Support for KIAM and kube2iam for S3 storage
In our Kubernetes environment, we don't put our AWS creds on our pods, but rather, we use KIAM and proxy all our AWS requests through it where it then attaches the correct creds based on the IAM role we annotate the pods with. I would think a good feature would be to specify an environment variable (something like `USE_KUBE_IAM_ROLE`) that when set to true bypasses the AWS cred provider initialization in s3.go

1) specify an environment variable (something like `USE_KUBE_IAM_ROLE`)
2) When the above env variable is set to true, bypass the AWS cred provider initialization in s3.go and go straight to creating the AWS Session
3) Add a field in the helm template to specify a pod annotation for the IAM Role and the new environment variable.

**Additional context**
I have a fork with this change but I was having issues with github creds locally, I can post it when I have resolved these.

aaqel-s:
Adding support for AWS IAM proxies in Kubernetes

Currently, S3 support in S3s only supports AWS configuration using Hardcoded credentials. Our application of K8s uses KIAM avoid using hardcoded creds for AWS by having AWS requests proxied through a sidecar that injects credentials based on the IAM role of the pod.

**How is the fix applied?**

Added an environment variable to enable support for IAM proxies and then added an if statement in s3.go to bypass the AWS cred provider (which is not needed when using a proxy.

Added two flags to Helm to configure IAM support and the particular role a user wants the proxy attached to.

Verified with `make docker` and running in our k8s cluster.

Fixes #1452


arschles:
@aaqel-s If you're still interested in this PR, let me know and I can give it a review. It looks really good!

aaqel-s:
Will, do! There's some things I need to take care of on my end, and once I do, I'll reopen it.

arschles:
Awesome! We'll look out for it when you reopen it 😄 💚

Issue #1460, opened 2019-11-13
Add some GitHub Actions


arschles:
**Is your feature request related to a problem? Please describe.**

Not related to a problem

**Describe the solution you'd like**

We opened #808 to discuss how we might use GitHub actions. At the time, this repo didn't have actions enabled. Now that it does, we should try them out for some of the things in that thread.

**Describe alternatives you've considered**

We could do nothing!

**Additional context**

teaguecole:
I will start working on `We could publish a tweet to the @gomodsio Twitter account when we publish a new GitHub release`.  Not sure if this is the best way to manage/track which of the ideas in #808 are being worked on.

arschles:
@teaguecole awesome. thanks!

arschles:
>Not sure if this is the best way to manage/track which of the ideas in #808 are being worked on.

@teaguecole I will rename this issue specific to the twitter action, so we can track just that here. We can open issues for other specific actions as we go, then, and track things more easily.

keremgocen:
Anyone is working on this? I’d like to have a go of not or if anyone needs help.

teaguecole:
@keremgocen I was going to wok on doing the twitter thing but ive been busy with work. If you get to it, go for it.

jtn7:
I have the action ready for this. I just have some questions before I make a pull request.

**Q:** Do you need a tweet for pre-releases or just releases? 

**Q:** What should the message format be? `v0.9.0 is out` `v0.9.0 pre-release is out`

**Q:** The workflow has to be in the branch where you're creating the release from so I'm guessing this will be a pull request into `main`?

PR #1462, opened 2019-11-13
Adding YAML for a GH Action

arschles:

None

**How is the fix applied?**

I added `.github/workflows/go.yml` to set up a basic GitHub action

Ref #1460


arschles:
Closing in favor of #1463, that seems to have triggered the action, not this one 🤷‍♂

PR #1463, opened 2019-11-13
Adding YAML for a GH Action (try 2)

arschles:

None

**How is the fix applied?**

I added `.github/workflows/go.yml` to set up a basic GitHub action

Ref #1460


marwan-at-work:
@arschles actions are down lol https://www.githubstatus.com/

arschles:
@marwan-at-work must be why all my previous PRs weren't running 😆 

I just updated the branch and looks like the action just passed. Yesterday GH opened up actions to everyone, do you think we should put this on hold until they stabilize?

marwan-at-work:
Probably, but also see if it provides any value that we are currently missing from our current CIs

arschles:
@marwan-at-work there are a few things that overall make GH actions a nicer experience for me. Most of the below points are subjective though, so I hope this doesn't sound too much like a sales pitch. My biggest overall argument for GH actions is that they're right next to the PR so you don't need to authenticate with another service or deal with another interface.

- You can see the build logs, restart a build, etc... [inside of the PR](https://github.com/gomods/athens/pull/1463/checks?check_run_id=320373296) instead of going to drone/circle/etc... and logging in
- They're a bit more modular than what I've seen in drone or circle. We can have N workflows in different yaml files, which run in parallel. We can also split our steps up into separate docker containers if they end up getting big
- We can choose to run docker commands natively from scripts or run commands inside a container
- We can [cache](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows) our dependencies. I know we could do it with circle, but not sure about drone?
- Scripts are already authenticated to GH and we can respond to tons of events
- We can do testing on windows VMs if we wanted to later ([docs](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources))

Let me know what you think. I'm totes cool if we stick with drone/circle/appveyor for a while or forever if people would rather

marwan-at-work:
@arschles that's pretty cool, a couple things: 

- I couldn't figure out how to restart a build from the link you sent me.
- Can we have background containers?

arschles:
@marwan-at-work I have pressed a restart button before, I know it, but I can't find it at the moment. Maybe it was taken out. Anyway, that might be a problem for us if it's not there.

Yes, you can run background containers with `docker run -d ...`

arschles:
Ah I see it on a different repo. There is a button called "re-run checks":

Screen Shot 2019-11-26 at 10 21 17 AM

I'm wondering if it has to do with repo permissions. I'll need to investigate this more

arschles:
@marwan-at-work I checked some other GH actions and my guess is that you can only restart checks if they failed. See [here](https://github.com/arschles/athens-azure/commit/c9571b89c1f3e074b30541f2c038f57550248a1c/checks?check_suite_id=332407104) for a failure example

DrPsychick:
closing as we have GitHub actions already merged: https://github.com/gomods/athens/tree/main/.github/workflows

PR #1466, opened 2019-11-20
Read basic auth credentials from a secret.

### LINKED ISSUE #1307: Health endpoint without basic authentication
We are currently running Athens with basic auth as an ECS Service on AWS. In addition there is an ALB which does a periodic health check on the root path. Since health checks don't support basic authentication, it gets a 401 response. In order to bypass this issue, we had to add the 401 status code to the allowed status codes but it is not ideal.

It would be great if there is an arbitary endpoint, that could return a 200 status code that is not affected by the enabled basic authentication.

bmcstdio:

Currently, the Helm chart requires the HTTP basic authentication credentials (if any) to be defined  in the `values.yaml` file or via the CLI. Both of these approaches are cumbersome when working with [Flux](https://fluxcd.io/) and the [Helm operator](https://github.com/fluxcd/helm-operator), as that would require these credentials to be checked (as plain text) into source control. This PR updates the Helm chart in order to make it possible to reference a secret instead. This secret can then be managed using [Sealed Secrets](https://github.com/bitnami-labs/sealed-secrets) or bby some other process. 

**How is the fix applied?**

* Replace `basicAuth.username` and `basicAuth.password` with `basicAuth.secretName`, `basicAuth.usernameSecretKey` and `basicAuth.passwordSecretKey`.
* Remove the need for authentication from `/healthz` and `/readyz` and adapt the liveness and readiness probes accordingly.

Closes #1316.

---
_Note from @arschles :
Fixes #1307
Since #1316 also fixes #1307_

### DIFF
--- a/charts/athens-proxy/templates/deployment.yaml
+++ b/charts/athens-proxy/templates/deployment.yaml
@@ -54,20 +54,10 @@ spec:
           periodSeconds: {{ .Values.livenessProbe.periodSeconds }}
           successThreshold: {{ .Values.livenessProbe.successThreshold }}
           timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
-            {{- if .Values.basicAuth.enabled }}
-            httpHeaders:
-              - name: Authorization
-                value: "Basic {{ (printf "%s:%s" (default "gomod" .Values.basicAuth.user) (default "gomod" .Values.basicAuth.password) | b64enc) }}"
-            {{- end }}
         readinessProbe:
           httpGet:
             path: "{{ template "readinessPath" . }}"
             port: 3000
-            {{- if .Values.basicAuth.enabled }}
-            httpHeaders:
-              - name: Authorization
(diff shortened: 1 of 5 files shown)


charts/athens-proxy/templates/deployment.yaml
```diff
@@ -54,20 +54,10 @@ spec:
...
           timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
-            {{- if .Values.basicAuth.enabled }}
-            httpHeaders:
-              - name: Authorization
-                value: "Basic {{ (printf "%s:%s" (default "gomod" .Values.basicAuth.user) (default "gomod" .Values.basicAuth.password) | b64enc) }}"
-            {{- end }}
         readinessProbe:
           httpGet:
-            path: "{{ template "readinessPath" . }}"
+            path: "/readyz"
```
arschles: @bmcstdio would you be ok with leaving this as-is with the `{{template ...}}`?

bmcstdio: @arschles sure thing, I've reverted that.

bmcstdio:
> Also, do you need any of the changes you made to the go.sum in this PR?

@archles I don't need them, no :) I've reverted said changes.

PR #1467, opened 2019-11-21
storage/s3: support custom endpoint url

ernado:
This is required to support hosted S3-compatible storages (e.g. OpenStack).

I've added `Endpoint` config field with `AWS_ENDPOINT` env and started passing it (if not blank) to `awsConfig.Endpoint`.

From aws-sdk-go docs:
```go
// An optional endpoint URL (hostname only or fully qualified URI)
// that overrides the default generated endpoint for a client. Set this
// to `""` to use the default generated endpoint.
//
// Note: You must still provide a `Region` value when specifying an
// endpoint for a client.
Endpoint *string
```
https://github.com/aws/aws-sdk-go/blob/98babde4e50a39648e6618b79d256074f451754d/aws/config.go#L44-L50

### DIFF
--- a/config.dev.toml
+++ b/config.dev.toml
@@ -356,7 +356,6 @@ SingleFlightType = "memory"
         Insecure = false
 
     [Storage.S3]
-
         ### The authentication model is as below for S3 in the following order
         ### If AWS_CREDENTIALS_ENDPOINT is specified and it returns valid results, then it is used
         ### If config variables are specified and they are valid, then they return valid results, then it is used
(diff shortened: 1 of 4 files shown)


ernado:
@arschles seems like yes, we can use minio for that purpose, but it is slightly counter-intuitive for users that are not familiar with minio. I've created this PR after checking the `pkg/storage/minio` package and it was still not 100% clear that implementation is identical to `aws/aws-sdk-go`.

It was not straightforward and will likely require comments like "use minio client for S3-compatible storage" everywhere in our infra, so I want to keep things as clear and possible.

I'm ok with that solution, but I think that an ability to optionally set the `aws.Config.Endpoint` field will be useful. Are there any cons?

ernado:
Gentle ping. 

Actually after re-checking the docs I've [found](https://docs.gomods.io/configuration/storage/#minio) that minio is mentioned as an interface for any S3-compatible storage. Sorry, I've missed that.

It is fine, but I still want to have a choice to use a standard S3 client.

ernado:
Thanks, @arschles! I've described new config option both in documentation and sample config file, please take a look. (I've reused wording from `aws-sdk-go`)

PR #1477, opened 2019-12-04
Bumping chart version in the Helm chart

arschles:

After #1428, the chart version needed to be updated.

@epk had increased the chart version enough times and we both
wanted to get that PR merged. Thank you for your patience @epk!

**How is the fix applied?**

I've increased the `version` field in `Chart.yaml`

Fixes #


arschles:
On hold until we fix https://github.com/gomods/athens/issues/1478. The build is failing but it isn't this PR's fault.

Issue #1478, opened 2019-12-04
Fix build flakes


arschles:
**Describe the bug**

We have had consistent build flakes in drone.io such as https://cloud.drone.io/gomods/athens/1283/1/5. Errors manifest themselves like this:

- [build step 5](https://cloud.drone.io/gomods/athens/1281/1/5) succeeds, but Athens can't connect to the Google Cloud Storage bucket because it doesn't have the proper credentials (I think, based on the error message). This step should fail
- [build step 6](https://cloud.drone.io/gomods/athens/1281/1/6) fails because the Athens server isn't actually running 😦 

**Error Message**

```
error getting storage configuration (could not create new storage client: dialing: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.)
```

**To Reproduce**

This is a flake, so not consistently reproducible, but https://cloud.drone.io/gomods/athens/1281/1/5 is a great example.

**Expected behavior**

Builds that should fail always fail, and builds that should pass always pass.

**Environment (please complete the following information):**

N/A - this is for CI/CD builds

**Additional context**
None

marwan-at-work:
@arschles 

The error you see in the logs I believe are coming from this line: https://github.com/golang/oauth2/blob/master/google/default.go#L117

But this line is only executed if there were no credentials provided. And when no credentials are provided, we attempt to go for the default ones. 

The athens image we're running takes these params: 

```
    ATHENS_STORAGE_GCP_JSON_KEY:
      from_secret: GCS_SERVICE_ACCOUNT
```

So it's one of the following possibilities: 

1. The secret got updated to an empty string in our ci settings 
2. Drone has a bug where it's not passing this correctly 
3. Athens has a bug where it's not picking up the environment variable correctly 


I almost thought 3 was not correct, but I think it is. I had an old version of the `canary` image, and it worked as expected, but when I updated canary, it broke. See here: 

```
MacBook-Pro:gcp marwansulaiman$ docker run --rm -e 'ATHENS_GOGET_WORKERS=5' -e 'ATHENS_PROTOCOL_WORKERS=10' -e 'ATHENS_STORAGE_TYPE=gcp' -e 'ATHENS_STORAGE_GCP_BUCKET=athens_ci_build' -e 'ATHENS_STORAGE_GCP_JSON_KEY=asdfasdf' gomods/athens:canary
[...]

PR #1479, opened 2019-12-04
Revert "Add GCS to helm chart (#1428)"

marwan-at-work:
This quickly reverts the last commit that caused a breaking change and our CI to fail

However, a nicer solution would be to only revert the breaking part since that PR had nice documentation as well as some Helm additions. 

That said, it's safer to revert this change then add the previous PR but only the parts that were non-breakable. 

If someone wants to open another PR that only fixes the breaking changes, I'm okay with that since we have tests that cover us (well only after we merged, but they covered us still) 

Fixes: https://github.com/gomods/athens/issues/1478

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -153,7 +153,7 @@ services:
   ports:
     - 6379
 - name: athens-proxy
-  image: gomods/athens:canary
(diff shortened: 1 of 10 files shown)


marwan-at-work:
@arschles im not familiar with Helm, but CircleCI asked me to bump a version so I bumped the minor one.

PR #1480, opened 2019-12-04
Fixing the build failures from a missing GCS Key

arschles:

When I merged https://github.com/gomods/athens/pull/1428,
I accidentally introduced a build failure where Athens couldn't find the GCS key,
and wouldn't start up. It silently failed though, because we start Athens up in
the background

**How is the fix applied?**

I have added backward compatibility to Athens, so that it falls back to the old
`ATHENS_STORAGE_GCP_SERVICE_ACCOUNT` env var if the new `ATHENS_STORAGE_GCP_SERVICE_ACCOUNT`
isn't found

Closes #1479
Fixes https://github.com/gomods/athens/issues/1478

In a follow up to this PR, the `.drone.yml` needs to be updated so that the `athens-proxy` step uses the `gomods/athens:canary` image

### DIFF
--- a/config.dev.toml
+++ b/config.dev.toml
@@ -307,8 +307,22 @@ SingleFlightType = "memory"
         # running Athens inside GCP, you will most
         # likely not need this as GCP figures out
         # internal authentication between products for you.
-        # Env override: ATHENS_STORAGE_GCP_JSON_KEY
+        # Env override: ATHENS_STORAGE_GCP_SERVICE_ACCOUNT
         ServiceAccount = ""
+
+        # JSONKey is a base64 encoded service account 
+        # key that allows Athens to be run outside of GCP
+        # but still be able to access GCS. If you are
+        # running Athens inside GCP, you will most
+        # likely not need this as GCP figures out
+        # internal authentication between products for you.
+        # 
+        # NOTE: This config value is deprecated in favor of
+        # ServiceAccount above. Athens will check for it,
+        # but please do not rely on it being available forever.
+        #
+        # Env override: ATHENS_STORAGE_GCP_JSON_KEY
+        JSONKey = ""
 
     [Storage.Minio]
         # Endpoint for Minio storage
--- a/config.dev.toml
+++ b/config.dev.toml
@@ -301,15 +301,6 @@ SingleFlightType = "memory"
         # Env override: ATHENS_STORAGE_GCP_BUCKET
         Bucket = "MY_GCP_BUCKET"
 
-        # ServiceAccount is a base64 encoded service account
-        # key that allows Athens to be run outside of GCP
-        # but still be able to access GCS. If you are
-        # running Athens inside GCP, you will most
-        # likely not need this as GCP figures out
-        # internal authentication between products for you.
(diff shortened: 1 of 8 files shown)


pkg/storage/gcp/gcp_test.go
```diff
@@ -51,6 +51,9 @@ func getStorage(t testing.TB) *Storage {
 	bucketName := randomBucketName(os.Getenv("DRONE_PULL_REQUEST"))
 	cfg := getTestConfig(bucketName)
 	if cfg == nil {
+		// Don't fail if there's no test config, so that these tests don't
+		// fail when you run them locally
+		t.Log("No GCS Config found")
```
marwan-at-work: the `t.Log` won't show up in stdout (or stderr) unless you pass the `-v` flag. If you pass the `-v` flag, then you'll also see that the test got skipped like so: `SKIP TestXXX` -- so this seems unnecessary unless I'm missing something

config.dev.toml
```diff
@@ -307,9 +307,23 @@ SingleFlightType = "memory"
...
+        # running Athens inside GCP, you will most
+        # likely not need this as GCP figures out
+        # internal authentication between products for you.
+        # 
+        # NOTE: This config value is deprecated in favor of
+        # ServiceAccount above. Athens will check for it,
+        # but please do not rely on it being available forever.
+        #
+        # Env override: ATHENS_STORAGE_GCP_JSON_KEY
+        JSONKey = ""
```
marwan-at-work: any reason why introduce a new name/deprecate the old one and force people to migrate? A Service Account is technically a JSON Key so it's not too ambiguous IMO

arschles:
@marwan-at-work what do you think about this PR instead of https://github.com/gomods/athens/pull/1479/files?

I tried to retain the new stuff that @epk added in #1428 and keep backward compatibility with the old JSONKey stuff. I might have missed somewhere...

marwan-at-work:
@arschles I'd be more in favor of not forcing people to migrate. We can still keep the GCP addition to Helm but change the name back to JSONKey. Since this change is from 1 commit ago, I can't imagine breaking anyone (maybe only @epk ? unless he hasn't started using it yet) 

My opinion is that any time we tell users they need to change stuff when something is working well, it takes away from the happy user experience :)

arschles:
@epk can you let us know if you're already using the new `ATHENS_STORAGE_GCP_SERVICE_ACCOUNT` env var? If not, we'd like to change it back to `ATHENS_STORAGE_GCP_JSON_KEY`. We'll update the helm chart accordingly as well.

arschles:
@marwan-at-work I'm all good with what you said. I'll do this:

- Change this PR back to using the old env var, but keep the helm stuff that https://github.com/gomods/athens/pull/1428 added
- Ask you for another review
- If @epk wants the new env var, we can add it back in a follow-up

pkg/storage/gcp/gcp_test.go
arschles: Good point. You're not missing anything - I put this log in so that we'd see something specific about the GCS config in our drone logs, if we mess up the env var in the future

arschles:
@marwan-at-work when you have time, could I get another review? I think I got everything 😄

PR #1482, opened 2019-12-05
Going back to using the canary image in drone

arschles:

In #1480, we had to temporarily use a commit-specific athens image
to fix the build at the time. In that PR, we wrote a TODO to change
it back.

**How is the fix applied?**

I've changed the image we use in drone back to `canary`

This is a follow-up to the follow-on work described in 
https://github.com/gomods/athens/pull/1480#issue-349109197

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -154,7 +154,7 @@ services:
   ports:
     - 6379
 - name: athens-proxy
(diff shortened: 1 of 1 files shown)


arschles:
@marwan-at-work let me know what you think about this, when you get some time

PR #1484, opened 2019-12-05
Add support for IAM proxies for S3, like Kube2IAM and KIAM by:

### LINKED ISSUE #1452: Support for KIAM and kube2iam for S3 storage

aaqel-s:
What is the problem I am trying to address?
Adding support for AWS IAM proxies in Kubernetes

Currently, S3 support in S3s only supports AWS configuration using Hardcoded credentials. Our application of K8s uses KIAM avoid using hardcoded creds for AWS by having AWS requests proxied through a sidecar that injects credentials based on the IAM role of the pod.

How is the fix applied?

Added an environment variable to enable support for IAM proxies and then added an if statement in s3.go to bypass the AWS cred provider (which is not needed when using a proxy.

Added two flags to Helm to configure IAM support and the particular role a user wants the proxy attached to.

Verified with make docker and running in our k8s cluster.

Mention the issue number it fixes or add the details of the changes if it doesn't have a specific issue.

Fixes #1452


charts/athens-proxy/templates/deployment.yaml
```diff
@@ -24,6 +24,9 @@ spec:
         checksum/upstream: {{ include (print $.Template.BasePath "/config-upstream.yaml") . | sha256sum }}
         checksum/ssh-config: {{ include (print $.Template.BasePath "/config-ssh-git-servers.yaml") . | sha256sum }}
         checksum/ssh-secret: {{ include (print $.Template.BasePath "/secret-ssh-git-servers.yaml") . | sha256sum }}
+        {{- if .Values.storage.s3.iamRole }}
+        iam.amazonaws.com/role: {{ .Values.storage.s3.iamRole | quote }}
```
arschles: @aaqel-s will a controller/operator need to be present in the cluster to detect this annotation?

go.sum
```diff
@@ -11,15 +11,21 @@ github.com/Azure/azure-pipeline-go v0.2.1 h1:OLBdZJ3yvOn2MezlWvbrBMTEUQC72zAftRZ
 github.com/Azure/azure-pipeline-go v0.2.1/go.mod h1:UGSo8XybXnIGZ3epmeBw7Jdz+HiUVpqIlpz/HKHylF4=
```
arschles: @aaqel-s did you add/remove/change any dependencies that caused the `go.sum` to change? If not, I'd like to take out these changes

charts/athens-proxy/templates/deployment.yaml
aaqel-s: Yup, here's an example https://github.com/uswitch/kiam

robjloranger:
@aaqel-s just wanted to check in and see if you saw the other changes requested.

aaqel-s:
Sorry folks, I got sidetracked with other stuff at work, I'll try to get back to this asap.

arschles:
Oh, one more thing @aaqel-s - can you please change the `version` field in `Chart.yaml` to `0.4.4`?

matt0x6F:
Just doing a quick review since this is in the post 0.13.x milestone:

- This probably requires an update to the docs: https://docs.gomods.io/configuration/storage/#configuration-4
- This branch needs a rebase since it was ~4 years ago

I'm willing to take this change over if @aaqel-s doesn't have time. I don't really have a way to test it however as this requires a specific Kubernetes setup.

Issue #1491, opened 2019-12-12
Audit validation tags for configuration properties


bluekeyes:
**Is your feature request related to a problem? Please describe.**

For some properties (I hit it on `DownloadMode`, but may apply to others), the zero value is invalid and the property is effectively required, but is not marked as required for validation.

**Describe the solution you'd like**

Audit the configuration type definitions to make sure the validation tags are correct for properties. I think the `required` tag is the most relevant, since a missing required property gives a clear error at startup, while the zero value for the type may give a confusing error later on.

**Describe alternatives you've considered**

Introduce another method to validate configuration. This seems non-ideal, since validation tags are already used for some properties.

**Additional context**

This issue is split out of #1336, which has some additional information.

arschles:
@bluekeyes We should certainly get better at validating configurations. Since there are a lot of options, it's easy to forget something and make a mistake, so I think we should look holistically at the configuration. For example, if `DownloadMode` is set to `"redirect"`, then `DownloadURL` needs to be set. We should check that and report it and disallow Athens from starting.

What do you think?

bluekeyes:
Yeah, it makes sense to me that the application would need a way to validate relationships between properties and that that would provide a better user experience than just saying if an option is required or not.

From a pure code perspective, I personally find it a bit confusing when there are multiple validation methods at play in a project, so long term, I think it could be valuable to decide on when to use the `validator.v9` library (if at all) vs. when to use any new method introduced here. But I'm also not very familiar with the Athens code base or development style.

arschles:
@bluekeyes I don't have any experience with validator.v9, but it looks nice. The idea I have in mind is to centralize all validation of configuration in one place in the code, regardless of where the configuration values came from - file, environment variables, or anywhere else we define later on.

Ideally, the validations would come with good error messages too, when Athens is misconfigured. I'll check out validator.v9 to see if I can get it working in https://github.com/gomods/athens/pull/1492 too - thanks!

PR #1492, opened 2019-12-12
Adding a better error message for download mode

arschles:

There are a lot of configuration options and it's easy to forget something, misconfigure
Athens, and get stuck when you make a configuration mistake.

`DownloadMode` is a good example because it cannot be empty, and when it is, we just say
"it can't be empty" without any extra context.

See https://github.com/gomods/athens/issues/1491
and https://github.com/gomods/athens/issues/1336 for more discussion on `DownloadMode`

**How is the fix applied?**

I've added a nicer error message in cases where `DownloadMode` is empty, or it is set
to an invalid value.

If this code is acceptable, I'd like to apply it to a few other config values 
(in a follow-up PR) that are easy to get wrong.

ref https://github.com/gomods/athens/issues/1491, since this checks existence and 
validates values for download mode, and also reports errors in a more human-friendly way.

cc/ @bluekeyes

### DIFF
--- a/cmd/proxy/actions/app.go
+++ b/cmd/proxy/actions/app.go
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	"github.com/gomods/athens/pkg/config"
+	"github.com/gomods/athens/pkg/errors"
 	"github.com/gomods/athens/pkg/log"
 	mw "github.com/gomods/athens/pkg/middleware"
 	"github.com/gomods/athens/pkg/module"
@@ -132,6 +133,9 @@ func App(conf *config.Config) (http.Handler, error) {
 		lggr,
 		conf,
 	); err != nil {
+		if errors.IsConfigErr(err) {
+			return nil, err
+		}
 		err = fmt.Errorf("error adding proxy routes (%s)", err)
 		return nil, err
 	}
--- a/cmd/proxy/actions/app_proxy.go
+++ b/cmd/proxy/actions/app_proxy.go
@@ -11,6 +11,7 @@ import (
 	"github.com/gomods/athens/pkg/download"
 	"github.com/gomods/athens/pkg/download/addons"
 	"github.com/gomods/athens/pkg/download/mode"
+	"github.com/gomods/athens/pkg/errors"
 	"github.com/gomods/athens/pkg/log"
 	"github.com/gomods/athens/pkg/module"
 	"github.com/gomods/athens/pkg/stash"
@@ -98,7 +99,10 @@ func addProxyRoutes(
 	}
 	st := stash.New(mf, s, stash.WithPool(c.GoGetWorkers), withSingleFlight)
 
-	df, err := mode.NewFile(c.DownloadMode, c.DownloadURL)
+	df, err := mode.NewFile(c.DownloadMode, c.DownloadURL, errors.ConfigError(
+		"DownloadMode",
+		"https://docs.gomods.io/configuration/download/",
+	))
 	if err != nil {
 		return err
 	}
--- a/pkg/download/mode/mode.go
+++ b/pkg/download/mode/mode.go
@@ -18,13 +18,11 @@ type Mode string
 
 // DownloadMode constants. For more information see config.dev.toml
 const (
-	Sync            Mode = "sync"
-	Async           Mode = "async"
-	Redirect        Mode = "redirect"
-	AsyncRedirect   Mode = "async_redirect"
-	None            Mode = "none"
-	downloadModeErr      = "download mode is not set"
-	invalidModeErr       = "unrecognized download mode: %s"
+	Sync          Mode = "sync"
+	Async         Mode = "async"
+	Redirect      Mode = "redirect"
+	AsyncRedirect Mode = "async_redirect"
+	None          Mode = "none"
 )
 
 // DownloadFile represents a custom HCL format of
@@ -50,11 +48,11 @@ type DownloadPath struct {
 // you can either point to a file path by passing
 // file:/path/to/file OR custom:<base64-encoded-hcl>
 // directly.
-func NewFile(m Mode, downloadURL string) (*DownloadFile, error) {
+func NewFile(m Mode, downloadURL string, errFn errors.ConfigErrFn) (*DownloadFile, error) {
 	const op errors.Op = "downloadMode.NewFile"
 
 	if m == "" {
-		return nil, errors.E(op, downloadModeErr)
+		return nil, errFn("You have to pass a value.")
 	}
 
 	if strings.HasPrefix(string(m), "file:") {
@@ -76,7 +74,7 @@ func NewFile(m Mode, downloadURL string) (*DownloadFile, error) {
 	case Sync, Async, Redirect, AsyncRedirect, None:
 		return &DownloadFile{Mode: m, DownloadURL: downloadURL}, nil
 	default:
-		return nil, errors.E(op, errors.KindBadRequest, fmt.Sprintf(invalidModeErr, m))
+		return nil, errFn(fmt.Sprintf("%s isn't a valid value.", m))
 	}
 }
 
--- a/pkg/download/mode/mode_test.go
+++ b/pkg/download/mode/mode_test.go
@@ -3,6 +3,8 @@ package mode
 import (
 	"fmt"
 	"testing"
+
+	"github.com/gomods/athens/pkg/errors"
 )
 
 var testCases = []struct {
@@ -119,6 +121,10 @@ func TestMode(t *testing.T) {
 }
 
 func TestNewFile_err(t *testing.T) {
+	dlModeErrFn := errors.ConfigError(
+		"DownloadMode",
+		"https://docs.gomods.io/configuration/download/",
+	)
 	tc := []struct {
 		name     string
 		mode     Mode
@@ -127,12 +133,12 @@ func TestNewFile_err(t *testing.T) {
 		{
 			name:     "empty mode",
 			mode:     "",
-			expected: downloadModeErr,
+			expected: dlModeErrFn("You have to pass a value.").Error(),
 		},
 		{
 			name:     "invalid mode",
 			mode:     "invalidMode",
-			expected: fmt.Sprintf(invalidModeErr, "invalidMode"),
+			expected: dlModeErrFn(fmt.Sprintf("%s isn't a valid value.", "invalidMode")).Error(),
 		},
 	}
 	for _, c := range tc {
--- a/pkg/errors/config.go
+++ b/pkg/errors/config.go
@@ -1,0 +1,41 @@
+package errors
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ConfigErrFn is a function that can create a new configuration error. You pass it a
+// message specific to the error you found when you were validating configuration,
+// and it knows how to print out the actual configuration name and other helpful information.
+type ConfigErrFn func(string) error
+
+type configErr struct {
+	str string
+}
+
+func (c configErr) Error() string {
+	return c.str
+}
+
+// ConfigError returns a function that creates a configuration error to be printed
+// to the terminal. Call this function and pass its return value down the call stack to
+// functions that validate configuration fields.
+//
+// The function that ConfigError returns
+//
+// For example:
+//
+//	downloadModeFn := ConfigError("DownloadMode (ATHENS_DOWNLOAD_MODE)")
+//	err := doThingsWithDownloadMode(configStruct, downloadModeFn)
+func ConfigError(field string, url string) ConfigErrFn {
+	return func(helpText string) error {
+		slc := []string{
+			fmt.Sprintf("There was a configuration error with %s. %s", field, helpText),
+		}
+		if url != "" {
+			slc = append(slc, fmt.Sprintf("See %s for more information.", url))
+		}
+		return &configErr{str: strings.Join(slc, "\n")}
+	}
+}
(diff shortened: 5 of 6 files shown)


pkg/errors/kinds.go
```diff
@@ -4,3 +4,9 @@ package errors
 func IsNotFoundErr(err error) bool {
 	return Kind(err) == KindNotFound
 }
+
+// IsConfigErr returns true if the given err is a configuration error
+func IsConfigErr(err error) bool {
```
marwan-at-work: we probably don't need this if we just do `errors.Is(err, errors.ConfigErrorKind)` or whatever kind name we'll eventually pick. 

But furthermore, I don't think we need a new config error kind, because we don't programmatically do anything about it anyway, we just log the error and exit.

cmd/proxy/actions/app.go
```diff
@@ -132,6 +133,9 @@ func App(conf *config.Config) (http.Handler, error) {
 		lggr,
 		conf,
 	); err != nil {
+		if errors.IsConfigErr(err) {
```
marwan-at-work: so all that we're doing here is choosing *not* to wrap the error with "error adding proxy routes (message)". I think, instead of doing that, we can just pick a better wrapper-message and then we don't need this exposed API at all, we can just say: `"error initializing Athens: %s"` below.

cmd/proxy/actions/app_proxy.go
```diff
@@ -98,7 +99,10 @@ func addProxyRoutes(
 	}
 	st := stash.New(mf, s, stash.WithPool(c.GoGetWorkers), withSingleFlight)
 
-	df, err := mode.NewFile(c.DownloadMode, c.DownloadURL)
+	df, err := mode.NewFile(c.DownloadMode, c.DownloadURL, errors.ConfigError(
```
marwan-at-work: though this functional pattern is pretty awesome, it's a bit of an overkill just to add 2 lines of text to an error: 
1. We never have to customize this line
2. We can just add those two lines into the errors themselves inside `mod.NewFIle` 
3. We can still keep it functional by keeping the `mode.NewFile` signature exactly the same, but having this functional pattern internal and private to the package itself so that you don't have to manually do it a few times

pkg/errors/config.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ConfigErrFn is a function that can create a new configuration error. You pass it a
+// message specific to the error you found when you were validating configuration,
+// and it knows how to print out the actual configuration name and other helpful information.
+type ConfigErrFn func(string) error
```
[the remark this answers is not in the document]
arschles: @marwan-at-work I wanted to but can't because the config package imports mode which needs all the config error stuff. Do you think it should be somewhere besides there?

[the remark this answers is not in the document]
marwan-at-work: Sorry, I meant the `mode` package...which makes me realize `ConfigErrFn` makes no sense here, since it's not a Config package error, it's a `mode` error :) 

So that said, all of my suggestion are meant for keeping this functionality local to wherever the error is happening, in this case inside `mode.go` 

Thanks @arschles :v:

arschles:
@marwan-at-work I made some changes to (hopefully) address your comments. let me know what you think ✌️

cmd/proxy/main.go
```diff
@@ -32,6 +33,9 @@ func main() {
 	if err != nil {
 		log.Fatalf("could not load config file: %v", err)
 	}
+	if err := conf.Validate(errors.Op("main")); err != nil {
```
marwan-at-work: We already have a `validateConfig` function here: https://github.com/gomods/athens/blob/master/pkg/config/config.go#L238

It gets automatically called when you call `config.Load` 

So this is a bit confusing :)

In the same fashion, the `mode.Validate` function in this case is getting called twice, once here and once in `mode.NewFile`.

I think we can just:
1. remove this line here 
2. remove `config.Validate` function 
3. make `mode.Validate` private

pkg/download/mode/mode.go
```diff
@@ -18,15 +18,29 @@ type Mode string
...
+	Sync          Mode = "sync"
+	Async         Mode = "async"
+	Redirect      Mode = "redirect"
+	AsyncRedirect Mode = "async_redirect"
+	None          Mode = "none"
 )
 
+// Validate ensures that m is a valid mode. If this function returns nil, you are
+// guaranteed that m is valid
+func (m Mode) Validate(op errors.Op) error {
```
marwan-at-work: Instead of it taking an `op`, just make a `mode.Validate` operation under this line and wrap the error with it. This way, people can get the entire call stack by the time the server fails to boot up.

pkg/download/mode/mode.go
```diff
@@ -44,17 +58,30 @@ type DownloadPath struct {
...
+}
+
 // NewFile takes a mode and returns a DownloadFile.
 // Mode can be one of the constants declared above
 // or a custom HCL file. To pass a custom HCL file,
 // you can either point to a file path by passing
 // file:/path/to/file OR custom:<base64-encoded-hcl>
 // directly.
 func NewFile(m Mode, downloadURL string) (*DownloadFile, error) {
+	const downloadModeURL = "https://docs.gomods.io/configuration/download/"
```
marwan-at-work: I would make this a package level constant (private) as it is being used 3 more times in the pr

pkg/download/mode/mode_test.go
```diff
@@ -127,12 +129,12 @@ func TestNewFile_err(t *testing.T) {
 		{
 			name:     "empty mode",
 			mode:     "",
-			expected: downloadModeErr,
+			expected: Mode("").Validate(op).Error(),
```
marwan-at-work: hmm, we probably shouldn't do string matching here: either 
1. change `expected string` to `hasErr bool` 
2. or, change `expected string` to `expected errors.Kind`

pkg/errors/errors.go
```diff
@@ -18,6 +18,7 @@ const (
 	KindRateLimit      = http.StatusTooManyRequests
 	KindNotImplemented = http.StatusNotImplemented
 	KindRedirect       = http.StatusMovedPermanently
+	KindConfigError    = -1
```
marwan-at-work: I believe the HTTP protocol gives you some room for custom codes
In this case this is a client (user) error so maybe we can give it a code between 451 and 499: https://golang.org/pkg/net/http/#pkg-constants

It doesn't really matter but just being a bit more conventional :)

bluekeyes: 422 / `http.StatusUnprocessableEntity` is a common response when a client provide syntactically valid but _semantically invalid_ objects to an API and could be appropriate here, if you want to stick with valid HTTP codes.

arschles: @marwan-at-work fair enough! I'll take @bluekeyes's suggestion and go with 422, just so we can use a constant.

cmd/proxy/main.go
arschles: @marwan-at-work I added a `Validate` function on config structs so that we can nest validations. I'll try to move all of the logic into [`validate`](https://pkg.go.dev/gopkg.in/go-playground/validator.v9) struct tags though

arschles:
@marwan-at-work I made a few fixes in here. Regarding https://github.com/gomods/athens/pull/1492#discussion_r360176799, I put all the calls to `Validate` inside `validateConfig`. I think that the validate struct tags and custom validators are useful in some cases, but the nested `Validate` functions seem to be a lot more understandable than [validator](https://pkg.go.dev/gopkg.in/go-playground/validator.v9@v9.20.2?tab=doc). Are you cool with leaving this pattern in?

pkg/errors/config.go
```diff
@@ -0,0 +1,22 @@
...
+	"strings"
+)
+
+// Config returns an error specifically tailored for reporting errors with configuration
+// values. You can check for these errors by calling errors.Is(err, KindConfigError)
+// (from the github.com/gomods/athens/pkg/errors package).
+//
+// Generally these kinds of errors should make Athens crash because it was configured
+// improperly
+func Config(op Op, field, helpText, url string) error {
```
marwan-at-work: will this ever be called from outside the `config` package? why not just define it there and define it locally to minimize the API surface of this package?

pkg/errors/config.go
```diff
@@ -0,0 +1,22 @@
...
+	slc := []string{
+		fmt.Sprintf("There was a configuration error with %s. %s", field, helpText),
+	}
+	if url != "" {
+		slc = append(slc, fmt.Sprintf("Please see %s for more information.", url))
+	}
+	return E(op, KindConfigError, strings.Join(slc, " "))
```
marwan-at-work: I'm actually not sure, but would it be more readable if we strings.Join'd by "\n\t"? I haven't looked at what all of those errors will look like on the terminal yet, but will git-checkout your branch to check it out 👍

pkg/download/mode/mode.go
```diff
@@ -72,12 +104,11 @@ func NewFile(m Mode, downloadURL string) (*DownloadFile, error) {
 		return parseFile(bts)
 	}
 
-	switch m {
-	case Sync, Async, Redirect, AsyncRedirect, None:
-		return &DownloadFile{Mode: m, DownloadURL: downloadURL}, nil
-	default:
-		return nil, errors.E(op, errors.KindBadRequest, fmt.Sprintf(invalidModeErr, m))
+	retFile := &DownloadFile{Mode: m, DownloadURL: downloadURL}
```
marwan-at-work: Just a nit: what does retFile stand for? maybe we can just call it `df` for DownloadFile

pkg/download/mode/mode.go
```diff
@@ -53,8 +85,8 @@ type DownloadPath struct {
 func NewFile(m Mode, downloadURL string) (*DownloadFile, error) {
 	const op errors.Op = "downloadMode.NewFile"
 
-	if m == "" {
-		return nil, errors.E(op, downloadModeErr)
+	if err := m.Validate(); err != nil {
```
marwan-at-work: I think this error is now less readable, because before it used to say `download mode is not set` but now it will say ` isn't a valid value` (notice the space in the beginning because there's nothing) 

I think in the Sprintf of the validate function we should pass a `%q` so it would say `"" isn't a valid value` or we should keep the old error here for better messaging.

pkg/download/mode/mode.go
```diff
@@ -72,12 +104,11 @@ func NewFile(m Mode, downloadURL string) (*DownloadFile, error) {
 		return parseFile(bts)
 	}
 
-	switch m {
-	case Sync, Async, Redirect, AsyncRedirect, None:
-		return &DownloadFile{Mode: m, DownloadURL: downloadURL}, nil
-	default:
-		return nil, errors.E(op, errors.KindBadRequest, fmt.Sprintf(invalidModeErr, m))
+	retFile := &DownloadFile{Mode: m, DownloadURL: downloadURL}
+	if err := retFile.Validate(); err != nil {
```
marwan-at-work: Correct me if I'm wrong but: 

The old code that was removed here no longer works because in the old code we are checking the top level Mode, but in this validate function only the internal `d.Paths[n].Mode` is checked.

However, since top level Mode is now already checked above, then this Validate function is not necessary because `retFile` has no `Path`s to begin with.

pkg/config/validator.go
```diff
@@ -0,0 +1,8 @@
+package config
+
+// Validator can validate a config struct. If you implement this,
+// validate all of the configuration in your struct. It will
+// automatically be called when Athens starts
+type Validator interface {
```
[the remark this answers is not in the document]
arschles: You're right, I got ahead of myself here. My intention was to have the `config.Validate` function look for `Validator`s in each of its fields, calling validate on each of them. Right now, `Validate` is called as necessary on each field that implements it, in various other parts of the code.

I'll leave `Validator` out of this PR

pkg/errors/config.go
arschles: Nope. I put it here just because I think it makes more sense to group a configuration error under the `errors` package. I'm willing to be convinced otherwise 😄

pkg/errors/config.go
arschles: They'll be a long-ish single line most of the time. For example: `There was a configuration error with somefield. You have to do something else. Please see docs.gomods.io for more information`.

I agree that multiple lines would be more readable. I'll make that change.

pkg/download/mode/mode.go
arschles: a heavily abbreviated "file to return". I see `df` in a lot of other places, so happy to change it to that.

pkg/download/mode/mode.go
arschles: How about if we pass `%q` as you said, and also prefix the error with "Download mode ..."? 
The result would then be `Download mode "" isn't a valid value`

pkg/download/mode/mode.go
arschles: @marwan-at-work you're right. I left a few things out of this `Validate` and some tangential functions

- `DownloadFile.Validate` needs to check `DownloadURL` for well-formedness
- ^^ also needs to call `DownloadPath.Validate` for each `d.Paths[n]` - that does more than just validate its mode
- `Mode.Validate` does not check for `custom:` and `file:` values

I've made changes to address all this

I've fixed all of that

arschles:
@marwan-at-work I think I changed code or responded to all your comments. Let me know what you think or if I missed something! 💚

pkg/download/mode/mode.go
marwan-at-work: That sounds great

marwan-at-work:
@arschles thank you for fixing all the comments 💯 💯 will take a look as soon as possible 🙏

arschles:
Take your time @marwan-at-work! This PR is a "slow burn" 😄

Issue #1495, opened 2019-12-13
Change 'make build' to not cd into ./cmd/proxy


arschles:
**Is your feature request related to a problem? Please describe.**

The Makefile runs `cd cmd/proxy && go build` when you run `make build`, so you get a built Athens binary but you end up in a different directory than where you started.

**Describe the solution you'd like**

I'd like the makefile to run `go build -o cmd/proxy/proxy ./cmd/proxy` when you run `make build`. This has the exact same effect, except you stay in the directory you started in.

**Describe alternatives you've considered**

We could do the same thing as we do now, but run `cd ../../` at the end of the command. That seems like more complexity than we need.

**Additional context**

I keep not noticing that I get moved to a different directory 😆

PR #1497, opened 2019-12-13
Making 'make build' not cd into the cmd/proxy directory

arschles:

`make build` runs `cd cmd/proxy` before doing the `go build`. That means that you end up in a different directory after you run this command.

**How is the fix applied?**

I've changed the `make build` from `cd cmd/proxy && go build` to `go build -o ./cmd/proxy/proxy ./cmd/proxy`, so that the exact same thing happens as before, except you stay in the directory you started at 

Fixes https://github.com/gomods/athens/issues/1495

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ endif
 
 .PHONY: build
 build: ## build the athens proxy
-	cd cmd/proxy && go build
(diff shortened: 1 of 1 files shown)


arschles:
@robjloranger I'm gonna break my own rules and merge my own PR 😬

PR #1498, opened 2019-12-19
Azure blob test

arschles:

`_` is an invalid char in Azure blob store container names.

**How is the fix applied?**

Using `-` instead.

I'm picking up where @marpio left off in #1366. I've only added [one](https://github.com/gomods/athens/commit/763532bad4458756328ffc670fa09ed7f6980142) extra tiny commit to finish his PR off. @marpio when you get a chance, can you let me know if this looks ok to you?

Closes #1366


manugupt1:
thanks for the changes. The changes look good to me; but the build for this test is failing @arschles. Seems like a key needs to be updated from what I understand, would you be able to fix it when you get a chance.

arschles:
@manugupt1 I am about to go on vacation for the rest of the year, but I'll bookmark this and get back to it in the new year!

Issue #1363, opened 2019-08-28
Proposal: adding docs on how to set GONOSUMDB


arschles:
**Is your feature request related to a problem? Please describe.**

We have [docs](https://docs.gomods.io/configuration/sumdb/) on how to make Athens proxy the global sum DB, but no docs on how to configure your toolchain to ignore the sum db for private modules.

**Describe the solution you'd like**

I think we should add docs on how to configure `GONOSUMDB` and why you'd want to do that.

**Describe alternatives you've considered**

We could just send people to the modules wiki I suppose?

**Additional context**

We won't need this until the sum DB is turned on by default (1.13 at the earliest if I remember correctly off the top of my head)

If we do this work, we can see if it ends up being difficult for folks to set GONOSUMDB, and add guidance on how to make it easier to set that env var. [direnv](https://direnv.net/) makes this easier, for example.

mschneider82:
docs would be nice! I just tried go 1.13rc2 with athens and private repo and failed because of the sum's..

marwan-at-work:
@mschneider82 this is fixed in Athens v0.6.1, feel free to give it a try. You can pass GOPRIVATE or GONOSUMDB through the new `GoBinaryEnvVars` config, or through the existing `NoSumPatterns` config. 

This issue should stay open open and would be great to add documentation for people unfamiliar with the new Go 1.13 behavior 

Thanks

mschneider82:
hi @marwan-at-work 
I tried it, but same nearly the same problem:


my athens config:

```
GoBinaryEnvVars = ["GOPROXY=direct", "GOPRIVATE=git.internal", "GONOSUMDB=git.internal/*"]
```

Athens log:
```
INFO[0140] exit status 1: go list -m: git.internal@latest: unrecognized import path "git.internal" (parse https://git.internal/?go-get=1: no go-import meta tags ())  http-method=GET http-path=/git.internal/@v/list http-url=/git.internal/@v/list kind="Not Found" module= operation=download.ListHandler ops="[download.ListHandler pool.List protocol.List vcsLister.List]" version=
handler: GET /git.internal/@v/list [404]
[...]

PR #1504, opened 2019-12-20
Adding docs for setting the GONOSUMDB env var on the Athens side

arschles:

There are not good documentation for now to set `NoSumPatterns` / `ATHENS_GONOSUM_DB` config values

**How is the fix applied?**

I've added some documentation for that.

Ref https://github.com/gomods/athens/issues/1363

Not marking this "`fixes`" because there's still ongoing troubleshooting going on in #1363

cc/ @mschneider82 @marwan-at-work


docs/content/configuration/sumdb.md
```diff
@@ -5,41 +5,47 @@ weight: 2
...
 ```
 
-This will make sure that Go does not send any requests to the Checksum DB for your private import paths. 
-However, how can you ensure that all of your employees are building private code with the right configuration? 
+This will make sure that Go does not send any requests to the Checksum DB for your private import paths. However, how can you ensure that all of your employees are building private code with the right configuration?
+
+Athens, in this case can help ensure that all private code flowing through it is never checked against the checksum DB. That means that as long as your employees are using Athens, then they will get a helpful reminder to ensure Their GONOSUMDB is rightly configured.
 
-Athens, in this case can help ensure that all private code flowing through it never goes to the Checksum DB. So as long as your employees are using Athens, then they will get a helpful reminder to ensure Their GONOSUMDB is rightly configured. 
+As the Athens company maintainer, you can run Athens with the following configuration in your `config.toml` file:
```
twexler: "Athens company maintainer" seems odd here, as it may not always be a company running  Athens. Let's just keep this simple: 
```suggestion
Athens allows the user to specify a list of patterns to not lookup via the upstream sum provider:
```

marwan-at-work: Agree with @twexler , although an "Athens user" is a bit ambiguous because it can mean one of two things: 
1. Athens server maintainer (the person who configures and runs Athens)
2. Athens server user (the person pinging Athens directly or through the Go command) 

So maybe we can just say "Athens allows its operator to specify a list of patterns..."
or "Athens allows its maintainer to specify a list of patterns..."

arschles: I applied @twexler's commit. I'll change it to add the "operator" qualifier

arschles:
Sorry for the delay. @twexler and @marwan-at-work can you take one last look?

Issue #1501, opened 2019-12-20
Update the upstream proxy & filter config documentation


arschles:
**Is your feature request related to a problem? Please describe.**

[The upstream config docs](https://docs.gomods.io/configuration/upstream/) and [filter docs](https://docs.gomods.io/configuration/filter/) still reference the deprecated filter file

**Describe the solution you'd like**

These docs should be updated to show the [download mode](https://docs.gomods.io/configuration/download/) file as the preferred way configure upstreams. The filter file docs should stay as legacy documentation.

**Describe alternatives you've considered**

Removing the page altogether in favor of https://docs.gomods.io/configuration/download/. I think we should keep the deprecated documentation, though, because we haven't officially dropped support for the filter file.

**Additional context**
N/A

PR #1508, opened 2019-12-21
Revamping the download mode file documentation

arschles:

The [upstream](https://docs.gomods.io/configuration/upstream/) and [filter](https://docs.gomods.io/configuration/filter/) configuration documents reference the deprecated filter file. They shouldn't.

**How is the fix applied?**

I've added a note that these two configuration documents are deprecated, and added a link to the download file documentation.

Additionally:

- I've updated the download mode file to be a bit more clear
- I've changed the order of the configuration documentation to put the deprecated documents at the bottom of the config section
- I've added a note to the disk storage documentation about pre-filling the disk cache

Fixes https://github.com/gomods/athens/issues/1501


arschles:
I'm going to merge this w/o a maintainer review since @twexler has recently run into confusion because the deprecation notice was missing. Others have run into the same confusion in the past. @gomods/maintainers please leave comments here and I will make changes in follow-ups

marwan-at-work:
LGTM :)

PR #1514, opened 2019-12-30
Replace the current e2e script with a test suite running e2e tests.

### LINKED ISSUE #1474: Expand the e2e tests
Right now the e2e tests we have are provided the test_e2e.sh file. We want to extend them, ideally using a code driven approach replacing the bash script.

Having the script replaced by a test suite that can run e2e tests.

None

**Additional context**
This was discussed in https://github.com/gomods/athens/pull/1449

fedepaol:
Right now the e2e tests are performed by a non trivial bash script, difficult to extend with new tests.
This pr replace it with a go - baked e2e test suite, right now adding the same tests we have in the bash script plus one using a wrong `GOPROXY` variable.

This should be easier to maintain / extend.

**How is the fix applied?**

By implementing a new test suite that builds athens and interacts directly with go.
 
Fixes #1474


robjloranger:
Assigned myself to remember to try this out and go over it in more detail. But I like it.

fedepaol:
@robjloranger @arschles any chance to give a push to this?

robjloranger:
I'm so sorry @fedepaol, I missed it again. I've been working too much at my day job. I will do this on the weekend at the latest.

fedepaol:
@robjloranger no problem! I was just going through blocked PRs of mine and I remembered about this. Take your time, and thanks!

robjloranger:
testing now, finally

appveyor.yml
```diff
@@ -11,4 +11,4 @@ environment:
 stack: go 1.13
 
 test_script:
-  - go test ./...
\ No newline at end of file
+  - go test --tags unittests ./...
```
robjloranger: silly newline missing

fedepaol: Done

fedepaol:
> Looks good to me, other than one line ending smiley

Thanks for your review :+1:

.drone.yml
```diff
@@ -28,7 +28,7 @@ steps:
   # test
   - ./scripts/check_gofmt.sh
   - go vet ./...
-  - go test -v ./...
+  - go test --tags unittests -v ./...
```
marwan-at-work: Just checking: this means that `go test ./...` will always try to run e2e tests no? 
In other words, everytime you want to run the unit tests you have to pass that tag. 

If that's true, then it would be more correct to say `go test --tags e2e` instead

robjloranger: Good call, if true then defaulting to unit tests and hiding e2e behind a flag would be a better UX.

fedepaol: Good point, and I agree. I will change the build tag to be selective for e2e tests, so the default one is for plain unit tests.

e2etests/run_athens.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"fmt"
+	"net/http"
+	"os/exec"
+	"path"
+	"path/filepath"
+	"time"
+)
+
+func buildAthens(goBin string, destPath string, env []string) (string, error) {
```
marwan-at-work: nit: `goBin, destPath string` -- no need to declare the type twice.

e2etests/run_athens.go
```diff
@@ -0,0 +1,56 @@
...
+	target := path.Join(destPath, "athens-proxy")
+	binFolder, err := filepath.Abs("../cmd/proxy")
+	if err != nil {
+		return "", fmt.Errorf("Failed to get athens source path %v", err)
+	}
+
+	cmd := exec.Command(goBin, "build", "-o", target, binFolder)
+	cmd.Env = env
```
marwan-at-work: I'd add a bytes.Buffer to cmd.Stderr and return it in case of error because then it will explain why go build failed.

e2etests/run_athens.go
```diff
@@ -0,0 +1,56 @@
...
+	err = cmd.Run()
+	return target, err
+}
+
+func stopAthens() error {
+	cmd := exec.Command("pkill", "athens-proxy")
+	err := cmd.Run()
```
marwan-at-work: i'd either just `return cmd.Run()` or properly populate stdout/stderr

e2etests/run_athens.go
```diff
@@ -0,0 +1,56 @@
...
+	return err
+}
+
+func runAthensAndWait(athensBin string, env []string) error {
+	cmd := exec.Command(athensBin)
+	cmd.Env = env
+
+	go func() {
```
marwan-at-work: why not just `go cmd.Run()` ?

marwan-at-work: and if you are trying to run the command in the background, it is more appropriate to use cmd.Start(): https://golang.org/pkg/os/exec/#Cmd.Start

e2etests/run_athens.go
```diff
@@ -0,0 +1,56 @@
...
+		cmd.Run()
+	}()
+
+	ticker := time.NewTicker(time.Second)
+	timeout := time.After(2 * time.Minute)
```
marwan-at-work: two minutes seems a lil crazy to wait for athens to start 😄 -- let's make it 20 seconds?

e2etests/run_athens.go
```diff
@@ -0,0 +1,56 @@
...
+		return "", fmt.Errorf("Failed to get athens source path %v", err)
+	}
+
+	cmd := exec.Command(goBin, "build", "-o", target, binFolder)
+	cmd.Env = env
+	err = cmd.Run()
+	return target, err
+}
+
+func stopAthens() error {
```
marwan-at-work: I think a better way to "stopAthens" is to have the runAthensAndWait take a context that you can cancel whenever you want to stop athens and just pass that context into https://golang.org/pkg/os/exec/#CommandContext

which should render this whole function unnecessary.

e2etests/all_test.go
```diff
@@ -0,0 +1,135 @@
...
+	runAthensAndWait(athensBin, m.getEnv())
+	setupTestRepo(m.sampleRepoPath, "https://github.com/athens-artifacts/happy-path.git")
+}
+
+func (m *E2eSuite) TearDownSuite() {
+	err := stopAthens()
+	if err != nil {
+		m.Fail("Failed to stop athens", err)
+	}
+	chmodR(m.tmpDir, 0777)
```
marwan-at-work: I think a cleaner way to do this is to just call clean modcache -- I think we should probably do it on the go_get_fetcher as well. But that's another topic :)

e2etests/all_test.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (m *E2eSuite) SetupSuite() {
+	var err error
+	m.tmpDir, err = mkTemp()
+	if err != nil {
+		m.Fail("Failed to make temp dir", err)
+	}
+
+	m.goModCache = path.Join(m.tmpDir, "pkg", "mod")
```
marwan-at-work: I would probably make two temp directories to maintain a nicer separation: one for GOPATH and one outside of GOPATH for happy-path. This way you don't to actually care about the "pkg/mod" part here, and cleaning the cache only needs to take the GOPATH env var and nothing else

e2etests/run_athens.go
fedepaol: > and if you are trying to run the command in the background, it is more appropriate to use cmd.Start(): https://golang.org/pkg/os/exec/#Cmd.Start

Ah, nice!

e2etests/run_athens.go
fedepaol: Sorry, I am getting deviated to k8s timescales

e2etests/run_athens.go
fedepaol: Did not know about it, thanks!

fedepaol: Well, yes and no, as I am using stopAthens also to remove any dangling instance before running the real tests. So it's ok for the teardown phase, but I'll keep this around as well.

e2etests/all_test.go
fedepaol: I do agree to clear the cache instead of brutally erasing in `SetupTest`, as there we care only about the cache to be empty.
In `TearDown` though, I'd like to remove any additional folder that was created by the test suite, and I am afraid clearing the cache is not enough.

e2etests/all_test.go
fedepaol: Done

fedepaol:
@arschles I think I addressed all the comments in my latest push of yesterday.
@marwan-at-work would you mind giving another look?

e2etests/all_test.go
marwan-at-work: @fedepaol right you can still remove the folder, but there won't be a need to `chmod` anything :)

e2etests/all_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+	"github.com/gobuffalo/envy"
+	"github.com/stretchr/testify/suite"
+)
+
+type E2eSuite struct {
+	suite.Suite
+	goBinaryPath   string
+	env            []string
+	tmpDir         string
```
marwan-at-work: i'd call it `goPath` or `GOPATH`  to know what the tmpDir is for :)

e2etests/all_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+	athensBin, err := buildAthens(m.goBinaryPath, m.tmpDir, m.env)
+	if err != nil {
+		m.Fail("Failed to build athens ", err)
+	}
+	stopAthens() // in case a dangling instance was around.
+	// ignoring error as if no athens is running it fails.
+
+	ctx := context.Background()
+	ctx, m.stopAthens = context.WithCancel(ctx)
```
marwan-at-work: Form the context package docs: 
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter. 

I realize storing the context is necessary here because of testify/suite, but I'd rather we followed best practices instead of having the framework steer us away form that. 

That said, we can definitely do it in a follow up PR because it means removing the test suite from the e2e tests :)

fedepaol: Ah right. That would make things less easy. Not even sure if it's worth reverting to the previous `stopAthens`, as even if we remove testify having a propagated context will bloat the code a bit.

e2etests/all_test.go
fedepaol: Yes I left it there as I did not remember why I added originally. I'll try to remove it and see if it works :-)

fedepaol:
@marwan-at-work I applied the change suggested in https://github.com/gomods/athens/pull/1514#discussion_r379846877 .
I am not sure about the `context` thing though. We need to spin up athens when the "suite" (or the set of tests) start, and to stop it when the "suite" is complete, so I can't see a clean way to have the context not stored and propagated through functions. 
I feel like the best approach is to revert to the previous "stop" function. Wdyt?

marwan-at-work:
@fedepaol IMHO the best way is to just use the standard library because the std lib wouldn't get in your way the same way this test suite does :) 

That said, we can keep things the way they are because it works, then remove them and use regular table tests a follow up pr

fedepaol:
> @fedepaol IMHO the best way is to just use the standard library because the std lib wouldn't get in your way the same way this test suite does :)
> 
> That said, we can keep things the way they are because it works, then remove them and use regular table tests a follow up pr

Ok, we can discuss those details later then. Thanks again for your review!

fedepaol:
So this can be merged right?

arschles:
@fedepaol thank you * 💯 for your hard work on this PR!

mplachter:
Heads up this did break the `docker-e2e-tests`

arschles:
@mplachter would it help if I go ahead and fix `docker-e2e-tests` in a separate PR?

PR #1517, opened 2020-01-08
Display warning log message if validation fails

### LINKED ISSUE #1513: Improve ValidationHook responses
The ValidationHook responses only respond with true/false depending on the HTTP status code response after validation. It would be useful to have context around validation failures.

Optionally add JSON message to the response body that could include parameters such as:

> {
  “reason”: “not found”,
  “description”: “module and namespace does not exist”
}

Any JSON message received should get logged providing easier remediation.

doddi:

Provide details on the reason for the validation failure.

**How is the fix applied?**

As part of the response message for the ValidationHook the response body can optionally contain a payload

If a payload is detected a message is logged similar to that shown below:

`WARN[0152] Module github.com/go-gitea/gitea:v1.5.1 Reason: Reason for warning Description: Detailed description of warning  http-method=GET http-path=/github.com/go-gitea/gitea/@v/v1.5.1.mod http-url=/github.com/go-gitea/gitea/@v/v1.5.1.mod
handler: GET /github.com/go-gitea/gitea/@v/v1.5.1.mod [403]`

It is the responsibility of the ValidationHook endpoint to determine if a build should fail or continue (403/200)

Fixes #1513


pkg/middleware/validation.go
```diff
@@ -45,31 +47,60 @@ func NewValidationMiddleware(validatorHook string) mux.MiddlewareFunc {
 	}
 }
 
+func maybeLogValidationReason(valid ValidationResponse, r *http.Request, mod string, version string) {
+	message := valid.ValidationMessage
+	if message.Reason != "" {
+		entry := log.EntryFromContext(r.Context())
+		entry.Warnf("Module %s:%s Reason: %s, Description: %s", mod, version, message.Reason, message.Description)
```
marwan-at-work: It's a bit of an unwritten convention to do `mod@ver` instead of `mod:ver`

pkg/middleware/validation.go
```diff
@@ -45,31 +47,60 @@ func NewValidationMiddleware(validatorHook string) mux.MiddlewareFunc {
...
 }
 
-func validate(hook, mod, ver string) (bool, error) {
+type ValidationMessage struct {
+	Reason      string
+	Description string
+}
+
+type ValidationResponse struct {
+	Success           bool
```
marwan-at-work: what is this boolean used for? Conventionally, status OK means Success=true while status FORBIDDEN means Success=false. Does this serve another purpose?

pkg/middleware/validation.go
marwan-at-work: As for `Reason: %s, Description: %s`, couldn't the server return that message formatting and so we would just log it out without having to restrict all Validation Hook Servers to follow the same exact shape? In other words we can just print the http response body as is and the validation server is responsible for formatting the message however it wants.

pkg/middleware/validation.go
doddi: It is essentially still achieving the same result. I was just making it more explicit on the meaning that forbidden == success.
Here I am setting `Success` based on the response code https://github.com/gomods/athens/pull/1517/files#diff-7522a4893602f93e19a3c36e780166e2R101 . I can revert this if you would prefer to pass around the status code instead.

pkg/middleware/validation.go
doddi: Changed the format: https://github.com/gomods/athens/pull/1517/commits/248709afc8f032bee8a6cdd4818ffc02663566fd

pkg/middleware/validation.go
doddi: I have changed the success to carry the status code instead https://github.com/gomods/athens/pull/1517/commits/933d4430dadc1cfd1b8a34d85162930e503434b3

pkg/middleware/validation.go
doddi: Now just a simple message retrieved from the body https://github.com/gomods/athens/pull/1517/commits/933d4430dadc1cfd1b8a34d85162930e503434b3

pkg/middleware/validation.go
```diff
@@ -45,31 +48,50 @@ func NewValidationMiddleware(validatorHook string) mux.MiddlewareFunc {
...
+}
+
 type validationParams struct {
 	Module  string
 	Version string
 }
 
-func validate(hook, mod, ver string) (bool, error) {
+type ValidationResponse struct {
+	Status  int
```
marwan-at-work: I think it might better to change this to `Valid bool`

pkg/middleware/validation.go
```diff
@@ -35,7 +36,9 @@ func NewValidationMiddleware(validatorHook string) mux.MiddlewareFunc {
 					return
 				}
 
-				if !valid {
+				maybeLogValidationReason(valid, r, mod, version)
+
+				if valid.Status != http.StatusOK {
```
marwan-at-work: Given the previous comment, we can then just check `if valid.Valid` -- or maybe rename the variable and it will read better: `if response.Valid`

pkg/middleware/validation.go
```diff
@@ -45,31 +48,50 @@ func NewValidationMiddleware(validatorHook string) mux.MiddlewareFunc {
 	}
 }
 
+func maybeLogValidationReason(valid ValidationResponse, r *http.Request, mod string, version string) {
```
marwan-at-work: this signature can be slightly improved, since you don't need to use any underlying variables from an http.Response or a ValidationResponse: 

`func maybeLogValidationReason(ctx context.Context, msg, mod, version string)  {...}` -- or maybe even take a `log.Entry` instead of context, but happy either way.

pkg/middleware/validation.go
```diff
@@ -45,31 +48,50 @@ func NewValidationMiddleware(validatorHook string) mux.MiddlewareFunc {
...
+	if len(valid.Message) > 0 {
+		entry := log.EntryFromContext(r.Context())
+		entry.Warnf("Module %s@%s %s", mod, version, valid.Message)
```
marwan-at-work: I think it would read better if you changed the message to: `entry.Warnf("error validating %s@%s: %s", mod, version, valid.Message)`

pkg/middleware/validation.go
```diff
@@ -45,31 +48,50 @@ func NewValidationMiddleware(validatorHook string) mux.MiddlewareFunc {
...
 	resp, err := http.Post(hook, "application/json", bytes.NewBuffer(jsonVal))
 	if err != nil {
-		return false, errors.E(op, err)
+		return ValidationResponse{Status: http.StatusForbidden}, errors.E(op, err)
 	}
 
 	switch {
 	case resp.StatusCode == http.StatusOK:
-		return true, nil
+		return maybeReadResponseReason(resp), nil
```
marwan-at-work: This switch statement can be factored out to read a bit better: 

```
switch resp.Status {
  case http.StatusOK:
    return maybeReadResponseReason(true), nil
  case http.StatusForbidden:
    return maybeReadResponseReason(false), nil
  default:
    return ValidationResponse{Status: http.StatusForbidden}, errors.E(op, "Unexpected status code ", resp.StatusCode)
}
```

While maybeReadResponseReason can just take a bool to populate `Valid` from above

doddi: The response is required to be passed through in order to fill in the message parsed from the response.Body https://github.com/gomods/athens/blob/933d4430dadc1cfd1b8a34d85162930e503434b3/pkg/middleware/validation.go#L95

doddi: made the changes to switch statement https://github.com/gomods/athens/pull/1517/commits/12dba10897b22c735db29b02d50fd64a85bba2ea

pkg/middleware/validation.go
doddi: https://github.com/gomods/athens/pull/1517/commits/12dba10897b22c735db29b02d50fd64a85bba2ea

pkg/middleware/validation.go
doddi: https://github.com/gomods/athens/pull/1517/commits/12dba10897b22c735db29b02d50fd64a85bba2ea

pkg/middleware/validation.go
doddi: https://github.com/gomods/athens/pull/1517/commits/12dba10897b22c735db29b02d50fd64a85bba2ea

pkg/middleware/validation.go
doddi: https://github.com/gomods/athens/pull/1517/commits/12dba10897b22c735db29b02d50fd64a85bba2ea

doddi:
@marwan-at-work I dont appear to have the ability to merge. Could you do it on my behalf?

pkg/middleware/validation.go
marwan-at-work: Ah right, thanks!

marwan-at-work:
@doddi my comment was mainly for one more maintainer to take a look before merging :) 

cc: @arschles

pkg/middleware/validation.go
```diff
@@ -45,31 +49,50 @@ func NewValidationMiddleware(validatorHook string) mux.MiddlewareFunc {
 	}
 }
 
+func maybeLogValidationReason(context context.Context, message string, mod string, version string) {
+	if len(message) > 0 {
+		entry := log.EntryFromContext(context)
+		entry.Warnf("error validating %s@%s %s", mod, version, message)
```
geototti21: I am not very familiar with the codebase yet, but does it not make more sense to log this at `info` level? usually warning means that this can be potential harmfull forthe application 🤔

pkg/middleware/validation.go
```diff
@@ -45,31 +49,50 @@ func NewValidationMiddleware(validatorHook string) mux.MiddlewareFunc {
...
+		return maybeReadResponseReason(resp), nil
+	case http.StatusForbidden:
+		return maybeReadResponseReason(resp), nil
 	default:
-		return false, errors.E(op, "Unexpected status code ", resp.StatusCode)
+		return ValidationResponse{Valid: false}, errors.E(op, "Unexpected status code ", resp.StatusCode)
 	}
 }
+
+func maybeReadResponseReason(resp *http.Response) ValidationResponse {
```
geototti21: 💡 rename to something like `validationResponseFromRequest`?or something similar

pkg/middleware/validation.go
doddi: I think a warning is sufficient given that if the component was allowed to be retrieved it could have harmful implications to the client, i.e. it is not a valid package (for whatever reason)

geototti21: make sense now :)

pkg/middleware/validation.go
```diff
@@ -45,31 +49,50 @@ func NewValidationMiddleware(validatorHook string) mux.MiddlewareFunc {
...
+	}
+}
+
 type validationParams struct {
 	Module  string
 	Version string
 }
 
-func validate(hook, mod, ver string) (bool, error) {
+type ValidationResponse struct {
```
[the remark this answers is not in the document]
robjloranger: I agree here

pkg/middleware/validation.go
robjloranger: I like this change as well

pkg/middleware/validation.go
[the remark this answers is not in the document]
doddi: https://github.com/gomods/athens/pull/1517/commits/85dc8ac212290e6041826c44a737c0d68d3f85c5

pkg/middleware/validation.go
doddi: https://github.com/gomods/athens/pull/1517/commits/f578bf9872c7ff65b8682e8734320725d8154ebf

marwan-at-work:
Thank you @doddi 🙏

Issue #1521, opened 2020-01-10
Always Use Request Logger


marwan-at-work:
One of the main confusions is that GoEnv disabled our Request Logger Middleware if the GoEnv=production. 

That is mainly because we just fmt.Println the log line where we should be using our proper logger. 

We should make it so that the middleware is always in use, and if an Athens user wants to not see these logs, then they can just set their LogLevel to something higher than INFO 

In a related note: maybe the Request Logger Middleware should always log DEBUG instead of INFO?

PR #1522, opened 2020-01-10
always use request logging middleware

marwan-at-work:
Fixes https://github.com/gomods/athens/issues/1521

Also adds a new dev formatter so things look prettier than Logrus'es TextFormatter...I'm tempted to make it even more fancy but for now it works.

### DIFF
--- a/pkg/log/format.go
+++ b/pkg/log/format.go
@@ -1,6 +1,13 @@
 package log
 
 import (
+	"bytes"
+	"fmt"
+	sort "sort"
+	"strings"
+	"time"
+
+	"github.com/fatih/color"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/log/format.go
```diff
@@ -15,7 +22,44 @@ func getGCPFormatter() logrus.Formatter {
...
+}
+
+type devFormatter struct{}
+
+func (devFormatter) Format(e *logrus.Entry) ([]byte, error) {
+	var buf bytes.Buffer
+	var sprintf func(format string, a ...interface{}) string
+	switch e.Level {
+	case logrus.DebugLevel:
+		sprintf = color.New(0xffccc).Sprintf
```
robjloranger: I'm just curious what color this is

marwan-at-work: It's light grey. Good point though @robjloranger, I'll make a constant and give it the color name :)

PR #1524, opened 2020-01-13
Fixing the link to the download config docs from the deprecated filter docs

arschles:

The filter docs (now deprecated) had a broken link to the download mode documentation

**How is the fix applied?**

I fixed the link

N/A


arschles:
Since this is a docs fix, I'm merging right away

PR #1527, opened 2020-01-16
Adds detailed docker usage instructions

### LINKED ISSUE #1526: Include Documentation about our Dev Releases
On every commit, we push a new image to docker hub. This means that people can lock their Athens deployment to a commit sha without worrying about it changing from under their feet (i.e. Canary builds) as well as without having to wait for our official releases. 

The dev releases are kept in a separate docker hub repository here: https://hub.docker.com/r/gomods/athens-dev/tags

However, this is not mentioned anywhere, and it would be great to include in our docs, maybe a new section called "Docker" under Install Athens, or just on the home page.

twexler:

Adds documentation on Athens' docker images

**How is the fix applied?**

Added documentation on Athens' docker images, moves some documentation from the walkthrough to the new page.

Fixes #1526

### DIFF
--- a/docs/content/install/using-docker.md
+++ b/docs/content/install/using-docker.md
@@ -1,0 +1,70 @@
+---
+title: Using the Athens Docker images
+description: Information about Athens' Docker images
+weight: 1
+---
+
+Whether setting Athens up using [Kubernetes](./install-on-kubernetes) or using the [Walkthrough](/Walkthrough). You'll most likely be using one of the images that the Athens projects produces. This document details what images are available, and has a recap from the Walkthrough of how to use them on their own.
+
+---
+
+## Available Docker images
+
+The Athens project produces two docker images, available via [Docker Hub](https://hub.docker.com/) 
+
+1. A release version as `gomods/athens`, each tag corresponds with an Athens [release](https://github.com/gomods/athens/releases), e.g. `v0.7.1`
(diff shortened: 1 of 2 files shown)


docs/content/install/using-docker.md
```diff
@@ -0,0 +1,70 @@
...
+Whether setting Athens up using [Kubernetes](./install-on-kubernetes) or using the [Walkthrough](/Walkthrough), you'll most likely be using one of the images that the Athens project produces. This document details what images are available, and has a recap from the Walkthrough of how to use them on their own.
+
+---
+
+## Available Docker images
+
+The Athens project produces two docker images, available via [Docker Hub](https://hub.docker.com/) 
+
+1. A release version as [`gomods/athens`](https://hub.docker.com/gomods/athens), each tag corresponds with an Athens [release](https://github.com/gomods/athens/releases), e.g. `v0.7.1`
+2. A tip version, as [`gomods/athens-dev`](https://hub.docker.com/r/gomods/athens-dev), tagged with every commit to `master`, e.g. `1573339`
```
marwan-at-work: I think it's worth adding point `3` here (or addition to the first point since it's the same docker hub repo) which is 

- a `canary` version that is frequently mutated to be the latest commit from master.  To install run `docker pull gomods/athens:canary`

twexler: Is there some sort of schedule for `canary` releases?

marwan-at-work: @twexler on every merge to master

twexler: Done.

PR #1533, opened 2020-02-03
Add forcepathstyle config for s3

vapod:

Right now you cant use path style url with s3 storage

**How is the fix applied?**

Add ForcePathStyle to S3Config


vapod:
Guys, there is a bug in your master branch. I copied it by the first commit and resolved by the last one. 
Environment variable AWS_ENDPOINT is not injecting because of typo in s3 config struct - 'evnconfig'.

chriscoffee:
LGTM, weird that we missed this one. You'd have thought virtual hosting buckets was something that would be a cool thing to use with athens.

PR #1535, opened 2020-02-04
Add installation instructions for BOSH

s4heid:

Documentation for installing Athens on BOSH.

**How is the fix applied?**

The documentation will be appended to the `docs/content` folder.

This PR adds an installation guide, which briefly describes how to deploy an Athens Proxy with BOSH (https://bosh.io).

/cc @arschles


docs/content/install/install-on-bosh.md
```diff
@@ -0,0 +1,95 @@
+---
+title: Install Athens with BOSH
+description: Installing an Athens Instance with BOSH
+weight: 11
```
arschles: @s4heid I think this should go above the ["Managing private repos with .netrc files"](https://docs.gomods.io/install/managing-private-repos/) document, so the weight would need to be below 10. Are you ok with that?

docs/content/install/install-on-bosh.md
```diff
@@ -0,0 +1,95 @@
...
+    --vars-file=credentials.yml manifests/athens.yml
+```
+
+This will deploy a single Athens instance in the `athens` deployment with a persistent disk of 1024MB. The IP address of that instance can be obtained with
+
+```console
+bosh -d athens instances
+```
+
+which is useful for targeting Athens, e.g. with the `GOPROXY` variable.
```
arschles: @s4heid could you add a link to https://docs.gomods.io/try-out/, since that page explains how to use `GOPROXY`?

docs/content/install/install-on-bosh.md
s4heid: absolutely, I didn't know there is a limit. I changed it to 8, is that okay?

docs/content/install/install-on-bosh.md
s4heid: totally agree. I added a link to the try-out document, thanks.

s4heid:
@arschles I applied the requested changes, can you have look?

arschles:
Thanks again for this @s4heid !!

PR #1542, opened 2020-02-11
Starting for for encryption feature for issue # 1539 

### LINKED ISSUE #1539: Add option to set encryption key for storage
I have two use cases where data needs to be:
1. "encrypted at rest" (for file storage backend)
2. "encrypted in the runtime/process" even though TLS encrypts HTTP traffic and even though some storage providers may provide encryption at rest (ex Amazon's S3), storage providers still have my plaintext data in their servers and I just trusting providers is not a sufficient level of security

Provide an `ATHENS_STORAGE_ENCRYPTION_KEY` option that if set will wrap a `storage.Backend` and handle encrypting and decrypting of data.

Write an encryption backend pkg that wraps `storage.Backend` and created my own proxy app that uses most of the `gomods/athens/pkg/*` internals. I've done this, but I think writing a PR back to `gomods/athens` with the encryption would be a better solution.

wburningham:

- Adding storage/encryption pkg that wraps a `storage.Backend` to encrypt and decrypt data.
- Adding a `StorageEncryptionKey` config option to set an encryption key and enable encryption.

Fixes #1539

### DIFF
--- a/cmd/proxy/actions/app.go
+++ b/cmd/proxy/actions/app.go
@@ -28,8 +28,15 @@ func App(conf *config.Config) (http.Handler, error) {
 	ENV := conf.GoEnv
 	store, err := GetStorage(conf.StorageType, conf.Storage, conf.TimeoutDuration())
 	if err != nil {
-		err = fmt.Errorf("error getting storage configuration (%s)", err)
-		return nil, err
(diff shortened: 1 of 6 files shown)


arschles:
@wburningham Since you put "starting" in the title, I didn't do a thorough review - just quickly skimmed the code. It looks like a good start. Would you like a complete review now?

PR #1544, opened 2020-02-16
\#1500 - prototype for using testcontainers-go for dependencies

### LINKED ISSUE #1500: Is it worth using testcontainers to set up our unit, integration or e2e tests?
[Testcontainers](https://github.com/testcontainers/testcontainers-go) is a convenience library to set up containers from Go code. We could use it to set up pre-requisite services like Redis, Mongo, Minio, etc...

ClaytonNorthey92:

This is meant to be a prototype for using [testcontainers](https://github.com/testcontainers/testcontainers-go) for dependencies such as redis, minio, and mongo.

**How is the fix applied?**
In the tests, rather than checking if an env variable is set to run tests against these dependencies, I start docker containers with testcontainers in the tests themselves

This prototype was mentioned in #1500 , this should not be merged (yet), it is just a proof of concept.

Fixes #1500

_Edit by @arschles - added `Fixes` to this PR description so that it closes the issue when it's merged_


e2etests/all_test.go
```diff
@@ -0,0 +1,175 @@
+// +build e2etests
```
ClaytonNorthey92: remove this file

ClaytonNorthey92:
Please do not review this yet, I opened this draft for the prototype to see what I needed to update in CI, it is not ready for review.

.drone.yml
```diff
@@ -63,6 +63,8 @@ steps:
   volumes:
   - name: cache
     path: /go
+  - name: dockersock
```
ClaytonNorthey92: this (the droneio pipeline) is not working, I am not sure containers in the build are given privileges to access the docker daemon, are they?

appveyor.yml
```diff
@@ -1,14 +1,27 @@
...
 test_script:
-  - go test ./...
\ No newline at end of file
+  - docker version # test to make sure docker daemon is running
+  - go test -v ./...
+
+# in case we have dangling images, the testcontainers reaper image does not
+# support windows for now, so we have to skip it
+# https://github.com/testcontainers/testcontainers-go/issues/156
+after_test:
```
ClaytonNorthey92: since we are using ubuntu in this PR, we don't need this any more

bsideup:
cc @gianarb

.drone.yml
arschles: No, I don't believe so. You get access to the daemon in GitHub actions, though (#1460). Would it make sense to go that route?

appveyor.yml
```diff
@@ -1,14 +1,27 @@
+# need to use Ubunut (linux) because the images we use in tests
```
arschles: let's not run the tests in appveyor at all, then. cool?

ClaytonNorthey92: sounds good, was only doing it because I could easily communicate with the Docker daemon and wanted to prove this could work in CI, I will move stuff over to droneio

.drone.yml
ClaytonNorthey92: Let me do a little investigating here 🤔 will get back to you shortly

ClaytonNorthey92: @arschles I have added the workflows directory, based off of your gh-actions branch, here is the latest action running the tests (with testcontainers) https://github.com/ClaytonNorthey92/athens/actions/runs/42204895

ClaytonNorthey92:
@arschles I have added the workflows directory, based off of your gh-actions branch, [here is the latest action](https://github.com/ClaytonNorthey92/athens/actions/runs/42204895 ) running the tests (with testcontainers)

ClaytonNorthey92:
> @ClaytonNorthey92 thanks for this! it's looking very promising so far 🎉. One note - we will need to change some of the local development documentation, the Makefile and the docker compose file - which is also used primarily for local development. Since the tests themselves are starting up all of the service dependencies, we won't need the Makefile/docker compose to do it before running tests.
> 
> cc/ @fedepaol because you were in the original discussion as well

okay I will address this as well @arschles 👍

ClaytonNorthey92:
hey @arschles I have removed the dependencies from the `testunit` service in the `docker-compose.yml` file, though I don't think I want to remove them from the `dev` service.  I think it is still valid to have them all running locally when developing.  plus, when you test, testcontainers will start them mapped to random ports, so they shouldn't clash with any locally running containers

let me know what you think

arschles:
>I have added the workflows directory, based off of your gh-actions branch, here is the latest action running the tests (with testcontainers)

@ClaytonNorthey92 great, I missed that before. We probably need to turn these tests off in drone then. I'd say the best way to do that is with an environment variable that defaults to "on". Then, we can set it to off in drone.

>let me know what you think

what you said makes sense for the docker-compose file. I'm glad that things won't conflict!

>edit: I think I still need to update the action to fail a "check" if the tests fail in the action...I don't think it is doing that now

I was thinking the same thing. When we merge this PR, tests will probably start failing in GH actions because there are some tokens missing that we need for connecting to various cloud services in our tests. I'll be sure to put those into GH actions secrets 👍

ClaytonNorthey92:
> > I have added the workflows directory, based off of your gh-actions branch, here is the latest action running the tests (with testcontainers)
> 
> @ClaytonNorthey92 great, I missed that before. We probably need to turn these tests off in drone then. I'd say the best way to do that is with an environment variable that defaults to "on". Then, we can set it to off in drone.
> 
> > let me know what you think
> 
> what you said makes sense for the docker-compose file. I'm glad that things won't conflict!
> 
> > edit: I think I still need to update the action to fail a "check" if the tests fail in the action...I don't think it is doing that now
> 
> I was thinking the same thing. When we merge this PR, tests will probably start failing in GH actions because there are some tokens missing that we need for connecting to various cloud services in our tests. I'll be sure to put those into GH actions secrets 👍

Hey @arschles thanks for the feedback 👍 I should have time to get back to this later this weekend, just wanted to let you know I saw your comments

ClaytonNorthey92:
hey @arschles 
If you look in the "checks" section, I have added the tests action to be run on `pull_requests` in addition to `push`, so if the tests fail, it will fail the check.

I believe that these are the environment variables we need for cloud services tests?
* GCS_SERVICE_ACCOUNT
* GOOGLE_CLOUD_PROJECT
* ATHENS_AZURE_ACCOUNT_NAME
* ATHENS_AZURE_ACCOUNT_KEY

I notice that in the tests, they are skipped if we do not have the environment variable needed to run them.  Would you be able to add these environment variables to Github actions? Then we can re-run the checks and those tests _should_ run and we can ensure that they pass.

I have also removed testing from the `.drone.yml` file, so all tests are run in Github actions now.  I think I would prefer adding those env variables to actions now, rather than the on/off environment variable in drone and test before merging.  That seems less invasive.  What do you think?

docker-compose.yml
```diff
@@ -22,15 +22,12 @@ services:
       args:
         GOLANG_VERSION: 1.13
     command: ["./scripts/test_unit.sh"]
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock # test this on windows, it might be different
```
ClaytonNorthey92: @arschles do you have a Windows machine?  Could you ensure that the `testunit` service works on Windows via the make command to run unit tests?  If it fails, could you paste the error here?

ClaytonNorthey92: I believe it is `make test-unit-docker`

appveyor.yml
```diff
@@ -1,15 +0,0 @@
...
-  GOPATH: c:\gopath
-  GO111MODULE: on
-  GOPROXY: https://proxy.golang.org
-  SKIP_UNTIL_113: true
-
-stack: go 1.13
-
-test_script:
-  - go test ./...
-
```
arschles: @ClaytonNorthey92 we still need to run at least unit tests on windows. One idea would be to add an env var to turn off the testcontainers tests and set it in appveyor but not in GH actions. What do you think?

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,28 @@
...
+      uses: actions/checkout@v1
+
+    - name: Test
+      run: go test -v ./...
+
+    - name: Build
+      run: go build cmd/proxy/*.go
+      env:
+        GOPROXY: https://proxy.golang.org
+        GO111MODULE: "on"
```
arschles: @ClaytonNorthey92 can you add a newline (with no indent) at the end of this file?

docker-compose.yml
arschles: I can try, but let's try to get testcontainers-less appveyor tests back 😄

arschles:
@ClaytonNorthey92 I haven't forgotten about this. Things got busy this week. I can set up the secrets in GH actions early next week. I hope that's ok! 😄

ClaytonNorthey92:
> @ClaytonNorthey92 I haven't forgotten about this. Things got busy this week. I can set up the secrets in GH actions early next week. I hope that's ok! 😄

@arschles yes of course, I see there are a few other comments here, I was going to address them this weekend, I have been a little too busy to look this week, but I will this weekend

appveyor.yml
ClaytonNorthey92: so the issue here came from trying to start the reaper containers (the containers that run and ensure that the containers we start get killed at the end of the tests), but you can optionally skip the reaper containers.

I think that since, according to appveyor's documentation, they don't preserve any VMs between builds, we can safely skip the reaper in appveyor tests.  So I think we should add that env variable like `TESTCONTAINERS_SKIP_REAPER`, if that is set to `1` then we skip the reaper builds and we should be okay (_I think_)

.github/workflows/integration-tests.yml
ClaytonNorthey92: just did 👍

appveyor.yml
ClaytonNorthey92: I am testing this theory out now...I accidentally got a build stuck debugging...so I am going to wait for that to timeout and fail to keep on testing

ClaytonNorthey92: hey @arschles how do you feel about putting integration tests into their own package? the using a build tag to ignore them by default, then you can pass in the tag to enable them?  that should be able to be done just from the code and the config files and wouldn't require a change to the environment

arschles: Fine with me RE the different package. I'd rather have them enabled by default and pass a tag to disable them. Are you cool with that?

arschles:
@ClaytonNorthey92 Ok, all the GH Actions secrets are set up! Sorry for that delay 😄 

Screen Shot 2020-03-06 at 4 33 23 PM

One problem is that the secrets are not passed to PRs from a fork, for security's sake. The best option I can think of is to check for those env vars in the tests and skip if they're not there.

What do you think?

appveyor.yml
ClaytonNorthey92: yeah that makes sense to me, just need a way to cleanly include/not include them

I can work on this

ClaytonNorthey92:
> One problem is that the secrets are not passed to PRs from a fork, for security's sake. The best option I can think of is to check for those env vars in the tests and skip if they're not there.

yeah that makes sense to me, I will make sure that I log out a warning when we skip though,  sorry I haven't had time to look back at this, I will have time Saturday

appveyor.yml
arschles: That's great! We had an idea in #1500 to not turn off the tests completely, but just to turn off the testcontainers logic based on a tag. I laid it out [here](https://github.com/gomods/athens/issues/1500#issuecomment-597381000)

What do you think about that?

ClaytonNorthey92: just left a comment

ClaytonNorthey92:
hey @arschles I started work on moving integration tests over to their own package...I am starting to not like the idea so much.  The integration tests where they are use a lot of private attributes to test.  For example, the mongo storage, the attribute `db` is private and are used in testing. So in order to move integration tests to their own package, I would have to either: make these public or expose them via public methods.  I have two concerns with this: they would be major changes and I don't know if there are any security risks involved with this.  So I propose we include integration tests based on an environment variable, and each test will check if that variable is set, if not, log out a warning and skip the test.

what do  you think?

arschles:
>So I propose we include integration tests based on an environment variable, and each test will check if that variable is set, if not, log out a warning and skip the test.

@ClaytonNorthey92 that sounds great. Except on the last part - if the variable is not set, can it just continue with the test and assume the service is already running?

ClaytonNorthey92:
> > So I propose we include integration tests based on an environment variable, and each test will check if that variable is set, if not, log out a warning and skip the test.
> 
> @ClaytonNorthey92 that sounds great. Except on the last part - if the variable is not set, can it just continue with the test and assume the service is already running?

Yeah I can do that.  I would also like to make their connection configurations (hosts, ports, etc) configurable via env variables.  So then you could run them however you wanted (docker, just running the process locally, etc.) and set those environment variables so your test could connect to them, however you have them running.

arschles:
@ClaytonNorthey92 sounds great! We'll be around when you're ready 😄 ✌️

ClaytonNorthey92:
hey @arschles I apologize I somewhat abandoned this pull request, ultimately, I couldn't find a great way past [this build error when compiling the project on windows](https://ci.appveyor.com/project/gomods/athens/builds/36571500#L561), the issue has since [been fixed](https://github.com/moby/moby/commit/c3a0a3744636069f43197eb18245aaae89f568e5#diff-31c7821daf0eebb430bf1ac78f405f70e92c63e39de3cb603982d4cd1986be75) and I [opened a pull request to testcontainers-go to ensure they don't use a version that can't compile on windows](https://github.com/testcontainers/testcontainers-go/pull/270).  This _should_ fix the compilation issue that I mentioned, since that dependency is coming from the testcontainers-go project.  

However, even if that gets merged and this PR works and gets merged, I don't think I would have time to support this after it is merged.  If you believe testcontainers-go is still worth pursuing I hope my work here will help jump-start whoever picks it up next.  If I can help at all, please reach out 👍

arschles:
@ClaytonNorthey92 thanks for your reply here - and no worries at all! 😄

Issue #1387, opened 2019-09-19
SingleFlight.Redis endpoint does not support setting passwords


ZHRen:
**Describe the bug**
SingleFlight.Redis endpoint does not support setting passwords. I have tried "redis://:password@host:6379", but it didn't work. 

**Error Message**
2019-09-19 11:52:37.207681 I | error adding proxy routes (NOAUTH Authentication required.)
or
2019-09-19 14:42:06.649731 I | error adding proxy routes (dial tcp: address redis://:passrord@host:6379: too many colons in address)

**Expected behavior**
SingleFlight.Redis endpoint support setting passwords.

**Environment (please complete the following information):**
 - OS: Linux 64bit
 - Go version : 1.13
 - Proxy version : 0.6.0
 - Storage (fs/mongodb/s3 etc.) : disk

arschles:
@ZHRen thanks for this catch! I think the best fix for doing this is adding a `[SingleFlight.Redis.Password]` field to the `config.toml` file and a corresponding `ATHENS_REDIS_PASSWORD` environment variable override, and then using that variable in https://github.com/gomods/athens/blob/master/pkg/stash/with_redis.go#L17.

Would you be up for submitting a PR for that? No pressure at all to do so - let me know and I'll do it if not.

PR #1545, opened 2020-02-17
add config options for redis password

mplachter:

SingleFlight.Redis endpoint does not support setting passwords

**How is the fix applied?**

Added **[SingleFlight.Redis.Password]** to config.toml and Global Variable **ATHENS_REDIS_PASSWORD** to allow when making the redis connection to pass the password if defined if it wasn't defined it is treated as an empty string which redis client accepts as a connection without a password.

Fixes https://github.com/gomods/athens/issues/1387


pkg/stash/with_redis.go
```diff
@@ -14,11 +14,12 @@ import (
...
 // using an redis cluster. If it cannot connect, it will return an error.
-func WithRedisLock(endpoint string, checker storage.Checker) (Wrapper, error) {
+func WithRedisLock(endpoint string, password string, checker storage.Checker) (Wrapper, error) {
 	const op errors.Op = "stash.WithRedisLock"
 	client := redis.NewClient(&redis.Options{
-		Network: "tcp",
-		Addr:    endpoint,
+		Network:  "tcp",
+		Addr:     endpoint,
+		Password: password,
```
arschles: @mplachter if there is no password set in the redis server config, will passing `""` here be ok?

Sidenote: I'm not sure if we test this case in `with_redis_test.go` below, but we can add that in a follow-up

mplachter: Hey @arschles,

No problem happy to help :)

So i'm pretty sure this is being testing in `with_redis_test.go` already as there is no env variable set for `ATHENS_REDIS_PASSWORD`, this should cause os.Getenv to return an empty string which is passed into WithRedisLock.

What still needs to be tested is the ability to test with `password` working correctly,  the docker `redis` image does not have a straight forward interface for setting this password, instead they support copying a redis [conf](http://download.redis.io/redis-stable/redis.conf) over to the container.

```
FROM redis
COPY redis.conf /usr/local/etc/redis/redis.conf
CMD [ "redis-server", "/usr/local/etc/redis/redis.conf" ]
```

I do not mind doing this work just would like to know where you prefer the config to be stored.

arschles: @mplachter I missed that about `with_redis_test.go` - thanks for setting me straight 😄 

Thanks for volunteering to expand the tests for password, too. You can put the `redis.conf` file in the `test/` directory. And if you'd like to get this merged, it's fine to do it in a follow-up PR.

Just let me know what you'd like to do

mplachter: Hey @arschles lets add it in the same PR.

mplachter:
Thanks @marwan-at-work, the tests are done will be adding them to this PR shortly.

mplachter:
New Tests added

pkg/stash/with_redis_test.go
```diff
@@ -49,6 +50,63 @@ func TestWithRedisLock(t *testing.T) {
...
+	strg, err := mem.NewStorage()
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = WithRedisLock(endpoint, password, strg)
+	if err == nil {
+		t.Fatal("Expected Connection Error")
+	}
+
+	if err.Error() != "NOAUTH Authentication required." {
```
marwan-at-work: Is there a Redis guarantee that this message will always stay the same? Maybe we can type assert to whatever client error and check something more concrete like an error number or a specific error type.

test/redis.conf
```diff
@@ -0,0 +1,32 @@
+# Redis2 testing configuration file.
```
marwan-at-work: do we need all of these variables? Can it possibly just be the `requirepass` field?

mplachter: I'll try to limit yo just `requirepass` and `port`

pkg/stash/with_redis_test.go
mplachter: This is a good idea, let me check if we can do something other `string matching`

.drone.yml
```diff
@@ -154,6 +156,13 @@ services:
   image: redis
   ports:
     - 6379
+- name: redis2
```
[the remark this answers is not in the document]
mplachter: yep. Can rename to `protected_redis`

pkg/stash/with_redis_test.go
arschles: looks like there's no guaranteed error specified in the docs either, so +1 from me on string matching

mplachter: Yeah i double checked. It has a strange `cmd` handler for the result as well.

test/redis.conf
mplachter: Was able to limit the redis.conf to a minimalistic conf.

.drone.yml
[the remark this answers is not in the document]
mplachter: went with `protectedredis` for Docker DNS reasons but all testing `ENV` variables read `PROTECTED_REDIS`

mplachter:
FYI fixing some `drone` stuff thought it was like `docker` but it isn't :) first time using it give me a bit to fix.

marwan-at-work:
@mplachter no problem, take your time. Here are the docs for Drone https://docs.drone.io/
If you need any help with it let me know.

.drone.yml
```diff
@@ -154,6 +156,20 @@ services:
...
+- name: protectedredis
+  image: redis
+  ports:
+    - 6380
+  volumes:
+    - name: RedisConf
+      path: "/usr/local/etc/redis/redis.conf"
+  commands:
+    - "redis-server /usr/local/etc/redis/redis.conf"
+volumes:
```
marwan-at-work: this needs to be added to be moved to the bottom because now `athens-proxy` is being considered as part of the volume :)

.drone.yml
```diff
@@ -167,6 +179,10 @@ services:
     ATHENS_STORAGE_GCP_BUCKET: "athens_ci_build"
     ATHENS_STORAGE_GCP_JSON_KEY:
       from_secret: GCS_SERVICE_ACCOUNT
+volumes:
+- name: RedisConf
+  host:
```
marwan-at-work: `host` is disallowed in Drone because it's a security risk to expose the underlying machine: https://docs.drone.io/pipeline/docker/syntax/volumes/

You'll need to use a temporary volume if you can https://docs.drone.io/pipeline/docker/syntax/volumes/temporary/

.drone.yml
mplachter: Yeah its very strange to be honest.... I'm just testing some stuff since the `drone` documentation is meh when it comes to mounting `host volumes`.....

mplachter: But it also doesn't like `./test` it wants a full path....

marwan-at-work: It wouldn't work anyway given their highlighted doc: 

>This setting is only available to trusted repositories, since mounting host machine volumes is a security risk.

Also note that each image in the pipeline gets its own volume where the WORKDIR is the repository root dir, so you can just immediately use that without having to volume anything, as far as i know.

.drone.yml
mplachter: Not sure how that will help me mount the `redis.conf` from the `./test` directory of the `repo`.

marwan-at-work: @mplatcher try having no volumes at all, and just point the command to `test/redis.conf` as the working directory should be the same. 

If Drone doesn't give the volume to a `service` we can try to have it as a detached step: https://docs.drone.io/pipeline/docker/syntax/services/

Finally, if you'd like to disable test so we can merge feel free to do so. As a last resort, we can publish our own image on docker/gomods that already has the config file :)

.drone.yml
mplachter: This i didn't know... I don't think they really have that documented anywhere when looking at volume or command stuff :/

.drone.yml
mplachter: @marwan-at-work thanks so much didn't know they mount the root volume of the project to each service. that was super easy once I knew that.

mplachter:
so the commits got kind of crazy. I'm going to rebase to squash really quick.... I have met my next `enemy` drone :D... Just kidding its actually a really cool CI solution.

.drone.yml
marwan-at-work: @mplachter no worries, let me know when you're all set and I'll merge. Thanks for all the updates!

pkg/stash/with_redis_test.go
arschles: Unfortunate, but that's what we're left with. I don't think string matching will be materially better unfortunately.

If the test breaks because the error message changed, I don't think this will be that hard to track down.

PR #1551, opened 2020-02-21
Implementing an external storage driver

### LINKED ISSUE #1110: Plugin architecture for storages

### LINKED ISSUE #1130: Implement storage.Backend for Artifactory storage

arschles:

Over time, requests have come in for a storage driver for "X". Over time, the Athens codebase and config has grown as we've added more drivers

**How is the fix applied?**

I've added a new "external" storage driver that uses a [gRPC](https://grpc.io) API to communicate with any external server for storage. This is so that anybody can generate gRPC stubs for their language of choice, build their own storage driver, and then configure Athens to use it. This can be done without compiling additional code into Athens.

We have had considerable discussion and PRs on this topic (external storage drivers) over time, and I am hoping that this PR can move us toward a solution 😄 

Closes #1353
Fixes #1110 
Fixes #1130
Fixes #1131


arschles:
Ref #1539, where the discussion of storage drivers came up most recently.

cc/ @marwan-at-work @wburningham since you've been in discussion on this issue
cc/ @Kunde21 @marpio @tylerchr since you've all been in discussions about this topic sometime in the past

proto/external_storage.proto
```diff
@@ -0,0 +1,52 @@
+syntax = "proto3";
+// purposely leaving out delete. that is up to individual external
+// storage server operators to do on their end
+
+service ProxyStorage {
+
+    rpc GetModule (ModuleMetadata) returns (ModuleData) {}
```
xytan0056: Not an expert in proto. But does this mean `@info` or `@mod` will download `zip_file` too?

proto/external_storage.proto
```diff
@@ -0,0 +1,52 @@
...
+    // this is redundant for GetModuleCatalogResponse but not for 
+    // GetModule and others. Implementations can return this and Athens
+    // may choose to ignore it.
+    string version = 2;
+}
+
+message ModuleData {
+    ModuleMetadata metadata = 1;
+    string mod_file = 2;
+    string info = 3;
```
xytan0056: Should info be a json?
```
message ModuleInfo {
    string version = 1;
    string timstamp =2;
}
```

proto/external_storage.proto
Kunde21: Yes, that matched the interface, at the time (before #406 landed).  
This service definition would need to be changed to match the storage interface.

arschles: I think that Athens never downloads module@version metadata without also downloading the zip file. That said, does it make more sense to split this int `GetModuleZip` and `GetModuleMetadata`?

proto/external_storage.proto
arschles: Indeed it should. Thanks @xytan0056 !

arschles:
@Kunde21 @xytan0056 I updated the interface according to your comments. Do you mind taking another look?

pkg/storage/external/external_storage.pb.go
```diff
@@ -0,0 +1,555 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
geototti21: 💡 Should we add a CI check that `diffs` the commited auto-generated with the one that is getting generated by the proto file on CI. So this way we know that auto-generate code that we commit is up to date with what is expected?

marwan-at-work:
@arschles I see this is still in draft, lemme know if you want me to give this a review now or later :)

pkg/storage/external/external_storage.pb.go
arschles: @geototti21 good idea, yes I think we should

arschles:
@marwan-at-work yes please! I'd really like your input on the interface

proto/external_storage.proto
```diff
@@ -0,0 +1,72 @@
...
+// ModuleInfo is the JSON metadata that gets returned from
+// the @v/v1.2.3.info endpoint
+message ModuleInfo {
+    string timestamp = 1;
+    string version = 2;
+}
+
+// ModuleZip is the zip file of a module's source code
+message ModuleZip {
+    bytes zip_file = 1;
```
marwan-at-work: Wouldn't this be a big performance problem? Go by default loads 10 dependencies concurrently as far as I recall, and if you have multiple go clients, this can quickly become a problem. 

Would be interesting to test it against a couple of concurrent `go mod download`s and see how it performs.

proto/external_storage.proto
```diff
@@ -0,0 +1,72 @@
...
+// storage server operators to do on their end
+
+service ProxyStorage {
+
+    rpc GetModuleModfile (ModuleMetadata) returns (ModuleMod) {}
+    rpc GetModuleInfo (ModuleMetadata) returns (ModuleInfo) {}
+    rpc GetModuleZip (ModuleMetadata) returns (ModuleZip) {}
+    rpc SaveModule (ModuleData) returns (ModuleSaveResponse) {}
+    rpc ListVersions (ListVersionsRequest) returns (ListVersionsResponse) {}
+    rpc CheckModuleAndVersion(ModuleMetadata) returns (CheckModuleAndVersionResponse) {}
```
marwan-at-work: just my 2cents, i think this was a design flaw from the good old days, a nicer way to keep the interface small is to just interpret a `NotFound`  error as "not exists" :)

proto/external_storage.proto
```diff
@@ -0,0 +1,72 @@
+syntax = "proto3";
+// purposely leaving out delete. that is up to individual external
+// storage server operators to do on their end
+
+service ProxyStorage {
+
+    rpc GetModuleModfile (ModuleMetadata) returns (ModuleMod) {}
```
marwan-at-work: I think `GetModuleModfile` is a little verbose, maybe `Modfile` or `GetModfile`? Same for the methods below.

proto/external_storage.proto
arschles: @marwan-at-work if the problem you're talking about is how many concurrent calls to `GetZip` Athens is making, I think we can use `GoGetWorkers` and the accompanying machinery to limit the number of times we hit that endpoint. The other side could also decide how many concurrent calls that it wants to serve concurrently.

I don't think we get any worse from a memory usage perspective if we limit concurrent calls to the `GetZip` like we do with `go mod download`. Streaming definitely doesn't hurt us though.

proto/external_storage.proto
arschles: good point, updated

proto/external_storage.proto
arschles: good point! I reduced redundancy in this and all the other functions

arschles:
@marwan-at-work I have some answers 😄 

For (1), I chose gRPC because its code generation (for servers) & versioning come for free. If we end up streaming per https://github.com/gomods/athens/pull/1551#discussion_r385474989, then that will be another reason (we wouldn't need to roll a websocket solution ourselves). There are other code generation options, but gRPC is the most widely used and supported that I know of. Second to gRPC would be Swagger/OpenAPI, which Kubernetes uses and, from my understanding, has had to hack a bit to get it to work for its purposes.

I haven't seen any major cloud load balancers that don't support HTTP/2. If someone has on-prem LB's that don't, we can include instructions in our docs on how to use [grpc-web](https://github.com/grpc/grpc-web) with their server. 

For (2), the major reason I didn't roll a custom solution is because gRPC more widely supported and understood than our homegrown HTTP client/server would be. And at the risk of sounding redundant, the code generation will end up saving us a lot of time when people start building servers 😄

proto/external_storage.proto
marwan-at-work: > if the problem you're talking about is how many concurrent calls to GetZip Athens is making

That's just one of the problems :) I think the biggest one is the fact that we'd have to completely load the module zip in memory no its way in and on its way out. Even if we limit save/get operations, then the Go command will time out because it would simply take too long to try and process less than 10 gets at a time.

marwan-at-work:
@arschles 

> we wouldn't need to roll a websocket solution ourselves 

We wouldn't need to do websockets for streaming. You can still stream with pure http in the same way that the Go Download Protocol does for zips, and we can use multi-part for sending files up. 

Definitely don't want you to change this PR though to completely get rid of gRPC, we can potentially jam on both ideas and see which one provides a better user experience :) 

My hope is that users wouldn't have to learn something completely new to them just in order to implement a storage for Athens. So what we can also do, whether we go gRPC or http, is provide some helper functions where we'd hide away the transport itself.

arschles:
Long response ahead, sorry in advance 😄 

>You can still stream with pure http in the same way that the Go Download Protocol does for zips

Oh, I misunderstood your suggestion. I'd be cool with doing the the as-is HTTP chunked streaming as we do now, yes!

I'm also cool with streaming server -> client with gRPC (the [docs for client streaming](https://grpc.io/docs/tutorials/basic/go/#client) show roughly how we could do it) if we go this route.

>My hope is that users wouldn't have to learn something completely new to them just in order to implement a storage for Athens

I'm with you there. I'll ditch gRPC in an instant if it ends up being hard on server authors (I am going to be one of those authors right after this gets merged 😉)

I just auto-generated the server stub and checked it into `./pkg/storage/external_storage` package. At the moment, you'd have to follow the ["how to write a gRPC server"](https://grpc.io/docs/tutorials/basic/go/#server) docs to run your server. They're pretty well written but verbose. We could mostly replace those docs with a few paragraphs on our site and some helpers. Writing a server would then boil down to implementing an interface and then calling a function.

The downside of all of this is that gRPC is fairly easy to learn and the tooling ecosystem is decent, but at the end of the day, it's still new underlying technology for most people. Obviously I think the tradeoff is worth it. 

@Kunde21 @tylerchr @marpio @xytan0056 @geototti21 you have all either built alternative backends for Athens, or are involved with this topic in some other way. Would any of you be up for giving feedback on how comfortable you'd be writing and running your own gRPC-based Athens backend?

> we can potentially jam on both ideas and see which one provides a better user experience :)

I'm cool with that. Would you be up with PRing something small showing what you're thinking for the HTTP protocol?

I appreciate you all talking this through with me! 💚

marwan-at-work:
@arschles here ya go: https://github.com/gomods/athens/pull/1570/files

PR #1554, opened 2020-02-24
Adds redis sentinel support

### LINKED ISSUE #1553: SingleFlight should support Redis Sentinel
In some environments (including my current employer's) another group/team provides Redis, but only via redis-sentinel. Athens doesn't currently support connecting to redis via a sentinel 

Add another SingleFlight provider `redis-sentinel`

None.

twexler:

Redis sentinel support

**How is the fix applied?**

Adds a `WithRedisSentinelLock` to `stash` package, adds various configuration bits needed.

Fixes #1553


twexler:
Thanks @arschles!

Right now, I'm actually in progress of upgrading the redis module to v7 to support authenticated sentinels, this PR may expand in scope a bit. I'll keep it updated!

pkg/stash/with_redis_sentinel.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// WithRedisSentinelLock returns a distributed singleflight
+// with a redis cluster that utilizes sentinel for quorum and failover
+func WithRedisSentinelLock(endpoints []string, master, password string, checker storage.Checker) (Wrapper, error) {
+	const op errors.Op = "stash.WithRedisSentinelLock"
+	// The redis client constructor does not return an error when no endpoints
+	// are provided, so we check for ourselves.
+	if len(endpoints) == 0 {
+		return nil, ErrNoEndpoints
```
marwan-at-work: This error is marked as exported but doesn't get used anywhere. I'd rather keep it private until there's a use case to make it public :) 

Also, let's just keep our error objects uniform across the database by replacing the instantiation with `errors.E(op, "no endpoints provided")`

pkg/stash/with_redis_sentinel.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+var (
+	// ErrNoEndpoints is returned when no endpoints were provided
+	ErrNoEndpoints = errs.New("no endpoints provided")
+)
+
+// WithRedisSentinelLock returns a distributed singleflight
+// with a redis cluster that utilizes sentinel for quorum and failover
+func WithRedisSentinelLock(endpoints []string, master, password string, checker storage.Checker) (Wrapper, error) {
```
twexler: This could alternatively be compacted into `WithRedisLock`, but I didn't want to add _too_ much complexity. Thoughts?

config.dev.toml
```diff
@@ -290,6 +290,17 @@ SingleFlightType = "memory"
         # TODO(marwan): enable multiple endpoints for redis clusters.
         # Env override: ATHENS_REDIS_ENDPOINT
         Endpoint = "127.0.0.1:6379"
+    [SingleFlight.RedisSentinel]
```
marwan-at-work: Looks like the options here are the same except for `MasterName` -- would it be possible to just re-use SingleFlight.Redis and add the MasterName field there? If MasterName exists, then use `NewFailoverClient` otherwise just use `NewClient` 

In the documentation here, we'd mention this logic. 

This would keep our large configuration file from becoming larger. 

But if there's a problem with this definitely let me know. 

Thanks

marwan-at-work: I see `Endpoint` is `Endpoints`, maybe we can combine those and make it backwards incompatible..it would get rid of the TODO I put there :)

pkg/stash/with_redis_sentinel.go
twexler: Bad habit of making the error exported, I'll resolve that. I wasn't aware that `errors.E()` accepted a string though, need to RTFM.

config.dev.toml
twexler: When it comes to accepting multiple endpoints on a non-sentinel redis setup, what would the expected behavior be? Looking at `go-redis`' implementation, I don't think having multiple endpoints would satisfy the need of having a consistent lock at least based on the implementation of `redis.ClusterClient`

cmd/proxy/actions/app_proxy.go
```diff
@@ -133,6 +133,16 @@ func getSingleFlight(c *config.Config, checker storage.Checker) (stash.Wrapper,
 			return nil, fmt.Errorf("Redis config must be present")
 		}
 		return stash.WithRedisLock(c.SingleFlight.Redis.Endpoint, checker)
+	case "redis-sentinel":
+		if c.SingleFlight == nil || c.SingleFlight.Redis == nil {
+			return nil, fmt.Errorf("Redis config must be present")
```
arschles: Could you mention the names of both the variables in `config.dev.toml` so people know what they need to set? 

(I know we don't get this detailed in many of our error messages, but I'm trying to make things more helpful)

pkg/stash/with_redis_sentinel.go
arschles: I like that you've separated sentinel from redis itself. sentinel is treated differently enough in the redis docs that I think we should treat it differently here from a docs and code perspective.

pkg/stash/with_redis_sentinel.go
arschles: @twexler in case you didn't find it, `errors.E()` can take a variety of args after the `Op`, including a string. I most commonly use a string myself, and I'm pretty sure it's most widely used that way. Here's the code that shows the list of things it accepts: https://github.com/gomods/athens/blob/71c57ef7eeed364c77cb25aec89654b5518a5be4/pkg/errors/errors.go#L93-L104

config.dev.toml
arschles: Coming from a docs / usability perspective, I think that we should keep sentinel config separate from plain old redis config. Even though configs are similar, the underlying systems are different, so I think we should reflect that as much as possible.

pkg/stash/with_redis_sentinel.go
twexler: I did find it, but thanks for the pointer! I've got the change for this locally, still want to resolve the conversation about config/sentinel below.

config.dev.toml
twexler: @arschles's point is spot on, too. Regardless, since I'm tooling around in here, it may be beneficial either here, or in a quick-follow up to resolve @marwan-at-work's `TODO`...but there's a pretty big implementation detail that needs to be sussed out, as mentioned in my comment.

cmd/proxy/actions/app_proxy.go
```diff
@@ -133,6 +133,16 @@ func getSingleFlight(c *config.Config, checker storage.Checker) (stash.Wrapper,
 			return nil, fmt.Errorf("Redis config must be present")
 		}
 		return stash.WithRedisLock(c.SingleFlight.Redis.Endpoint, checker)
+	case "redis-sentinel":
+		if c.SingleFlight == nil || c.SingleFlight.Redis == nil {
```
marwan-at-work: I think you meant to use `RedisSentinel` here

config.dev.toml
marwan-at-work: I'm happy either way, but we should just be mindful of adding more configs because it will very hard to remove them :) 

I'm not familiar with Redis Sentinel, @arschles how is it a different system? It looks like the client object is exactly the same.

arschles: @marwan-at-work I bet @twexler can fill in more info, but last I used it, sentinel was a cluster of servers that cooperate. You can then issue a write or read to any of them, and they will load balance and failover.

Your point about removing configs is very well taken, and I am reconsidering my point on that. I wonder if it's understandable enough to merge the configs & keep backward compatibility like you said, and then document how to use sentinel _really_ well 😄

arschles:
@twexler @marwan-at-work I think we should decide if we want to make sentinel support configurable as part of the existing redis config, or a new config section in the TOML. I thought it over and I vote for a separate config (even though it's the same client in the code).

It's more work for us on almost all fronts, but I believe it'll be harder to write docs, but easier to document. And I think a separate documentation section for sentinel vs. non-sentinel will be easier to understand for users.

Can you 👍 on this if you share my vote, and 👎 on this if you want to merge the configs?

twexler:
I'm okay with keeping it separate, but in that case...let me take the docs question on in this PR as well, so we can have full agreement on how to document this

arschles:
sounds good @twexler

docs/content/configuration/storage.md
```diff
@@ -335,3 +335,94 @@ It assumes that you already have the following:
             # Name of container in the blob storage
             # Env override: ATHENS_AZURE_CONTAINER_NAME
             ContainerName = "MY_AZURE_BLOB_CONTAINER_NAME"
+
+## Running multiple Athens pointed at the same storage
+
+Athens has the abilit to run concurrently pointed at the same storage medium, using
```
chriscoffee: s/abilit/ability

docs/content/configuration/storage.md
```diff
@@ -335,3 +335,94 @@ It assumes that you already have the following:
...
+#### Direct connection to redis
+
+Using a direct connection to redis is simple, and only requires a single `redis-server`.
+You can also optionally specify a password to connect to the redis server with
+
+  SingleFlightType = "redis"
+
+  [SingleFlight]
+      [SingleFlight.Redis]
+          # Endpoint is the redis endpoint for the single flgiht mechanism
```
chriscoffee: s/flgiht/flight

docs/content/configuration/storage.md
```diff
@@ -335,3 +335,94 @@ It assumes that you already have the following:
...
+          # Env override: ATHENS_REDIS_PASSWORD
+          Password = ""
+
+#### Connecting to redis via redis sentinel
+
+**NOTE**: redis-sentinel requires a working knowledge of redis and is not recommended for
+everyone.
+
+redis sentinel is a high-availability set up for redis, it provides automated monitoring, replication,
+failover and configuration of multiple redis servers in a master-follower setup. It is more
```
chriscoffee: leader/follower rather than master/follower so we're not mixing leader/follower, master/slave (though to be honest either is fine just a question given might trigger master/slave issues)

PR #1558, opened 2020-02-26
Update builds for Go 1.14

twexler:

Our CI process should use Go 1.14

**How is the fix applied?**

Updated `.drone.yml` et. al. for Go 1.14

Fixes #


chriscoffee:
>What’s the worst that can happen
**Ted Wexler**

LGTM 👍

marwan-at-work:
we should probs also update our Dockerfiles: 

https://github.com/gomods/athens/blob/master/Dockerfile.test
https://github.com/gomods/athens/blob/master/cmd/proxy/Dockerfile
https://github.com/gomods/athens/blob/master/scripts/build-image/Dockerfile

PR #1561, opened 2020-02-27
fix issue with jaeger apiVersion

thienbao92:
When running Helm according to documentation, There's an error
```
Error: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "Deployment" in version "extensions/v1beta1"
```

## How is the fix applied?
`charts/athens-proxy/templates/jaeger-deploy.yaml`

Update apiVersion from  `extensions/v1beta1` to `apps/v1`

Fixes #


gnanderson:
Hi @arschles deprecated `extensions/v1beta1` for Deployment kind was removed in 1.16, `apps/v1` is available since 1.9 so I believe that satisfies the minimum 1.10 supported version - documented here:

https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

DrPsychick:
@arschles this can be closed as it's already in master.

PR #1562, opened 2020-03-03
Remove chart version from deployments and services selectors

rabadin:
Label selectors are immutable; having the chart version part of the label selectors means that upgrades require a full recreation of the deployments and services instead of a simple
update.
This branch removes the chart version from the selectors to allow smooth upgrades.

With the current code chart upgrades (without using `--force`) break with (this is an upgrade from 0.3.3 to 0.4.3):
```
Upgrading charts/vendor/athens-proxy
UPGRADE FAILED
ROLLING BACK
Error: Deployment.apps "athens-proxy-release" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"chart":"athens-proxy-0.4.3", "release":"athens-proxy-release", "app":"athens-proxy-release"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

See https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates for details.

See an example of a high-quality chart (prometheus): https://github.com/helm/charts/blob/master/stable/prometheus/templates/_helpers.tpl#L19-L22 / https://github.com/helm/charts/blob/master/stable/prometheus/templates/_helpers.tpl#L74-L77

### DIFF
--- a/charts/athens-proxy/templates/deployment.yaml
+++ b/charts/athens-proxy/templates/deployment.yaml
@@ -12,7 +12,6 @@ spec:
   selector:
     matchLabels:
       app: {{ template "fullname" . }}
-      chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
       release: "{{ .Release.Name }}"
   template:
     metadata:
--- a/charts/athens-proxy/Chart.yaml
+++ b/charts/athens-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: athens-proxy
-version: 0.4.3
+version: 0.4.4
 appVersion: 0.7.0
 description: The proxy server for Go modules
 icon: https://raw.githubusercontent.com/gomods/athens/master/docs/static/banner.png
--- a/charts/athens-proxy/templates/jaeger-deploy.yaml
+++ b/charts/athens-proxy/templates/jaeger-deploy.yaml
@@ -13,7 +13,6 @@ spec:
   selector:
     matchLabels:
       app: {{ template "fullname" . }}-jaeger
-      chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
       release: "{{ .Release.Name }}"
   template:
     metadata:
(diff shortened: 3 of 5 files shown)


marwan-at-work:
cc @arschles  -- I'm not familiar with Helm :)

charts/athens-proxy/templates/deployment.yaml
```diff
@@ -12,7 +12,6 @@ spec:
   selector:
     matchLabels:
       app: {{ template "fullname" . }}
-      chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
```
arschles: @rabadin I don't recall the reason we put this in to begin with. to be conservative, is it worth leaving in the chart name?

arschles (changes requested):
@rabadin thanks so much for this 😄. Just one question and it's ready to go

🚀 💚

charts/athens-proxy/templates/deployment.yaml
rabadin: We could leave the chart name, yes. As long as it doesn't change.
That being said, looking at a bunch of charts from `git@github.com:helm/charts.git` it seems the practice is to have the selector use `app` and `release` so I suggest following this pattern and remove `chart` altogether like I'd doing here.

rabadin: @arschles You okay with the removal of the `chart` bit in the selectors?

arschles: >@arschles You okay with the removal of the chart bit in the selectors?

@rabadin yes, that's ok with me. thanks for the explanation!

arschles (approved):
thanks for the explanation @rabadin, and the PR too! :shipit: 🎉 🚀 💚

CarAuto34:
Ty

PR #1563, opened 2020-03-03
Adding "Powered by Netlify" text

arschles:

We have to pay for the Netlify service powering docs.gomods.io

## How is the fix applied?

I am going to apply for the [open source plan](https://www.netlify.com/legal/open-source-policy/), and it requires that we add a powered by link on the page. This PR adds it

Fixes #


chriscoffee:
Although @marwan-at-work's point is totally valid. The emails about netlify needing $45 are daily and this solves the initial problem.

If we want to move to using already existing infra then I'm happy to make a ticket for someone to solve.

arschles:
@marwan-at-work it was about $40/month and it started in December iirc. Netlify changed their pricing. Anyway, Netlify is way easier to deal with than just about anything so I'd like to stick with it, if we can get it for free

marwan-at-work:
@arschles i see, so Netlify OSS *is* free?

arschles:
@marwan-at-work sorry for the delay. Yes, Netlify OSS is free 😄

PR #1569, opened 2020-03-10
"sumdbPoxy" ==> "sumdbProxy"

arschles:

The function to create the new checksum DB proxy was called
`sumdbPoxy`. It should be called `sumdbProxy`

## How is the fix applied?

I've renamed the function and its usage everywhere.

None


arschles:
@marwan-at-work if this looks cool, can you merge after netlify is done?

PR #1570, opened 2020-03-11
WIP - comparing external storage to grpc

marwan-at-work:
Things to note: 
We re-used the download protocol itself for all of the endpoints except save/delete. 

A super interesting fact: by adding those two endpoints to Athens, Athens itself can become an external storage server :meta:

Even more interesting, this would also enable direct uploads to Athens (registry style)


pkg/storage/external/server.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		zip, err := strg.Zip(r.Context(), params.Module, params.Version)
+		if err != nil {
+			http.Error(w, err.Error(), errors.Kind(err))
+			return
+		}
+		defer zip.Close()
+		io.Copy(w, zip)
+	})
+	r.HandleFunc("/{module:.+}/@v/{version}.upload", func(w http.ResponseWriter, r *http.Request) {
```
arschles: where is this called on the client? I see `.save` but not `.upload`

pkg/storage/external/client.go
```diff
@@ -0,0 +1,185 @@
...
+			err = nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+func (s *service) Save(ctx context.Context, module, version string, mod []byte, zip io.Reader, info []byte) error {
+	const op errors.Op = "external.Save"
+	url := s.url + "/" + module + "/@v/" + version + ".save"
```
arschles: why not make this a `POST /module/@v/version`? same with deleting below when it gets implemented (`DELETE /module/@v/version`)

pkg/storage/external/server.go
marwan-at-work: Yep, that's a typo :) -- we should stick with either `.save` or `.upload` - any preference?

pkg/storage/external/client.go
marwan-at-work: @arschles i think it's better to be explicit and to take the `.ext` pattern. For example the download protocol explicitly does `GET /module/@v/list` and not just `GET /module` -- and

marwan-at-work:
closing in favor of https://github.com/gomods/athens/pull/1587

PR #1577, opened 2020-03-17
Sample pr template change

arschles:

The section for how to write `Fixes #` was unclear. Specifically, when I wrote "issue", that might not have been very clear to people not as familiar with GitHub.

## How is the fix applied?

I added clarity to the commented above the "What GitHub issue(s) does this PR fix or close?" section


arschles:
@twexler would you mind merging this? I don't want to merge my own PR 😄

Issue #1578, opened 2020-03-18
[Proposal] Remove storage.Checker from storage.Backend


marwan-at-work:
The storage.Backend interface has an `Exists` method which is part of the storage.Checker interface. 

This is used to check if a module exists. However, the Go command itself uses the `Info` endpoint to check if a module exists, it even calls it [Stat](https://github.com/golang/go/blob/971f8a2f9a5beb0473f82d7299613c86d2b4a5b9/src/cmd/go/internal/modfetch/proxy.go#L354)

There are a couple of good reasons to remove `Exists` from our storage API: 

1. Keeping the interface smaller makes implementing new storages easier. 
2. With the external storage that will land soon, we should make it as easy as possible for people to integrate their own storage solutions. 
3. The Info endpoint will always be a small json blob that should not be much of an overhead to use instead of the Exists call. 

Storage implementations right now can still use Exists to validate for whatever logic they need, but they certainly can be optimized to just return KindNotFound instead of making extra calls.

arschles:
@marwan-at-work sounds fine to me

I definitely like #1580 better than #1579 (looks like that is the alternative) because external storages can opt into using `Exists` if that that's more efficient for them. I wrote a few more details about that in https://github.com/gomods/athens/pull/1579#pullrequestreview-377128503

PR #1579, opened 2020-03-18
pkg/storage: remove Exists from exported Backend interface

marwan-at-work:
Prerequisite to External Storage

Fixes https://github.com/gomods/athens/issues/1578

### DIFF
--- a/pkg/stash/stasher.go
+++ b/pkg/stash/stasher.go
@@ -58,12 +58,12 @@ func (s *stasher) Stash(ctx context.Context, mod, ver string) (string, error) {
 	}
 	defer v.Zip.Close()
 	if v.Semver != ver {
-		exists, err := s.storage.Exists(ctx, mod, v.Semver)
-		if err != nil {
-			return "", errors.E(op, err)
-		}
-		if exists {
+		_, err = s.storage.Info(ctx, mod, v.Semver)
+		if err == nil {
 			return v.Semver, nil
+		}
+		if !errors.Is(err, errors.KindNotFound) {
+			return "", errors.E(op, err)
 		}
 	}
 	err = s.storage.Save(ctx, mod, v.Semver, v.Mod, v.Zip, v.Info)
--- a/pkg/stash/stasher_test.go
+++ b/pkg/stash/stasher_test.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/gomods/athens/pkg/errors"
 	"github.com/gomods/athens/pkg/storage"
 )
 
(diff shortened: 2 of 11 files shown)


pkg/stash/with_redis.go
```diff
@@ -57,13 +57,13 @@ func (s *redisLock) Stash(ctx context.Context, mod, ver string) (newVer string,
...
-	ok, err := s.checker.Exists(ctx, mod, ver)
-	if err != nil {
-		return ver, errors.E(op, err)
-	}
-	if ok {
+	_, err = s.getter.Info(ctx, mod, ver)
+	if err == nil {
 		return ver, nil
 	}
+	if errors.Is(err, errors.KindNotFound) {
```
marwan-at-work:
```suggestion
	if !errors.Is(err, errors.KindNotFound) {
```

pkg/stash/stasher.go
```diff
@@ -58,13 +58,13 @@ func (s *stasher) Stash(ctx context.Context, mod, ver string) (string, error) {
...
-			return "", errors.E(op, err)
-		}
-		if exists {
+		_, err = s.storage.Info(ctx, mod, v.Semver)
+		if err == nil {
 			return v.Semver, nil
 		}
+		if !errors.Is(err, errors.KindNotFound) {
+			return "", errors.E(op, err)
+		}
```
arschles: @marwan-at-work can you remind me why we never had to check for other errors?

pkg/stash/stasher_test.go
```diff
@@ -90,9 +91,12 @@ func (ms *mockStorage) Save(ctx context.Context, module, version string, mod []b
 	return nil
 }
 
-func (ms *mockStorage) Exists(ctx context.Context, mod, ver string) (bool, error) {
+func (ms *mockStorage) Info(ctx context.Context, mod, ver string) ([]byte, error) {
 	ms.existsCalled = true
-	return ms.existsResponse, nil
+	if ms.existsResponse {
+		return []byte{'e', 'x', 'i', 's', 't', 's'}, nil
+	}
```
arschles: can we change these to `infoCalled` and `infoResponse` respectively? seems important to change the names to indicate that `Info` was called to test existence

pkg/stash/stasher.go
marwan-at-work: @arschles there are no other errors, just not found and 500..and the not found is not really smart either thanks to `go mod download` not parsing VCS errors (which i understand is tricky)

marwan-at-work:
@arschles 

>Also, sorry for the dumb question, but how is mockStorage a storage.Backend? I only see Save and Info defined on it

Because mockStorage embeds the storage.Backend interface, it doesn't need to declare all the methods, just the ones we need to use.

marwan-at-work:
Closing the issue in favor of the alternative

PR #1581, opened 2020-03-18
Move SingleFlight.Redis.Password back under SingleFlight.Redis in config

twexler:

I goofed in #1545, this fixes that goof

## How is the fix applied?

Moved the Password field for `SingleFlight.Redis` back to where it was in the example config


marwan-at-work:
@twexler @arschles 
I imagine this warrants a patch release?

arschles:
@marwan-at-work yup, it does

Issue #1416, opened 2019-10-09
404 from @latest and @v/list when pre-filling the disk cache


brian73286:
**Describe the bug**
I get 404 from @v/list and @latest endpoints for modules when using athens on a computer that has never cached internet resources, after following instructions for pre-filling the disk cache. 

**Error Message**
When I GET $HOST:$PORT/github.com/julienschmidt/httprouter/@v/list my client receives 404 and kubectl logs <POD_NAME> shows:
<pre>
```console
time="2019-10-09T12:34:56Z" level=info msg="exit status 1: go list -m: github.com/julienschmidt/httprouter@latest: git ls-remote -q origin in /tmp/athens<DECIMAL>/pkg/mod/cache/vcs/<HEXADECIMAL>: exit status 128:\n\tfatal: unable to access 'https://github.com/julienschmidt/httprouter/': Could not resolve host: github.com\n" http-method=GET http-path=/github.com/julienschmidt/httprouter/@v/list http-url=/github.com/julienschmidt/httprouter/@v/list kind="Not Found" module= operation=download.ListHandler ops="[download.ListHandler pool.List protocol.List vcsLister.List]" version=
```
</pre>
When I GET $HOST:$PORT/github.com/julienschmidt/httprouter/@latest my client receives 404 and kubectl logs <POD_NAME> shows:
<pre>
```console
time="2019-10-09T12:34:56Z" level=info msg="exit status 1: go list -m: github.com/julienschmidt/httprouter@latest: git ls-remote -q origin in /tmp/athens<DECIMAL>/pkg/mod/cache/vcs/<HEXADECIMAL>: exit status 128:\n\tfatal: unable to access 'https://github.com/julienschmidt/httprouter/': Could not resolve host: github.com\n" http-method=GET http-path=/github.com/julienschmidt/httprouter/@latest http-url=/github.com/julienschmidt/httprouter/@latest kind="Not Found" module= operation=download.LatestHandler ops="[download.LatestHandler pool.Latest protocol.Latest vcsLister.List]" version=
```
</pre>

**To Reproduce**
Using a computer that has never connected to the internet, follow the instructions at https://github.com/gomods/athens/pull/1414 as they were on 2019-10-08 using the v0.2.0 tag of github.com/plexsystems/pacmod to pack up the v1.3.0 tag of github.com/julienschmidt/httprouter for athens. That creates 3 files: `go.mod`, `v1.3.0.info`, `v1.3.0.zip`.
[...]

PR #1583, opened 2020-03-20
Fixing pacmod command

arschles:

The `pacmod pack` command on the [pre-filling docs](https://docs.gomods.io/configuration/prefill-disk-cache/) was wrong

## How is the fix applied?

I fixed it

Ref https://github.com/gomods/athens/issues/1416, because that issue is related to a problem related to using `pacmod` to "side-load" disk storage

### DIFF
--- a/docs/content/configuration/prefill-disk-cache.md
+++ b/docs/content/configuration/prefill-disk-cache.md
@@ -25,7 +25,7 @@ The `source.zip` file has a specific directory structure and the `$VERSION.info`
 To install the `pacmod` tool, run `go get` like this:
 
 ```console
-$ go get github.com/plexsystems/pacmod
+$ go get github.com/plexsystems/pacmod@v0.4.0
 ```
 
 This command will install the `pacmod` binary to your `$GOPATH/bin/pacmod` directory, so make sure that is in your `$PATH`.
--- a/docs/content/configuration/prefill-disk-cache.md
+++ b/docs/content/configuration/prefill-disk-cache.md
@@ -45,7 +45,7 @@ $ export VERSION="v1.0.0"
 Next, navigate to the top-level directory of the module source code, and run `pacmod` like this:
 
 ```console
-$ pacmod pack $VERSION .
+$ pacmod pack github.com/my/module $VERSION .
 ```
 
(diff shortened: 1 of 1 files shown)


marwan-at-work:
I'd say it's best to completely remove documentation on pacmod or fix it upstream :)

arschles:
@marwan-at-work what's wrong with it?

marwan-at-work:
@arschles it does not guarantee consistent checksums as it does not properly use the zip constructor: https://github.com/plexsystems/pacmod/blob/master/pack/pack.go

It must use https://pkg.go.dev/golang.org/x/mod@v0.2.0/zip?tab=doc

arschles:
Ah, didn't realize that. I created https://github.com/plexsystems/pacmod/issues/4 to ask about fixing it upstream.

I know this method does work for at least some modules. Until we fix it upstream, are you ok with putting a notice at the top indicating that it's an "alpha" (or otherwise unreliable) feature?

jpreese:
v0.4.0 now uses the https://pkg.go.dev/golang.org/x/mod@v0.2.0/zip?tab=doc package for archiving the Go module!

arschles:
@jpreese that's great to hear. I've updated the docs to specify the version. @marwan-at-work do you mind taking another look?

DrPsychick:
`pacmod` either also needs to be updated regularly and maintained, or this guide should be updated.

I'm curious to know how many people out there use Athens in a "offline" scenario, to better judge if we should consider putting effort into this. @jpreese are you still using it in an offline scenario?

jpreese:
> `pacmod` either also needs to be updated regularly and maintained, or this guide should be updated.
> 
> I'm curious to know how many people out there use Athens in a "offline" scenario, to better judge if we should consider putting effort into this. @jpreese are you still using it in an offline scenario?

I'm not personally, no. Are there updates needed for `pacmod` ?

DrPsychick:
> I'm not personally, no. Are there updates needed for `pacmod` ?

I would argue, that setting up dependabot (or similar) to keep it usable with recent go versions & dependencies would be the minimum. Can be done within a week, I'd say.

DrPsychick:
for the meantime, I'll schedule this small fix for the next patch.

Issue #1584, opened 2020-03-20
Pin the alpine base image to a specific version


arschles:
**Is your feature request related to a problem? Please describe.**

The final Athens image is currently based on just `alpine`. That means that on each build, we're uncertain what version of alpine is being used.

**Describe the solution you'd like**

I'd like a Docker build arg that exposes the alpine image version.

**Describe alternatives you've considered**

None

**Additional context**

@dylan-bourque and I discussed this on slack

PR #1585, opened 2020-03-20
Adding build arg for the alpine version the docker image is based on

arschles:

The final Athens docker image is based on `alpine`, which is the latest. We've had requests to pin it to a specific version

## How is the fix applied?

I've added a Docker build arg and pinned it to 

Fixes https://github.com/gomods/athens/issues/1584

cc/ @dylan-bourque


cmd/proxy/Dockerfile
```diff
@@ -16,7 +16,8 @@ ARG VERSION="unset"
 
 RUN DATE="$(date -u +%Y-%m-%d-%H:%M:%S-%Z)" && GO111MODULE=on CGO_ENABLED=0 GOPROXY="https://proxy.golang.org" go build -ldflags "-X github.com/gomods/athens/pkg/build.version=$VERSION -X github.com/gomods/athens/pkg/build.buildDate=$DATE" -o /bin/athens-proxy ./cmd/proxy
 
-FROM alpine
+ARG ALPINE_VERSION=3.9.5
+FROM alpine:3.9.5
```
dylan-bourque: Shouldn't this line be `FROM alpine:${ALPINE_VERSION}` to make use of the build arg?

cmd/proxy/Dockerfile
```diff
@@ -16,7 +16,8 @@ ARG VERSION="unset"
 
 RUN DATE="$(date -u +%Y-%m-%d-%H:%M:%S-%Z)" && GO111MODULE=on CGO_ENABLED=0 GOPROXY="https://proxy.golang.org" go build -ldflags "-X github.com/gomods/athens/pkg/build.version=$VERSION -X github.com/gomods/athens/pkg/build.buildDate=$DATE" -o /bin/athens-proxy ./cmd/proxy
 
-FROM alpine
+ARG ALPINE_VERSION=3.9.5
```
dsexton: This `ARG` here is unused, however I would just drop it and leave the image tag explicit as you do on the next line `FROM alpine:3.9.5` or `FROM alpine:3.9`. 

Unless you are building a factory for building docker images I would just be explicit :)

cmd/proxy/Dockerfile
arschles: Thanks for that catch @dylan-bourque

cmd/proxy/Dockerfile
twexler: Agreed, this should just be removed.

arschles: @dsexton I'm gonna leave in the arg (after fixing https://github.com/gomods/athens/pull/1585#discussion_r395838521) on the off chance someone wants to do a custom build of the image. Some folks have wanted to do the same for the [go version](https://github.com/gomods/athens/blob/e89cf0feb2155f60ab2b0a808a75c1c84ccf8e34/cmd/proxy/Dockerfile#L8)

marwan-at-work: might as well upgrade it to 3.11?

arschles: @marwan-at-work done, I upgraded to 3.11.5

PR #1587, opened 2020-03-20
pkg/storage: add External implementation

### LINKED ISSUE #1130: Implement storage.Backend for Artifactory storage

marwan-at-work:
This PR adds an http external storage implementation. It re-uses a lot of what we have already and optimizes "save" operations by uploading all 3 files (info/mod/zip) in one endpoint as a piped multi-writer. 

Basic documentation and example is included, but I'll add more docs in follow ups. 

Fixes https://github.com/gomods/athens/issues/1110
Fixes #1353
Fixes #1130
Fixes #1131
Closes #1551

### DIFF
--- a/cmd/proxy/actions/storage.go
+++ b/cmd/proxy/actions/storage.go
@@ -9,6 +9,7 @@ import (
 	"github.com/gomods/athens/pkg/errors"
 	"github.com/gomods/athens/pkg/storage"
 	"github.com/gomods/athens/pkg/storage/azureblob"
+	"github.com/gomods/athens/pkg/storage/external"
 	"github.com/gomods/athens/pkg/storage/fs"
 	"github.com/gomods/athens/pkg/storage/gcp"
 	"github.com/gomods/athens/pkg/storage/mem"
@@ -60,6 +61,12 @@ func GetStorage(storageType string, storageConfig *config.StorageConfig, timeout
 			return nil, errors.E(op, "Invalid AzureBlob Storage Configuration")
 		}
 		return azureblob.New(storageConfig.AzureBlob, timeout)
+	case "external":
+		if storageConfig.External == nil {
+			return nil, errors.E(op, "Invalid External Storage Configuration")
+		}
+		// TODO(marwan-at-work): add client tracing
+		return external.NewClient(storageConfig.External.URL, nil), nil
 	default:
 		return nil, fmt.Errorf("storage type %s is unknown", storageType)
 	}
--- a/pkg/storage/external/client.go
+++ b/pkg/storage/external/client.go
@@ -23,7 +23,7 @@ type service struct {
 // NewClient returns an external storage client
 func NewClient(url string, c *http.Client) storage.Backend {
 	if c == nil {
-		c = http.DefaultClient
+		c = &http.Client{}
 	}
 	url = strings.TrimSuffix(url, "/")
 	return &service{url, c}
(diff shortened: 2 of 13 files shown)


arschles:
@marwan-at-work this PR will close these issues in addition to #1110 :

- #1353
- #1130
- #1131

Can you put those 3 in the OP?

cmd/proxy/actions/storage.go
```diff
@@ -60,6 +61,12 @@ func GetStorage(storageType string, storageConfig *config.StorageConfig, timeout
 			return nil, errors.E(op, "Invalid AzureBlob Storage Configuration")
 		}
 		return azureblob.New(storageConfig.AzureBlob, timeout)
+	case "external":
+		if storageConfig.External == nil {
+			return nil, errors.E(op, "Invalid External Storage Configuration")
+		}
+		// TODO(marwan-at-work): add client tracing
```
arschles: If this is a TODO for later after this PR (totally cool with me), could you make sure there's an issue for it?

pkg/storage/external/client.go
```diff
@@ -0,0 +1,183 @@
...
+type service struct {
+	url string
+	c   *http.Client
+}
+
+// NewClient returns an external storage client
+func NewClient(url string, c *http.Client) storage.Backend {
+	if c == nil {
+		c = http.DefaultClient
+	}
```
arschles: Can you force a specific client, so we are not using the default? It's a well-configured client but I'm always concerned that some other part of Athens will use it later and there will be connection contention

pkg/storage/external/external_test.go
```diff
@@ -0,0 +1,22 @@
...
+func TestExternal(t *testing.T) {
+	strg, err := mem.NewStorage()
+	if err != nil {
+		t.Fatal(err)
+	}
+	handler := NewServer(strg)
+	srv := httptest.NewServer(handler)
+	defer srv.Close()
+	externalStrg := NewClient(srv.URL, nil)
+	clear := strg.(interface{ Clear() error }).Clear
```
arschles: worth having an explicit named `Clearer` (or something) interface?

marwan-at-work: @arschles this method is only defined/used inside `_test.go`  files and you cannot import it if you define it in a `_test.go` file anyway. So I'd keep it this way, but if we find ourselves doing this casting in other places, then it would def be worth it to define it in a none-test file and use it from within tests similar to our compliance package.

marwan-at-work:
@arschles addressed all the comments, let me know if it looks good. thanks!

pkg/storage/external/external_test.go
arschles: What about [here](https://github.com/gomods/athens/pull/1587/files#diff-9ea50d8556888305dac9befb25ed6069R42)?

marwan-at-work: @arschles that is where the method is defined, sorry i meant the interface-casting is only defined/used in _test.go 

The reason that method is defined (and got moved from _test.go) to non-test Go file, is exactly because _test.go files in other packages don't get imported even during tests.

arschles: I got ya, this is good for me then 👍

arschles:
I'm looking forward to this going in! I'll start writing some storage servers after it does 😄 

Let's plan to do some docs after this is merged, and before it's released 🛩 🚀

PR #1588, opened 2020-03-20
go.mod: update/use google cloud storage sub module

marwan-at-work:
Google cloud client has divided up its repository into sub modules, this PR uses the only one we need.


go.mod
```diff
@@ -1,9 +1,10 @@
 module github.com/gomods/athens
 
-go 1.12
+go 1.14
```
marwan-at-work:
```suggestion
go 1.12
```

arschles:
@marwan-at-work are you looking to get this merged before we do the patch release?

marwan-at-work:
@arschles patch releases should be tagged from the release branch after cherry-picking the hot fix, so merging to master should be irrelevant :)

PR #1591, opened 2020-03-23
Enable multiple access modes for PVC

timrcoulson:

For my particular setup, I want Athens to RW to the volume but enable CI workers to access the volumes in RO mode.


charts/athens-proxy/values.yaml
```diff
@@ -48,6 +48,8 @@ storage:
       ## Note if you use disk.persistence.enabled, replicaCount should be set to 1 unless your access mode is ReadWriteMany
       enabled: false
       accessMode: ReadWriteOnce
+      ## Enable multiple access modes
+      accessModes:
```
theomjones: Is empty ok?

timrcoulson:
Realised a better way of solving my problem

arschles:
@timrcoulson if you don't mind, what was the solution?

PR #1595, opened 2020-03-27
Updating the helm chart for release v0.8.1

arschles:
## What is the problem I am trying to address?

I just did the v0.8.1 release and the helm chart is outdated, it still points to the v0.8.0 docker image tag.

## How is the fix applied?

I've updated the values.yaml and the Chart.yaml files according to [these documentation](https://github.com/gomods/athens/blob/master/DEVELOPMENT.md#finishing-up)

## What GitHub issue(s) does this PR fix or close?

N/A

PR #1597, opened 2020-03-29
charts: allow to set deployment update strategy

bclermont:

If [`storage.disk.persistence.enabled`](https://github.com/gomods/athens/blob/6b5b62c0fceb97e8c86816fcf82a93f730a330e3/charts/athens-proxy/values.yaml#L49) is set to `true` and the `accessMode` is not `ReadWriteMany`. the default strategy must be switched to `type: Recreate`. if not it will fail as the other pod is using the PVC.

### DIFF
--- a/charts/athens-proxy/templates/deployment.yaml
+++ b/charts/athens-proxy/templates/deployment.yaml
@@ -9,6 +9,13 @@ metadata:
     heritage: "{{ .Release.Service }}"
 spec:
   replicas: {{ .Values.replicaCount }}
+{{- if .Values.strategy }}
+  strategy:
+{{ toYaml .Values.strategy | indent 4 }}
+  {{- if eq .Values.strategy.type "Recreate" }}
+  rollingUpdate: null
+  {{- end }}
+{{- end }}
   selector:
     matchLabels:
       app: {{ template "fullname" . }}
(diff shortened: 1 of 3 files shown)


charts/athens-proxy/templates/deployment.yaml
```diff
@@ -9,6 +9,13 @@ metadata:
     heritage: "{{ .Release.Service }}"
 spec:
   replicas: {{ .Values.replicaCount }}
+{{- if .Values.strategy }}
+  strategy:
+{{ toYaml .Values.strategy | indent 4 }}
+  {{- if eq .Values.strategy.type "Recreate" }}
```
twexler: This clause seems odd. I assume it's a bit of defensive programming, but if a user is changing the deployment strategy they should probably not be specifying values that aren't compatible.

bclermont: @twexler 

[this is the reason why I did that](https://github.com/helm/charts/commit/663bd1f7484877f4fcfa7ae9a9139cb198d7e659)

if you're still uncomfortable with that, I can remove it

twexler: Ah, okay. This is a common pattern, interesting.

twexler:
@bclermont I'm going to wait to merge this until we merge #1595 (as it also updates this chart). We're working through some issues with Netlify right now and want to make sure those are solved before that merges.

arschles:
@twexler @bclermont I've merged #1595, so after conflicts are resolved this is ok to go. The conflicts will just be to update the `version` field in the chart.yaml

bclermont:
@twexler done

Issue #1596, opened 2020-03-27
Fix the "How to release" section in the contributing docs


arschles:
**Describe the bug**

The release docs have a few issues.

First, they don't have information on how to do patch releases. There should be some things added:

- Code freeze is not required for patch releases
- You should branch from the previous release branch and cherry pick (e.g. `git cherry-pick $COMMIT`) commits from `master` into the new branch
- Before you tag the branch / do the GH release, you should update the helm chart version
- In the [creating a new release](https://github.com/gomods/athens/blob/master/DEVELOPMENT.md#creating-the-new-release-in-github) section, the branch compare link (e.g. `https://github.com/gomods/athens/compare/master...v0.7.2`) should compare between the old branch (not `master`) and the new one
- In the final [Finishing up](https://github.com/gomods/athens/blob/master/DEVELOPMENT.md#finishing-up) section, you should only PR to the `master` branch

**Error Message**

N/A

**To Reproduce**

N/A

**Expected behavior**

N/A

**Additional context**
I ran into these issues when doing the [v0.8.1 patch release](https://github.com/gomods/athens/releases/tag/v0.8.1)

PR #1603, opened 2020-04-13
Updating the release notes

arschles:

The [release notes](https://github.com/gomods/athens/blob/master/DEVELOPMENT.md#for-people-doing-a-release) are incomplete because they don't include specific instructions for patch releases.

## How is the fix applied?

I added instructions that differentiate between patch and minor releases.

Fixes https://github.com/gomods/athens/issues/1596

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -246,46 +246,75 @@ This section is written primarily for maintainers, so we don't forget how to ini
 
 Take a look at our [releases page](https://github.com/gomods/athens/releases). This is how we release official new builds of Athens. Almost all of the release process is automatic.
 
+## Notes about the release number
+
+Before we continue, please be sure you're familiar with our release numbering scheme.
+
+We follow the [semver](https://semver.org) convention for our release numbers. At the moment, we're not at `v1`, so release versions are `v0.x.y`. `x` is the `MINOR` number, and `y` is the `PATCH` number. You'll need to decide which one to update.
+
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -246,50 +246,81 @@ This section is written primarily for maintainers, so we don't forget how to ini
...
-Make sure the Drone job finished, and check in Docker Hub to make sure the new release showed up in the [tags](https://hub.docker.com/r/gomods/athens/tags) section.
+Make sure the Drone CI/CD job finished, and check in Docker Hub to make sure the new release showed up in the [tags](https://hub.docker.com/r/gomods/athens/tags) section.
 
 ## Finishing up
 
 The Drone job will do everything except:
 
 - Tweet out about the new release
-- Update the helm chart
+- Update the helm chart in the `master` branch
```
twexler, line 313: I wonder if this should be done as part of a pre-release commit, that way it also ends up on the release branch

arschles: @twexler works for me. so it would be 

- update helm chart in master
- cut release branch 
- release etc...

is that what you had in mind?

twexler: Yeah, although my one concern there is that without the latest docket image, the helm chart build may fail.

arschles: Same here. That's why I'd prefer to make the releaser to wait until the new docker image is up before they manually submit the PR. Is that still cool with you? Let me know what you think

twexler: Yeah, that’s fine.

PR #1605, opened 2020-04-15
docs: typo for ATHENS_DOWNLOAD_MODE custom

BradErz:
Fixed a missing `:` when configuring `custom:$BASE_64`.

## How is the fix applied?

Change the docs.

Signed-off-by: Bradley Wilson-Hunt <bwilsonhunt@gmail.com>

### DIFF
--- a/docs/content/configuration/download.md
+++ b/docs/content/configuration/download.md
@@ -18,7 +18,7 @@ This document will outline how to use this file - called the download mode file 
 First, once you've created your download mode file, you tell Athens to use it by setting the `DownloadMode` configuration parameter in the `config.toml` file, or setting the `ATHENS_DOWNLOAD_MODE` environment variable. You can set this configuration value to one of two values to tell Athens to use your file:
 
 1. Set its value to `file:$FILE_PATH`, where `$FILE_PATH` is the path to the HCL file
-2. Set its value to `custom$BASE_64` where `$BASE_64` is the base64 encoded HCL file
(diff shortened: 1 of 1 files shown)


docs/content/configuration/download.md
```diff
@@ -72,6 +72,8 @@ If a user has their `GOPROXY` environment variable set with a comma separated li
 
 The last block specifies that any module matching `github.com/pkg/*` (such as `github.com/pkg/errors`) will always redirect the `go` tool to https://gocenter.io. In this case, Athens will never persist the given module to its storage.
 
+_NOTE_: https://pkg.go.dev/path/?tab=doc#Match is used to parse the contents of `download` if you are working with a package that contains `-` then you should escape that character, e.g. `test\\-pkg`
```
marwan-at-work, line 75: Hmm, if this is true, I'd imagine it's more of a bug. 

If I understand correctly, a module named `github.com/pkg/hello-world` must be written as `github.com/pkg/hello\\-world` in the Download File HCL? 

Will definitely give it a test soon

twexler: Yeah, we encountered this today on Slack (gophers/#athens), I haven't tested myself but it seemed a reasonable explanation for what @BradErz was encountering.

BradErz: So i just tried to replicate this again for the past hour and now i can't replicate it at all..

I also added a test to check in the `download/mode.DownloadFile.Match()` but it also passed, maybe i was just being ultimatley special.

I will remove this `__NOTE__`

PR #1606, opened 2020-04-20
pkg/storage/external: remove debug lines

marwan-at-work:


twexler:
Looks like Circle is hung up on something? It looks like it never got the webhook. Might want to try pushing again.

twexler:
looks like the merge got it.

PR #1609, opened 2020-04-27
[WIP] Building a CLI for migrating between storage backends

### LINKED ISSUE #1604: Tool for data migration between storage backend
Hello!

It seems that it is quite hard to migrate athens local files backed to GCS or other cloud storage. Seems that such process requires changing paths and names of Athens objects.

It would be awesome to have some kind of CLI which is able to push data from local storage to cloud one.

Regards

arschles:

It's hard to move modules from one storage backend to another. You have to understand the internals of how Athens lays out zip/mod/info files in each backend and then you have to write a script or manually copy.

## How is the fix applied?

I've started a CLI that can migrate between any 2 storages that Athens supports. It uses built-in catalog, get and set methods, so any new storages automatically become supported by the CLI next time it's built.

Fixes #1604


cmd/migrator/cmd/catalog.go
```diff
@@ -0,0 +1,74 @@
...
+	to storage.Saver,
+) error {
+	token := ""
+	// TODO: parallelize this
+	for {
+		pathParams, newToken, err := cataloger.Catalog(ctx, token, 100)
+		if err != nil {
+			return err
+		}
+		for _, pathParam := range pathParams {
```
marwan-at-work, line 45: moving storages is likely a very timely process, we should do a couple of things: 

1. Parallelize reads/writes 
2. Check if the destination storage has the module@version first,  instead of always overriding it. 
3. Further enhancement: make the migration process resumable, though with 2 it should be good enough i think.

miecio45:
Thanks for this PR! Great job! In my opinion having everything in single binary is good idea and I agree for parallelization as migrating my test repository took quite a long time.

arschles:
@miecio45 I'm glad you like this 😄 

@marwan-at-work regarding the `athens` CLI, adding it to the existing `cmd/proxy` is going to mean a fairly large refactor of `main.go`. Can we save that for a follow-up?

I'll just add it to `cmd/proxy`. is that cool with you?

cmd/migrator/cmd/catalog.go
arschles: @marwan-at-work for (1), I [agree that it should be parallelized ](https://github.com/gomods/athens/pull/1609/files#diff-ef0e160aad5c050733a77e59a54b3807R39). I was waiting to see if the general structure of this looks ok. 

If you're ok with [my suggestion for a follow-up](https://github.com/gomods/athens/pull/1609#issuecomment-627635998), I can continue on with parallelizing and doing reads before writes in here. Let me know ✌️

Issue #1619, opened 2020-05-23
sumdb proxy always returns 404 error with a global path prefix.


elliotmr:
**Describe the bug**
If you set a global path prefix for the athens install using the config file or ATHENS_PATH_PREFIX, the sumdb reverse proxy will never be called. This is because the call to `http.StripPrefix` in `addProxyRoutes()` (app_proxy.go) will fail if the path-prefix does not match the request path. 

https://golang.org/pkg/net/http/#StripPrefix

PR #1620, opened 2020-05-24
Fix sumdb/* paths when config.PathPrefix is set

elliotmr:
Fixes: #1619

http.StripPrefix will look at the entire request path when called,
if we do not include config.PathPrefix then the StripPrefix call
will never receive a valid path from the application and the user
will always get a 404 error.

There were no test where I could easily check this regression so
I also added a few endpoint tests, the last test will fail with
a 404 instead of 403 if this change in not applied.

## How is the fix applied?

Fixes #

### DIFF
--- a/cmd/proxy/actions/app_proxy.go
+++ b/cmd/proxy/actions/app_proxy.go
@@ -47,7 +47,7 @@ func addProxyRoutes(
 		sumHandler := sumdbProxy(sumdbURL, c.NoSumPatterns)
 		pathPrefix := "/sumdb/" + sumdbURL.Host
 		r.PathPrefix(pathPrefix + "/").Handler(
-			http.StripPrefix(strings.TrimRight(c.PathPrefix, "/")+pathPrefix, sumHandler),
+			http.StripPrefix(strings.TrimSuffix(c.PathPrefix, "/")+pathPrefix, sumHandler),
 		)
 	}
 
--- a/cmd/proxy/actions/app_proxy.go
+++ b/cmd/proxy/actions/app_proxy.go
@@ -47,7 +47,7 @@ func addProxyRoutes(
 		sumHandler := sumdbProxy(sumdbURL, c.NoSumPatterns)
 		pathPrefix := "/sumdb/" + sumdbURL.Host
 		r.PathPrefix(pathPrefix + "/").Handler(
-			http.StripPrefix(pathPrefix, sumHandler),
+			http.StripPrefix(strings.TrimRight(c.PathPrefix, "/")+pathPrefix, sumHandler),
 		)
 	}
 
--- a/cmd/proxy/actions/app_proxy_test.go
+++ b/cmd/proxy/actions/app_proxy_test.go
@@ -13,7 +13,6 @@ import (
 	"github.com/gomods/athens/pkg/log"
 	"github.com/gomods/athens/pkg/storage/mem"
 	"github.com/gorilla/mux"
-	testlog "github.com/sirupsen/logrus/hooks/test"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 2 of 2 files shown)


cmd/proxy/actions/app_proxy.go
```diff
@@ -47,7 +47,7 @@ func addProxyRoutes(
 		sumHandler := sumdbProxy(sumdbURL, c.NoSumPatterns)
 		pathPrefix := "/sumdb/" + sumdbURL.Host
 		r.PathPrefix(pathPrefix + "/").Handler(
-			http.StripPrefix(pathPrefix, sumHandler),
+			http.StripPrefix(strings.TrimRight(c.PathPrefix, "/")+pathPrefix, sumHandler),
```
marwan-at-work, line 50: Out of curiosity, did you consider using `strings.TrimSuffix`? If a user has multiple slashes at the end of the PathPrefix, that is either a mistake they should realize themselves or maybe it's on purpose for whatever reason so we might not wanna magically remove all of them and replace them with the one "/". 

But happy to keep this as well 👍 

```suggestion
			http.StripPrefix(strings.TrimSuffix(c.PathPrefix, "/")+pathPrefix, sumHandler),
```

cmd/proxy/actions/app_proxy_test.go
```diff
@@ -0,0 +1,94 @@
...
+	test   func(t *testing.T, resp *http.Response)
+}
+
+func TestProxyRoutes(t *testing.T) {
+	r := mux.NewRouter()
+	s, err := mem.NewStorage()
+	require.NoError(t, err)
+	lggr, hook := testlog.NewNullLogger()
+	require.NotNil(t, hook)
+	l := &log.Logger{Logger: lggr}
```
marwan-at-work, line 34: I believe we already have a NoOp logger you can use. 

```suggestion
	l := log.NoOpLogger()
```

cmd/proxy/actions/app_proxy_test.go
```diff
@@ -0,0 +1,94 @@
...
+	c, err := config.Load("")
+	require.NoError(t, err)
+	c.NoSumPatterns = []string{"*"} // catch all patterns with noSumWrapper
```
marwan-at-work, line 37:
```suggestion
	c.NoSumPatterns = []string{"*"} // catch all patterns with noSumWrapper to ensure the sumdb handler doesn't make a real http request to the sumdb server.
```

I was a bit confused as to why this was here, so the comment should help future code readers realize what the star is for

cmd/proxy/actions/app_proxy_test.go
```diff
@@ -0,0 +1,94 @@
...
+			err := json.NewDecoder(resp.Body).Decode(&details)
+			require.NoError(t, err)
+			assert.EqualValues(t, build.Data(), details)
+		}},
+
+		// Default sumdb is sum.golang.org
+		{"GET", "/sumdb/sum.golang.org/supported", "", func(t *testing.T, resp *http.Response) {
+			assert.Equal(t, http.StatusOK, resp.StatusCode)
+		}},
+		{"GET", "/sumdb/sum.rust-lang.org/supported", "", func(t *testing.T, resp *http.Response) {
```
marwan-at-work, line 73: lol

cmd/proxy/actions/app_proxy_test.go
```diff
@@ -0,0 +1,94 @@
...
+	l := &log.Logger{Logger: lggr}
+	c, err := config.Load("")
+	require.NoError(t, err)
+	c.NoSumPatterns = []string{"*"} // catch all patterns with noSumWrapper
+	c.PathPrefix = "/prefix"
+	subRouter := r.PathPrefix(c.PathPrefix).Subrouter()
+	err = addProxyRoutes(subRouter, s, l, c)
+	require.NoError(t, err)
+
+	baseURL := "https://athens.azurefd.net/prefix"
```
marwan-at-work, line 43:
```suggestion
	baseURL := "https://athens.azurefd.net" + c.PathPrefix
```

This way, someone else reading the code will recognize that `/prefix` is an intentional part of the test, and that if we needed to change it, we'll just have to change it once on line 38

elliotmr:
All changes looked good to me, I used TrimRight instead of TrimSuffix because I forgot that it existed.

PR #1623, opened 2020-05-26
github actions + goreleaser to build and publish binaries

### LINKED ISSUE #1607: Published Binary Distributions with Releases

I'm currently in the process of setting up a homelab system, and as a result of choice of hypervisor (Proxmox), the most efficient way for me to deploy Athens is within an LXC container. The existing Docker images produced during releases are helpful in building scripts to create these, but there's no direct migration tool that I can use to convert them. The one thing that I need to automate the setup is pre-built binaries of Athens.

It would be fantastic to have binary builds available for Windows, macOS, and Linux in addition to the existing Docker images (don't get me wrong—the Docker images are awesome 👍 )

I considered converting the docker image, for which no tool seems to exist. Also, I looked at doing a docker-in-docker-esque thing but with LXC instead... which gets messy. I also looked at extracting the binary from the image, which results in a strange Rube Goldberg machine that still requires Docker. Finally, I built the binaries on my own, which could work (since the `go` toolchain is required anyway), but it's extra steps that would be nice to eliminate.

**Additional context**
None that I can think of now, but please feel free to ask away :)

arothste-blk:

I want to deploy/use athens, not own the build process.

## How is the fix applied?

A little bit of github action grease + goreleaser

Fixes #1607


arothste-blk:
Bueller?

arschles:
Sorry about the delay @arothste-blk. I'll review this today!

arschles (approved):
@arothste-blk we'll see how this works after merge. thanks for this, and sorry for the delay

Issue #1625, opened 2020-06-14
Add client tracing to external storage


marwan-at-work:
The new external storage uses http to make requests to the backend storage layer. 

We should instrument those http calls so that users can have debug bottlenecks in their systems.

PR #1626, opened 2020-06-14
Add client tracing to external storage requests

marwan-at-work:
This PR adds the ochttp Transport plugin into an http client in order to propagate tracing. 

Fixes https://github.com/gomods/athens/issues/1625


marwan-at-work:
@twexler thank you for the review 🙌

PR #1628, opened 2020-06-15
Switching to using the main default branch

arschles:

The previous default branch, `master`, is now called `main`. Although it's the default out of the box, the former is inappropriate and offensive.

## How is the fix applied?

I've run [this tool](https://www.npmjs.com/package/github-default-branch) to do most of the work to switch the default branch. This PR changes some things over - like CI/CD - that the tool couldn't get.

None


twexler:
@marwan-at-work 

> Will git clone https://github.com/gomods/athens know to pull the main branch instead of master?

Yes! This is actually nothing that the client needs to know about on `git clone`, and you can just `git checkout main` if you already have a clone. I think the hardest part of this transition will be the muscle memory!

> Is it okay if we kept the master branch for a little bit of time before removing it entirely just in case CI/CD or other things break?

I'm torn here, I think we should do a hard break so we know right away what breaks, but I also don't like breaking people's workflows.

Issue #1546, opened 2020-02-18
Integrate with pkg.go.dev


marwan-at-work:
Now that pkg.go.dev will be open sourced (more here https://groups.google.com/forum/#!msg/golang-dev/mfiPCtJ1BGU/OZ9cU3SgBgAJ) 

This is a place holder ticket to see if/how we can integrate Athens with it :v:

arschles:
One idea - we could add it to some of our deployment artifacts. For example, you could install the helm chart and it would optionally have a doc server all set up alongside your Athens server.

Obviously depends on if the server is easily packaged up in an image and easily deployed

twexler:
This feels a bit out of scope for Athens, but could be part of a larger initiative to provide non-Google-hosted Go development resources

arschles:
Agreed. If `pkg.go.dev` is like moddoc, you would be able to point it at Athens and it would "just work".

If that is not the case, I hope there would be a way we can make them work together. I would advocate we build the software to do that in a different repo. Or at least in a different binary.

twexler:
I wasn't familiar with `moddoc`, nice work @marwan-at-work! 

That would be the most ideal. It's a bit disappointing `go doc`'s http server was deprecated, as having access to the on-disk cache that Athens uses could also be a great excuse to run a docs server with it.

Issue #1186, opened 2019-04-13
Improve the catalog endpoint format


arschles:
**Is your feature request related to a problem? Please describe.**

Right now, utilities that want to use the `/catalog` endpoint often need to aggregate each module's versions into a list. For example, the `/catalog` endpoint returns this:

```json
{"modules": [
  {
    "module":"github.com/arschles/assert",
    "version": "v1.0.0"
  },
  {
    "module": "github.com/arschles/assert",
    "version: "v1.1.0"
  }
}]
```

Many programs will find it more convenient to get a response similar to this: 

```json
{
    "modules": {
        "github.com/arschles/assert": ["v1.0.0", "v1.1.0"]
    }
}
```

**Describe the solution you'd like**

I'd like us to add a V2 of the catalog API that responds this way. This could possibly be exposed as a `/catalog/v2` endpoint.

**Describe alternatives you've considered**

Clients can always aggregate in their own code

**Additional context**

There will be non-trivial work on the storage systems to achieve this

MaruHyl:
@arschles It looks simple, can I try it?

marpio:
@MaruHyl sure, it's yours!
One thing you should keep in mind is that we have paging in place currently. This means that the last record we're getting from the storage doesn't have to be the last version of that module.
page1:
mod1@v1.00
mod1@v1.0.0
mod2@v2.0.0
Page2:
mod2@v2.1.0
etc.
So we can't just simply group by module the page records.

MaruHyl:
@marpio Thanks!
I will use `module` as the key for sorting and paging. For example, pagesize is 3, then the result is m1(v1,v2..), m2(v1,v2..),m3(v1,v2..).

MaruHyl:
@marpio 
In `fs.CatalogV2`, the token will be `module`, not `module|version`.For example, pagesize is 3, then the result is m1(v1,v2..), m2(v1,v2..),m3(v1,v2..), the next token is m4
In `mongo.CatalogV2`, the token will be `module`, not `ID`, and the next token is m4 too.
...

manugupt1:
Hi,

My preference is not to sort as the order is not guaranteed by any storage
providers. The two storages may not necessarily be the same.


Cheers,

Manu Gupta

marpio:
@MaruHyl
Sorry for not being clear, I was thinking about the object storages (S3,GCS etc.)
[...]

PR #1630, opened 2020-06-22
implement /index endpoint

marwan-at-work:
This is the first of possibly a couple of more PRs to try and integrate pkg.go.dev (pkgsite) with Athens. 

Pkgsite uses https://index.golang.org to discover new module versions so that they are searchable and are parsed ahead of time. More on pkgsite's design [here](https://github.com/golang/pkgsite/blob/master/doc/design.md). 

This PR adds a `/index` endpoint that tries to adhere to the same logic and format in `index.golang.org/index`

To support a /index endpoint we create a new Indexer interface: 

```
type Indexer interface {
	Index(ctx context.Context, mod, ver string) error
	Lines(ctx context.Context, since time.Time, limit int) ([]*Line, error)
}
```

This is different (and separate) from a storage.Backend for one important reason: 

- A storage backend might not store modules in a chronological/append-only way. For example, blob storage stores modules based on path names, and not in an append only fashion. 

Furthermore, a storage backend stores a module's go.mod file and the entire zip while an index backend only cares about indexing a module's metadata. This allows us to add index backends that make sense for indexing and storing metadata but would not make sense for storing large zips, such as SQL stores. 

This PR adds 3 implementations of an index backend: memory, MySQL, and Postgres, defaulting to none to keep backwards compatibility (though ideally, I'd love to have /index be turned on and using in-memory by default so people can play with the endpoint immediately). 

More index backends can be added as seen fit but ideally we'll also add an external index backend similar to pkg/storage/external so users can bring in their own solution without the maintenance overhead. 

Lastly, this PR also adds compliance tests to ensure all implementations comply to the interface expectations but I will probably need to add more test cases, handler tests, and integration tests as follow up PRs.

Updates https://github.com/gomods/athens/issues/1186
Updates https://github.com/gomods/athens/issues/1546

### DIFF
--- a/cmd/proxy/actions/index.go
+++ b/cmd/proxy/actions/index.go
@@ -1,0 +1,52 @@
+package actions
+
+import (
+	"encoding/json"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/index"
+)
+
+// indexHandler implements GET baseURL/index
+func indexHandler(index index.Indexer) http.HandlerFunc {
+	const op errors.Op = "actions.IndexHandler"
+	return func(w http.ResponseWriter, r *http.Request) {
+		var (
+			err   error
+			limit int
+			since time.Time
+		)
+		if limitStr := r.FormValue("limit"); limitStr != "" {
+			limit, err = strconv.Atoi(limitStr)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusBadRequest)
+				return
+			}
+		}
+		if sinceStr := r.FormValue("since"); sinceStr != "" {
+			since, err = time.Parse(time.RFC3339, sinceStr)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusBadRequest)
+				return
+			}
+		}
+		if limit <= 0 {
+			limit = 2000
+		}
+		list, err := index.Lines(r.Context(), since, limit)
+		if err != nil {
+			http.Error(w, err.Error(), 500)
+			return
+		}
+		enc := json.NewEncoder(w)
+		for _, meta := range list {
+			if err = enc.Encode(meta); err != nil {
+				http.Error(w, err.Error(), 500)
+				return
+			}
+		}
+	}
+}
--- a/pkg/index/mysql/mysql.go
+++ b/pkg/index/mysql/mysql.go
@@ -1,0 +1,105 @@
+package mysql
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"time"
+
+	"github.com/go-sql-driver/mysql"
+	"github.com/gomods/athens/pkg/config"
+	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/index"
+)
+
+func New(cfg *config.MySQL) (index.Indexer, error) {
+	dataSource := getMySQLSource(cfg)
+	db, err := sql.Open("mysql", dataSource)
+	if err != nil {
+		return nil, err
+	}
+	if err = db.Ping(); err != nil {
+		return nil, err
+	}
+	_, err = db.Exec(schema)
+	if err != nil {
+		return nil, err
+	}
+	return &indexer{db}, nil
+}
+
+const schema = `
+	CREATE TABLE IF NOT EXISTS indexes(
+	id INT
+		AUTO_INCREMENT
+		PRIMARY KEY
+		COMMENT 'Unique identifier for a module line',
+
+	module VARCHAR(255)
+		NOT NULL
+		COMMENT 'Name of the module',
+
+	version VARCHAR(255)
+		NOT NULL
+		COMMENT 'Name of the module',
+
+	timestamp TIMESTAMP(6)
+		COMMENT 'Date and time when the module was first created',
+
+	INDEX (timestamp),
+	UNIQUE INDEX module_version (module, version)
+	) CHARACTER SET utf8;
+`
+
+type indexer struct {
+	db *sql.DB
+}
+
+func (i *indexer) Index(ctx context.Context, mod, ver string) error {
+	const op errors.Op = "mysql.Index"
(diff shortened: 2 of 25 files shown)


cmd/proxy/actions/index.go
```diff
@@ -0,0 +1,48 @@
...
+		)
+		if limitStr := r.FormValue("limit"); limitStr != "" {
+			limit, err = strconv.Atoi(limitStr)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusBadRequest)
+				return
+			}
+		}
+		if sinceStr := r.FormValue("since"); sinceStr != "" {
+			since, err = time.Parse(time.RFC3339, sinceStr)
```
twexler, line 30: Looks like you set this error, but never check it

cmd/proxy/actions/index.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+		if limit <= 0 {
+			limit = 2000
```
twexler, line 33: Do we want the default limit to be this high? Should the default be configurable?

pkg/index/mysql/mysql.go
```diff
@@ -0,0 +1,106 @@
...
+	UNIQUE INDEX module_version (module, version)
+	) CHARACTER SET utf8;
+`
+
+type indexer struct {
+	db *sql.DB
+}
+
+func (i *indexer) Index(ctx context.Context, mod, ver string) error {
+	const op errors.Op = "sql.Index"
```
twexler, line 60: should the op mention the type of db, rather than just `sql`?

pkg/index/mysql/mysql.go
```diff
@@ -0,0 +1,106 @@
...
+	"context"
+	"database/sql"
+	"fmt"
+	"time"
+
+	"github.com/go-sql-driver/mysql"
+	"github.com/gomods/athens/pkg/config"
+	"github.com/gomods/athens/pkg/errors"
+	"github.com/gomods/athens/pkg/index"
+	_ "github.com/lib/pq"
```
twexler, line 13: this import seems superfluous?

pkg/index/mysql/mysql.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, err
+	}
+	_, err = db.Exec(mysqlTable)
+	if err != nil {
+		return nil, err
+	}
+	return &indexer{db}, nil
+}
+
+const mysqlTable = `
```
twexler, line 32: nit: you called this `schema` in the postgres driver, might be helpful to keep the terminology the same between the two implementations

cmd/proxy/actions/index.go
marwan-at-work: Nice catch, will definitely write handler tests after this PR 😄

cmd/proxy/actions/index.go
marwan-at-work: This default is modeled after the https://index.golang.org mentioned on its home page: 

>'limit': the maximum length of the returned list. Default = 2000, Max = 2000

So far, I don't think we need to make it configurable because of 2000 was a bit too high for someone, they can always just pass a query string like `?limit=500`

pkg/index/mysql/mysql.go
marwan-at-work: It should probably read `mysql.Index`, because the pattern has mostly been `<pkg>.<method>`

pkg/index/mysql/mysql.go
marwan-at-work: This is so that the Postgres driver is registered with the `sql/db` standard library. See the package doc here: https://pkg.go.dev/github.com/lib/pq?tab=doc

I will add a comment on top of the import to help with readability a little bit. Thanks for pointing it out 👍

pkg/index/mysql/mysql.go
marwan-at-work: Yep!

marwan-at-work:
@twexler thank you for the review 🙏 -- All comments should be addressed, let me know if it looks good to you :v:

cmd/proxy/actions/index.go
twexler: Hmm, I'm on the fence. I was wondering if it was a spec-related thing. I think its fine for now.

pkg/index/mysql/mysql.go
twexler: Sure, but this is in the `mysql` package? I don't think it needs to be imported in here.

marwan-at-work: Ohh good call. I had both packages under one at some point while I was developing. Will fix it up

drewwells:
What is needed to populate the index backend? I've enabled the env variable and get a 200 response, but the response is empty.

marwan-at-work:
@drewwells you will need to set 
1. https://github.com/gomods/athens/blob/0c78ebd6308329713a2d2dd9bc007ca9c36d98cc/config.dev.toml#L299
2. As well as https://github.com/gomods/athens/blob/0c78ebd6308329713a2d2dd9bc007ca9c36d98cc/config.dev.toml#L495 if you are not using an in-memory index. 

If that still doesn't work, can you pls open an issue with reproduction steps?

drewwells:
I got it working today, thanks it's great! It appears the pkgsite worker is actually looking for a /poll endpoint. Is there a ticket tracking that feature?

This happens when I click 'Poll Module Index' from the worker dashboard.
```
ERROR: handlePollIndex("/poll"): index.Client.GetVersions(ctx, 0001-01-01 00:00:00 +0000 UTC, 10): decoding JSON: json: cannot unmarshal string into Go value of type internal.IndexVersion
```

marwan-at-work:
@drewwells from what I understand, the pkgsite worker is the server that implements the `/poll` endpoint and not Athens. 

It looks like the above error is happening here: https://github.com/golang/pkgsite/blob/88e7b2bf3152210697fadc36ce66eb8ce0030111/internal/index/index.go#L76

It might be worth running pkgsite locally and reading the entire response body and log out to see what is being returned vs what pkgsite expects.

PR #1633, opened 2020-06-22
Upgrade AWS SDK to v1.32.7

### LINKED ISSUE #1632: AWS SDK upgrade for working with AWS EKS IRSA
**Describe the bug**
I am not sure if this is a bug or a feature request :) Looks like Athens uses AWS SDK v1.15.24 but using IAM Roles for Service Accounts in AWS EKS (Kubernetes) requires at least v1.23.13: https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html.

**Error Message**

/readyz throws a 500 without any log messages but it starts to return a 200 if I use an in memory store instead of AWS S3. So I am assuming this is the issue but happy to provide more logs if it's helpful.

**To Reproduce**
1. Run Athens in EKS with an S3 storage backend using IRSA.

**Expected behavior**
It should support IRSA.

Using the image from the Docker Hub: `gomods/athens:v0.9.0`

sidprak:

My attempt at trying to fix #1632.

## How is the fix applied?

N/A

Fixes #1632

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,10 +13,11 @@ require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/DataDog/datadog-go v0.0.0-20180822151419-281ae9f2d895 // indirect
 	github.com/DataDog/opencensus-go-exporter-datadog v0.0.0-20180917103902-e6c7f767dc57
-	github.com/aws/aws-sdk-go v1.15.24
+	github.com/aws/aws-sdk-go v1.32.7
 	github.com/bsm/redislock v0.4.2
 	github.com/codegangsta/negroni v1.0.0 // indirect
 	github.com/fatih/color v1.7.0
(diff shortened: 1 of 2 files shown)


sidprak:
Thanks for the quick review! After this is merged, do you know if an image will automatically be built and pushed to Docker Hub?

marwan-at-work:
@sidprak yes, we have two image repositories: 

1. [gomods/athens](https://hub.docker.com/r/gomods/athens) for master builds (canary) and releases 
2. [gomods/athens-dev](https://hub.docker.com/r/gomods/athens-dev) for building against every commit in master.

So you should be able to use the canary version or an explicit commit sha. 

It would be great if you can confirm if it works or not in the issue you opened. Thanks!

PR #1638, opened 2020-06-26
Init github actions with helm

marwan-at-work:
Testing github actions with helm charts

We don't completely understand how this `ct` tool is working. Docs for that are here: https://github.com/helm/chart-testing/blob/master/doc/ct_lint.md


chriscoffee:
I've had to do the following:
- Rather than using `with: fetch-depth: 0` it requires `git fetch --prune --unshallow` instead
- Because I bumped the version of kind/helm/helm-actions etc I then had to alter the `apiVersion` for the Deployment resources. I also had to add `apiVersion` to Chart.yaml as this is enforced now.
- I had to add `tag` back in and remove `registry`. Also removed this from `deployment.yaml`
- `{{ default .Chart.appVersion }}` on ln 56 should be `{{ default .Chart.AppVersion }}` so I sorted this.

twexler:
Things to do _after_ this merges:
1. Remove CircleCI integration
2. Remove requirements on CircleCI

PR #1640, opened 2020-06-28
Support downloadURLs with paths

### LINKED ISSUE #1639: downloadURL path is truncated
**Describe the bug**
If a downloadURL with a path such https://internal.domain/repository/gonexus is configured then this is truncated to https://internal.domain when redirecting to the upstream proxy, resulting in failures.

**Expected behavior**
It should be possible to use a downloadURL that has a path.

erwbgy:

If a downloadURL with a path such https://internal.domain/repository/gonexus is configured then this is truncated to https://internal.domain when redirecting to the upstream proxy, resulting in failures.

## How is the fix applied?

In getRedirectURL append the path to the downloadURL path rather than replacing it.

Fixes #1639   

Signed-off-by: Keith Burdis <keith.burdis@gs.com>

### DIFF
--- a/pkg/download/handler.go
+++ b/pkg/download/handler.go
@@ -3,6 +3,7 @@ package download
 import (
 	"net/http"
 	"net/url"
+	"path/filepath"
 
 	"github.com/gomods/athens/pkg/download/mode"
 	"github.com/gomods/athens/pkg/log"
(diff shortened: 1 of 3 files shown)


pkg/download/handler.go
```diff
@@ -60,6 +61,6 @@ func getRedirectURL(base, path string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	url.Path = path
+	url.Path = filepath.Join(url.Path, path)
```
marwan-at-work, line 64: I believe `filepath` will end up joining with backslashes instead of forward slashes, so you might want to use `path.Join` instead of `filepath.Join`

erwbgy: Right you are - good catch.

PR #1646, opened 2020-07-03
add mysql and postgres single flight locks

WillAbides:

This adds single flight implementations for mysql and postgres.

## How is the fix applied?

This uses advisory locks from mysql and postgres as SingleFlight locks.


WillAbides:
I'm having trouble figuring out how appveyor works in order to make mysql run on it.

pkg/stash/with_locker.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func (s *lockerLock) Stash(ctx context.Context, mod, ver string) (string, error) {
+	const op errors.Op = "lockerLock.Stash"
+	ctx, span := observ.StartSpan(ctx, op.String())
+	defer span.End()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	mv := fmt.Sprintf("%s@%s", mod, ver)
+	releaseErrs, err := s.locker.lock(ctx, mv)
```
marwan-at-work, line 54: This interaction of this function API seems a bit weird. We ask it to lock for us, and gives us an immediate error plus an asynchronous error that we have to wait on in a goroutine. 

Why not instead of this, we make `s.locker.lock` take a context (which it already does) but instead return a new context. 

This will mimic a lot of std lib behavior like `context.WithTimeout` and `context.WithCancel`. 

This way, users of this API won't be confused as to how they are supposed to use this function. They can just do this: 

```
ctx, err = s.locker.lock(ctx, mv)
if err != nil { ... }
// pass new ctx that will handle lock cancellation...
```

pkg/stash/with_azureblob.go
```diff
@@ -142,9 +93,10 @@ func (s *azblobLock) acquireLease(ctx context.Context, blobURL azblob.BlockBlobU
 	if err != nil {
 		return "", errors.E(op, err)
 	}
+	ttlSeconds := int32(ttl / time.Second)
 	for {
-		// acquire lease for 15 sec (it's the min value)
-		res, err := blobURL.AcquireLease(tctx, leaseID.String(), 15, azblob.ModifiedAccessConditions{})
+		// acquire lease for 20 sec
```
marwan-at-work, line 98: Given that ttl is passed by the caller, we can't know for sure it's 20 seconds. Also given the previous comment that 15 seconds is the minimum, it might be worth checking whether azure will return an error or whether it implicitly ensures it is 15 seconds. If the former, we wanna make sure we bump it so we don't error out.

pkg/stash/with_etcd.go
```diff
@@ -38,48 +37,41 @@ func WithEtcd(endpoints []string, checker storage.Checker) (Wrapper, error) {
...
-	if ok {
-		return ver, nil
-	}
-	newVer, err = s.stasher.Stash(ctx, mod, ver)
-	if err != nil {
-		return ver, errors.E(op, err)
+	holder := &lockHolder{
+		ttl: ttl,
+		release: func() error {
+			_ = mu.Unlock(ctx)  // don't care about this error because the lock will expire when the session is closed below
```
marwan-at-work, line 65: Out of curiosity, what happens if Etcd is in a bad state and is returning errors we should know about? This looks like it might get swallowed?

pkg/stash/with_locker.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+type locker interface {
+	// lock obtains a lock to be used for single flight.
+	// The lock will be released when ctx is canceled. When the lock is released whether intentionally or not,
+	// releaseErrs will receive an error. If that error is nil, the lock was successfully released as a response to ctx.
+	// Returns:
+	//	 releaseErrs:	A channel that emits an error when the lock is released.
+	//   err: 			Any error creating the lock. If err is non-nil, no lock is created.
+	lock(ctx context.Context, name string) (releaseErrs <-chan error, err error)
```
marwan-at-work, line 26: see my comment below. I think it would be a much easier API to work with if lock's signature was this: 

`lock(ctx context.Context, name string) (ctx context.Context, err error)` 

If we care about what the releaseErr is, we can mimic a mutext behavior by having both a `lock` and a `unlock` method and do the `defer mu.Unlock()`  similar to how the etcd implementation has been doing it. 

Let me know if im missing something :v:

arschles:
@WillAbides I'm gonna close this for now. Feel free to re-open if you're still interested in having these new locks

PR #1647, opened 2020-07-04
demo: move to actions

WillAbides:
After looking at https://github.com/gomods/athens/discussions/1556, I wanted to have a go at moving from circleci and drone onto actions. This is the result of a timeboxed session. I got all the circleci workflows and most of drone moved over. There shouldn't be much of an issue moving the rest of drone over.

In order for the publish workflow to work, credentials for hub.docker.com will need to be added to actions secrets. That will mean that everybody with contributor access to `gomods/athens` will have access to those credentials. That may be a reason to keep the workflows that require credentials on drone.

Adding a Windows integration test on github actions will be challenging. Because you can't mix Linux and Windows runners, it will require finding Windows builds for all the dependencies.


WillAbides:
Maybe I shouldn't have deleted .circleci yet. I hate seeing those  ❌.

DrPsychick:
Closing this as "gone cold" and we have moved to GitHub actions.
Don't be discouraged from contributing to Athens through this experience @WillAbides, a new group of maintainers has joined and is committed to keeping the project alive. Thank you also for the effort you put into this PR.

Issue #1649, opened 2020-07-20
Implement Authentication Propagation


marwan-at-work:
Athens currently supports a couple of types of authentication: 

1. [Basic Authentication](https://github.com/gomods/athens/blob/408fd74a9c95c019264982f020f6eaf31a1eeabe/config.dev.toml#L152): authenticates clients to the Athens server itself. If the user has the basic auth, then he/she can access the Athens server. 

2. [.netrc file](https://github.com/gomods/athens/blob/408fd74a9c95c019264982f020f6eaf31a1eeabe/config.dev.toml#L192): authenticates Athens with upstream modules. This is so that Athens can download and store private modules. 

I propose implementing a third option: 

**Propagated Authentication**: This will allow Athens to accept an Authorization header. However, it will not check it against a static username/password. Instead, it will dynamically create a `.netrc` file with the given authentication and with `machine` set to the requested module hostname. 

This will allow Go developers to provide their own identity for Athens which will then allow them access to their own private modules. 

There are a few things to be aware of: 

1. `go mod download` can only implicitly authenticate with git-clones through the .netrc file in the HOME directory. So we'd have to dynamically create/destroy home dirs for every go mod download that carries an Auth header. If there is enough demand, we can provide a feature in the future that copies certain files from HOME to TempHome (i.e. .gitconfig)

2. Once a Module is retrieved and stored, Athens won't know if the next user who requests it is actually _authorized_ to access it or not. This can be solved by providing a ValidationHook or putting Athens behind an auth proxy that can authorize a auth header against the requested module. Note that a ValidationHook at the moment does not include list/latest, so a proxy is likely more appropriate.

PR #1650, opened 2020-07-20
Allow Athens to Propagate Authentication to Mod Download

marwan-at-work:
This PR allows Athens to create dynamic authentication for "go mod downloads" so that a server can serve multiple modules visible to only those who request it. 

Fixes https://github.com/gomods/athens/issues/1649


marwan-at-work:
@twexler this feature is most likely accompanied by a ValidationHook or an auth proxy, which can do that part of the work for us. 

Do you think it's worth including here as well? Happy to do so :v:

twexler:
> @twexler this feature is most likely accompanied by a ValidationHook or an auth proxy, which can do that part of the work for us.

Right, but since you're using the `.netrc` file any `go mod download` call, this has the potential to leak credentials beyond the `ValidationHook` endpoint...no? In my opinion, users should be able to decide where these credentials are actually sent to rather than sending them everywhere

marwan-at-work:
>Right, but since you're using the .netrc file any go mod download call, this has the potential to leak credentials beyond the ValidationHook endpoint...no?

So each call gets its own $HOME variable with a .netrc file. This means no two calls would ever share the same .netrc file. This ensures one call's credentials is never leaked to another call's go mod download. See [this block](https://github.com/gomods/athens/blob/91a789b83524e4aeb3d850c83f0fe51d19be0110/pkg/module/go_get_fetcher.go#L122:L126) for where it happens for the download part (same with the lister). 

That said, your suggestion is definitely a nice insurance to make sure a user's auth header is never propagated to the wrong Hostname even if the Athens admin has this feature turned on.

@twexler I added the pattern matching to the latest commit. Let me know if it looks good to you, thanks! 🙌

config.dev.toml
```diff
@@ -166,6 +166,35 @@ BasicAuthUser = ""
...
+# Env override: ATHENS_PROPAGATE_AUTH
+PropagateAuth = false
+
+# PropagateAuthPatterns can be used to limit which modules will have
+# the credentials accompanied when calling "go mod download". Note that
+# this is only when a module is not in storage. See PropagateAuth above
+# for more details.
+# 
+# Env override: ATHENS_PROPAGATE_AUTH_PATTERNS
+PropagateAuthPatterns = ["*"]
```
twexler, line 196: I'm not sure if I like the default here, from a security perspective. Perhaps just `["localhost"]`?

pkg/auth/auth.go
```diff
@@ -0,0 +1,64 @@
...
+		return errors.E(op, fmt.Errorf("netrcFromToken: could not write to file: %v", err))
+	}
+	return nil
+}
+
+// WriteTemporaryNETRC writes a netrc file to a temporary directory, returning
+// the directory it was written to.
+func WriteTemporaryNETRC(host, user, password string) (string, error) {
+	const op errors.Op = "auth.WriteTemporaryNETRC"
+	dir, err := ioutil.TempDir("", "netrcp")
```
twexler, line 45: is the `p` here a typo?

pkg/module/all_test.go
```diff
@@ -13,6 +13,9 @@ const (
 	// github.com/NYTimes/gizmo is a example of a path that needs to be encoded so we can cover that case as well
 	repoURI = "github.com/NYTimes/gizmo"
 	version = "v0.1.4"
+
+	privateRepoURI     = "github.com/athens-artifacts/private"
```
twexler, line 17: random, but why is this in a different org?

config.dev.toml
marwan-at-work: @twexler you're right. I removed this whole functionality in favor of static hosts. This should eliminate the possibility of accidentally leaking credentials to any host other host than the pre declared one.

pkg/auth/auth.go
marwan-at-work: Nope, it's short for "netrc path" :v:

pkg/module/all_test.go
marwan-at-work: @twexler https://github.com/athens-artifact is a dedicated organization for testing Athens where we put repositories with an exact state (i.e. number of semver versions, major SIV upgrades, private/public etc) and see how Athens would behave against them. This makes it so that testing against live repositories is predictable because if we tested against 3rd party libraries, they can end up being deleted or mutated in ways that break our tests. Most of the tests against this organization lives here: https://github.com/gomods/athens/blob/main/pkg/download/protocol_test.go
hope this helps :v:

twexler: Gotcha, makes sense. Not _quite_ sure why it should be a different org, but the rest seems logical

config.dev.toml
```diff
@@ -166,7 +166,7 @@ BasicAuthUser = ""
 # Env override: BASIC_AUTH_PASS
 BasicAuthPass = ""
 
-# PropagateAuth, when set to true, will pass the Basic Authentication
+# PropagateAuth, when set to a hostname such as "github.com", will pass the Basic Authentication
```
twexler, line 169:
```suggestion
# PropagateAuthHost, when set to a hostname such as "github.com", will pass the Basic Authentication
```

pkg/module/all_test.go
marwan-at-work: @twexler to keep `github.com/gomods` clean and only containing meaningful repositories for users (and not test artifacts)

PR #1656, opened 2020-07-27
pkg/stash: fix redis failing test

marwan-at-work:
I noticed that the Redis test is failing due to some spacing issues in the error message. 

String matching when it comes to error handling is usually not great, but Redis doesn't provide us with error codes here so we should make the string matching a bit looser as there is no compatibility guarantee.


twexler:
merging because circle shouldn't be breaking things

PR #1658, opened 2020-08-03
Implement an optional /config endpoint that outputs the used configuration.

### LINKED ISSUE #1503: Add an endpoint that will output Athens configuration

Athens has many configuration values and it's easy to get them mixed up. We are taking several measures to make it easier to make sure you don't configure it with anything obviously bad.

Along with [other strategies](https://github.com/gomods/athens/pull/1492) to make this better, it would be helpful to opt-in to an HTTP endpoint that an operator can `curl` and that outputs the current configuration. That would be helpful to help with debugging.

We could have Athens log everything on startup.

**Additional context**

N/A

fedepaol:

By setting a ConfigPort parameter, athens will expose a /config endpoint returning the
configuration currently used.
This is useful for debugging purposes. The port is different as the configuration
may contain sensitive informations, and we want to make sure it's exposed only locally.

## How is the fix applied?

Expose the configfile  via a /config endpoint.

Fixes #1503

### DIFF
--- a/config.dev.toml
+++ b/config.dev.toml
@@ -99,6 +99,12 @@ EnablePprof = false
 # PprofPort specifies the port on which pprof will be exposed if activated.
 PprofPort = ":3001"
 
+# ConfigPort specifies the port on which the /config endpoint will listen for
+# providing the internal configuration.
+# Note if you have secrets in your configuration, this endpoint will return them in plaintext.
+# This is why they are exposed in a different port and should be used for debugging purposes only.
+ConfigPort = ""
+
 # The filename for the include exclude filter.
 # Env override: ATHENS_FILTER_FILE
 #
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -60,6 +60,7 @@ type Config struct {
 	SingleFlight      *SingleFlight
 	Storage           *Storage
 	Index             *Index
+	ConfigPort        string `envconfig:"ATHENS_CONFIG_PORT"`
 }
 
 // EnvList is a list of key-value environment
--- a/cmd/proxy/actions/config.go
+++ b/cmd/proxy/actions/config.go
@@ -1,0 +1,19 @@
+package actions
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/gomods/athens/pkg/config"
+)
+
+// ConfigHandler returns the current configuration as json when hitting the /config url
+func ConfigHandler(c *config.Config) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path != "/config" {
+			w.WriteHeader(http.StatusForbidden)
+		}
+		w.Header().Set("Content-Type", "application/json")
+		json.NewEncoder(w).Encode(c)
+	}
+}
--- a/cmd/proxy/actions/config_test.go
+++ b/cmd/proxy/actions/config_test.go
@@ -1,0 +1,118 @@
+package actions
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/gomods/athens/pkg/config"
+)
+
+var testStorage = &config.Storage{
+	Disk: &config.DiskConfig{
+		RootPath: "/path/on/disk",
+	},
+	GCP: &config.GCPConfig{
+		ProjectID: "MY_GCP_PROJECT_ID",
+		Bucket:    "MY_GCP_BUCKET",
+	},
+	Minio: &config.MinioConfig{
+		Endpoint:  "127.0.0.1:9001",
+		Key:       "minio",
+		Secret:    "minio123",
+		EnableSSL: false,
+		Bucket:    "gomods",
+	},
+	Mongo: &config.MongoConfig{
+		URL:                   "mongodb://127.0.0.1:27017",
+		CertPath:              "",
+		InsecureConn:          false,
+		DefaultDBName:         "athens",
+		DefaultCollectionName: "modules",
+	},
+	S3: &config.S3Config{
+		Region: "MY_AWS_REGION",
+		Key:    "MY_AWS_ACCESS_KEY_ID",
+		Secret: "MY_AWS_SECRET_ACCESS_KEY",
+		Token:  "",
+		Bucket: "MY_S3_BUCKET_NAME",
+	},
+}
+
+var testConfig = &config.Config{
+	GoEnv:           "development",
+	LogLevel:        "debug",
+	GoBinary:        "go",
+	GoProxy:         "direct",
+	GoGetWorkers:    10,
+	ProtocolWorkers: 30,
+	CloudRuntime:    "none",
+	TimeoutConf: config.TimeoutConf{
+		Timeout: 300,
+	},
+	StorageType:      "memory",
+	GlobalEndpoint:   "http://localhost:3001",
+	Port:             ":3000",
+	EnablePprof:      false,
+	PprofPort:        ":3001",
+	BasicAuthUser:    "",
+	BasicAuthPass:    "",
+	Storage:          testStorage,
+	TraceExporterURL: "http://localhost:14268",
+	TraceExporter:    "",
+	StatsExporter:    "prometheus",
+	SingleFlightType: "memory",
+	GoBinaryEnvVars:  []string{"GOPROXY=direct"},
+	SingleFlight:     &config.SingleFlight{},
+	SumDBs:           []string{"https://sum.golang.org"},
+	NoSumPatterns:    []string{},
+	DownloadMode:     "sync",
+	RobotsFile:       "robots.txt",
+}
+
+func TestConfigHandler(t *testing.T) {
+	req, err := http.NewRequest("GET", "/config", nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	h := ConfigHandler(testConfig)
+	rr := httptest.NewRecorder()
+
+	h.ServeHTTP(rr, req)
+
+	// Check the status code is what we expect.
+	if status := rr.Code; status != http.StatusOK {
+		t.Errorf("handler returned wrong status code: got %v want %v",
+			status, http.StatusOK)
+	}
+
+	toCheck := &config.Config{}
+	if err := json.Unmarshal(rr.Body.Bytes(), toCheck); err != nil {
+		t.Error("Failed to unmarshal the reply")
+	}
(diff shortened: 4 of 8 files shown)


pkg/config/gcp.go
```diff
@@ -2,7 +2,7 @@ package config
 
 // GCPConfig specifies the properties required to use GCP as the storage backend
 type GCPConfig struct {
-	ProjectID string `envconfig:"GOOGLE_CLOUD_PROJECT"`
-	Bucket    string `validate:"required" envconfig:"ATHENS_STORAGE_GCP_BUCKET"`
-	JSONKey   string `envconfig:"ATHENS_STORAGE_GCP_JSON_KEY"`
+	ProjectID string `envconfig:"GOOGLE_CLOUD_PROJECT" json:"projectID"`
```
arschles, line 5: @fedepaol does this affect any of the config decoding we do? Guessing no because it's TOML but just want to check

pkg/config/s3.go
```diff
@@ -2,14 +2,14 @@ package config
...
-	Token                              string `envconfig:"AWS_SESSION_TOKEN"`
-	Bucket                             string `validate:"required" envconfig:"ATHENS_S3_BUCKET_NAME"`
-	UseDefaultConfiguration            bool   `envconfig:"AWS_USE_DEFAULT_CONFIGURATION"`
-	ForcePathStyle                     bool   `envconfig:"AWS_FORCE_PATH_STYLE"`
-	CredentialsEndpoint                string `envconfig:"AWS_CREDENTIALS_ENDPOINT"`
-	AwsContainerCredentialsRelativeURI string `envconfig:"AWS_CONTAINER_CREDENTIALS_RELATIVE_URI"`
-	Endpoint                           string `envconfig:"AWS_ENDPOINT"`
+	Region                             string `validate:"required" envconfig:"AWS_REGION" json:"region"`
+	Key                                string `envconfig:"AWS_ACCESS_KEY_ID" json:"key"`
+	Secret                             string `envconfig:"AWS_SECRET_ACCESS_KEY" json:"secret"`
```
arschles, line 7: @fedepaol I don't think this config value should be returned. And the same for other secrets. Maybe have a struct tag that indicates it should be hidden, and then read that when you encode it for JSON?

fedepaol: As part of the discussion in [https://github.com/gomods/athens/issues/1503#issuecomment-604677096](https://github.com/gomods/athens/issues/1503#issuecomment-604677096) we agreed that it was fine to expose (also) the sensitive data on a different port, as the endpoint was reachable only with direct machine / container access.

pkg/config/gcp.go
fedepaol: I assumed it doesn't, as the annotation is for json. Also, looking [at the lib we use](https://github.com/BurntSushi/toml), the annotation syntax is `toml:foo`.

fedepaol: also also, if this was affecting decoding, tests would've failed

twexler:
Could you explain how this makes debugging easier? I'm a little weary on adding something that could expose credentials.

fedepaol:
> Could you explain how this makes debugging easier? I'm a little weary on adding something that could expose credentials.

@twexler the rationale is describe in the related issue https://github.com/gomods/athens/issues/1503.
The key to safety is having the endpoint exposed under a different point, as discussed with @marwan-at-work  in https://github.com/gomods/athens/issues/1503#issuecomment-604677096 
Hope I did not misunderstand it.

pkg/config/handler.go
```diff
@@ -0,0 +1,17 @@
+package config
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+// Handler returns the current configuration as json when hitting the /config url
+func Handler(c *Config) http.HandlerFunc {
```
marwan-at-work, line 9: Let's keep `pkg/config` non-http-aware, and move this file to `cmd/proxy/actions`. We can call it `actions.ConfigHandler`. 

This helps keep all of our handler in one place (of course with exception to the download.Protocol handlers since they are tightly coupled with the logic of a GOPROXY)

pkg/config/config.go
```diff
@@ -21,45 +21,46 @@ const defaultConfigFile = "athens.toml"
...
-	NoSumPatterns     []string  `envconfig:"ATHENS_GONOSUM_PATTERNS"`
-	DownloadMode      mode.Mode `envconfig:"ATHENS_DOWNLOAD_MODE"`
-	DownloadURL       string    `envconfig:"ATHENS_DOWNLOAD_URL"`
-	SingleFlightType  string    `envconfig:"ATHENS_SINGLE_FLIGHT_TYPE"`
-	RobotsFile        string    `envconfig:"ATHENS_ROBOTS_FILE"`
-	IndexType         string    `envconfig:"ATHENS_INDEX_TYPE"`
-	SingleFlight      *SingleFlight
-	Storage           *Storage
-	Index             *Index
+	GoEnv             string        `validate:"required" envconfig:"GO_ENV" json:"goenv"`
```
marwan-at-work, line 24: I don't think we need to add any `json:""` tags in this struct. We can just keep the outputted JSON with upper cases. It's quite common for example the GOPROXY itself outputs is `.info` file with upper case fields: https://proxy.golang.org/github.com/pkg/errors/@v/v0.9.1.info

pkg/config/config.go
```diff
@@ -21,45 +21,46 @@ const defaultConfigFile = "athens.toml"
...
+	NoSumPatterns     []string      `envconfig:"ATHENS_GONOSUM_PATTERNS" json:"noSumPatterns"`
+	DownloadMode      mode.Mode     `envconfig:"ATHENS_DOWNLOAD_MODE" json:"downloadMode"`
+	DownloadURL       string        `envconfig:"ATHENS_DOWNLOAD_URL" json:"downloadUrl"`
+	SingleFlightType  string        `envconfig:"ATHENS_SINGLE_FLIGHT_TYPE" json:"singleFlightType"`
+	RobotsFile        string        `envconfig:"ATHENS_ROBOTS_FILE" json:"robotsFile"`
+	IndexType         string        `envconfig:"ATHENS_INDEX_TYPE" json:"indexType"`
+	SingleFlight      *SingleFlight `json:"singleFlight"`
+	Storage           *Storage      `json:"storage"`
+	Index             *Index        `json:"index"`
+	ConfigPort        *string       `envconfig:"ATHENS_CONFIG_PORT" json:"configPort"`
```
marwan-at-work, line 63: No need to make this a pointer to a string, we can just make it a string and check in main.go for `!= ""` instead of `!= nil`. 

It's pretty common and we do it a lot for the other config fields :v:

pkg/config/handler_test.go
```diff
@@ -0,0 +1,114 @@
...
+			status, http.StatusOK)
+	}
+
+	toCheck := &Config{}
+	if err := json.Unmarshal(rr.Body.Bytes(), toCheck); err != nil {
+		t.Error("Failed to unmarshal the reply")
+	}
+	if !reflect.DeepEqual(toCheck, testConfig) {
+		t.Error("Returned config different from original")
+	}
```
marwan-at-work, line 96: Let's use `cmp` to check for equality, but more importantly to output a friendly Diff so that if the test fails a user knows why it failed. 

[See this line of code for an example(https://github.com/gomods/athens/blob/23672f78c933b1d7ad4b3353ce628ccfb1b988b1/pkg/config/config_test.go#L28:L32)

config.dev.toml
```diff
@@ -99,6 +99,10 @@ EnablePprof = false
 # PprofPort specifies the port on which pprof will be exposed if activated.
 PprofPort = ":3001"
 
+# ConfigPort specifies the port on which the /config endpoint will listen for
+# providing the internal configuration
```
marwan-at-work, line 103: Let's add a note about secrets: "Note if you have secrets in your configuration, this endpoint will return them in plaintext. This is why they are exposed in a different port and should be used for debugging purposes only".

config.dev.toml
```diff
@@ -99,6 +99,10 @@ EnablePprof = false
...
+#ConfigPort = ":3002"
```
marwan-at-work, line 104: Per my comment about making `*string` into a `string`. You can uncomment this line and just do `ConfigPort = ""` similar to a lot of our configuration fields

marwan-at-work:
@twexler  @arschles I hear your point about the security concerns. My idea was that by making this endpoint off-by-default and by forcing the Athens admin to expose it on a different port, that would help mitigate credentials exposure. 

That said, I'm also pretty happy with the alternative: expose the config on the same port but obscure all credentials. 

Happy either way :v:

pkg/config/handler_test.go
fedepaol: Did not know about it, awesome!

fedepaol:
@marwan-at-work I think I addressed all the comments.

Also, @arschles @twexler let me know if you prefer me to take the other route, and not publish (or obscure) the "sensitive" configuration parameters.

twexler:
> Also, @arschles @twexler let me know if you prefer me to take the other route, and not publish (or obscure) the "sensitive" configuration parameters.

I'm still struggling to ascertain the usefulness of this feature, could you provide a concrete example of how one might use this? All of the scenarios I'm aware of where this level of debugging would be needed, I'd have access to the configuration provided in one form or another...

marwan-at-work:
@tnextday 

Say you are deploying Athens to a cloud provider and you want to pass specific configs. You have two options: 

1. Create your own image that extends `gomods/athesn:v0.10.0`, and add your own config file. 
2. Use Environment Variables to override some config that are either the default or were passed in your custom config file. 

Then, you realize something is not quite working. Maybe you misspelled an env var and didn't realize it

How can you see that your configs were correctly read by Athens? If you misspelled an Env Var, Athens won't know and won't tell you because it won't go through env vars and guess what you were trying to say of course. 

Asking the server for `/config` to see "the final results" of the config file (the combination of the file + env var overrides), would at least give you the peace of mind that your configuration was successfully read by Athens. 

Now, if you are deploying this to a serveless platform like Cloud Run, you can only ever access on port...so my suggestion fails here and a redacted implementation might be better. 

PS. If you don't think that's a good use case, I'm always happy to remove features instead of add them. The simpler we can keep Athens the easier it will be to maintain :v:

twexler:
> Now, if you are deploying this to a serveless platform like Cloud Run, you can only ever access on port...so my suggestion fails here and a redacted implementation might be better.

This is really the case I can think of, any of these "hey just deploy a docker container" things wouldn't allow more than one port to be exposed (usually). So having this operate on another port for security doesn't make a lot of sense, which leaves this as a security issue...and therefore a not particularly useful feature.

fedepaol:
Closing this, we may reopen in case we change our mind

PR #1660, opened 2020-08-07
go.mod: update github.com/mattn/go-ieproxy to v0.0.1

qbit:
This allows athens to build on OpenBSD.


qbit:
@marwan-at-work without the updated version `proxyMiddleman` is undefined for OpenBSD:

```
# github.com/mattn/go-ieproxy
/home/qbit/go/pkg/mod/github.com/mattn/go-ieproxy@v0.0.0-20190610004146-91bb50d98149/GetProxyFunc.go:10:9: undefined: proxyMiddleman
```

I am not sure if adding it as a dep is the correct fixe. Maybe using `go mod edit -replace` would be better.

qbit:
Switched to using replace as it makes things more clear as to what is being done.

twexler:
I'm wondering if this is the right fix, since we don't directly depend on this. Where is the dependency actually coming from?

go.mod
```diff
@@ -64,3 +64,5 @@ require (
 	gopkg.in/go-playground/validator.v9 v9.20.2
 	honnef.co/go/tools v0.0.1-2020.1.3 // indirect
 )
+
+replace github.com/mattn/go-ieproxy v0.0.0-20190610004146-91bb50d98149 => github.com/mattn/go-ieproxy v0.0.1
```
marwan-at-work, line 68: I don't think we want a replace here, since it will make Athens not go get-able.

qbit: Will it really? I could have sworn I read somewhere in the module code that replace was respected by `go get`.. maybe not though..

qbit: https://github.com/golang/tools/blob/master/gopls/go.mod#L13 `go get golang.org/x/tools/gopls` still works, for example.

jrick: It will still be `go get`-able (assuming the build works), but replaces are currently ignored.  IIRC there is a Go proposal or issue which would cause `go get` to respect the replaces when it is run outside of any module and is being used to install an executable, but only for exact version replaces (like here) and not local filepath replaces (beginning with . or ..).

marwan-at-work: Yep, for those reasons and since the jury is out on how replaces will work (i believe there's another proposal to error out on replace clauses), we should shy away from having replace clauses in our go.mod file. 

@qbit I'm happy to keep this dependency as an explicit requirement like before. 

I think the only thing we were wondering about is details: how does this dependency work? And how come it works by just updating go.mod file without any code in our Athens repository that explicitly calls it?

qbit: @marwan-at-work the reason it works now is the `+build` like in `proxyMiddleman_unix.go` was (a1ba01f) `darwin unix linux`.  Later in https://github.com/mattn/go-ieproxy/commit/91bb50d981495aef1c208d31be3d77d904384f20 the definition of `proxyMiddleman()` moved to `proxy_middleman_unix.go` and the `+build` was changed to `!windows` - thus including OpenBSD.

qbit:
Switched back to `indirect` - let me know if I should sqash it!

qbit:
> I'm wondering if this is the right fix, since we don't directly depend on this. Where is the dependency actually coming from?

@twexler  : according to `go mod why`:
```
# github.com/mattn/go-ieproxy
github.com/gomods/athens/pkg/stash
github.com/Azure/azure-storage-blob-go/azblob
github.com/Azure/azure-pipeline-go/pipeline
github.com/mattn/go-ieproxy
```

marwan-at-work:
> go mod why github.com/mattn/go-ieproxy
>github.com/gomods/athens/pkg/stash
>github.com/Azure/azure-storage-blob-go/azblob
>github.com/Azure/azure-pipeline-go/pipeline
>github.com/mattn/go-ieproxy

It seems to me the proper fix is to actually update the `github.com/Azure/azure-storage-blob-go` go.mod file and not the Athens one. This way, Athens can just update its own dependency properly and not bypass Azure's blob library. 

This can help catch any dependency errors that `go-ieproxy` might have with Azure. 

That being said, we can merge this PR to unblock users and can always revert back if there's a problem. 

LGTM

qbit:
Looks like just bumping `github.com/Azure/azure-storage-blob-go` to v0.10.0 pulls in the latest ieproxy. It also removes `github.com/Azure/go-autorest/autorest/adal v0.6.0 // indirect`.

marwan-at-work:
@qbit let's do that instead!

PR #1662, opened 2020-08-10
Fix missing chart apiVersion

hypnoglow:

`apiVersion` field is Chart.yaml is required but missing for some reason.

## How is the fix applied?

Added `apiVersion: v1` to the Chart.yaml


gw0:
I have a feeling that everyone forgot about this PR. I believe it would make sense to merge that in and also upgrade the Athens version to 0.10.0?

arschles:
@gw0 haven't forgotten about it, just trying to work through the backlog of reviews. sorry about the delay!

arschles (changes requested):
@hypnoglow can you increment the PATCH version number in the `Chart.yaml`? once that's done, this is ready to merge

arschles (approved):

arschles:
thanks @hypnoglow !

Issue #1532, opened 2020-01-28
Availability vs. correctness in list API


linzhp:
**Is your feature request related to a problem? Please describe.**
When Athens is serving list API, it issues request to both storage and VCS and merge two lists. It [fails if there is a storage error](https://github.com/gomods/athens/blob/master/pkg/download/protocol.go#L97). I understand this is to guarantee that Athens returns tags that are deleted from VCS but cached locally can be returned.

However, this approach compromise the availability of Athens server: either VCS or storage is down, Athens list API will be down too. 

**Describe the solution you'd like**
Specifically to this list API, I think availability is more important than correctness. Given that Go toolchain relies on list api to get the latest version of a module, we can think about two scenarios:

* storage is down, but VCS is up. If the deleted but cached tag is the latest, then people won't be able to use `go get -u` or `go get <something>@latest` to get that version, but they can still use `go get <something>@<cached tag>`. If the deleted tag is not the latest, no impact.
* storage is up, but VCS is down. People will only get the latest from storage instead of from VCS, but Go commands will not break with 500 errors.

**Describe alternatives you've considered**
With current Athens implementation, Go commands will break with 500 errors in above scenarios. I agree that neither situation is ideal, but when that happens, instead of having people not able to run `go mod` or `go get` at all, I would prefer limiting the capability of getting the latest version.

@marwan-at-work @arschles thoughts?

marwan-at-work:
@linzhp 

> storage is down, but VCS is up. If the deleted but cached tag is the latest, then people won't be able to use go get -u or go get <something>@latest to get that version, but they can still use go get <something>@<cached tag>. 

If storage is down, pretty much nothing will work. Athens always serves from the storage, never directly from `go mod download` 


>storage is up, but VCS is down. People will only get the latest from storage instead of from VCS, but Go commands will not break with 500 errors.

1. If VCS is down, and you have a resolved go.mod file, then everything should still work. This is because the list endpoint never gets hit. 

2. If VCS is up, but the repo got deleted, we actually continue to work by just "listing" the versions from the storage. 

3. If VCS is down, and you don't have a resolved go.mod file, or maybe you're doing a `go get -u` etc. Then this is the only scenario where things break. 

If you're okay with all of the above, I think the only thing we should focus on is issue number `3` right above^. 

For this, Athens chose "consistency" over "availability".  
Say you have module `x` with 2 versions in storage: `v0.1.0` and `v0.2.0` 
[...]

PR #1668, opened 2020-09-28
begin offline mode implementation

arschles:

In a few places like https://github.com/gomods/athens/issues/1532, people have asked for a mode of Athens that never hits the internet (the current one _will_ hit the internet in various circumstances).

## How is the fix applied?

I began adding support for a mode in Athens that just reads from storage, not any upstream or anywhere else.

Ref https://github.com/gomods/athens/issues/1532

### DIFF
--- a/cmd/proxy/actions/app_proxy.go
+++ b/cmd/proxy/actions/app_proxy.go
@@ -127,7 +127,7 @@ func addProxyRoutes(
 		handlerOpts := &download.HandlerOpts{Protocol: dp, Logger: l, DownloadFile: df}
 		download.RegisterHandlers(r, handlerOpts)
 	} else {
-		handlerOpts := download.OfflineHandlerOpts{Logger: l, Storage: s}
+		handlerOpts := &download.OfflineHandlerOpts{Logger: l, Storage: s}
 		download.RegisterOfflineHandlers(r, handlerOpts)
 	}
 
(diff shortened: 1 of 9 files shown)


cmd/proxy/actions/app_proxy.go
```diff
@@ -86,45 +86,50 @@ func addProxyRoutes(
...
+		dp := download.New(dpOpts, addons.WithPool(c.ProtocolWorkers))
 
-	handlerOpts := &download.HandlerOpts{Protocol: dp, Logger: l, DownloadFile: df}
-	download.RegisterHandlers(r, handlerOpts)
+		handlerOpts := &download.HandlerOpts{Protocol: dp, Logger: l, DownloadFile: df}
+		download.RegisterHandlers(r, handlerOpts)
+	} else {
+		handlerOpts := &download.OfflineHandlerOpts{Logger: l, Storage: s}
+		download.RegisterOfflineHandlers(r, handlerOpts)
+	}
```
Kunde21, line 132: Drive-by look, but this could be:
```
if c.Mode == config.ModeOffline {
		handlerOpts := &download.OfflineHandlerOpts{Logger: l, Storage: s}
		download.RegisterOfflineHandlers(r, handlerOpts)
		return nil
}
// online handling
```

r-ashish:
Solved by: https://github.com/gomods/athens/pull/1717

PR #1671, opened 2020-10-18
Add artifactory storage backend

### LINKED ISSUE #1130: Implement storage.Backend for Artifactory storage

bored-engineer:
Add the option to use [Artifactory](https://jfrog.com/artifactory/) as a storage backend for athens.
The implementation it is very similar to the existing `external` provider with a few tweaks specific to the [Artifactory REST API](https://www.jfrog.com/confluence/display/JFROG/Artifactory+REST+API) and specific configuration options added for usability.
Note: This client uses artifactory as a generic object storage backend, it may or may not be compatible with the folder structure created by official [Artifactory Go Registry](https://www.jfrog.com/confluence/display/JFROG/Go+Registry) plugin.

## How is the fix applied?
Implements a new `storage.Backend` which uses the [Artifactory REST API](https://www.jfrog.com/confluence/display/JFROG/Artifactory+REST+API) to read/write objects.
The upload action is accomplished by writing a zip file (generated on the fly) and using the `X-Explode-Archive: true` and `X-Explode-Archive-Atomic: true` so the objects are extracted atomically into a folder by artifactory.
Lister functionality is accomplished by the [FolderInfo](https://www.jfrog.com/confluence/display/JFROG/Artifactory+REST+API#ArtifactoryRESTAPI-FolderInfo) API call. 

Fixes #1130 (which was already closed/abandoned but is the closest match)

### DIFF
--- a/cmd/proxy/actions/storage.go
+++ b/cmd/proxy/actions/storage.go
@@ -9,6 +9,7 @@ import (
 	"github.com/gomods/athens/pkg/config"
 	"github.com/gomods/athens/pkg/errors"
 	"github.com/gomods/athens/pkg/storage"
+	"github.com/gomods/athens/pkg/storage/artifactory"
 	"github.com/gomods/athens/pkg/storage/azureblob"
 	"github.com/gomods/athens/pkg/storage/external"
 	"github.com/gomods/athens/pkg/storage/fs"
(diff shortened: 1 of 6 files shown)


bored-engineer:
```
redis: 2020/10/18 19:45:13 sentinel.go:393: sentinel: GetMasterAddrByName master="redis-1" failed: dial tcp: lookup redis-sentinel on 127.0.0.11:53: no such host
--
429 | with_redis_sentinel_test.go:31: redis: all sentinels are unreachable
```
Seems like the test failure is unrelated to the PR

arschles:
@bored-engineer yea, I think that's a flake. I restarted it. One question though - would it not be better to just use Artifactory as an upstream? iirc you can use it as a Go module server as well

bored-engineer:
> One question though - would it not be better to just use Artifactory as an upstream? iirc you can use it as a Go module server as well

Yes, that would absolutely make sense if that's an available in your environment. JFrog offers a [Go Registry](https://www.jfrog.com/confluence/display/JFROG/Go+Registry) that should be the first option to use.

In our environment that was not an option but using Artifactory as a storage backend was a business requirement (due to in-built compliance/security controls).

arschles:
@bored-engineer makes sense. Ok I will review this right now.

arschles (changes requested):
@bored-engineer this looks very well done, thanks for submitting it. The only thing left would be to write some tests.

PR #1674, opened 2020-10-27
Fix broken link to docker hub

trenton42:

A broken link to the docker hub image in the documentation.

## How is the fix applied?

Link fixed!


marwan-at-work (approved):
Thank you for the fix :v:

PR #1676, opened 2020-11-14
[WIP] Support for @v/list and @latest endpoints in an offline environment

cobalt77:

I am using Athens to use go modules in an offline, air-gapped environment. With download mode set to none, the @v/list and @latest api endpoints are still trying to go out to the internet to retrieve version information.

## How is the fix applied?

When download mode is set to none and disk storage is used, @v/list will list only the modules available on disk, and @latest will return the latest version from the modules available on disk. It will not try to go out to the internet, since nothing can be downloaded anyway with download mode set to none.

So far these changes are working perfectly in my environment. I am able to develop using go modules as if I was online with no workflow changes or workarounds. 

TODO: the implementation of @latest (sorting all the versions) could probably be moved somewhere else and/or simplified.

For reference, I am using the following system to synchronize the disk storage between my online and offline networks:

- On the online machine, keep an empty go project with a go.mod listing all the direct dependencies you need
- Run [this command](https://gist.github.com/cobalt77/b0d6951e3b142137f03937304f33bc16) to download all dependencies recursively into the local module download cache (~/go/pkg/mod/cache/download). 
    - This is important, and a simple `go mod download` doesn't really work, because (as far as I can tell) outdated semver-compatible dependencies need to be found before the go module system can decide that they can be updated. For example: your package directly requires foo v1.0.2. A dependency of yours requires foo v1.0.1. When online, go finds v1.0.1, but also knows that v1.0.2 is available, and the dependency's foo is updated to v1.0.2 automatically since it's semver compatible. But when offline, go still needs to be able to locate v1.0.1, which means we need both versions in our Athens disk cache. `go mod download` will only give you v1.0.2.
- Copy the download cache to the offline machine 
- Rename the files/folders into the format that Athens wants, copy everything into the Athens disk cache, and restart it.

This works and it's relatively easy. It would be nice to be able to configure Athens to serve the default download cache disk format instead of needing to re-structure the files. Maybe that could be a potential improvement for this PR.


arschles:
@cobalt77 we're not ignoring this. this feature has a relatively long history of discussion in the project, so I need to research all the past PRs and issues so I can properly review it. I'll do that and come review this early next week.

pkg/download/protocol.go
```diff
@@ -74,19 +77,27 @@ func (p *protocol) List(ctx context.Context, mod string) ([]string, error) {
 	ctx, span := observ.StartSpan(ctx, op.String())
 	defer span.End()
 
+	// If the download mode is None, we won't be downloading anything from
+	// the internet anyway, so we should only examine local storage.
+	localOnly := p.df.Match(mod) == mode.None
```
arschles, line 82: @cobalt77 I think we need a different signal to determine when to not look to the internet. `mode.None` means that a 404 should be returned for this module, unless I'm misunderstanding something?

pkg/download/protocol.go
```diff
@@ -145,12 +156,38 @@ func (p *protocol) Latest(ctx context.Context, mod string) (*storage.RevInfo, er
 	const op errors.Op = "protocol.Latest"
 	ctx, span := observ.StartSpan(ctx, op.String())
 	defer span.End()
-	lr, _, err := p.lister.List(ctx, mod)
-	if err != nil {
-		return nil, errors.E(op, err)
+
+	localOnly := p.df.Match(mod) == mode.None
```
arschles, line 160: same as my comment above

arschles (changes requested):
@cobalt77 except for my comment about `mode.None`, I think this is a good step toward enabling an "offline" mode that @nathanhack and others talked about in [this thread](https://github.com/gomods/athens/issues/1532#issuecomment-605214341)

pkg/download/protocol.go
cobalt77: Thought a lot about this. Let's say I have the following setup:
- Offline environment, and a self-hosted git server with some go modules
- Athens with file based storage and a dynamic config where all modules except my.git.server.com/* get download mode "None"

If I go get github.com/something, that will use download mode None and treat the contents of the athens cache as ground truth, and get me the latest of the available modules.
If I go get my.git.server.com/somethingelse, athens will go out to my git server for the module, and not assume that what's in cache is the latest.

This will work as expected with this change. But it changes what None does, and maybe there is a scenario where a 404 would be preferable. Or in online environments, athens treating the cache as ground truth _sometimes_ but not all the time could lead to nondeterministic behavior or other weird issues. 

What do you think about keeping None as-is (always 404 if it doesn't exist), and adding a new separate download mode that will do this?

PR #1672, opened 2020-10-19
Fix K8s service account usage

owenhaynes:
## What is the problem I am trying to address?

For the K8s service account to be used in the deployment, fix legacy resource,  and allow annotations on service account.

## How is the fix applied?

added the service account to the deployment

PR #1677, opened 2020-11-16
Allow using existing service accounts

bnrjee:

Right now, there is no way to use existing service accounts to deploy athens using helm chart. It always tries to create a new one. But the requirement here is to use a service account that has already been created.
## How is the fix applied?

Tested with dry-run and on kubernetes cluster. With this change, athens can use an exiting service account.

Fixes #

### DIFF
--- a/charts/athens-proxy/Chart.yaml
+++ b/charts/athens-proxy/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: athens-proxy
-version: 0.4.15
+version: 0.4.16
 appVersion: 0.9.0
 description: The proxy server for Go modules
 icon: https://raw.githubusercontent.com/gomods/athens/main/docs/static/banner.png
--- a/charts/athens-proxy/templates/service-account.yaml
+++ b/charts/athens-proxy/templates/service-account.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.serviceAccount.create }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
@@ -11,3 +12,4 @@ metadata:
   annotations:
 {{ toYaml . | indent 4 }}
   {{- end }}
+{{- end }}
--- a/charts/athens-proxy/values.yaml
+++ b/charts/athens-proxy/values.yaml
@@ -147,6 +147,7 @@ sshGitServers: {}
 goGetWorkers: 3
 
 serviceAccount:
+  create: true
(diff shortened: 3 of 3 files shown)


bnrjee:
@arschles,
Would it be possible for you to take a look?

arschles:
@bnrjee yes, I will look today ASAP

arschles (changes requested):
@bnrjee don't you need to change the serviceAccount parameter in the [deployment.yaml](https://github.com/gomods/athens/blob/2428d133b7cb39c552d87b72563528418030e009/charts/athens-proxy/templates/deployment.yaml#L37) file and parameterize the name?

bnrjee:
> @bnrjee don't you need to change the serviceAccount parameter in the [deployment.yaml](https://github.com/gomods/athens/blob/2428d133b7cb39c552d87b72563528418030e009/charts/athens-proxy/templates/deployment.yaml#L37) file and parameterize the name?

With the existing code(https://github.com/gomods/athens/pull/1672), a service account does get created and used. This PR will make sure, we will use the existing SAs and not create one. So, that kind of solves the problem.

bnrjee:
> > @bnrjee don't you need to change the serviceAccount parameter in the [deployment.yaml](https://github.com/gomods/athens/blob/2428d133b7cb39c552d87b72563528418030e009/charts/athens-proxy/templates/deployment.yaml#L37) file and parameterize the name?
> 
> With the existing code(#1672), a service account does get created and used. This PR intends to make sure that we will be able to use the existing SAs and not create one. So, that solves the problem we are facing.

@arschles , Would it be possible for you to take a took, based on my update?

abhay-krishna:
@arschles Is it possible to take a look at this and let us know your thoughts? Right now, we use a fork with the change that this PR adds. It would be helpful to have this feature in the upstream Athens so that we can remove forks and use charts directly from `gomods/athens`.

arschles:
@abhay-krishna yup, I'll take a look tomorrow - sorry for the long delay

arschles (approved):
@bnrjee this is great. sorry for the delay! 🚀

arschles:
@abhay-krishna @bnrjee you'll likely need to wait until we release a new version of Athens to take advantage of the helm chart changes. I'll be doing a release this week 👍

bnrjee:
Hi @arschles,
Please let us know, when can we expect a new release with these new changes.

PR #1681, opened 2020-11-21
Add Content-Length to .zip requests

marwan-at-work:
This PR enables Athens to return an accurate byte count of a module@version's size via the Content-Length header. 

It also allows HEAD requests to get said header without returning the content of the zip file. 

This is needed for Athens to integrate with pkgsite, particularly here: https://github.com/golang/pkgsite/blob/fd89b15b08609658e2b6d1363f52fd2f691db311/internal/proxy/client.go#L97:L116

### DIFF
--- a/pkg/storage/module.go
+++ b/pkg/storage/module.go
@@ -6,10 +6,11 @@ import (
 
 // Module represents a vgo module saved in a storage backend.
 type Module struct {
+	// TODO(marwan-at-work): ID is a mongo-specific field, it should not be
+	// in the generic storage.Module struct.
 	ID      primitive.ObjectID `bson:"_id,omitempty"`
 	Module  string             `bson:"module"`
 	Version string             `bson:"version"`
 	Mod     []byte             `bson:"mod"`
-	Zip     []byte             `bson:"zip"`
 	Info    []byte             `bson:"info"`
 }
--- a/pkg/storage/s3/getter.go
+++ b/pkg/storage/s3/getter.go
@@ -3,7 +3,6 @@ package s3
 import (
 	"context"
 	"fmt"
-	"io"
 	"io/ioutil"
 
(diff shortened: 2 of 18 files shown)


pkg/storage/module.go
```diff
@@ -6,10 +6,11 @@ import (
 
 // Module represents a vgo module saved in a storage backend.
 type Module struct {
+	// TODO(marwan-at-work): ID is a mongo-specific field, it should not be
```
arschles, line 9: https://github.com/gomods/athens/issues/1683

pkg/storage/s3/getter.go
```diff
@@ -100,6 +100,9 @@ func (s *Storage) open(ctx context.Context, path string) (io.ReadCloser, error)
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
-
-	return goo.Body, nil
+	var size int64
+	if goo.ContentLength != nil {
```
arschles, line 104: I'm assuming that `ContentLength` is never expected to be nil, but if it is, is it ok that we return a 0 length to the doc server? looks like it will be more likely to [process the module](https://github.com/golang/pkgsite/blob/fd89b15b08609658e2b6d1363f52fd2f691db311/internal/fetch/fetch.go#L173) even if it's big

arschles (commented):
LGTM @marwan-at-work

pkg/storage/s3/getter.go
marwan-at-work: @arschles yep, I think we should be fine. I contemplated putting an `if zip.Size() > 0` on the handler level so we only write the Content-Length header when size is greater than zero, but I figured if S3 is not returning content-length there are much more important things to worry about (since S3 is down or buggy) than the content-length header?

marwan-at-work:
@arschles thanks for the quick review. Addressed all the comments :v:

PR #1682, opened 2020-11-21
Implement CodeQL GitHub Action

chriscoffee:

CodeQL's GitHub integration is out of invite only and now available to public repositories. We should have some security scanning in place as a check.

## How is the fix applied?

Added the CodeQL action which autobuilds, scans and reports issues. It currently runs against main and on a schedule of 0 15 * * 0 or 15:00 every Sunday.


marwan-at-work (approved):
👌🏼 looking forward to see what it shows

PR #1687, opened 2020-12-15
Add possibility to specify image pull secrets

sarunask:

Deploy athens proxy from private Docker repository

## How is the fix applied?

It can be applied, when invoked with helm chart values like:
```
image:
  pullSecrets: your-secret-in-k8s-name
```
If this variable is not defined, helm chart will not emit any changes to final K8S object

None

### DIFF
--- a/charts/athens-proxy/Chart.yaml
+++ b/charts/athens-proxy/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: athens-proxy
-version: 0.4.15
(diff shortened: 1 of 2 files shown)


arschles (changes requested):
@sarunask looks great, sorry for the long delay here. Can you resolve the conflicts? Then this will be ready to merge.

DrPsychick:
fixed in https://github.com/gomods/athens-charts

Issue #1506, opened 2019-12-20
Use Athens as repository in private (offline) network


RonAmihai:
I'm working in an offline environment which consists of private (offline) network. 

I can bring packages (3rd party) which I've downloaded in an online machine (moving them to the offline network over USB media storage), but the machines in the offline network don't have direct access to the worldwide internet.

In Node/JavaScript project, I'm using [Verdaccio] (https://github.com/verdaccio/verdaccio)  as private repository *running in my private offline network*. 
In an online machine I'm downloading packages with it's dependencies using the [npm-offline-packager](https://www.npmjs.com/package/npm-offline-packager)  tool, then moving the downloaded tarball files to my private offline network. Afterwards I'm using npm-offline-packager or npm-publish to publish the packages to my Verdaccio. 

In Python projects, I don't have proxy/repository solution (like JFrog Artifactory), so what I'm doing is to download packages+dependencies online using the "pip download", than moving the files to our offline private network, than installing the files locally using "pip install --no-index --find-links <path_to_packages_folder>". 

I'm looking for applying one of the above approaches with Go project. 
In this case - Athens should act similarly to Verdaccio, and I need a tool to download Go packages *with all of it's dependencies* online, then moving files to the offline network and publish them to Athens in my offline private network. 

Is this approach possible? Does there any tool that can download go packages+dependencies (even from a list of packages) online and than publish them to Athens in an offline environment?

arschles:
@RonAmihai can you check out our [documentation on pre-filling the disk cache](https://docs.gomods.io/configuration/prefill-disk-cache/) and let me know if that will work for you?

RonAmihai:
> @RonAmihai can you check out our [documentation on pre-filling the disk cache](https://docs.gomods.io/configuration/prefill-disk-cache/) and let me know if that will work for you?

Thank you for this very helpful information. 
I guess I've skipped it while reading the documentation.

I'll give it a try in the next few days and share the results here

arschles:
@RonAmihai great. I'm going to leave this issue open for now 😄

RonAmihai:
> @RonAmihai can you check out our [documentation on pre-filling the disk cache](https://docs.gomods.io/configuration/prefill-disk-cache/) and let me know if that will work for you?

Are there convenient alternatives to pacmod for this method? 

Pacmod doesn't seem to support packaging with all dependencies. 
I want to PR this feature soon, but till than I'm wondering if I can bring package will all of it's dependencies to Athens in other way.

[...]

Issue #1378, opened 2019-09-13
Athens airgapped mode


LukeMauldin:
I would like to run Athens in a completely airgapped mode. I have it setup with disk storage. I use a separate Athens server on a machine with internet access to download modules and then manually transfer the storage directory to the Athens server. This works as long as I specify the exact version in go get. Example: 
‘go get -v github.com/foo/bar@v1.0.1’ however if I run ‘go get -v github.com/foo/bar’ then it fails because Athens tries to do a ‘go list -m’ internally which fails due to lack of internet access.

I would prefer for that ‘go list’ command just to return the versions that I have uploaded into Athens for a specific module. If that is not possible, I would at least like to configure it not to make the external calls which will always fail.

I have tried the filtering modes and download mode options but have not been able to get the configuration correct. Is this even possible with the current version?

Using version v0.6.0

arschles:
Hi @LukeMauldin - at the moment, Athens tries to fetch the list of versions from storage and from the upstream, and then merge them ([here](https://github.com/gomods/athens/blob/master/pkg/download/protocol.go#L72) is where it does that)

When you do a `go get` without the version number, iirc it's doing a request to the `@latest` endpoint. We could add a config flag to Athens to make it just look at the latest version in storage instead of going upstream. @marwan-at-work what do you think?

marwan-at-work:
I think it makes sense to add a new configuration option where a "more advanced" Athens admin would "disable using `go list`, or `go mod download`" 

However, I do want to understand your use case a little more...couldn't you just solve this on the client side by doing `GOPROXY=athens1,athens2,direct,etc go get ...`??

LukeMauldin:
I like the idea of a configuration option to tell Athens to use the packages in storage and not check upstream for completely airgapped scenarios.
As far as I know, I cannot resolve this client side. Use case: I import github.com/foo/bar v1.0.0 and v1.0.1 manually into Athens storage. I then have a module that depends on v1.0.0. I would like to be able to do ‘go get -u github.com/foo/bar’ and for Athens to return v1.0.1 and not try to check upstream in github.com because that will fail.

arschles:
Sorry for the delay @LukeMauldin. In your use-case, when you do the `go get -u github.com/foo/bar`, are you looking to prevent Athens from lookup upstream, even if it can access the public internet?

[...]

Issue #395, opened 2018-08-03
Proxy use-case for offline, secure development environments


nicknezis:
Will the proxy support the use case in which Go development happens in a totally offline environment? An enterprise could setup a process to manually sync modules to the offline repo in which an instance of the proxy is able to serve up the modules to internal Go developers.

arschles:
The proxy can work completely behind a firewall, without access to the
public internet. The current limitation in that case is that it can only
serve modules that exist in another internal proxy, or a vcs.

We don’t currently have functionality to “manually” sync - other than
exposing the proxy to the public internet and running a bunch of “go get”
commands against it.

Can you give some more details on what you’re trying to do?

nicknezis:
@marpio Yes I believe I'm good for now. Thanks.

PR #1691, opened 2021-01-07
Offline mode

chrisFrodo:
Hi,

I've been trying to use Athens on an air-gaped machine (offline) to fetch all my Golang dependencies. I've been able to make Athens work by copying the cache folder from an internet machine to the offline one.

However, Golang can only download the needed modules if the exact version of the dependency is specified since the `@latest` listing always tries to connect to internet.

This seems to be a recurrent issue as demonstrated by https://github.com/gomods/athens/issues/1506 https://github.com/gomods/athens/issues/1378 https://github.com/gomods/athens/issues/395

I saw the following PR https://github.com/gomods/athens/pull/1668 which seems to address this issue by adding a `MODE=offline` configuration. 

This PR only fix a few issues on the draft above. It has been tested and works perfectly fine in my offline setup.

Thank you

### DIFF
--- a/cmd/proxy/actions/app.go
+++ b/cmd/proxy/actions/app.go
@@ -23,6 +23,8 @@ const Service = "proxy"
 // should be defined. This is the nerve center of your
 // application.
 func App(conf *config.Config) (http.Handler, error) {
+	fmt.Printf("Athens is running in %s mode\n", conf.Mode)
+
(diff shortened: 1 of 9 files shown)


arschles:
@chrisFrodo I'll take a look at this in detail in the next few days. sorry for the delay here

PR #1699, opened 2021-03-07
Fix permissions for /config/config.toml so Athens can run as non-root

### LINKED ISSUE #1695: Docker: change /config/config.toml permissions so Athens can be ran as non-root user
**Problem**

Currently the `/config/config.toml` has permissions set to `0700`.
This makes it impossible to run Athens with a non-root user, and this is in turn needed so you can mount the filesystem storage directory with your UID/GID.

**Possible solution**
Changing the permissions to something sane, like `0644` (or `0755` and add a `#!/bin/athens-proxy` hashbang.)

**Alternative**
Building a wrapper image is indeed an option, yet it's not as simple as `docker run gomods/athens:latest -u $UID:$GID -v $PWD/data:/data -e...`

**Additional context**
I couldn't find the Dockerfile used to build https://hub.docker.com/r/gomods/athens, otherwise I would've sent a PR (even though it's a 1-line change.)

tzvetkoff:

Current permissions of `/config/config.toml` make Athens impossible to run as non-root user.

## How is the fix applied?

Changed the file permissions to `0644`.

Fixes #1695


arschles (approved):
@tzvetkoff looks good, thanks for this. there was a flake in the CI/CD, so updated code from `main` and waiting for the new CI/CD run to finish

arschles:
Ok, this looks like an unrelated test error, might be a flake but I'm not sure.

I'll do this as soon as I can @tzvetkoff but I can't today unfortunately. @marwan-at-work any chance you have an idea what's going on here?

marwan-at-work:
@arschles looks like redis-sentinel can time out on start up, I restarted the build and it's good now.

arschles:
nice, thanks @marwan-at-work . I restarted a few times but it didn't fix the flake. I guess drone got in a better mood.

abh:
This errors out when using the `canary` docker container under kubernetes. There is a (weird, IMO) check to disallow these permissions in https://github.com/gomods/athens/commit/5eba6f2e47728ddeebef4387dd909df9dfdbc484#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R170

manugupt1:
Hello @abh that's correct. The problem with 644 is that tokens stored in the config can be read by the world. Is 644 absolutely required, or would 0640 suffice?

Thanks!

abh:
@manugupt1 I didn't look at it carefully beyond noticing that installed in kubernetes with the helm chart the canary image would error out with the message about the permissions being too lax and at a glance it seems like the change to make them lax was deliberate (but not followed up with relaxing the other check).

manugupt1:
@abh Will you be able to do a quick check please? Does Athens work on latest image (not canary)

manugupt1:
@arschles @abh What do you think of the following:

1. Remove the code to allow users too modify the permissions: https://github.com/gomods/athens/commit/5eba6f2e47728ddeebef4387dd909df9dfdbc484#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R170
2. Change the Dockerfile permissions to 0700 (so that secrets are not visible to the world)
3. Add an FAQ to tell them how to change the permissions; Dockerfile, initContainers as possible options?

PR #1700, opened 2021-03-11
Reverting passthrough auth

arschles:

https://github.com/gomods/athens/pull/1650 was a big change to the authentication/authorization code, which we have decided to pull out and potentially move into a separate process/project

## How is the fix applied?

@marwan-at-work and I (pair programming) partially reverted the code in #1650 while selectively keeping some of the improvements in there.

N/A


marwan-at-work (approved):
:v:

PR #1710, opened 2021-04-06
Removing the known limitations file

### LINKED ISSUE #1709: maybe remove known-limitations.md
The first paragraph is about go modules being an experimental feature which isn't true anymore and the other two are not really about limitations at all.

arschles:

As @thomasf pointed out, these limitations no longer apply

## How is the fix applied?

I simply removed the `known-limitations.{md, zh.md}` files

Fixes #1709


marwan-at-work (approved):
👌🏼

arschles:
That was fast, thanks @marwan-at-work !

PR #1712, opened 2021-04-20
INF-13413

nickhyoti:
Extended to support mTLS for connection between a reverse proxy and Athens

Enabled by adding TLSCACERTFILE to the configuration file.

Enabling mTLS to harden network connectivity between a reverse proxy eg. nginx and Athens.

## How is the fix applied?

Code as been extended adding a configuration option TLSCACERTFILE.  If present, the http server will be configured to require the certificate from the reverse proxy on connection requestion

N/A


arschles:
hey @nickhyoti - I took a glance at this and it looks reasonable. do you have a use case for it in your environment where you run Athens though?

nickhyoti:
We have put this forward as we want to

1. Reduce the attack surface of the service.  By mandating a connection through nginx reverse proxy (our test config), connections to Athens are restricted.
2. Ensure connectivity is encrypted within the interconnecting components
3. Eliminate the risk of incorrect interconnection of components

To support demonstration/testing, I have created a directory mtls which contains a docker-compose override to create a nginx container dependent on Athens dev container and also some test certificate directories.
The frontend directory certificates are deployed on the reverse proxy, the backend are used by the proxied connection to Athens.
Configuration files are provided to support this in the nginx/conf.d directory.

To enable and test, run
docker-compose -f docker-compose.yml -f mtls/mtls.yaml up --build

docker ps will show that the athens_dev_1 and nginx_1 containers are running,listening on ports 80 & 443 for nginx, and port 3000 for athens

Testing connections using curl to port 3000 on Athens will return 
`curl: (56) OpenSSL SSL_read: error:14094412:SSL routines:ssl3_read_bytes:sslv3 alert bad certificate, errno 0`

To connect, 
`curl -k   --cert web.cert.pem --key web.key.pem https://localhost:3000`

**Note that the certificates provided are only for demonstration purposes.**

nickhyoti:
Hi,

Is there any feedback on this?

thanks

cmd/proxy/main.go
```diff
@@ -38,15 +41,37 @@ func main() {
 		log.Fatal(err)
 	}
 
-	cert, key, err := conf.TLSCertFiles()
+	cert, key, mtlsCert, err := conf.TLSCertFiles()
```
nrwiersma, line 44:
```suggestion
	cert, key, caCert, err := conf.TLSCertFiles()
```
It is not really an MTLS cert, but the CA cert.

cmd/proxy/main.go
```diff
@@ -38,15 +41,37 @@ func main() {
...
+		srv = &http.Server{
+			Addr:      conf.Port,
+			Handler:   handler,
+			TLSConfig: tlsConfig,
+		}
+	} else {
+		srv = &http.Server{
+			Addr:    conf.Port,
+			Handler: handler,
+		}
```
nrwiersma, line 72: This `else` is not needed if you initialise the server with these properties set. It also means that L63 can just be setting the server `TLSConfig`.

cmd/proxy/main.go
```diff
@@ -38,15 +41,37 @@ func main() {
...
 	if err != nil {
 		log.Fatal(err)
 	}
+	var srv *http.Server
 
-	srv := &http.Server{
-		Addr:    conf.Port,
-		Handler: handler,
+	if mtlsCert != "" {
+		clientCACert, err := ioutil.ReadFile(conf.TLSCACertFile)
```
nrwiersma, line 51: `ioutil` was depreciated in Go 1.16 and is not safe to use. Use `os.ReadFile` instead.

pkg/config/config.go
```diff
@@ -213,26 +214,36 @@ func (c *Config) BasicAuth() (user, pass string, ok bool) {
 
 // TLSCertFiles returns certificate and key files and an error if
 // both files doesn't exist and have approperiate file permissions
-func (c *Config) TLSCertFiles() (cert, key string, err error) {
+func (c *Config) TLSCertFiles() (cert, key string, cacert string, err error) {
```
nrwiersma, line 217:
```suggestion
func (c *Config) TLSCertFiles() (cert, key, cacert string, err error) {
```

nrwiersma (changes requested):
Hi. I have added some comments. In addition the `mtls` folder does not really belong in the project. I would be better added in the PR comments for testing purposes.

PR #1716, opened 2021-06-03
Update GCP Storage authentication comment

zdebra:

I am suggesting replacing a vague phrasing about how GCP out-of-box authentication works. If Athens is used in a GCP product like Cloud Run it will work without any extra work because `GOOGLE_APPLICATION_CREDENTIALS` variable is automatically propagated by GCP so the previous comment was not wrong but this is more helpful for custom deployments where `GOOGLE_APPLICATION_CREDENTIALS` approach can also be used.

## How is the fix applied?

A comment in the config file.

### DIFF
--- a/config.dev.toml
+++ b/config.dev.toml
@@ -353,10 +353,10 @@ IndexType = "none"
 
         # JSONKey is a base64 encoded service account 
         # key that allows Athens to be run outside of GCP
-        # but still be able to access GCS. If you are
-        # running Athens inside GCP, you will most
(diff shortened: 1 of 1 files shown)


arschles (approved):
Thanks @zdebra!

arschles:
This PR is an improvement to documentation in the `config.dev.toml` file and has no helm chart changes. Although the CircleCI jobs for linting and testing charts are hanging, I am merging this anyway.

PR #1717, opened 2021-06-03
Enable Athens to work in offline mode

marwan-at-work:
This PR adds a new "NetworkMode" configuration that allows Athens to operate in an air-gapped environment or to be more forgiving in the case of upstream VCS flakiness. 

See config.dev.toml for further explanation on the behavior and the new options for NetworkMode. 

Fixes https://github.com/gomods/athens/issues/1506
Fixes https://github.com/gomods/athens/issues/1532


pkg/download/protocol.go
```diff
@@ -42,8 +42,16 @@ type Opts struct {
 	Stasher      stash.Stasher
 	Lister       module.UpstreamLister
 	DownloadFile *mode.DownloadFile
+	NetworkMode  string
```
arschles, line 45: maybe make this its own type? obv not required, but could be helpful for readability

pkg/download/protocol.go
```diff
@@ -76,17 +85,31 @@ func (p *protocol) List(ctx context.Context, mod string) ([]string, error) {
...
+		TODO: potential refactor:
+
+		Storage Lister: just return stuff from storage, or error otherwise.
+
+		FallbackVCS lister: list from VCS, return empty list if error.
+
+		StrictVCS Lister: list from VCS, error if it doesn't succeed.
+
+		UnionLister(listers ...Lister): combines any number of listers.
+	*/
```
arschles, line 99: @marwan-at-work did you decide to do this refactor? could maybe do this in a follow-up

pkg/storage/mem/mem.go
```diff
@@ -2,36 +2,24 @@ package mem
...
-
 // NewStorage creates new in-memory storage using the afero.NewMemMapFs() in memory file system
 func NewStorage() (storage.Backend, error) {
 	const op errors.Op = "mem.NewStorage"
-	l.Lock()
-	defer l.Unlock()
-
-	if memStorage != nil {
-		return memStorage, nil
-	}
```
arschles, line 26: super happy you got rid of the globals. thanks!

arschles (changes requested):
@marwan-at-work only a few small things, otherwise this looks good!

I think this feature also warrants something on the docs site. Thoughts?

pkg/download/protocol.go
marwan-at-work: @arschles definitely planning on doing it 👍🏼

marwan-at-work:
@arschles I can def add something to the docs in a follow up PR

pkg/download/protocol.go
marwan-at-work: I think it's easier as a string since it's only used internally and no need to introduce cyclical imports since this is being passed from the config into the download protocol

pkg/download/protocol.go
arschles: ok. in this PR or later?

arschles:
> @arschles I can def add something to the docs in a follow up PR

@marwan-at-work sounds great!

arschles (approved):

PR #1719, opened 2021-06-18
Use actual content length when it is not available in headers

linzhp:

In some cases, Amazon S3 may not always return a Content-Length header. In this case, the size will not be set and take the zero value, and `http.ResponseWriter` will complain:
```
http: wrote more than the declared Content-Length
```
The data will be ignored

## How is the fix applied?

Skip  `Content-Length` in response when the size is 0, so at least the data can be sent.

### DIFF
--- a/pkg/download/version_zip.go
+++ b/pkg/download/version_zip.go
@@ -44,15 +44,14 @@ func ZipHandler(dp Protocol, lggr log.Entry, df *mode.DownloadFile) http.Handler
 		defer zip.Close()
 
 		w.Header().Set("Content-Type", "application/zip")
+		size := zip.Size()
+		if size >= 0 {
+			w.Header().Set("Content-Length", strconv.FormatInt(size, 10))
+		}
 		if r.Method == http.MethodHead {
 			return
 		}
-		written, err := io.Copy(w, zip)
-		size := zip.Size()
-		if size < 0 {
-			size = written
-		}
-		w.Header().Set("Content-Length", strconv.FormatInt(size, 10))
+		_, err = io.Copy(w, zip)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/download/version_zip.go
```diff
@@ -44,7 +44,10 @@ func ZipHandler(dp Protocol, lggr log.Entry, df *mode.DownloadFile) http.Handler
 		defer zip.Close()
 
 		w.Header().Set("Content-Type", "application/zip")
-		w.Header().Set("Content-Length", strconv.FormatInt(zip.Size(), 10))
+		size := zip.Size()
+		if size >= 0 {
```
marwan-at-work, line 48: Can we change this to `if size > 0` (instead of >=) so that we don't have to set `-1` on every storage implementation? 

There's very little to 0 percent chance that we have a proper module that happens to be 0 bytes...so I think it's safe to assume that a 0 size is just an "unknown size" as opposed to an actually 0 byte zip file.

pkg/storage/s3/getter.go
```diff
@@ -100,7 +100,7 @@ func (s *Storage) open(ctx context.Context, path string) (storage.SizeReadCloser
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
-	var size int64
+	var size int64 = -1
```
marwan-at-work, line 103: Given the comment above, would it be okay to revert this change and keep it at zero? This way all storage implementations have consistent behavior hopefully.

marwan-at-work (changes requested):
Thank you for the fix! Just a quick comment and should be good.

marwan-at-work (approved):
🙌🏼

PR #1720, opened 2021-06-23
Using directory as prefix for S3

linzhp:

Listing objects on S3 with prefixes like `github.com/golang/mock/@v/v1.4.4.` with return empty list, because prefixes are directories on S3 (https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-prefixes.html).

## How is the fix applied?

Taking the directory name from the path returned by `config.PackageVersionedName(module, version, "")`, so  it  can return all files for the module. Because of this, all files from all versions will be returned and filtered later.


pkg/storage/s3/checker.go
```diff
@@ -23,7 +24,7 @@ func (s *Storage) Exists(ctx context.Context, module, version string) (bool, err
 
 	lsParams := &s3.ListObjectsInput{
 		Bucket: aws.String(s.bucket),
-		Prefix: aws.String(config.PackageVersionedName(module, version, "")),
+		Prefix: aws.String(path.Dir(config.PackageVersionedName(module, version, ""))),
```
marwan-at-work, line 27: I think you can just use `config.FmtModVer` instead of taking path.Dir(config.PackageVersionedName(...)).

marwan-at-work (commented):
Thanks for the fix. I'm surprised this hasn't come up before. 

I think we could really use some integration testing here but unfortunately I don't have an AWS account. 

cc: @arschles

pkg/storage/s3/checker.go
linzhp: `config.FmtModVer` would return something like `github.com/golang/mock@1.4.4`, but we need `github.com/golang/mock/@v`

linzhp:
@marwan-at-work Can you take another look?

pkg/storage/s3/checker.go
marwan-at-work: @linzhp in that case adding a version then removing just to use existing helpers seems a little unnecessary. What if we just did this: `aws.String(fmt.Sprintf("%s/@v", module))`

marwan-at-work:
@linzhp just a nit comment and looks good 👍🏼

marwan-at-work (approved):

PR #1721, opened 2021-06-26
Better handling of unauthorized git for private repos

### LINKED ISSUE #1718: Better error message for invalid github token
We are running Athens in the company as a goproxy for both private and public repositories. It was working just fine until Github token was rotated and goproxy lost permissions for company private repositories. 

This is obfuscated output of the service:
```
INFO[12:14PM]: github.com/xxx/yyy@v0.0.0-20210604075338-e5bf7b8f2f37: invalid version: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /tmp/athens384359910/pkg/mod/cache/vcs/a224f30cc579fe6b9e50f9d309bdb2d22bbe0edc7b0b1db39bc5ae02dcbcd21d: exit status 128:
	fatal: could not read Username for 'https://github.com': terminal prompts disabled	http-method=GET http-path=/github.com/xxx/yyy/@v/v0.0.0-20210604075338-e5bf7b8f2f37.mod kind=Not Found module= operation=download.VersionModuleHandler ops=[download.VersionModuleHandler pool.GoMod protocol.GoMod protocol.processDownload gcslock.Stash stash.Pool stasher.Stash stasher.fetchModule goGetFetcher.Fetch module.downloadModule] request-id=a0b21e1d-1e78-4b84-912b-e8c46eb667c5 version=
INFO[12:14PM]: incoming request	http-method=GET http-path=/github.com/xxx/yyy/@v/v0.0.0-20210604075338-e5bf7b8f2f37.mod http-status=404 request-id=a0b21e1d-1e78-4b84-912b-e8c46eb667c5
```
I don't think this error message is clear.

The service should be very clear in the logs that the github token is not valid.

-

**Additional context**
Now, when the issue is uncovered, the message is more clear but it took me a day to debug since there was a strong lead that the GCP Storage, not Github was the issue.

P.S. I would be happy to do it if you give me a green light.

zdebra:

Clear error message when dealing with failed authorization when listing / fetching from private repositories.

## How is the fix applied?

Parsing error output from the go command + new error kind. It's not the best practise, but it's all we have since go command won't give us an error code. See https://github.com/gomods/athens/pull/1299.

Fixes #1718


pkg/module/go_vcs_lister.go
```diff
@@ -89,3 +94,7 @@ func (l *vcsLister) List(ctx context.Context, module string) (*storage.RevInfo,
 	}
 	return &rev, lr.Versions, nil
 }
+
+func gitAuthFailed(errCmd error) bool {
+	return strings.Contains(errCmd.Error(), "fatal: could not read Username for 'https://github.com'")
```
manugupt1, line 99: thanks for the PR. Apologies for the delay.

This is very specific to github. There are other providers that use git (eg: gitlab). Is it possible to update the code to account for other providers as well.

arschles: perhaps something like:

```suggestion
	return strings.Contains(strings.ToLower(errCmd.Error()), "could not read username")
```

arschles (changes requested):

pkg/module/go_vcs_lister.go
marwan-at-work: I don't think this is generic enough either. For example many times I've seen a "terminal prompt disabled" error being the one that really means "unauthorized". 

It's super unlikely we'll ever get a consistent way to figure out the errors mainly because "git" itself doesn't return error codes and even if it did, the Go command doesn't propagate them. 

My opinion (and it's just an opinion) is that I think it's best to just always return 404 while logging the error and let the admin maintainer decide the best route to re-configure Athens. Consistency here is better than returning 401 _only sometimes_ which could lead to maintainers trusting that 404 always means 404 when it could just be a different unauthorized message. 

Happy to hear other thoughts.

manugupt1: Hi 

I think you are correct. 

```
{
        "Path": "github.com/manugupt1/emptygo",
        "Version": "v0.0.0-20220219214136-6d3fdc1d3313",
        "Error": "github.com/manugupt1/emptygo@v0.0.0-20220219214136-6d3fdc1d3313: verifying go.mod: github.com/manugupt1/emptygo@v0.0.0-20220219214136-6d3fdc1d3313/go.mod: reading https://sum.golang.org/lookup/github.com/manugupt1/emptygo@v0.0.0-20220219214136-6d3fdc1d3313: 410 Gone\n\tserver response:\n\tnot found: github.com/manugupt1/emptygo@v0.0.0-20220219214136-6d3fdc1d3313: invalid version: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /tmp/gopath/pkg/mod/cache/vcs/56e82eb92b5dc61306dd3b906d375337ebe84c7f277c0bd376d8d62bb5680b40: exit status 128:\n\t\tfatal: could not read Username for 'https://github.com': terminal prompts disabled",
        "Info": "/Users/manu/go/pkg/mod/cache/download/github.com/manugupt1/emptygo/@v/v0.0.0-20220219214136-6d3fdc1d3313.info"
}
```
Would it be better to return the entire error to the user from git?

manugupt1: Would it be better to do the same idea above; but for lister.

We can expand listResp


```
https://pkg.go.dev/cmd/go/internal/list

type listResp struct {
	Path     string
	Version  string
	Versions []string `json:",omitempty"`
	Time     time.Time
	Error *ModuleError // new field.
}

type ModuleError struct {
	Err string // the error itself
}
```
I got the reference from: https://pkg.go.dev/cmd/go/internal/list

When there is an error; We will still try to decode the responded json and send it as a response in the proxy.

@zdebra @marwan-at-work thoughts??

marwan-at-work: Wouldn't the Error field still be a string message that we'd have to guess at? Sorry if I'm missing something.

zdebra: This issue is more about misleading error message. As explained [here](https://github.com/gomods/athens/issues/1718), the error message  `fatal: could not read Username for 'https://github.com': terminal prompts disabled` was given after we rotated `ATHENS_GITHUB_TOKEN` value for a token with insufficient capabilites.

I agree that this probably isn't the best solution to deal with this issue. Maybe some pre-startup verification of github token would be better. What do you think?

marwan-at-work: @zdebra a pre-startup verification is basically the same as as making sure to do `curl <athens>/<private-module>/@latest` right after a deploy. But maybe it helps people running Athens to not forget to do that. So I'd prefer it over trying to interpret error messages because as much as I'd love to have robust error handling, we can never trust error message strings to cover all cases or to never change.

zdebra: @marwan-at-work agreed, I will close this PR then

PR #1723, opened 2021-07-08
proxy sumdb with first not-direct GOPROXY

### LINKED ISSUE #1468: Proxying sumdb behind a proxy or upstream
Now `SumDBs` will proxy sumdb direct to the accually sumdb server insteads of behine the value of `GlobalEndpoint`

```
SumDBs = ["https://sum.golang.org"]  => the sumdb id is sum.golang.org
SumDBs = ["https://goproxy.cn/sumdb/sum.golang.org"]  => the sumdb id is goproxy.cn
```

What 's the best way to describe?

1. if the `SumDB` match `https://(?<proxyHost>.*)/sumdb/(?<sumdb id>.*)`, use id `sumdb id`
2. all sumdb will try if upstream server support, then proxy to `upstream/sumdb/id`

jan4984:

athens is not working for people who can not access https://sum.golang.org. A go mod proxy should also proxy sumdb requests.

## How is the fix applied?

redirect sum.golang.org sumdb to  first not-direct proxy set in GOPROXY, like https://goproxy.cn

Fixes #1468


arschles (commented):
I haven't thought about this problem for a while and I know there are a lot of considerations to think about here. @marwan-at-work @manugupt1 do either of you have thoughts?

PR #1724, opened 2021-07-08
Support download mode file

YaoC:

Since the module filter file is deprecated, the download mode file should be support in helm chart.

## How is the fix applied?

By adding a download mode file as a configmap, and mount it into pod if it is enabled.

Fixes #


DrPsychick:
Closing this PR as it needs to be reopened in Chart repo anyways. Proposed changes are linked and not lost.

PR #1732, opened 2021-09-30
Update alpine version to 3.14.2

guoshimin:
Alpine 3.11 which the athens images is built on will reach end of support soon.

## How is the fix applied?

Use a recent version of alpine.

None


marwan-at-work (approved):
Thank you!

PR #1739, opened 2021-11-13
Fix typos and comments in config.toml

abhay-krishna:
Just fixing typos and clarifying instructions.


pkg/config/config.go
```diff
@@ -26,7 +26,7 @@ type Config struct {
 	GoProxy          string    `envconfig:"GOPROXY"`
 	GoBinaryEnvVars  EnvList   `envconfig:"ATHENS_GO_BINARY_ENV_VARS"`
 	GoGetWorkers     int       `validate:"required" envconfig:"ATHENS_GOGET_WORKERS"`
-	GoGetDir         string    `envconfig:"ATHENS_GOGOET_DIR"`
+	GoGetDir         string    `envconfig:"ATHENS_GOGET_DIR"`
```
marwan-at-work, line 29: This is unfortunately a breaking change. But I think it's a necessary one. Thanks for the catch!

marwan-at-work (approved):
Thanks for the contribution 🎉

pkg/config/config.go
abhay-krishna: Yes, I was thinking it might affect the users of Athens who have configured their environment variables to use the older name in their scripts/automation.

abhay-krishna: You're most welcome! Athens is a great project and we@AWS have had a great experience using it. Thank you!

PR #1742, opened 2021-12-04
Deprecate bzr as it is unmaintained.

manugupt1:

Athens does not build on main. See example build:
https://cloud.drone.io/gomods/athens/2273/1/10

## How is the fix applied?

Alpine moved bzr to unmaintained status. See
https://gitlab.alpinelinux.org/alpine/abuild/-/issues/10000

Deprecate bzr from the Dockerfile as it is deprecated from alpine.
Also, update alpine version.


marwan-at-work (approved):
Thanks for this!

PR #1744, opened 2021-12-09
Set GO111MODULE to auto in golint script– #1743

### LINKED ISSUE #1743: `golint` checks in `make verify` throw GOROOT error
**Describe the bug**

Linting during `verify` Makefile task returns error for each module in Athens as "not in GOROOT" with the `GO111MODULE=off` setting.

**Error Message**

<pre>
```package _/Users/me/path/to/athens/cmd/proxy is not in GOROOT (/usr/local/Cellar/go/1.17.3/libexec/src/_/Users/me/path/to/go/athens/cmd/proxy)
```
</pre>

**To Reproduce**

`make verify`

**Expected behavior**

No errors thrown by `golint` and status 0 returned

 - OS: Darwin Kernel Version 20.6.0: Tue Oct 12 18:33:42 PDT 2021; root:xnu-7195.141.8~1/RELEASE_X86_64 x86_64 (macOS Big Sur 11.6.1)
 - Go version : 1.17.3
 - Proxy version : N/A

**Additional context**

The current verify script also returns status 0 but clogs up output with lines that look like errors. I am possibly not familiar enough with the workings of `GO111MODULE` but setting this to `on` seems to completely clear up the issue, and I don't understand why we would run listing with the modules option deliberately off.

rbrtl:
As mentioned in #1743 the Go modules environment flag is set to `off`
in the script which appears to cause a warning message for each module
of the codebase that it is "not in GOROOT".

Set to `auto` as this allows the same build to be run the original way
should someone choose to delete the `go.mod` file from the project root.

Golint warning for each submodule of Athens "package not in GOROOT" when running Make target `verify`

## How is the fix applied?

Set `GO111MODULE=auto` in `./scripts/check_golint.sh` 

Fixes #1743

### DIFF
--- a/scripts/check_golint.sh
+++ b/scripts/check_golint.sh
@@ -4,4 +4,4 @@
 # Run the linter on everything except generated code
 set -euo pipefail
 
-golint -set_exit_status $(GO111MODULE=off go list ./... | grep -v '/mocks')
+golint -set_exit_status $(GO111MODULE=auto go list ./... | grep -v '/mocks')
--- a/scripts/check_golint.sh
+++ b/scripts/check_golint.sh
@@ -4,4 +4,4 @@
 # Run the linter on everything except generated code
 set -euo pipefail
 
-golint -set_exit_status $(GO111MODULE=auto go list ./... | grep -v '/mocks')
+golint -set_exit_status $(go list ./...)
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ setup-dev-env:
 .PHONY: verify
 verify: ## verify athens codebase
 	./scripts/check_gofmt.sh
-	./scripts/check_golint.sh
+	./scripts/check_govet.sh
 	./scripts/check_deps.sh
 	./scripts/check_conflicts.sh
 
--- a/pkg/stash/with_etcd.go
+++ b/pkg/stash/with_etcd.go
@@ -29,10 +29,13 @@ func WithEtcd(endpoints []string, checker storage.Checker) (Wrapper, error) {
 	}
 	var eg errgroup.Group
 	for _, ep := range endpoints {
-		eg.Go(func() error {
-			_, err := c.Status(ctx, ep)
-			return err
-		})
+		epStat := func(ep string) func() error {
+			return func() error {
+				_, err := c.Status(ctx, ep)
+				return err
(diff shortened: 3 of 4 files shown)


scripts/check_golint.sh
```diff
@@ -4,4 +4,4 @@
 # Run the linter on everything except generated code
 set -euo pipefail
 
-golint -set_exit_status $(GO111MODULE=off go list ./... | grep -v '/mocks')
+golint -set_exit_status $(GO111MODULE=auto go list ./... | grep -v '/mocks')
```
manugupt1, line 7: Thanks for this change. I believe we have now started to build for go1.17.

Go1.17 will ignore this flag: https://go.dev/blog/go116-module-changes.
I believe: https://github.com/gomods/athens/pull/815 was an explicit choice though to make the CI faster

I am not sure if this will work anymore. Maybe we should just remove it: cc @marwan-at-work ?

marwan-at-work: Yes I think it's worth removing. Also, do we even have `/mock`? I wonder if we can do `golinst ./...` or `golint go list ./...`

rbrtl: I'll do it right and raise another PR!

rbrtl: I checked `golint ./...` vs. `golint $(go list ./...)`... I am struggling to prove it definitively prove that the former is linting all the files covered by the latter. The `go list` subshell command version takes noticeably longer, but that's not to say, necessarily, that the `golint` command isn't examining all the correct files. Any input on that much appreciated.

manugupt1: Thank you for doing it right. I just noticed that even golint is deprecated. 
https://github.com/golang/lint
Would you be able to update the code to use `staticcheck` and `go vet`

We already use govet
https://github.com/gomods/athens/blob/fd54babfae1d5fea4f089789e27b23956532f84c/.drone.yml#L32

I would probably just start with govet, open an issue and then do a separate PR for staticcheck. I hope this helps.

manugupt1:
Thanks for making this change. Also congrats on your first PR 🎉

manugupt1 (approved):

arschles:
@rbrtl this is awesome - thank you!

PR #1745, opened 2021-12-10
Update droneci to use go1.17

manugupt1:

Athens now uses go1.17 which is latest. Update drone ci to do the
same for better indication of our CI results.

## How is the fix applied?

Update droneci yaml definition.

N/A

Fixes #


marwan-at-work (approved):
👌🏼

PR #1747, opened 2021-12-15
a new way to pre-fill disk storage

alex-user-go:

Making the process of filling Athens disk storage (if it is run cut off from the internet) more easy and fast

## How is the fix applied?
Use docker-compose file as described in the docs do download go modules

None

### DIFF
--- a/docs/content/configuration/prefill-disk-cache.md
+++ b/docs/content/configuration/prefill-disk-cache.md
@@ -18,7 +18,9 @@ You'll need to produce the following assets from module source code:
 
 The `source.zip` file has a specific directory structure and the `$VERSION.info` has a JSON structure, both of which you'll need to get right in order for Athens to serve up the right dependency formats that the Go toolchain will accept.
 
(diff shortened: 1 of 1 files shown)


docs/content/configuration/prefill-disk-cache.md
```diff
@@ -28,7 +30,7 @@ $ go get github.com/plexsystems/pacmod
 
 This command will install the `pacmod` binary to your `$GOPATH/bin/pacmod` directory, so make sure that is in your `$PATH`.
 
-# Next, run `pacmod` to create assets
+### Next, run `pacmod` to create assets
```
manugupt1, line 33: Small nit, would you be able to change this to a bold'ed text rather than a headline for a bit better readability.

manugupt1:
This is awesome! thanks for this!

manugupt1:
Would you be able to squash the commits and I will merge it immediately; thanks for being patient.

PR #1748, opened 2021-12-18
fix flake on redis-sentinel by depending on redis container / step.

manugupt1:

Redis tests have flakes, this leads to a bad CI.

## How is the fix applied?

Redis tests have flakes, by using depends_on flag; I am trying to reduce these flakes.

I tested by force-pushing multiple commits.

Fixes #

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -69,6 +69,9 @@ steps:
   volumes:
   - name: cache
     path: /go
+  depends_on:
+    - redis
+    - redis-sentinel
 
 - name: docker-latest
   image: plugins/docker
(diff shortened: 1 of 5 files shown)


arschles (approved):
thanks a ton @manugupt1 !

arschles:
@manugupt1 it looks like there's a new redis sentinel related error now: https://cloud.drone.io/gomods/athens/2308/1/9

manugupt1:
thanks @arschles for review.

Adding dependencies did not seem to have fixed the issue; I will see what else can be done. This was what I was trying to fix.

manugupt1:
@arschles do you think adding retries in this case would be acceptable

Issue #1571, opened 2020-03-11
Update documentation


DylanMeeus:
Recently I deployed an instance of athens to Google Cloud, and I believe the docs could be updated a bit. 

1) Versioning is out-of-date, blindly following the docs at first I ended up with `v0.6.0` rather than `v0.7.2`

2) A `Dockerfile` is needed when you use `gcloud submit`, but this step wasn't found in the docs for `Google Cloud Run`, which I eventually did see in the `Google  App Engine` docs.

3) A minor one, and this is perhaps more my mistake than the docs, but copy-pasting the command:
```$ gcloud beta run deploy \
    --image gcr.io/$GOOGLE_CLOUD_PROJECT/athens:v0.6.0 \
    --platform managed \
    --region $ATHENS_REGION \
    --allow-unauthenticated \
    --set-env-vars=ATHENS_STORAGE_TYPE=gcp \
    --set-env-vars=GOOGLE_CLOUD_PROJECT=$GOOGLE_CLOUD_PROJECT \
    --set-env-vars=ATHENS_STORAGE_GCP_BUCKET=$GCS_BUCKET \
    athens
```

does not work, as the region specifier of GCP is missing. (`gcr.io` -> `[region].gcr.io`). Someone with more experience using GCP would probably have seen this faster, but perhaps the docs should mention it?

There's also some typos on those pages. :sweat_smile:

(If these changes are OK, I can update the Hugo markdown files. I suppose that's all that needs to happen?)

marwan-at-work:
@DylanMeeus thank you for reporting this issue. A PR would be great! 

Though I've run Athens on cloud run a number of times, I didn't know we had docs on it 😄 

> Versioning is out-of-date

I think we should put `x` instead of the version number, otherwise we'd always have to update this document. We can also say `check the latest upgrade, below is just an example`

> A Dockerfile is needed 

I'm not sure what you mean, the docs mention that you can just pull the gomods/athens image from Docker Hub, and re-tag it with `gcr.io` and push it back...in that case, there's no need for a Dockerfile at all no? 

> gcr.io -> [region].gcr.io

You should be able to just use `gcr.io`, I believe it's the global one. But let me know why that fails

DylanMeeus:
@marwan-at-work 

If you submit to GCR with this command:
```
$ gcloud builds submit --tag gcr.io/$GOOGLE_CLOUD_PROJECT/athens:vx.y.z
```

The exact error is: 
```
[...]

PR #1751, opened 2022-01-18
Update Cloud Run version and docker steps per #1571

StevenACoffman:
see https://github.com/gomods/athens/issues/1571#issuecomment-599221107

The google cloud commands have changed, as has the version numbers.

## How is the fix applied?

This was already predicted in a comment on an older, closed issue, so I just followed those steps.


marwan-at-work (approved):
Thanks for the update!

PR #1756, opened 2022-02-15
Update etcd

manugupt1:

Updating etcd to the latest release.

## How is the fix applied?

By using go get -u and also update the package path.

Fixes #


marwan-at-work (approved):
Thanks for this 👍🏼 

The changes are quite significant so I'll try to spend some time testing this but feel free to merge.

PR #1758, opened 2022-02-16
Revert retries in redis-flake as builds are failing again.

manugupt1:

Revert retries as main is failing again. There is no point in
slowing down the flaky tests.

## How is the fix applied?

Reverting the previous test

Fixes #


arschles (approved):

manugupt1:
I will merge this as the flake is the same 😕

PR #1564, opened 2020-03-04
bump img version && remove GoProxy

chriscoffee:
## What is the problem I am trying to address?

1. This should fix the issue #1560 which requires the release to be bumped in the chart.
2. Also removes some deprecated code that we don't need anymore since v0.7.0

## How is the fix applied?

Minor change to the `chart` folder and subsequent removals of wherever GoProxy environment was mentioned.

Have a quick peek through and make sure I've not done something stupid and if I need to alter any tests for move to `GoBinaryEnvVars` will y'.

I also gave it a quick test and it seems to workTM

## What GitHub issue(s) does this PR fix or close?

Fixes #1560

PR #1759, opened 2022-02-17
Remove deprecated GOPROXY in favor of GoBinaryEnvVars

manugupt1:

Addressing a TODO to remove GOPROXY.

Follow up from
https://github.com/gomods/athens/pull/1564

## How is the fix applied?

Deleted GOPROXY code.

Fixes #

### DIFF
--- a/charts/athens-proxy/Chart.yaml
+++ b/charts/athens-proxy/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: athens-proxy
-version: 0.5.0
-appVersion: 0.11.0
+version: 0.5.1
+appVersion: 0.11.1
 description: The proxy server for Go modules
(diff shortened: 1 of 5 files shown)


arschles (commented):
Technically this seems fine to remove, but I wonder if it will break folks who were unaware of the deprecation. I trust your judgement either way

manugupt1:
@arschles Since we are updating the chart; only folks who are point to the latest image should face this problem right? Images that are tagged should not have this problem or older versions of chat should not have this problem of breaking users.

Correct me if I am wrong.

arschles:
@manugupt1 yes, I think that's correct. I don't know who or how many people will use new images with old charts, but I assume it's a nonzero, if trivial number. I'm ok with this PR. @marwan-at-work do you have any feelings on this?

arschles (approved):

PR #1768, opened 2022-03-19
Update to go1.18

manugupt1:

go1.18 is released; updating athens to it.

## How is the fix applied?

Updating to go1.18 by changing the docker files and 1.17 references.

Fixes #


arschles (approved):
LGTM

PR #1777, opened 2022-06-03
Resolve memory leaks when using Minio storage

jorng:

When using the Minio storage provider, there is an apparent memory or goroutine leak, causing memory usage to steadily increase in use. After profiling, it appears that it is related to `GetObject` calls on the `minioClient`. I noticed that the `minio.Object` that is returned has a `Close` method, and found [a similar issue in minio/warp](https://github.com/minio/warp/issues/113) that was [fixed by closing the objects](https://github.com/minio/warp/pull/115/files).

## How is the fix applied?

This PR adds deferred `Close` calls in the `storageImpl.Info` and `storageImpl.GoMod` functions so that the objects may be released and garbage collected. The `storageImpl.Zip` function does not use `defer`, since it has the object embedded into the return value, so I added the `Close` only on error.

### DIFF
--- a/pkg/storage/minio/getter.go
+++ b/pkg/storage/minio/getter.go
@@ -21,6 +21,7 @@ func (v *storageImpl) Info(ctx context.Context, module, vsn string) ([]byte, err
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
+	defer infoReader.Close()
 	info, err := ioutil.ReadAll(infoReader)
 	if err != nil {
 		return nil, transformNotFoundErr(op, module, vsn, err)
@@ -38,6 +39,7 @@ func (v *storageImpl) GoMod(ctx context.Context, module, vsn string) ([]byte, er
 	if err != nil {
 		return nil, errors.E(op, err)
 	}
+	defer modReader.Close()
 	mod, err := ioutil.ReadAll(modReader)
 	if err != nil {
 		return nil, transformNotFoundErr(op, module, vsn, err)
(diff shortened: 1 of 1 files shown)


pkg/storage/minio/getter.go
```diff
@@ -62,6 +65,7 @@ func (v *storageImpl) Zip(ctx context.Context, module, vsn string) (storage.Size
 	}
 	oi, err := zipReader.Stat()
 	if err != nil {
+		zipReader.Close()
```
manugupt1, line 68: Why are we not using a defer call here?

manugupt1: Can we also add zipReader.close when GetObject returns an error as well? This is the only one change that I require of this PR and we are good to go.

Thanks for making these changes.

manugupt1 (changes requested):

pkg/storage/minio/getter.go
jorng: > Why are we not using a defer call here?

We can't defer a `Close` for `zipReader`, since it is returned and read elsewhere. All instances I was able to fine appear to properly close the returned value, so the only time we need to close it is on error.

> Can we also add zipReader.close when GetObject returns an error as well?

I specifically avoided that to avoid calling `Close` on `nil`.

manugupt1: Thanks a lot. I realized after reading through the comment left in description. I did not realize GetObject can return nil as well on error (I wish that was not the case though)

manugupt1 (approved):

jorng:
Is there anything else needed to get this merged?

marwan-at-work (approved):
Thanks for the catch 👌🏼

PR #1781, opened 2022-07-27
regrouping go.mod

linzhp:

The direct and indirect deps in go.mod are mixed in two `require` blocks

## How is the fix applied?

Separate direct and indirect deps, which is the default behavior of Go 1.18


arschles (approved):
LGTM

manugupt1:
Thanks!

PR #1787, opened 2022-08-27
Register HTTP client views for stats

### LINKED ISSUE #446: [Discussion] Important metrics to be recorded for athens
Metrics can be used to measure how Athens is performing.  We should have metrics for both Olympus and Proxy. 

Some metrics can be storage, latency, how much data has been transferred

Along with that, we want to see how much do we want to measure. For example, 
A) Disk usage / day.
B) Errors / minute 
C) Time to process a request / second

and so on.

It will be great if we can have a clear idea of the metrics that we want and create them as sub tickets for this one and start measuring one.

linzhp:

HTTP client metrics are still not collected after #1786

## How is the fix applied?

Registering HTTP client views in additional to the server views. Also `ochttp.DefaultServerViews` is [deprecated](https://pkg.go.dev/go.opencensus.io/plugin/ochttp@v0.23.0#DefaultServerViews). The official doc recommends registering server views individually, which I did.

Partially Fixes #446

### DIFF
--- a/pkg/observ/stats.go
+++ b/pkg/observ/stats.go
@@ -5,11 +5,11 @@ import (
 	"net/http"
 	"time"
 
+	"contrib.go.opencensus.io/exporter/prometheus"
 	"contrib.go.opencensus.io/exporter/stackdriver"
 	datadog "github.com/DataDog/opencensus-go-exporter-datadog"
 	"github.com/gomods/athens/pkg/errors"
 	"github.com/gorilla/mux"
-	"contrib.go.opencensus.io/exporter/prometheus"
 	"go.opencensus.io/plugin/ochttp"
 	"go.opencensus.io/stats/view"
 )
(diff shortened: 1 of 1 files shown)


pkg/observ/stats.go
```diff
@@ -92,7 +92,15 @@ func registerStatsStackDriverExporter(projectID string) (func(), error) {
 // registerViews register stats which should be collected in Athens
 func registerViews() error {
 	const op errors.Op = "observ.registerViews"
-	if err := view.Register(ochttp.DefaultServerViews...); err != nil {
+	if err := view.Register(
```
manugupt1, line 95: Is it possible to keep the defaults and append new views, otherwise we will introduce a breaking change

Sorry for the delay

linzhp: Explained in the PR summary, the only two default views that are skipped are not very useful for Athens. I doubt any one wants metrics from them. But sure, I can add them back if we want Athens to be backward-compatible

manugupt1: yes please as we do not know if anyone has come to rely on them

manugupt1 (approved):

manugupt1:
Thanks

PR #1793, opened 2022-10-07
Handle duplicate keys returned by S3 ListObject

linzhp:

S3 API sometimes returns duplicate keys, causing the number of file for a specific module greater than 3.

## How is the fix applied?

This PR makes Athens handle that situation gracefully and log a warning message instead of treating the module as not found.


nrwiersma (approved):
LGTM 🎉 I ran the tests on both the current PR and rebased on main.

I imagine the issue that is seen here is a paging issue, where new keys are added in the middle of paging.

r-ashish (approved):

Issue #1801, opened 2022-11-18
s3.Exists returns 404 even if module is present in S3


ngshiheng:
**Describe the bug**

The current `s3.Exists` method in `checker.go` may return a 404 even if the module cache is present in S3.: 

https://github.com/gomods/athens/blob/cc496afbf1c23ac3cd20bce78ad6d318f62b66db/pkg/storage/s3/checker.go#L25-L30 

Under the hood, `ListObjectsWithContext` (same as `ListObjects`) can only [return up to 1000 objects](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#S3.ListObjects); hence truncating the contents array

> Using `github.com/aws/aws-sdk-go/@v/v1.40.59.mod` in the following examples. 

If S3 were to contain most (if not all) of the existing versions of this module, it would easily amount up to > 1000 objects. As a result, if the module version that the user requested doesn't fall within the 1000 returned objects, it causes Athens to return 404.

**Error Message**

Here are the only 2 error messages emitted in our logs.

```sh
{"http-method":"GET","http-path":"/github.com/aws/aws-sdk-go/@v/v1.40.59.mod","kind":"Not Found","level":"info","module":"","msg":"Not Found","operation":"download.VersionModuleHandler","ops":["download.VersionModuleHandler","pool.GoMod","protocol.GoMod","s3.GoMod"],"request-id":"3270447c-39f8-47af-b74c-14dda9e4cc6f","time":"2022-11-18T06:57:32Z","version":""}

{"http-method":"GET","http-path":"/github.com/aws/aws-sdk-go/@v/v1.40.59.mod","http-status":"404","level":"info","msg":"incoming request","request-id":"3270447c-39f8-47af-b74c-14dda9e4cc6f","time":"2022-11-18T06:57:32Z"}
```

**To Reproduce**

1. Fill Athens S3 disk storage with all releases of `github.com/aws/aws-sdk-go`
2. `curl https://goproxy.company.com/github.com/aws/aws-sdk-go/@v/v1.40.59.mod`
3. Athens will return 404. From the logs, you could see the error messages as shown above

**Expected behavior**

Athens to return response code 200 with the following:

```txt
module github.com/aws/aws-sdk-go

require (
	github.com/jmespath/go-jmespath v0.4.0
	github.com/pkg/errors v0.9.1
	golang.org/x/net v0.0.0-20210614182718-04defd469f4e
)

go 1.11
```

**Environment (please complete the following information):**
[...]

PR #1802, opened 2022-11-18
Update s3 checker to iterate through all objects pages

ngshiheng:

Athens using AWS S3 as storage could return 404 back to users even if the requested module is present in S3 as the page contents could be truncated (max 1000 objects). 

See details here https://github.com/gomods/athens/issues/1801

## How is the fix applied?

Iterate through all S3 pages for the S3 prefix input using https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#S3.ListObjectsPagesWithContext

Fixes https://github.com/gomods/athens/issues/1801

### DIFF
--- a/pkg/storage/s3/checker.go
+++ b/pkg/storage/s3/checker.go
@@ -11,10 +11,6 @@ import (
 	"github.com/gomods/athens/pkg/observ"
 )
 
-const (
-	s3ErrorCodeNotFound = "NotFound"
-)
-
 // Exists implements the (./pkg/storage).Checker interface
 // returning true if the module at version exists in storage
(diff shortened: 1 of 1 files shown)


ngshiheng:
I'm not sure why the CI failed (still looking into it). Locally, i was able to complete `make test-unit-docker` without any failure

r-ashish:
LGTM. Just tested and can confirm that this PR fixes the issues. Thanks @ngshiheng! 🚀

Sadly there's no better way currently other than iterating through all the pages until we find the relevant items.

manugupt1 (approved):

PR #1804, opened 2022-11-30
Update lib/pq to fix cert permissions issues

jorng:

When attempting to connect to a PostgreSQL database using certificate credentials, authentication may fail due to permissions issues on the certificate files. When using Athens in Kubernetes, this issue may be unavoidable when using secrets.

## How is the fix applied?

The `github.com/lib/pq` library has [resolved this issue](https://github.com/lib/pq/commit/8c6de565f76fb5cd40a5c1b8ce583fbc3ba1bd0e) as of v1.10.6, so this commit updates that library to the latest release version (v1.10.7) to resolve the issue in Athens.


nrwiersma (approved):
LGTM 🎉 Run the tests on the PR as it is and rebased version.

r-ashish (approved):
LGTM! 🎉

Issue #1734, opened 2021-10-27
A Call for new Maintainers


arschles:
>TL;DR - we are looking for Gophers who have the time and desire to become new core maintainers of the project. Send a message to me at `arschles` on the Gophers Slack if you're interested.

Since its launch in August 2018, Athens has grown to become an important project as an open source Go module proxy and datastore for the Go community. Over the past 3 years, we (the existing core maintainers) have focused on a stable set of useful features while ensuring Athens is extremely contributor-friendly. That work has paid off. We now have 100 contributors to date and the software is used by a diverse [range of organizations](https://github.com/gomods/athens/issues/1323) around the world.

As the Athens project has evolved, we've identified a need for new core maintainers who can contribute the time and bandwidth is deserves to truly thrive. I and other maintainers know the project deserves more time than we have to contribute today. The Go community depends on Athens, and we believe it's important to keep the project and healthy and vibrant into the future.

## Maintainership Qualifications

If you're interested in becoming a core maintainer for the project, we are looking for some minimum qualifications:

- You are familiar with the project.
- You have contributed to the project before. This could mean contributing a pull request, engaging in discussion on issues, helping with releases, etc…
- You have the time to contribute regularly

## Maintainership Responsibilities

Since the project's feature set has remained relatively stable over the past year or so, much of the initial work would involve the following tasks:

- Triaging issues
- Reviewing pull requests
- Fixing bugs
- Cutting new releases as necessary

The time commitment for this has historically been around 5 hours per week, and I would help you to ramp up. Once you are familiar with the above, you would then have a voice on whether to continue new feature work.

## What to do if You're Interested

**If you are interested in becoming a core maintainer for Athens, please send a direct message to me at `arschles` on the Gophers Slack to let me know you're interested**. I will reach out to you to have an initial discussion.

Maintainership of this project is rewarding for many reasons, but primarily because you get to build software that is core to so many developer workflows across the community and world. I hope to hear from you!

-- Aaron, on behalf of the Athens Core Team

Edit: thanks to @asw101 and @fedepaol for proofreading this post.

DrPsychick:
We'll chat in the beginning of next year and see how/what I can contribute and if we find other folks willing to join

https://www.linkedin.com/posts/pascal-tobias-sick-57436983_athens-go-go-activity-7011808413319843840-qttT/

manugupt1:
I would like to have @linzhp and @ngshiheng as additional maintainers if they would like to and agree to.
[...]

PR #1805, opened 2022-11-30
Handle SIGTERM on unix-like OS

jorng:

When running Athens on a unix-like OS (linux, darwin, etc), the TERM signal is not being handled to perform a graceful shutdown, only `INT`. This is especially problematic on Kubernetes, where the default signal sent to pods is `TERM`.

## How is the fix applied?

This commit adds a new package to get the appropriate signals based on the operating system, using `go:build` tags. For `unix`, it includes `os.Interrupt` AND `syscall.SIGTERM`, but only `os.Interrupt` for others, which may not support that signal.

### DIFF
--- a/internal/shutdown/signals.go
+++ b/internal/shutdown/signals.go
@@ -1,0 +1,15 @@
+//go:build unix
+
+package shutdown
+
+import (
+	"os"
+	"syscall"
+)
+
+// GetSignals returns the appropriate signals to catch for a clean shutdown, dependent on the OS.
+//
+// On Unix-like operating systems, it is important to catch SIGTERM in addition to SIGINT.
+func GetSignals() []os.Signal {
+	return []os.Signal{os.Interrupt, syscall.SIGTERM}
+}
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -13,6 +13,7 @@ import (
 	_ "net/http/pprof"
 
 	"github.com/gomods/athens/cmd/proxy/actions"
+	"github.com/gomods/athens/internal/shutdown"
 	"github.com/gomods/athens/pkg/build"
 	"github.com/gomods/athens/pkg/config"
 )
@@ -51,8 +52,9 @@ func main() {
 
 	go func() {
 		sigint := make(chan os.Signal, 1)
-		signal.Notify(sigint, os.Interrupt)
-		<-sigint
+		signal.Notify(sigint, shutdown.GetSignals()...)
+		s := <-sigint
+		log.Printf("Recived signal (%s): Shutting down server", s)
 
 		// We received an interrupt signal, shut down.
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second*time.Duration(conf.ShutdownTimeout))
--- a/internal/shutdown/signals_notunix.go
+++ b/internal/shutdown/signals_notunix.go
@@ -1,0 +1,10 @@
+//go:build !unix
+
+package shutdown
+
+import "os"
+
+// GetSignals returns the appropriate signals to catch for a clean shutdown, dependent on the OS.
+func GetSignals() []os.Signal {
+	return []os.Signal{os.Interrupt}
+}


cmd/proxy/shutdown/signals.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"os"
+	"syscall"
+)
+
+// GetSignals returns the appropriate signals to catch for a clean shutdown, dependent on the OS.
+//
+// On Unix-like operating systems, it is important to catch SIGTERM in addition to SIGINT.
+func GetSignals() []os.Signal {
```
nrwiersma, line 13: Does this really need to be a package? Could it not instead go into `cmd/proxy/signals_unix.go`? In that case it would not need to be exposed. Better yet, it could just be a var.

jorng: Sure thing, if that's how you want it. I was just attempting to keep the `main` package clean, since it was currently only a single `main.go` file. I've always avoided putting OS-specific things into `main`.

If we just want to avoid exporting, another option would be to put it into `internal/shutdown`, instead of `cmd/proxy/shutdown`.

jorng: Also... I'm a bit confused. Are you a maintainer? You don't seem to have any previous contributions to this repo...

nrwiersma: No, I am not a maintainer, I am just helping out with some PRs.

nrwiersma (approved):
LGTM 🎉

cmd/proxy/shutdown/signals.go
manugupt1: Hello, 
@nrwiersma is helping us out as we are looking for new maintainers.

https://github.com/gomods/athens/issues/1734

internal/shutdown/signals.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"os"
+	"syscall"
+)
+
+// GetSignals returns the appropriate signals to catch for a clean shutdown, dependent on the OS.
+//
+// On Unix-like operating systems, it is important to catch SIGTERM in addition to SIGINT.
+func GetSignals() []os.Signal {
+	return []os.Signal{os.Interrupt, syscall.SIGTERM}
```
manugupt1, line 14: Should this be https://pkg.go.dev/syscall#SIGINT? Should we add SIGKILL also?

So, far athens does not hav ea lot of linux specific code afaik? Does it make sense to find out the os on runtime rather than have build tags and have a single package.

nrwiersma: `os.Interrupt` is mapped to `syscall.SIGINT`. Also, containers will get a `syscall.SIGTERM` to indicate the container should shutdown, so in general it should be handled along side `os.Interrupt`.  I don't know that `syscall.SIGTERM` exists in all OSes, so it is safer to use build tags IMO. If it can be proven that it always exists, then the build tags would not be needed.

`syscall.SIGKILL` can in general not be caught, so there is not much point in trying.

manugupt1: I was suggesting something like this: https://github.com/gomods/athens/blob/a1553999bb8bfc57d9962ca3ee3ed992640b8f2b/pkg/module/prepare_env.go#L37

nrwiersma: The issue I see there, is that should the signal not exist in the `syscall` package for the OS, you would not be able to build on that platform anymore, where build tags would just exclude the code from the build process in non-supporting OSes.

nrwiersma: To be more clear, `SIGTERM` in the `syscall` package (and all signals in the syscall package) is protected by platform specific build tags, and so there is no guarantee that it exists for all OSes.

jorng: Right, that is why I have it protected by build tags. This is what I had to do to get it shutting down properly in k8s.

manugupt1: Thanks! Lgtm

manugupt1 (approved):

PR #1806, opened 2022-11-30
Make shutdown timeout configurable

jorng:

Currently, when shutting down the server (via `SIGINT`), the shutdown closes any open connections after only 10 seconds (via a `context.WithTimeout`). This does not provie a lot of time for longer operations, such as listing versions, or downloading a larger module zip file.

When running in Kubernetes, and scaling instances or changing config, this causes a lot of dropped connections and gateway errors. 10 seconds is arguably much too short, and should be configurable.

## How is the fix applied?

This commit increases that default to 60 seconds, and adds a config variable to allow users to specify their desired timeout.

### DIFF
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -55,7 +55,7 @@ func main() {
 		<-sigint
 
 		// We received an interrupt signal, shut down.
-		ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*time.Duration(conf.ShutdownTimeout))
 		defer cancel()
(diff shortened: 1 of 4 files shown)


nrwiersma (approved):

nrwiersma:
You obviously need up fix the failing test.

nrwiersma (changes requested):
Just realised you have not updated a test. Could you also rebase the PR please.

jorng:
Rebased and fixed the failing test.

nrwiersma (approved):
LGTM 🎉

manugupt1 (approved):

manugupt1:
thanks!

PR #1799, opened 2022-11-11
Update to go1.19

manugupt1:
Updating to go1.19. This will update canary images.

PR #1812, opened 2023-01-10
Adding new fields to RevInfo

linzhp:

All recent CI jobs of Athens are failing.

## How is the fix applied?

Go 1.9 added a new `Origin` field to the [RevInfo](https://pkg.go.dev/cmd/go/internal/modfetch@go1.19#RevInfo) struct. When a test tries to decode the .info file downloaded by go mod, the addition field cause the decoder to fail, because the decoder `DisallowUnknownFields()`. Copying adding the new fields introduced in Go 1.19 fixes the issue.

The changes in this PR is taken from #1799.

This would fix the failing test in #1793

### DIFF
--- a/pkg/download/protocol_test.go
+++ b/pkg/download/protocol_test.go
@@ -294,6 +294,7 @@ func TestInfo(t *testing.T) {
 
 			var info storage.RevInfo
 			dec := json.NewDecoder(bytes.NewReader(bts))
+			dec.DisallowUnknownFields()
 			err = dec.Decode(&info)
 			require.NoError(t, err)
 
--- a/pkg/download/protocol_test.go
+++ b/pkg/download/protocol_test.go
@@ -294,7 +294,6 @@ func TestInfo(t *testing.T) {
 
 			var info storage.RevInfo
 			dec := json.NewDecoder(bytes.NewReader(bts))
-			dec.DisallowUnknownFields()
 			err = dec.Decode(&info)
 			require.NoError(t, err)
 
--- a/pkg/download/protocol_test.go
+++ b/pkg/download/protocol_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"github.com/stretchr/testify/assert"
 	"io/ioutil"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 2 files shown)


pkg/download/protocol_test.go
```diff
@@ -294,7 +294,6 @@ func TestInfo(t *testing.T) {
 
 			var info storage.RevInfo
 			dec := json.NewDecoder(bytes.NewReader(bts))
-			dec.DisallowUnknownFields()
```
manugupt1, line 297: Should we update like this:
https://github.com/gomods/athens/pull/1799/files#diff-34d858fcb9045702dd09c49922588e96b3200e22e41dfc6c0a3201406489396b

linzhp: We can do that too, but I am not sure whether Athens is interested in those additional fields. If we `DisallowUnknownFields`, we have to update the `RevInfo` every time Go add new fields into it, which is an unnecessary maintenance cost if we don't use those new fields.

manugupt1: I think we should extend RevInfo to keep it compatible with go proxy. See:
```
➜  ~ curl https://proxy.golang.org/golang.org/x/mod/@v/v0.7.0.info
{"Version":"v0.7.0","Time":"2022-11-08T21:31:42Z","Origin":{"VCS":"git","URL":"https://go.googlesource.com/mod","Ref":"refs/tags/v0.7.0","Hash":"7c05a442b7c1d1a107879b4a090bb5a38d3774a1"}}% ➜  ~ 
```

linzhp: OK. I think #1799 should fix the current CI issue too. Let me close this one then.

Can you restart a CI build for #1799?

manugupt1: Sorry, that change was not complete since it was also trying to do a lot more things. Would it be possible for you to expand or takeover that PR as I have not had the time to finish it?

linzhp: I can take the RevInfo related part to a separate PR and land it first, so we can fix CI and unblock other PR. For rest of Go 1.19 upgrade, I will see how far I can get

manugupt1 (approved):
Thanks a lot. I will merge it.

PR #1816, opened 2023-01-25
feat: replace ioutil with io and os

nrwiersma:

`io/ioutil` package was deprecated in Go 1.16 and should not longer be used.

## How is the fix applied?

Replaces all occurrences of `io/ioutil` with functions on the `io` or `os` package.

The replacement functions can be found in the package docs for [`io/ioutil`](https://pkg.go.dev/io/ioutil) for reference.


r-ashish (approved):

r-ashish:
LGTM! 🎉

PR #1817, opened 2023-01-26
Pass Athens's logger to the Redis package

mikesep:

The Redis package used by `pkg/stash/with_redis*` occasionally logs bits of information, such as when a Redis Sentinel is discovered or changes, but it uses its own logger by default. If Athens is configured to log in JSON but Redis is logging in plain text, this creates problems when parsing the log output.

## How is the fix applied?

Make an adapter struct around `pkg/log.Logger` to implement the interface that Redis v8 expects, then pass the logger to the `WithRedis*` functions to call `redis.SetLogger`.

Haven't found an Issue specifically for this.

### DIFF
--- a/cmd/proxy/actions/app_proxy.go
+++ b/cmd/proxy/actions/app_proxy.go
@@ -1,6 +1,7 @@
 package actions
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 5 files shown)


r-ashish:
Thanks for the fix!

The Changes LGTM, just wondering whether you've tested this already to make sure it's working as expected?

r-ashish (approved):

mikesep:
Hi Ashish, thanks for reviewing!

> just wondering whether you've tested this already to make sure it's working as expected?

Yes, I've been running this change in my group's Athens instance for the last day or so, and Redis traces are definitely showing up as JSON rather than plain text. 😄

PR #1818, opened 2023-01-27
Rename test-only files to end in _test.go

mikesep:

This should help prevent accidentally mixing non-test code with test-only code.

## How is the fix applied?

Renamed test-only files to end in `_test.go`.

None.


mikesep:
Never mind, this totally doesn't work since these helper functions need to be exported so multiple packages can use them.

PR #1821, opened 2023-01-29
test(pkg/config): use `t.Setenv` to set env vars

Juneezee:

Ensure no leftover environment variables between tests in `pkg/config/config_test.go`.

## How is the fix applied?

Starting from Go 1.17, we can use `t.Setenv` to set environment variable in test. The environment variable is automatically restored to its original value when the test and all its subtests complete. This ensures that each test does not start with leftover environment variables from previous completed tests.

Reference: https://pkg.go.dev/testing#T.Setenv

```go
func TestFoo(t *testing.T) {
	// before
	oldVal := os.Getenv("key")
	defer os.Setenv("key", oldVal)
	os.Setenv(key, "new value")
	
	// after
	t.Setenv(key, "new value")
}
```

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -100,7 +100,7 @@ func TestEnvOverrides(t *testing.T) {
 
 	envVars := getEnvMap(expConf)
 	for k, v := range envVars {
-		os.Setenv(k, v)
+		t.Setenv(k, v)
 	}
 	conf := &Config{}
 	err := envOverride(conf)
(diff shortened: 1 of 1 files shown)


pkg/config/config_test.go
```diff
@@ -189,47 +187,21 @@ func TestStorageEnvOverrides(t *testing.T) {
...
-	}
-	// unset all environment variables
-	envVars := getEnvMap(emptyConf)
-	envVarBackup := map[string]string{}
-	for k := range envVars {
-		oldVal := os.Getenv(k)
-		envVarBackup[k] = oldVal
-		os.Unsetenv(k)
-	}
+	os.Clearenv()
```
r-ashish, line 204: CMIIW but this won't restore the env after the test is done, right?

Would something like this work?

```
for k := range envVars {
    t.Setenv(k, "")
    os.Unsetenv(k)
}
```

r-ashish (changes requested):

pkg/config/config_test.go
Juneezee: Yes, you are right.

For `TestParseExampleConfig`, it is not necessary to restore the environment variables. The aim of this PR is to ensure that none of the tests in `config_test.go` depend on one or more environment variables set by another test.

Actually, `os.Clearenv()` is also optional. None of our tests have leftover environment variables so I wrote `os.Clearenv()` here as a fail-safe.

r-ashish: Right, makes sense. Thanks!

r-ashish (approved):

PR #1826, opened 2023-02-16
Fix download mode leak for latest and list handlers

ngshiheng:

Despite being set to restrict download via the [download mode file](https://docs.gomods.io/configuration/download/#blocking-certain-modules), the `/list` and `/latest` handlers are still returning data to the user.

This presents a serious security issue, as sensitive module information could potentially be exposed to unauthorized parties.

## How is the fix applied?

Check if `DownloadMode` is set to `"none"` for a specific module. If so, we just return `errors.KindNotFound`.

Fixes # https://github.com/gomods/athens/issues/1774


marwan-at-work (commented):
The download mode is strictly for what to do "when a module is not found", which is only for `.zip/.info/.mod` endpoint, not for `/list` and `latest`. For those, you can check the network mode configuration. 

Maybe the docs were not clear enough?

ngshiheng:
Ah, the docs [did mention blocking certain modules as a use case](https://docs.gomods.io/configuration/download/#blocking-certain-modules). 

From within our organization, this is one of the primary use cases of Athens - to act as a vendor to supply a set of internal modules to our downstream:

```hcl
downloadURL = "https://proxy.golang.org"

mode = "sync"

download "company.gitlab.com/private/one*" {
    mode = "sync"
}

download "company.gitlab.com/private/two*" {
    mode = "sync"
}

download "company.gitlab.com/*" {
    mode = "none"
}
```

PR #1828, opened 2023-02-22
chore: lint code with golangci-lint

nrwiersma:

To make the codebase more uniform, and to catch additional security issues, a static analyser should be used.

## How is the fix applied?

golangci-lint was configured, added to the workflow and any existing issues where fixed.

As it is no longer considered "safe" to cast errors directly, and as the `pkg/errors` package is used basically everywhere, the functions `errors.IsErr` and `errors.AsErr` where added that wrap the std lib functions. The `Err` postfix was added as `errors.Is` already exists. If prefered, `errors.Is` can be renamed, and the wrapper function can better mimic the std lib.

**Note:** It would be best to only merge this once the CI tests are fixed.


.github/workflows/lint.yml
```diff
@@ -0,0 +1,40 @@
...
+      fail-fast: false
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
```
manugupt1, line 27: https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs
Can we have something similar to this?

manugupt1:
https://github.com/gomods/athens/pull/1823/files has a lint step in build; which one do we prefer?

Makefile
```diff
@@ -43,10 +43,12 @@ docs: ## build the docs docker image
 setup-dev-env:
 	$(MAKE) dev
 
+.PHONY: lint
+lint:
+	@golangci-lint run ./...
```
manugupt1, line 48: Can we not do everything through makefile so that the user's golangci version and gha's golangci's version is the same?
We will have to do go install though.
Perhaps we can re-use docker instruction to make it platform agonstic and independent of the users environment.
https://golangci-lint.run/usage/install/

.github/workflows/lint.yml
```diff
@@ -0,0 +1,35 @@
...
+        uses: actions/checkout@v2
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          check-latest: true
+          cache: true
+
+      - name: Lint Go Code
```
manugupt1, line 32: This step would be simply be make verify or make lint whatever we decide to.

Makefile
nrwiersma: > We will have to do go install though.

You cannot Go install golanci-lint, it does not work. And while I am good with providing a docker version, golangci-lint uses a cache to make it fast. Just running it in docker will be very slow and painful. The Github action allows this caching in the CI, and running locally uses the cache, which will change position between systems.

TLDR; You can, but I would not suggest it.

Makefile
```diff
@@ -43,10 +45,16 @@ docs: ## build the docs docker image
...
+
+.PHONY: lint-docker
+lint-docker:
+	@docker run -t --rm -v $(CURDIR):/app -w /app golangci/golangci-lint:$(GOLANGCI_LINT_VERSION) golangci-lint run ./...
```
manugupt1, line 54: What's the benefit of @ ?

Makefile
manugupt1: I will leave upto you to decide if you want to use Docker or not. 

However, I think we should consider that this file moves to ci.yml and have the Makefile rule invoke linting. There are some benefits:
1. Makefile will make it consistent between Github actions and local environment.
2. Common ci.yaml will help it easier to control go version and golangci-lint versions as well.

I am not convinced above caching and speed since other CI jobs will definitely take longer than golangci-lint.

Makefile
nrwiersma: It stops make from echoing the command run.

.github/workflows/ci.yml
```diff
@@ -53,8 +53,10 @@ jobs:
         with:
           go-version-file: 'go.mod'
           cache: true
-      - name: Run linter
```
manugupt1, line 56: Sorry for going back and forth. I meant we could have a new job in the same yaml.
https://github.com/containerd/nerdctl/blob/main/.github/workflows/test.yml is a good example with lint, project test-unit being different jobs.

manugupt1:
Now lint is slower I made a mistake; sorry! but let's merge this PR.
Thanks a lot!

manugupt1 (approved):

PR #1830, opened 2023-02-23
chore: improve docker testing

nrwiersma:

This PR fixes the docker e2e tests that are currently broken and additionally adds mod cache mounting to make testing in docker faster.

## How is the fix applied?

The mod cache location is determined from the go environment and volume mounted into test containers.
As the docker e2e testing seems to not have been updated when e2e tests changed, I fix that as well.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,7 +37,7 @@ services:
       dockerfile: Dockerfile.test
       args:
         GOLANG_VERSION: 1.19
-    command: ["./scripts/test_e2e.sh"]
+    command: "/bin/sh -c 'cd e2etests && go test --tags e2etests'"
   azurite:
     image: arafato/azurite:2.6.5
     ports:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,6 +28,8 @@ services:
       - ATHENS_MONGO_STORAGE_URL=mongodb://mongo:27017
       - TIMEOUT=20 # in case the mongo dependency takes longer to start up
       - ATHENS_STORAGE_TYPE=mongo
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -37,7 +39,9 @@ services:
       dockerfile: Dockerfile.test
       args:
         GOLANG_VERSION: 1.19
-    command: ["./scripts/test_e2e.sh"]
+    command: "/bin/sh -c 'cd e2etests && go test --tags e2etests'"
+    volumes:
+      - ${GO_MODCACHE}:/go/pkg/mod
```
manugupt1, line 44: Won't this fail if GO_MODCACHE is not set?

nrwiersma: No, you get a warning that it is not set, and then it just downloads the mods.

manugupt1: Hmm! I tried locally. I am getting the following error if GO_MODCACHE is not set or empty

```
WARNING: using --exit-code-from implies --abort-on-container-exit
Creating athense2e_teste2e_1 ... error

ERROR: for athense2e_teste2e_1  Cannot create container for service teste2e: create .: volume name is too short, names should be at least two alphanumeric characters

ERROR: for teste2e  Cannot create container for service teste2e: create .: volume name is too short, names should be at least two alphanumeric characters
ERROR: Encountered errors while bringing up the project.
make: *** [Makefile:76: test-e2e-docker] Error 1
➜  athens git:(fix-docker-e2e) ✗ 
```

nrwiersma: Fair

PR #1831, opened 2023-02-23
fix release workflow

r-ashish:
Follow up fix for https://github.com/gomods/athens/pull/1823


r-ashish:
Tested and confirmed that both canary and commit tags are being pushed: 

- https://hub.docker.com/r/gomods/athens/tags
- https://hub.docker.com/r/gomods/athens-dev/tags

manugupt1 (approved):
Thanks!

.github/workflows/release.latest.yml
```diff
@@ -21,3 +21,4 @@ jobs:
           file: cmd/proxy/Dockerfile
           build-args: VERSION=${{github.sha}}
           tags: gomods/athens:latest
+          push: true
```
nrwiersma, line 24: We are not pushing a tagged image here, only latest, which no one uses in production.

r-ashish: Right, will update later. Idk how it was working in drone though 🤔. (maybe I missed a step while migrating)

nrwiersma: https://github.com/gomods/athens/commit/3f26845cffd789688277c8a79d1cc11159a2c689#diff-b54b39f1afced2465e1f3641db9d5bbf4f3a7fcf890996dfedd3c197bcb7f8c7L123

r-ashish: Updated. Thanks for pointing out!

nrwiersma (approved):
LGTM 🎉

PR #1836, opened 2023-02-27
chore: use builtin action linter package

nrwiersma:

The CI linter is currently uses docker, which is very slow due to not having a cache.

## How is the fix applied?

Use the provided Golanci-lint actions package which has caching builtin.


manugupt1 (approved):

r-ashish (approved):

manugupt1:
https://github.com/gomods/athens/pull/1850 takes around the same time. 2m in both builds. Am I missing something?

nrwiersma:
Fresh caches, cause of the new Go version.

PR #1840, opened 2023-02-28
Update Exists method to use HeadObjectWithContext

ngshiheng:

The .mod endpoint for modules with a large number of tags (such as the [aws-sdk-go](https://github.com/aws/aws-sdk-go) module) is currently experiencing slow performance. 

This is due to a change in the S3 filtering prefix introduced during the implementation of "Using directory as prefix for S3 #1720."

When listing objects using the S3 prefix `"github.com/aws/aws-sdk-go/@v"`, which returns approximately 4000 S3 objects, it takes almost **6 seconds** to complete the .mod endpoint request.

### About PR #1720 

The original PR https://github.com/gomods/athens/pull/1720 mentioned that list objects on S3 with prefixes like `"github.com/golang/mock/@v/v1.4.4."` will return empty list, this is no longer the case:

Screenshot 2023-02-28 at 3 39 54 PM

```sh
❯ aws s3 ls s3://s3-cache/github.com/golang/mock/@v/v1.4.4.
2022-10-28 15:43:05         50 v1.4.4.info
2022-10-28 15:43:05        102 v1.4.4.mod
2022-10-28 15:43:05      95367 v1.4.4.zip

❯ aws s3 ls s3://s3-cache/github.com/golang/mock/@v/v1.4.4
2022-10-28 15:43:05         50 v1.4.4.info
2022-10-28 15:43:05        102 v1.4.4.mod
2022-10-28 15:43:05      95367 v1.4.4.zip
```

## Examples

1. Result when using the [current HEAD commit](https://github.com/gomods/athens/blob/af82a7a9cd7cc32db37dba102f1b1e18297702fc/pkg/storage/s3/checker.go#L24) (af82a7a9):
```sh
❯ time curl http://localhost:3000/github.com/aws/aws-sdk-go/@v/v1.40.58.mod
module github.com/aws/aws-sdk-go

require (
	github.com/jmespath/go-jmespath v0.4.0
	github.com/pkg/errors v0.9.1
	golang.org/x/net v0.0.0-20210614182718-04defd469f4e
)

go 1.11
curl http://localhost:3000/github.com/aws/aws-sdk-go/@v/v1.40.58.mo  0.02s user 0.01s system 0% cpu 5.986 total

❯ aws s3 ls s3://s3-cache/github.com/aws/aws-sdk-go/@v --recursive | wc -l
    4542
```

In the example above, we're listing objects using the S3 prefix `"github.com/aws/aws-sdk-go/@v"`, which returns **~4000** S3 objects. To iterate through that many objects, it took us ~5.9s to complete the `.mod` endpoint request.

2. Result when using [Athens v0.11.0](https://github.com/gomods/athens/blob/v0.11.0/pkg/storage/s3/checker.go#L26)
```sh
❯ time curl http://localhost:3000.com/aws/aws-sdk-go/@v/v1.40.58.mod
module github.com/aws/aws-sdk-go

require (
	github.com/jmespath/go-jmespath v0.4.0
	github.com/pkg/errors v0.9.1
	golang.org/x/net v0.0.0-20210614182718-04defd469f4e
)

go 1.11
curl http://localhost:3000.com/aws/aws-sdk-go/@v/v1.40.58.mod  0.02s user 0.01s system 13% cpu 0.246 total

❯ aws s3 ls s3://s3-cache/github.com/aws/aws-sdk-go/@v/v1.40.59. --recursive | wc -l
       3
```

In the example above, we're listing objects using the S3 prefix `"github.com/aws/aws-sdk-go/@v/v1.40.58."`, which returns only **3** S3 objects. As a result, it only took ~0.2s to complete the `.mod` request.

## How is the fix applied?

Essentially I propose to just undo the change of https://github.com/gomods/athens/pull/1720.

This will allow us to list objects on S3 with prefixes like `"github.com/aws/aws-sdk-go/@v/v1.40.58."`, which returns only 3 S3 objects in this example. This results in a much faster response time of only 0.2 seconds.

## References

- [Old AWS docs](https://web.archive.org/web/20210507070946/https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-prefixes.html)
- [Latest AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-prefixes.html)

I hope that this fix will improve the performance of the .mod endpoint for modules with a large number of tags. Thank you for considering this PR!

### DIFF
--- a/pkg/storage/s3/checker.go
+++ b/pkg/storage/s3/checker.go
@@ -2,12 +2,13 @@ package s3
 
 import (
 	"context"
+	goerr "errors"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/gomods/athens/pkg/config"
 	"github.com/gomods/athens/pkg/errors"
-	"github.com/gomods/athens/pkg/log"
 	"github.com/gomods/athens/pkg/observ"
 )
 
@@ -18,27 +19,25 @@ func (s *Storage) Exists(ctx context.Context, module, version string) (bool, err
 	ctx, span := observ.StartSpan(ctx, op.String())
 	defer span.End()
 
-	lsParams := &s3.ListObjectsInput{
-		Bucket: aws.String(s.bucket),
-		Prefix: aws.String(config.PackageVersionedName(module, version, "")),
+	fileKeys := []string{
+		config.PackageVersionedName(module, version, "info"),
+		config.PackageVersionedName(module, version, "mod"),
+		config.PackageVersionedName(module, version, "zip"),
 	}
-	found := make(map[string]struct{}, 3)
-	err := s.s3API.ListObjectsPagesWithContext(ctx, lsParams, func(loo *s3.ListObjectsOutput, lastPage bool) bool {
-		for _, o := range loo.Contents {
-			if _, exists := found[*o.Key]; exists {
-				log.EntryFromContext(ctx).Warnf("duplicate key in prefix %q: %q", *lsParams.Prefix, *o.Key)
-				continue
-			}
-			if *o.Key == config.PackageVersionedName(module, version, "info") ||
-				*o.Key == config.PackageVersionedName(module, version, "mod") ||
-				*o.Key == config.PackageVersionedName(module, version, "zip") {
-				found[*o.Key] = struct{}{}
-			}
+
+	for _, key := range fileKeys {
+		params := &s3.HeadObjectInput{
+			Bucket: aws.String(s.bucket),
+			Key:    aws.String(key),
 		}
-		return len(found) < 3
-	})
-	if err != nil {
-		return false, errors.E(op, err, errors.M(module), errors.V(version))
+
+		if _, err := s.s3API.HeadObjectWithContext(ctx, params); err != nil {
+			var s3Err awserr.Error
+			if goerr.As(err, &s3Err) && s3Err.Code() == "NotFound" {
+				return false, nil
+			}
+			return false, errors.E(op, err, errors.M(module), errors.V(version))
+		}
 	}
-	return len(found) == 3, nil
+	return true, nil
 }
--- a/pkg/storage/s3/checker.go
+++ b/pkg/storage/s3/checker.go
@@ -2,7 +2,6 @@ package s3
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/s3"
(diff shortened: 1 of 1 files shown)


linzhp:
I can take a look this week

linzhp:
OK. I think we can merge this for now. However, I feel we should change the directory layout to avoid putting too many files in one directory. Something like `/github.com/aws/aws-sdk-go/@v/v1.44.90/v1.44.90.mod`, `/github.com/aws/aws-sdk-go/@v/v1.44.90/v1.44.90.info`

 @r-ashish What do you think?

marwan-at-work:
> OK. I think we can merge this for now. However, I feel we should change the directory layout to avoid putting too many files in one directory. Something like /github.com/aws/aws-sdk-go/@v/v1.44.90/v1.44.90.mod, /github.com/aws/aws-sdk-go/@v/v1.44.90/v1.44.90.info

Wouldn't this be a breaking change?

linzhp:
It won't. The directory layout in a blob storage is implementation details of Athens and not visible to its clients. There is no API changes. The existing deployments will have a cache miss after the deploy the new layout. But the cache will be populated after that and no migration is needed.

marwan-at-work:
> The existing deployments will have a cache miss after the deploy the new layout. But the cache will be populated after that and no migration is needed.

That's the breaking change no? It won't get the old (already stored) module, it will refetch it from the VCS causing two things: 

1. Duplication of a module version, we need to clean up and delete the old path now? (Or ideally rename it)
2. Immutability is broken: if the module has changed in the VCS, the checksums will fail.

pkg/storage/s3/checker.go
```diff
@@ -21,7 +20,7 @@ func (s *Storage) Exists(ctx context.Context, module, version string) (bool, err
 
 	lsParams := &s3.ListObjectsInput{
 		Bucket: aws.String(s.bucket),
-		Prefix: aws.String(fmt.Sprintf("%s/@v", module)),
+		Prefix: aws.String(config.PackageVersionedName(module, version, "")),
```
manugupt1, line 23: Can we use HeadObject as @linzhp suggested if it works?

ngshiheng: I think `HeadObject` would work here 👍 

While I am happy to incorporate that into this PR, my concern is that we would be making 2 additional requests (3 in total for `.mod`, `.zip` and `.info`  files) using `HeadObject` if my understanding is correct. 

AFAIK there's no way we can batch the 3 calls using `HeadObject`, unfortunately.

WDYT?

manugupt1: Head does not scan the entire bucket like list. It does an exact match. List is charged by AWS while Head I do not think is.
If you can, have three go routines start with a wait group; I think your latency should be under a second unless AWS is doing something really wrong.

ngshiheng: After some research, I think you're right about the API pricing. My initial concern was more on the AWS S3 request rate limits (afaik it's at 5,500 req/s see [this doc](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html))

I've updated the `Exist` method for review.

Since we are only sending 3 requests in the loop, and they are not particularly computationally intensive, it is unlikely that using a Go routine would result in a significant performance improvement

PR #1842, opened 2023-03-02
Removing Exists() check from S3 getters

linzhp:

Athens first checks the existence of a module/version before:

1. Reading mod, info or zip file
2. Deleting a module
3. Writing a module/version fetched from VCS

Some of these checks are not necessary, because 

* we can know that from the returned error of 1
* For 2, S3 doesn't return errors when we tries to delete an non-existing blob, but users may not need to know that either. In fact, I don't see where Athens is deleting a blob, any pointers?

Removing these checks will improve the atomicity and performance of the operations, which would also address the performance issue described in #1840

3 is the only case when we need to check for existence to avoid duplicate uploads, but we can potentially make `HeadObject` calls in the [Uploader](https://pkg.go.dev/github.com/gomods/athens/pkg/storage/module#Uploader), which is called concurrently. It would make the check more granular: instead of uploading all 3 files again when one file is missing, we can upload only the missing file. 

## How is the fix applied?

* removing the `Exists()` check from getter.go and instead read the return error from `GetObjectWithContext` to determine whether it is not found.

If this approach is OK, I can apply similar changes to other storage clients.

@ngshiheng can you test this?

@marwan-at-work @r-ashish What do you think?

### DIFF
--- a/pkg/storage/s3/getter.go
+++ b/pkg/storage/s3/getter.go
@@ -2,11 +2,12 @@ package s3
 
 import (
 	"context"
+	errs "errors"
 	"fmt"
-	"github.com/aws/aws-sdk-go/aws/awserr"
 	"io"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/gomods/athens/pkg/config"
 	"github.com/gomods/athens/pkg/errors"
@@ -22,7 +23,8 @@ func (s *Storage) Info(ctx context.Context, module, version string) ([]byte, err
 
 	infoReader, err := s.open(ctx, config.PackageVersionedName(module, version, "info"))
 	if err != nil {
-		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == s3.ErrCodeNoSuchKey {
+		var aerr awserr.Error
+		if errs.As(err, &aerr) && aerr.Code() == s3.ErrCodeNoSuchKey {
 			return nil, errors.E(op, errors.M(module), errors.V(version), errors.KindNotFound)
 		}
 		return nil, errors.E(op, err, errors.M(module), errors.V(version))
@@ -44,7 +46,8 @@ func (s *Storage) GoMod(ctx context.Context, module, version string) ([]byte, er
 
 	modReader, err := s.open(ctx, config.PackageVersionedName(module, version, "mod"))
 	if err != nil {
-		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == s3.ErrCodeNoSuchKey {
+		var aerr awserr.Error
+		if errs.As(err, &aerr) && aerr.Code() == s3.ErrCodeNoSuchKey {
 			return nil, errors.E(op, errors.M(module), errors.V(version), errors.KindNotFound)
 		}
 		return nil, errors.E(op, err, errors.M(module), errors.V(version))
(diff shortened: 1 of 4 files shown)


manugupt1:
@linzhp @ngshiheng For performance issues: I think we can replace ListObjects call in S3 with HeadObject for all the 3 and it should significantly speed things up.

The key should be: `config.PackageVersionedName(module, version, "info")` when making the head call. wdyt?

manugupt1:
I think Delete is to help Delete specific modules. It is a path
1. /{module:.+}/@v/{version}.delete

pkg/storage/s3/getter.go
```diff
@@ -18,16 +20,13 @@ func (s *Storage) Info(ctx context.Context, module, version string) ([]byte, err
 	const op errors.Op = "s3.Info"
 	ctx, span := observ.StartSpan(ctx, op.String())
 	defer span.End()
-	exists, err := s.Exists(ctx, module, version)
```
manugupt1, line 21: If we do not ensure these checks; will this not break in private networks where modules are already cached; but cannot fallback.

linzhp: If the modules are already cached, the `s.open` below will succeed; if not, this method will return the same `errors.KindNotFound` like before. From the caller of this method, there is no difference.

linzhp:
> @linzhp @ngshiheng For performance issues: I think we can replace ListObjects call in S3 with HeadObject (as suggested; sorry did not read all of it) for all the 3 and it should significantly speed things up. If this works, we can keep the exists check around.
> 
> The key should be: `config.PackageVersionedName(module, version, "info")` when making the head call. wdyt?

Yes, this is a good option too. I can do it in a separate PR.

linzhp:
With this PR, the `Exists()` call would be pretty rare: it only happens before writing modules to S3. So 5500 req/s means Athens is allowed to write 1833 modules per second.

manugupt1:
You can’t predict, if this is a concern the S3 client can have some back offs. I would not worry though. In case, we are throttled, we can always serialize these requests

linzhp:
With #1844, I think this is still a good improvement to getters. @marwan-at-work Can you take a look too?

marwan-at-work (approved):
LGTM

manugupt1:
I am not sure though! Can we hold on to merging this please?

marwan-at-work:
@manugupt1 What is your concern here? You are more familiar with S3 and AWS than I am so I'm happy to hear your thoughts

marwan-at-work:
@manugupt1 What is your concern here? You are more familiar with S3 and AWS than I am so I'm happy to hear your thoughts

manugupt1:
I am not sure if we should download all the bytes for info, mod and version with a GET call. It’s just that the bytes transferred will introduce latency as well as cost.

marwan-at-work:
> I am not sure if we should download all the bytes for info, mod and version with a GET call. It’s just that the bytes transferred will introduce latency as well as cost.

From the diff of this PR, all it's doing is that it's not calling ".Check()" but instead using the 404 from the GET call as a way to check if something does not exist.  On the flip side, if a module does exist then we just proceed to serving like as if Check() return true. Am I missing something? 

Thanks!

manugupt1:
You are correct! I need sometime to think and understand. Let me get back! I am still not sure though

manugupt1:
@marwan-at-work @linzhp Sorry! I think this PR looks good to me. I think it is good to go. I did not read the commit properly as I was going through my phone.

manugupt1 (approved):

linzhp:
No worries. Thanks for the quick review!

PR #1843, opened 2023-03-02
feat: add non-root user to docker image

nrwiersma:

In many modern deployment environments, root containers are not allowed. While it is simple to switch the user on the fly, that user needs to have a home directory for this to work.

## How is the fix applied?

A non-root user `athens` uid/gid 1000 is created in the docker image, but not used by default. As existing users will still want to use `root` for this to be non-breaking, the new user is not used by default. Documentation on the non-root user has also been added to make this change discoverable.

### DIFF
--- a/cmd/proxy/Dockerfile
+++ b/cmd/proxy/Dockerfile
@@ -32,6 +32,9 @@ RUN chmod 644 /config/config.toml
 RUN apk add --update git git-lfs mercurial openssh-client subversion procps fossil tini && \
 	mkdir -p /usr/local/go
 
+ARG USER=athens
+RUN adduser -D -h /home/$USER $USER
+
 EXPOSE 3000
 
 ENTRYPOINT [ "/sbin/tini", "--" ]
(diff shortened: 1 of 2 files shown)


cmd/proxy/Dockerfile
```diff
@@ -32,6 +32,9 @@ RUN chmod 644 /config/config.toml
 RUN apk add --update git git-lfs mercurial openssh-client subversion procps fossil tini && \
 	mkdir -p /usr/local/go
 
+ARG USER=athens
+RUN adduser -D -h /home/$USER $USER
```
manugupt1, line 36: It might break k8s and docker-compose which provide external options to control non root access.

nrwiersma: k8s just lets you set the user id and group id to be run, much like docker does. Should you however need to use `.netrc` or `.gitconfig` you cannot without the user you are using having a home directory.

DrPsychick (changes requested):
Typo "authens"

cmd/proxy/Dockerfile
manugupt1: Makes sense! Should we create non root group as well

nrwiersma: This is already done by `adduser`. The group is 1000 called `athens`.

manugupt1 (approved):

DrPsychick (approved):

PR #1844, opened 2023-03-03
Calling HeadObjectWithContext() concurrently to check the existence of a module

linzhp:
This is an alternative implementation of #1840, with goroutines.

@manugupt1 @ngshiheng @r-ashish

### DIFF
--- a/pkg/storage/s3/checker.go
+++ b/pkg/storage/s3/checker.go
@@ -47,6 +47,7 @@ func (s *Storage) Exists(ctx context.Context, module, version string) (bool, err
 		}
 		var aerr awserr.Error
 		if errs.As(err, &aerr) && aerr.Code() == "NotFound" {
+			err = nil
 			exists = false
 		}
 		break
--- a/pkg/storage/s3/checker.go
+++ b/pkg/storage/s3/checker.go
@@ -35,11 +35,11 @@ func (s *Storage) Exists(ctx context.Context, module, version string) (bool, err
 			errChan <- err
 		}(file)
 	}
-	for _ = range files {
+	for range files {
 		err := <-errChan
 		if err != nil {
 			var aerr awserr.Error
-			if errs.As(err, &aerr) && aerr.Code() == s3.ErrCodeNoSuchKey {
+			if errs.As(err, &aerr) && aerr.Code() == "NotFound" {
 				return false, nil
 			}
 			return false, err
--- a/pkg/storage/s3/checker.go
+++ b/pkg/storage/s3/checker.go
@@ -39,11 +39,14 @@ func (s *Storage) Exists(ctx context.Context, module, version string) (bool, err
 		}(file)
 	}
 	exists := true
-	var err error
+	var firstErr error
 	for range files {
-		err = <-errChan
+		err := <-errChan
 		if err == nil {
 			continue
+		}
+		if firstErr == nil {
+			firstErr = err
 		}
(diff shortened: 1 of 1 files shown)


pkg/storage/s3/checker.go
```diff
@@ -19,27 +19,31 @@ func (s *Storage) Exists(ctx context.Context, module, version string) (bool, err
...
+	cancelingCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	for _, file := range files {
+		go func(file string) {
+			_, err := s.s3API.HeadObjectWithContext(
+				cancelingCtx,
+				&s3.HeadObjectInput{
+					Bucket: aws.String(s.bucket),
+					Key:    aws.String(config.PackageVersionedName(module, version, file)),
+				})
```
marwan-at-work, line 34: is that the implementation of the `Exists()` method, if so can we just call `.Exists()` instead? And if not, any reason not to make the Exists() method use this logic?

marwan-at-work (approved):
Nevermind my previous comment 👍🏼 PS. Not an AWS/S3 expert but the concurrency part looks good to me

pkg/storage/s3/checker.go
```diff
@@ -19,27 +19,31 @@ func (s *Storage) Exists(ctx context.Context, module, version string) (bool, err
...
-			}
-			if *o.Key == config.PackageVersionedName(module, version, "info") ||
-				*o.Key == config.PackageVersionedName(module, version, "mod") ||
-				*o.Key == config.PackageVersionedName(module, version, "zip") {
-				found[*o.Key] = struct{}{}
+	for range files {
+		err := <-errChan
+		if err != nil {
+			var aerr awserr.Error
+			if errs.As(err, &aerr) && aerr.Code() == "NotFound" {
```
marwan-at-work, line 42: As a side note, we should probably create a `function As` inside our `errors` package that just proxies to the standard library one, this way you don't need to import and rename to `errs` here.

pkg/storage/s3/checker.go
```diff
@@ -19,27 +19,31 @@ func (s *Storage) Exists(ctx context.Context, module, version string) (bool, err
 	ctx, span := observ.StartSpan(ctx, op.String())
 	defer span.End()
 
-	lsParams := &s3.ListObjectsInput{
-		Bucket: aws.String(s.bucket),
-		Prefix: aws.String(fmt.Sprintf("%s/@v", module)),
+	files := []string{"info", "mod", "zip"}
+	errChan := make(chan error, len(files))
+	defer close(errChan)
+	cancelingCtx, cancel := context.WithCancel(ctx)
```
manugupt1, line 26: I think this should go in the go routine so that each remote cal has its own ctx

manugupt1 (approved):

pkg/storage/s3/checker.go
linzhp: Having all goroutine sharing the same `cancelingCtx` allows one failing goroutine to cancel other calls to S3: if we already know one of the file don't exist, we can cancel other `HeadObject` calls.

linzhp:
All unit tests passed. Deployed to Uber's internal Go proxy and verified it worked.

pkg/storage/s3/checker.go
nrwiersma: A proxy function already exists: `errors.AsErr`.

PR #1845, opened 2023-03-03
New home for Athens-proxy helm chart

DrPsychick:
Separation of the helm chart from work on Athens and regular updates of the helm chart.

## How is the fix applied?
Agreed with maintainers to separate the chart and take over maintenance.

Fixes #1454


nrwiersma (approved):
LGTM

arschles:
Thank you for this 👍

PR #1849, opened 2023-03-09
Refactoring: Use errors.AsErr

linzhp:
Using `errors.AsErr` as suggested by @nrwiersma , so we don't need to import two errors packages.


marwan-at-work (approved):
Thanks!

PR #1850, opened 2023-03-13
Update alpine to 3.17 and build smaller binary.

manugupt1:
This PR updates alpine to 3.17 and builds smaller athens binary. By using these flags; we are saving around 14M in athens binary size.

### DIFF
--- a/cmd/proxy/Dockerfile
+++ b/cmd/proxy/Dockerfile
@@ -6,7 +6,7 @@
 # See project Makefile if using make.
 # See docker --build-arg if building directly.
 ARG GOLANG_VERSION=1.20
-ARG ALPINE_VERSION=3.15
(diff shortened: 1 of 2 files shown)


cmd/proxy/Dockerfile
```diff
@@ -6,7 +6,7 @@
 # See project Makefile if using make.
 # See docker --build-arg if building directly.
 ARG GOLANG_VERSION=1.20
-ARG ALPINE_VERSION=3.15
+ARG ALPINE_VERSION=3.17
```
nrwiersma, line 9: You are missing this `-s -w` flags on L19

nrwiersma (changes requested):

nrwiersma (approved):
LGTM

PR #1851, opened 2023-03-13
Remove duplicate runs of all the workflows

manugupt1:
All the WFs are running twice. See https://github.com/gomods/athens/pull/1850 for example. This fixes it.
pull_request takes in account for any activity


nrwiersma (approved):
LGTM 🎉

PR #1852, opened 2023-03-13
Remove duplicate runs of all the workflows #1851

manugupt1:


manugupt1:
I merged this to avoid appveyor to back up queued builds.

PR #1856, opened 2023-03-14
chore: run tests on other OSes

nrwiersma:

This adds Go tests for other OSes, primarily to remove AppVeyor, which is a bottleneck to the CI pipeline.


manugupt1:
I do not think there is a lot of value if we do not run e2e on windows

nrwiersma:
This is the same as AppVeyor was doing, so just unit tests. It removes a bottleneck in the pipeline.

nrwiersma:
Yip, never mind. The windows setup is too slow on actions. 3m in and it hasnt even setup Go yet, with caches.

PR #1866, opened 2023-04-13
fix: relax file permissions

### LINKED ISSUE #1007: config file permission
System ceontos 7.2
 I'm always frustrated when  I open with config.toml, the information show it need config file's permission is bit mask 077. Actually I use chmod 077 filename not work.
but chmod 700 work well.

the suggest information may need more friendly.
E.g ''the config's permission may error. you can use chmod 700 file name to solve this problem'

nrwiersma:

This relaxes the permissible config file permission to allow read for group and other (effectively 0o644).

For reference, the current docker build sets the file permissions to 0o644.

## How is the fix applied?

Changed the permission check and added a test for allowing 0o644.

Fixes #1007


DrPsychick (approved):
Looks good

PR #1870, opened 2023-05-03
feat: support arm64 and multi-arch docker images

### LINKED ISSUE #1869: Add builds for ARM64
## Describe the solution you'd like

I would like for there to be official builds of docker.io/gomods/athens for ARM64
Additionally, it would be nice to have a native build for MacOS M1/M2 chips too

In AWS (and perhaps other cloud providers); ARM CPUs are generally cheaper than Intel/x86_64 ones; however
the docker images and binaries built in this repository are for AMD64 architecture only.

## Describe alternatives you've considered

- I could use amd64 CPU instead; but ARM is starting to take over the server market due to its more favorable pricing.
- I could build from source to get arm64 binaries; but then no one else could benefit, unless I were to maintain a fork.

## Additional context

goreleaser supports building [Universal binaries](https://goreleaser.com/customization/universalbinaries/?h=universal) which packages a single binary for MacOS that supports Intel and Apple Silicon.

justenwalker:

Adds support for AMD64 binaries and builds a multi-arch docker image for linux/amd64/v1 + linux/arm64/v8

## How is the fix applied?

- modified the .goreleaser.yaml to add arm64 build target and macos universal binaries
- updated the docker build/push script to use docker buildx with both linux/amd64/v1 and linux/arm64/v8

Fixes #1869

## Additional Information

I've built an image and pushed it to [Docker Hub](https://hub.docker.com/layers/justenwalker/athens/v0.12.1-multiarch/images/sha256-33f5b96cdda5d6425ea7a9ecd7fb01dac6cf72c924d63d953e40a21fe892c707?context=explore)

`docker pull justenwalker/athens:v0.12.1-multiarch`

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -12,11 +12,14 @@ builds:
     - darwin
   goarch:
     - amd64
+    - arm64
   main: cmd/proxy/main.go
   binary: athens
   ldflags:
     - -X "github.com/gomods/athens/pkg/build.version={{ .Tag }}"
     - -X "github.com/gomods/athens/pkg/build.buildDate={{ .Env.DATE }}"
+universal_binaries:
+- replace: true
 checksum:
   name_template: 'checksums.txt'
 snapshot:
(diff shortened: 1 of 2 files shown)


justenwalker:
Closing. duplicates #1862

PR #1871, opened 2023-05-08
docs: update hugo-theme-learn version to latest master

ngshiheng:

Update Athens's outdated Hugo theme. I was using a newer version of `hugo` CLI and I couldn't get `hugo serve` working locally. After some debugging, I came to the conclusion that this is resolved by bumping the theme version.

_NOTE: All the changes in this PR under `/docs/themes/` are via `git clone`. Which is most of the changes in this PR 😅_ 
 
## How is the fix applied?

- Update Athens's existing Hugo theme to the latest master by cloning `https://github.com/matcornic/hugo-theme-learn` to `themes/` directory
- Updates the `HUGO_VERSION` to `0.72.0`
- This PR also moves Athens custom HTML to the Hugo custom layout directory. IMO any future customization made to the current Go Hugo theme _should_ be made at `/athens/docs/layouts` and **not** directly at `/athens/docs/themes/layouts`. This helps to each the theme updating process (we can just completely replace whatever that is under `/themes/` dir)

## How is this tested?

- cd `docs/` && `hugo serve`
- `make docs`
- `docker run -it --rm --name hugo-server -p 1313:1313 -v /home/athens/docs:/src:cached gomods/hugo`


r-ashish (approved):
LGTM 👍

PR #1872, opened 2023-05-08
test: storage catalog

### LINKED ISSUE #1213: Write a test for the continuation token on the minio storage driver

https://github.com/gomods/athens/pull/1212 fixes the continuation token on the minio cataloger but I don't believe we have a test for it.

Let's write a test for it 😁 

**Additional context**

https://github.com/gomods/athens/pull/1212#pullrequestreview-235358843

ofekatr:

- Creates a test for the `Catalog` method of the MinIO storage implementation, to ensure the authentication token is passed as expected.

## How is the fix applied?

- Creating a corresponding test that will fail if the change in 50a3a92 is reverted.

Fixes #1213


pkg/storage/minio/cataloger_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"context"
+	"testing"
+
+	"github.com/gomods/athens/pkg/paths"
+	"github.com/gomods/athens/pkg/storage/minio/mocks"
+	"github.com/minio/minio-go/v6"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCatalog(t *testing.T) {
```
manugupt1, line 13: Instead of mocks can we use a real minio set up like here

https://github.com/gomods/athens/blob/main/pkg/storage/minio/minio_test.go

PR #1873, opened 2023-05-17
docs: add docs about network mode

ngshiheng:

- Added some documentation for Athens' Network Mode configuration
- Added some use cases for each network mode

### DIFF
--- a/docs/content/configuration/network.md
+++ b/docs/content/configuration/network.md
@@ -1,0 +1,19 @@
+---
+title: Network mode
+description: Configures how Athens will return the results of the /list endpoint
+weight: 6
+---
+
+The NetworkMode configuration in Athens determines how results are returned by the `/list` endpoint. This endpoint can retrieve information from both Athens' own [storage](/configuration/storage) and the upstream version control system (VCS).
+
+> Note: The NetworkMode configuration can also affect the behavior of other endpoints by improving error messaging
+
+## Network mode keywords
+
+There are 3 modes available for the NetworkMode configuration. To configure the `NetworkMode` settings at `config.dev.toml`, set the `NetworkMode` to 1 of the 3 available modes:
+
+1. `strict`: In this mode, Athens will merge versions from the VCS and storage, but will fail if either of them fails. This mode provides the most consistent results.
+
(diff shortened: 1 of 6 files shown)


ngshiheng:
@marwan-at-work can you help to take a look at this PR for docs update? I think you have the most context about Athens NetworkMode

docs/content/configuration/network.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Network mode keywords
+
+There are 3 modes available for the NetworkMode configuration. To configure the `NetworkMode` settings at `config.dev.toml`, set the `NetworkMode` to 1 of the 3 available modes:
+
+1. `strict`: In this mode, Athens will merge versions from the VCS and storage but will fail if either of them fails. This mode provides the most consistent results.
+
+2. `offline`: This mode only retrieves versions from Athens' storage and never reaches out to the VCS.
+
+3. `fallback`: This mode retrieves versions from Athens' storage only if the VCS fails. Note that using this mode may result in inconsistent results since fallback mode does its best to provide the available versions at the time of the request.
```
marwan-at-work, line 19: > This mode retrieves versions from Athens' storage only if the VCS fails.

I don't believe that's true. We always retrieve versions from the Athens storage. We just fallback to only providing this list in case of an upstream outage

ngshiheng: > We always retrieve versions from the Athens storage

Hmm sorry, I don't exactly follow

1. This was taken from [this snippet](https://github.com/gomods/athens/blob/8e1581e10b0d3a70a30f45b10c24c3f992464d7a/config.dev.toml#LL271C6-L272C6).
2. That aside, I can attest this behavior is correct based on existing live deployments
3. Also referring to [/pkg/download/protocol.go ](https://github.com/gomods/athens/blob/8e1581e10b0d3a70a30f45b10c24c3f992464d7a/pkg/download/protocol.go#L137-L141)

maybe I could have worded it better? 🤔

ngshiheng: @marwan-at-work WDYT? can you recommend what can I change here to move this PR forward?

PR #1877, opened 2023-07-21
singleflight go commands

wozz:

see: https://github.com/gomods/athens/pull/1876

## How is the fix applied?

Use singleflight package to ensure identical go commands are not run simultaneously.

### DIFF
--- a/pkg/module/go_vcs_lister.go
+++ b/pkg/module/go_vcs_lister.go
@@ -36,7 +36,7 @@ func NewVCSLister(goBinPath string, env []string, fs afero.Fs) UpstreamLister {
 		goBinPath: goBinPath,
 		env:       env,
 		fs:        fs,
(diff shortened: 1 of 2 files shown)


pkg/module/go_vcs_lister.go
```diff
@@ -34,58 +36,74 @@ func NewVCSLister(goBinPath string, env []string, fs afero.Fs) UpstreamLister {
 		goBinPath: goBinPath,
 		env:       env,
 		fs:        fs,
+		sfg:       new(singleflight.Group),
```
Ullaakut, line 39:
```suggestion
		sfg:       &singleflight.Group{},
```

Ullaakut (approved):
Nitpick, besides that LGTM and seems to address the problem mentioned in the issue just fine.

nrwiersma (approved):

PR #1880, opened 2023-09-01
fix(proxy): don't check TLS file permission

### LINKED ISSUE #1879: Don't check TLS key file permissions
**Describe the bug**

Athens checks the permission bits of the TLS key file and fails to start if it fails the check.

https://github.com/gomods/athens/blob/cac9c754cb616fbeb6f25ad5e7a20b4cd6a89321/pkg/config/config.go#L235-L237

This is a bit annoying for environments where it doesn't matter, like in Kubernetes where secrets are written with 0o644. There is nothing else in the container - this check does not help.

**Error Message**

```console
2023/09/01 09:54:19 TLSKeyFile should not be accessible by others
```

**To Reproduce**

Run Athens with TLS.

**Expected behavior**

Athens should not check the permission bits of the file.

 - OS: Linux (Kubernetes)
 - Go version: N/A
 - Proxy version: v0.12.1
 - Storage: N/A

**Additional context**

I understand the intent behind the check, but I'm not sure it really helps and just makes deployment more fragile than it needs to be.

uhthomas:

Athens should not check the permission of TLS files, as it is not compatible with some automatic certificate issuers.

## How is the fix applied?

The config method TLSCertFiles has been removed as it was redundant after the change. The return values were never even used.

Fixes #1879

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -215,30 +215,6 @@ func (c *Config) BasicAuth() (user, pass string, ok bool) {
 	return user, pass, ok
 }
 
-// TLSCertFiles returns certificate and key files and an error if
-// both files doesn't exist and have approperiate file permissions.
-func (c *Config) TLSCertFiles() (cert, key string, err error) {
-	if c.TLSCertFile == "" && c.TLSKeyFile == "" {
-		return "", "", nil
-	}
-
-	certFile, err := os.Stat(c.TLSCertFile)
-	if err != nil {
-		return "", "", fmt.Errorf("could not access TLSCertFile: %w", err)
-	}
-
-	keyFile, err := os.Stat(c.TLSKeyFile)
-	if err != nil {
-		return "", "", fmt.Errorf("could not access TLSKeyFile: %w", err)
-	}
-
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -215,30 +215,6 @@ func (c *Config) BasicAuth() (user, pass string, ok bool) {
...
 }
 
-// TLSCertFiles returns certificate and key files and an error if
-// both files doesn't exist and have approperiate file permissions.
-func (c *Config) TLSCertFiles() (cert, key string, err error) {
-	if c.TLSCertFile == "" && c.TLSKeyFile == "" {
-		return "", "", nil
-	}
-
-	certFile, err := os.Stat(c.TLSCertFile)
```
manugupt1, line 225: Is there a reason to remove these checks?

uhthomas: They aren't really useful and don't guard against anything, the Go HTTP server will read these files before starting the HTTP server and return any errors.

https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/net/http/server.go;l=3119-3122

manugupt1 (approved):
Thanks!

uhthomas:
Thanks for your timely review @manugupt1, much appreciated!

PR #1883, opened 2023-09-08
test: unit test for testing KindUnexpected case in Mongo query function

### LINKED ISSUE #1327: Write tests for error paths in Mongo query funtion

This is a follow up to the conversation in #1314 

The tests for the query function handle all of the good paths but not the error cases. There should be tests to verify that the error paths behave as expected. 

N/A

**Additional context**

N/A

yatesliang:

Provided unit test for testing the error paths for Mongo query function, which will return the error type KindUnexpected

## How is the fix applied?

1. Insert error documents into the module collection
2. Getter will fail to  decode the error fields (in type integer or boolean) to []byte
3. Error KindUnexpected return

Fixes #1327


pkg/storage/mongo/mongo_test.go
```diff
@@ -97,6 +99,49 @@ func TestQueryKindNotFoundErrorCases(t *testing.T) {
...
+		Module  string      `bson:"module"`
+		Version string      `bson:"version"`
+		Mod     interface{} `bson:"mod"`
+		Info    interface{} `bson:"info"`
+	}{
+		{"model1", "v1.1.1", 12345678, "error document with integer mod"},
+		{"model2", "v2.0.0", true, "error document with boolean mod"},
+	}
+	// clear collection before insert
+	coll.Drop(ctx)
```
manugupt1, line 125: can you please remove this drop? It can interfere with other parallel running tests

yatesliang: Sure, I have removed the drop operation, and all the docs will be upserted instead.

manugupt1 (approved):

manugupt1:
Thank you!

Issue #1886, opened 2023-09-11
Athens leaks etcd leases


uhthomas:
**Describe the bug**

Athens [creates a new session (lease) for each stash request](https://github.com/gomods/athens/blob/e248d228921243998aca56dab01786e3cb938919/pkg/stash/with_etcd.go#L59), and then never closes the session. The leases are then held indefinitely, or until the process is restarted, which causes excessive load on the etcd cluster and eventually overloads it.

<img width="1448" alt="Screenshot 2023-09-11 at 22 31 10" src="https://github.com/gomods/athens/assets/9749173/fb8f298a-2e5f-4f8e-a6fd-446f2963bbef">

<img width="1462" alt="Screenshot 2023-09-11 at 22 32 27" src="https://github.com/gomods/athens/assets/9749173/7429715b-551f-482e-9624-bc0b1143db02">

The spikes are from restarting Athens.

**Error Message**

N/A

**To Reproduce**

1. Run Athens with etcd configured for singleflight.
2. Make a few thousands requests for modules which aren't already cached.
3. Observe the number of leases increases indefinitely.

**Expected behavior**

Athens should either create a single session for all locks, or close the session. I am not sure which is correct as I could not find great documentation on what is recommended by the etcd authors. I imagine it's probably best just to make sure the session is closed.

**Environment (please complete the following information):**
 - OS: Linux
 - Go version: N/A
 - Proxy version: https://github.com/gomods/athens/commit/e248d228921243998aca56dab01786e3cb938919
 - Storage (fs/mongodb/s3 etc.): etcd and s3

**Additional context**

santiagoPinzonD:
Hello, @uhthomas  wanted to know how did you solve this problem ?
I know that the fix is already in the code but not yet released.
this is my singleflight config, I have a lot of traffic and I have to restart constantly because it consumes the machine's memory.
![image](https://github.com/gomods/athens/assets/66101364/58d52597-dcf4-4c16-b684-c1ff819c0904)

uhthomas:
@santiagoPinzonD You will have to build Athens from source.

```Dockerfile
FROM golang:1.21.3-alpine AS build

RUN apk update && apk add git

RUN go install github.com/gomods/athens/cmd/proxy@main

FROM gomods/athens:v0.12.1

COPY --from=build /go/bin/proxy /bin/athens-proxy
```

manugupt1:
Please use canary tag which has more fixes

r-ashish:
[...]

PR #1887, opened 2023-09-11
fix(pkg/stash): close etcd sessions

uhthomas:

Athens creates a new session for each 'Stash' request and then never closes it. The sessions (etcd leases) are consequently held forever and eventually leads to resource exhaustion on the etcd cluster.

## How is the fix applied?

This has been fixed by closing the acquired session at the end of the request.

Fixes: #1886


uhthomas:
I've also excluded `(*go.etcd.io/etcd/client/v3/concurrency.Session).Close` from errcheck as there's not really any action which can be taken if a session fails to close - the session and lease should automatically expire.

uhthomas:
Question for reviewers: should we also remove the error check from `(*go.etcd.io/etcd/clientv3/concurrency.Mutex).Unlock`? There's not really anything which can be done about it. It may be a bit unfortunate to then return an error just because the mutex didn't unlock, despite successfully stashing the module.

uhthomas:
I've made the changes inline with my previous comment, I hope this is okay 😄

pkg/stash/with_etcd.go
```diff
@@ -56,33 +56,31 @@ func (s *etcd) Stash(ctx context.Context, mod, ver string) (newVer string, err e
...
-	err = mu.Lock(ctx)
-	if err != nil {
-		return ver, errors.E(op, err)
+	defer sess.Close()
+
+	m := concurrency.NewMutex(sess, config.FmtModVer(mod, ver))
+	if err := m.Lock(ctx); err != nil {
+		return "", errors.E(op, err)
 	}
-	defer func() {
```
manugupt1, line 69: I prefer keeping it the way it was.
https://go.dev/play/p/cq0DGy9OJES

Knowing if unlock / session close fails is a useful thing

uhthomas: The reason I made this change is because it seems a bit strange to fail the request because the mutex failed to unlock. There's not really any action to be taken, and everything else has already succeeded. The best thing we can do here is log, not fail the entire request - as it doesn't matter.

uhthomas:
We deployed this change yesterday and have seen significant improvement in both Athens and etcd.

## etcd performance

Screenshot 2023-09-13 at 16 44 56

Screenshot 2023-09-13 at 16 44 19

Screenshot 2023-09-13 at 16 43 40

gRPC filtered by LeaseRevoke:

Screenshot 2023-09-13 at 16 45 50


## Resource usage

### Athens

Screenshot 2023-09-13 at 16 33 51

Screenshot 2023-09-13 at 16 39 06

### etcd

Screenshot 2023-09-13 at 16 36 02

Screenshot 2023-09-13 at 16 37 37

pkg/stash/with_etcd.go
manugupt1: I agree; I think we should definitely log. Can we have some tests if possible please

uhthomas: @manugupt1 I've replaced the etcd tests. They should now be a lot more realistic and actually connects to an etcd server. The tests cover the main code paths and ensure that a lease is acquired, and then freed again.

manugupt1: Sorry for asking again, can we log if unlock or session close fails?

uhthomas: @manugupt1 There doesn't really seem to be a standard for logging in this repository. What should I use?

uhthomas: I also am not entirely sure it's worth logging. It _could_ be useful? Though neither of these conditions should have any long-lasting negative effects. If the connection to etcd is unhealthy then the lease will expire automatically with no hard done.

pkg/stash/with_etcd_test.go
```diff
@@ -2,82 +2,219 @@ package stash
...
+func startEtcdServer(t *testing.T) (endpoint string, e *etcdembed.Etcd) {
+	c := etcdembed.NewConfig()
+	c.Dir = t.TempDir()
+	c.ListenPeerUrls = []url.URL{{Host: "localhost:0"}}
+	c.ListenClientUrls = []url.URL{{Host: "localhost:0"}}
+
+	e, err := etcdembed.StartEtcd(c)
 	if err != nil {
-		t.Fatal(err)
+		log.Fatal(err)
```
uhthomas, line 67:
```suggestion
		t.Fatal(err)
```

pkg/stash/with_etcd.go
uhthomas: @manugupt1 What are your thoughts? I'm also keen to merge this one as it is quite a major issue for the etcd integration.

I am sort of surprised this issue wasn't raised before - I wonder how many people use Athens with etcd.

manugupt1 (approved):

Issue #1888, opened 2023-09-12
Athens should not check the status of all etcd endpoints on startup


uhthomas:
**Describe the bug**

etcd and the client are designed to be highly available and resilient, manually checking the status of each endpoint defeats these goals as Athens will crash if only 2/3 endpoints of an etcd cluster are available. This behaviour makes rolling updates much harder.

**Error Message**

N/A - Athens will print an error that it cannot connect to a member of the etcd cluster.

**To Reproduce**
1. Have multiple athens instance, and an etcd cluster.
2. Restart both at the same time.
3. Observe that Athens will not connect to etcd unless all endpoints are available.

**Expected behavior**

Athens should connect to the etcd cluster and defer connection management to the etcd client. It should automatically load balance and route to available members.

**Environment (please complete the following information):**
 - OS: Linux
 - Go version: N/A
 - Proxy version: https://github.com/gomods/athens/commit/e248d228921243998aca56dab01786e3cb938919
 - Storage: etcd and s3

**Additional context**

We run 5 Athens pods and 3 etcd pods in Kubernetes with high availability. We will update both images at the same time, and the Athens deployment will take many minutes to progress as it will crash loop until the etcd cluster is completely ready.

PR #1889, opened 2023-09-12
fix(pkg/stash): don't check status of all etcd endpoints on start

uhthomas:

Athens checks the status of all etcd endpoints when started, which can cause issues when some members of the etcd cluster are unavailable. It is perfectly okay for some members of an etcd cluster to be unavailable, as it's designed for high availability and fault tolerance.

## How is the fix applied?

The management of the connections is instead deferred to the etcd client, which will handle failures and load balancing as expected.

Fixes: #1888


uhthomas:
Thank you @manugupt1 😄

PR #1912, opened 2024-01-04
Send standard logger's output to logrus

mikesep:

Some packages like [`httputil`'s `ReverseProxy`](https://pkg.go.dev/http/httputil#ReverseProxy) [use the standard `log.Logger`](https://github.com/golang/go/blob/6db1102605f227093ea95538f0fe9e46022ad7ea/src/net/http/httputil/reverseproxy.go#L307) to emit log traces. This means that their output won't match the format of Athens's `logrus` output.

## How is the fix applied?

[`logrus.Logger.Writer`](https://pkg.go.dev/github.com/sirupsen/logrus#Logger.Writer)'s documentation states that it "can be used to override the standard library logger easily." It spawns a go routine, so instead of doing that each time we need to make `logrus` look like a (standard) `log.Logger` (and then remembering to `Close()` it right away), this calls `logrus.Writer()` once in `main`, `defer`s the `Close()`,  and sets it into the global `log.Logger`.

Related to the work done in #1819 to unify logging.

### DIFF
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -44,6 +44,16 @@ func main() {
 
 	logger := athenslog.New(conf.CloudRuntime, logLvl)
 
+	// logrus.Writer() spawns a go routine, so instead of doing that each time
+	// we need to make logrus look like a (standard) log.Logger (and then Close()ing it),
+	// we can do it once in main and have the global log.Logger use it.
+	{
+		logrusWriter := logger.Writer()
+		defer logrusWriter.Close()
+		stdlog.SetOutput(logrusWriter)
+		stdlog.SetFlags(stdlog.Flags() &^ (stdlog.Ldate | stdlog.Ltime))
+	}
+
 	handler, err := actions.App(logger, conf)
 	if err != nil {
 		logger.WithError(err).Fatal("Could not create App")
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -44,19 +44,15 @@ func main() {
 
 	logger := athenslog.New(conf.CloudRuntime, logLvl)
 
-	// logrus.Writer() spawns a go routine, so instead of doing that each time
-	// we need to make logrus look like a (standard) log.Logger (and then Close()ing it),
-	// we can do it once in main and have the global log.Logger use it.
-	{
-		logrusWriter := logger.WriterLevel(logrus.ErrorLevel)
-		defer func() {
-			if err := logrusWriter.Close(); err != nil {
-				logger.WithError(err).Warn("Could not close logrus writer pipe")
-			}
-		}()
-		stdlog.SetOutput(logrusWriter)
-		stdlog.SetFlags(stdlog.Flags() &^ (stdlog.Ldate | stdlog.Ltime))
-	}
+	// Turn standard logger output into logrus Errors.
+	logrusErrorWriter := logger.WriterLevel(logrus.ErrorLevel)
+	defer func() {
+		if err := logrusErrorWriter.Close(); err != nil {
+			logger.WithError(err).Warn("Could not close logrus writer pipe")
+		}
+	}()
+	stdlog.SetOutput(logrusErrorWriter)
+	stdlog.SetFlags(stdlog.Flags() &^ (stdlog.Ldate | stdlog.Ltime))
 
 	handler, err := actions.App(logger, conf)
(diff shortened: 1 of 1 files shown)


mikesep:
If maintainers would rather I figure out how to pass this down through the call chain to the sumdb `ReverseProxy`, I can try that approach, but it occurred to me that this global approach would catch anything else using the standard `log` too, and I wanted to avoid repeatedly spawning and closing the pipe-reading goroutine.

cmd/proxy/main.go
```diff
@@ -44,6 +44,20 @@ func main() {
 
 	logger := athenslog.New(conf.CloudRuntime, logLvl)
 
+	// logrus.Writer() spawns a go routine, so instead of doing that each time
+	// we need to make logrus look like a (standard) log.Logger (and then Close()ing it),
+	// we can do it once in main and have the global log.Logger use it.
+	{
+		logrusWriter := logger.Writer()
```
nrwiersma, line 51: Surely INFO is not the correct level here? The components you speak of generally only write to the log in case of Error, so passing the logs off as INFO seems incorrect.

nrwiersma (changes requested):
IMO, global loggers are not great, but given the cost (A go routine to mimic an io.Writer seems a little insane), this might be OK in this case.

cmd/proxy/main.go
mikesep: Fair point. I guess I wasn't sure if I was ready to say that everything traced via the standard logger would be an Error. What do you think about Warn as a "this trace might be concerning" compromise?

mikesep: Happy to start with Error. 😄

mikesep: Changed in 18f5e26.

nrwiersma (approved):
LGTM

cmd/proxy/main.go
```diff
@@ -44,6 +44,20 @@ func main() {
...
+	{
+		logrusWriter := logger.WriterLevel(logrus.ErrorLevel)
+		defer func() {
+			if err := logrusWriter.Close(); err != nil {
+				logger.WithError(err).Warn("Could not close logrus writer pipe")
+			}
+		}()
+		stdlog.SetOutput(logrusWriter)
+		stdlog.SetFlags(stdlog.Flags() &^ (stdlog.Ldate | stdlog.Ltime))
+	}
```
Ullaakut, line 59: Scoping the variables here seems unnecessary, unless I am missing something that isn't described by the comment.

```suggestion
	logrusWriter := logger.WriterLevel(logrus.ErrorLevel)
	defer func() {
		if err := logrusWriter.Close(); err != nil {
			logger.WithError(err).Warn("Could not close logrus writer pipe")
		}
	}()
	stdlog.SetOutput(logrusWriter)
	stdlog.SetFlags(stdlog.Flags() &^ (stdlog.Ldate | stdlog.Ltime))
```

WDYT of simply removing the scope?

cmd/proxy/main.go
```diff
@@ -44,6 +44,20 @@ func main() {
 
 	logger := athenslog.New(conf.CloudRuntime, logLvl)
 
+	// logrus.Writer() spawns a go routine, so instead of doing that each time
+	// we need to make logrus look like a (standard) log.Logger (and then Close()ing it),
+	// we can do it once in main and have the global log.Logger use it.
```
Ullaakut, line 49:
```suggestion
	// Use the logrus writer as the standard logger's output in order to avoid starting goroutines whenever logrus.Writer() is called.
```

WDYT of this simplification of the comment?

Ullaakut (changes requested):

cmd/proxy/main.go
mikesep: I did it just to logically group things together, and also to discourage/prevent using `logrusWriter` later on. I'm fine removing it if this feels too awkward/unusual to others.

cmd/proxy/main.go
mikesep: I think it's fair to say that this was too long a comment for code and better off in the PR/commit explanation. I went with something even shorter in 6eea0a6.

cmd/proxy/main.go
mikesep: Removed scope in 6eea0a6.

Ullaakut (approved):
LGTM, thanks!

PR #1926, opened 2024-03-21
Add log formatting settings

### LINKED ISSUE #1924: Coupling of CloudRuntime (ATHENS_CLOUD_RUNTIME) config var with log formatting is unexpected
**Describe the bug**
The only way to configure the formatter applied to Athens' logs is by setting the `CloudRuntime` / `ATHENS_CLOUD_RUNTIME` config var. The default value `"none"` configures unstructured text logging, which is undesirable outside of dev/debug contexts.

**Error Message**
N/A

**To Reproduce**
The maximal example config file [here](https://github.com/gomods/athens/blob/main/config.dev.toml) says:
```
# CloudRuntime is the Cloud Provider on which the Proxy/Registry is running.
# Currently available options are "GCP", or "none". Defaults to "none"
CloudRuntime = "none"
```

We don't run Athens in GCP, so we leave this as-is. When this value is `"none"`, the logger setup code uses `devFormatter`: https://github.com/gomods/athens/blob/d877af209911025713b65beb8dd409f649ca98f1/pkg/log/log.go#L22-L23

This formatter outputs unstructured text, but we are running Athens in a production context and would like JSON logs to send to our ingest pipeline. I think can work around this by setting `CloudRuntime = "anything else"`, but IMO this coupling is unintuitive.

**Expected behavior**
JSON logging should be configured by a dedicated config variable, independent of cloud provider.

 - OS: Ubuntu 22.04 64-bit
 - Go version: 1.22.0
 - Proxy version: 0.13.1
 - Storage (fs/mongodb/s3 etc.): S3

**Additional context**
N/A

matt0x6F:

Log formatting can only be configured via the `CloudRuntime` setting which does some niche formatting based on Google Cloud Platform logging norms. Athens settings, however, should favor configuration that is not reliant on cloud provider behavior. `ATHENS_LOG_FORMAT` introduces the ability to configure `plain` or `json` loggers when `CloudRuntime` is `none` (default).

## How is the fix applied?

Introducing a setting for `LogFormat` which can be `plain` or `json`. Defaults to `json` (original behavior).

Fixes #1924


nrwiersma (approved):
LGTM

PR #1927, opened 2024-03-21
Rework logging defaults

### LINKED ISSUE #1924: Coupling of CloudRuntime (ATHENS_CLOUD_RUNTIME) config var with log formatting is unexpected

matt0x6F:

#1926 made configuring the JSON logger more obvious, however, it omitted that there is a requirement to change `CloudRuntime` to something other than `none` in order for it to work. This uses the plain text logger (aka dev formatter) as the plaintext configuration.

## How is the fix applied?

Remove `none` case, and use devFormatter as the `plain` and blank option.

Closes #1924 

**Note**: This will cause anyone not configuring a cloudruntime or setting format to `plain` to by default see logs in `json`. If this is absolutely breaking or undesireable we can change the default to `plain` in the configuration.


matt0x6F:
@nrwiersma I have two commits here. One sets the JSON logger as a default when CloudRuntime is `none` and the other sets `json` as the default. I can't totally match the original behavior, because before you had to set CloudRuntime to something _other_ than `none` in order to get the JSON logger, but `plain` was what came with the default config.

Any thoughts or preferences?

nrwiersma:
@matt0x6F I would try keep the default behaviour of `plain`, as IMO changing this would affect the most people. From there the config changes are what they are.

matt0x6F:
Sounds good, the latest commit restores the default behavior

nrwiersma (approved):

PR #1936, opened 2024-04-08
should use errors.AsErr to extract and detect errors.Error

kkHAIKE:

When outputting JSON logs, there are no additional fields.
截屏2024-04-09 10 59 56

## How is the fix applied?

Output that fields

截屏2024-04-09 11 01 55

Fixes #

### DIFF
--- a/pkg/log/entry.go
+++ b/pkg/log/entry.go
@@ -36,7 +36,7 @@ func (e *entry) WithFields(fields map[string]any) Entry {
 
 func (e *entry) SystemErr(err error) {
 	var athensErr errors.Error
-	if !errors.IsErr(err, athensErr) {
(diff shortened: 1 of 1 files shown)


matt0x6F:
@kkHAIKE thanks for your contribution! Can you give me an example of what output looked like before versus after your proposed change?

kkHAIKE:
> @kkHAIKE thanks for your contribution! Can you give me an example of what output looked like before versus after your proposed change?

This is considered a relatively basic mistake; if you use `errors.IsErr`, the variable used in the subsequent `errFields` function parameter will always be empty.

https://github.com/gomods/athens/blob/a32e5f0a42408f01659963567504953b864d0540/pkg/log/entry.go#L38-L44

matt0x6F (approved):

PR #1945, opened 2024-04-18
Add an introduction page as the home page

### LINKED ISSUE #1944: Add a splash page

Add a splash page behind a feature flag that shows documentation at the root URL on how to connect and configure a client.

Athens knows all of the bits required for a client to connect. For instance, it knows its root URL that it's being accessed from. It knows what `NOSUMPATTERNS` are specified. The users of Athens often don't know these details and may be entirely unfamiliar with the configuration of a proxy. 

Continue showing nothing at the root URL

**Additional context**
N/A

matt0x6F:

The home page of Athens is just "Welcome to The Athens Proxy". While this is cute in an old internet kind of way, it isn't super informative to users of Athens. At my workplace we spent a decent amount of time building messaging for how to configure clients. It would be nice if users could land on the Athens page and know how to configure their clients.

## How is the fix applied?

A default homepage is baked into the server that uses the request host address, or in HTTP 2 the authority. This includes ports. It also checks for schema. The values are used to indicate to users how to configure their `go env`

Of course, this won't work on all installations - especially enterprise ones. For that, I've introduced `ATHENS_HOME_TEMPLATE_PATH` as an environment variable along with `HomeTemplatePath` in the config. This value defaults to `/var/lib/athens/home.html` but can be configured to any location that Athens can reliably read from. This is a Go HTML template so it should use Go HTML template formatting and logic.

Closes #1944


cmd/proxy/actions/home.go
```diff
@@ -1,9 +1,139 @@
...
+
+			// collapse the NoSumPatterns into a comma separated list
+			for i, pattern := range c.NoSumPatterns {
+				if i == len(c.NoSumPatterns)-1 {
+					templateData["NoSumPatterns"] += pattern
+					continue
+				}
+
+				templateData["NoSumPatterns"] += pattern + ","
+			}
```
nrwiersma, line 122: Would `strings.Join` not be simpler here?

cmd/proxy/actions/home.go
```diff
@@ -1,9 +1,139 @@
...
+			templateData := make(map[string]string)
+
+			// determine the host. If the proxy host is set, use that. Otherwise, use the request host, lastly use localhost + port
+			if c.ProxyHost != "" {
+				templateData["Host"] = c.ProxyHost
+			} else if r.Host != "" {
+				templateData["Host"] = r.Host
+			} else {
+				templateData["Host"] = "localhost" + ":" + c.Port
+			}
```
nrwiersma, line 101: This should be a `switch`.

cmd/proxy/actions/home.go
```diff
@@ -1,9 +1,139 @@
...
+
+			tmp, err := template.New("home").Parse(homepage)
+			if err != nil {
+				lggr.SystemErr(err)
+				w.WriteHeader(http.StatusInternalServerError)
+			}
+
+			w.Header().Add("Content-Type", "text/html")
+			w.WriteHeader(http.StatusOK)
+			tmp.ExecuteTemplate(w, "home", templateData)
```
nrwiersma, line 132: This error needs to be checked.

nrwiersma (changes requested):

matt0x6F:
@nrwiersma great ideas code wise. What do you think about the concept of the feature though? I have an alternative idea that I also listed here that I'm kind of leaning in the direction of, which is to allow the user to specify their own template and we just load it from disk.

nrwiersma:
I like the idea here of a fixed one to start, and should people need a customisable, then add the complexity, IMO.

matt0x6F:
> I like the idea here of a fixed one to start, and should people need a customisable, then add the complexity, IMO.

This is great feedback. I do think I'm going to move it to a file based Go-template the user can swap out, but I'll add something to the container image so that we ship a default one.

cmd/proxy/actions/home.go
matt0x6F: I decided to rely entirely on the request host. This _should be_ accurate for most users, but complicated networks can use the template override for their instructions.

matt0x6F:
@nrwiersma let me know what you think of this. Instead of adding a file to the container image I just kept the default baked in and simplified the logic.

cmd/proxy/actions/home.go
```diff
@@ -1,9 +1,138 @@
...
+		rawTemplateFileContents, err := os.ReadFile(c.HomeTemplatePath)
+		if errors.Is(err, os.ErrNotExist) {
+			templateContents = homepage
+		} else if err != nil {
+			// this is some other error, log it and revert to default
+			lggr.SystemErr(err)
+			templateContents = homepage
+		} else {
+			templateContents = string(rawTemplateFileContents)
+		}
```
nrwiersma, line 107: This is a slightly weird construct. Perhaps:
```suggestion
		templateContents, err := os.ReadFile(c.HomeTemplatePath)
		if err != nil {
		    if !errors.Is(err, os.ErrNotExist) {
		        // this is some other error, log it and revert to default
			lggr.SystemErr(err)
		    }
		    templateContents = homepage
		}
```

matt0x6F: That misses the byte slice to string conversion case. What do you think of this?

```
		templateContents := homepage

		// load the template from the file system if it exists, otherwise revert to default
		rawTemplateFileContents, err := os.ReadFile(c.HomeTemplatePath)
		if err != nil {
		    if !errors.Is(err, os.ErrNotExist) {
		        // this is some other error, log it and revert to default
				lggr.SystemErr(err)
		    }
		} else {
			templateContents = string(rawTemplateFileContents)
		}
```

nrwiersma: Also fine.

nrwiersma (approved):

nrwiersma (approved):

PR #1950, opened 2024-04-20
Upgrade to AWS SDK v2 (#1938)

### LINKED ISSUE #1938: Use AWS SDK v2 for Go

The [existing AWS SDK we use](https://github.com/aws/aws-sdk-go?tab=readme-ov-file) is in maintenance mode. We should migrate to [v2](https://aws.github.io/aws-sdk-go-v2/docs/) when we can. No official dates for migration have been issued at the time of writing.

Reimplement the AWS SDK with the new version.

- Not upgrading

**Additional context**

N/A

k124k3n:

The [existing AWS SDK we use](https://github.com/aws/aws-sdk-go?tab=readme-ov-file) is in maintenance mode. We should migrate to [v2](https://aws.github.io/aws-sdk-go-v2/docs/) when we can.

## How is the fix applied?

Refactor all AWS SDK v1 related code to v2 one.

Fixes #1938


.github/workflows/ci.yml
```diff
@@ -30,7 +30,7 @@ jobs:
   build:
     env:
       ATHENS_MONGO_STORAGE_URL: mongodb://localhost:27017
-      ATHENS_MINIO_ENDPOINT: localhost:9000
+      ATHENS_MINIO_ENDPOINT: http://localhost:9000
```
matt0x6F, line 33: This won't be breaking for existing Minio users because we trim the protocol later in the config, however, would this be breaking for existing AWS users who do not leverage a protocol in their configuration?

@nrwiersma what do you think? The context is that the newest AWS SDK requires a protocol.

matt0x6F: On second review, I think this is fine.

pkg/storage/minio/minio.go
```diff
@@ -53,3 +54,10 @@ func NewStorage(conf *config.MinioConfig, timeout time.Duration) (storage.Backen
 	}
 	return &storageImpl{minioClient, minioCore, bucketName}, nil
 }
+
+// TrimHTTP trims "http://" or "https://" prefix from input string.
```
matt0x6F, line 58: Can you expand on this comment, maybe even linking to the issue would help with context.

matt0x6F (changes requested):
Requested a minor documentation change. Everything looks good otherwise!

pkg/storage/minio/minio.go
k124k3n:
```suggestion
// TrimHTTP trims "http://" or "https://" prefix from input string.
// By default, Minio doesn't need to specify protocol in the URL so it should be trimmed. 
// Related issue: https://github.com/gomods/athens/issues/1938#issuecomment-2067590653
```

k124k3n: Is this okay?

matt0x6F: That looks great!

k124k3n: Thanks

matt0x6F (approved):
Looks good!

PR #1957, opened 2024-04-30
gcp/saver: Only return errors.KindAlreadyExists if all three exist

dwbuiten:

In #1124, a GCP lock type was added as a singleflight backend. As part of this work, the GCP storage backend's `Save()` was made serial, likely because `moduploader.Upload` requires a call to `Exists()` before it, rendering the GCP lock less useful, by doubling the calls to GCS.

However, by doing this, the existence check was now only checking the existence of the mod file, and not the info or zip. This meant that if during a `Save()`, the zip or info uploads failed, on subsequent rquests, that when using the GCP singleflight backend, Athens woul assume everything had been stashed and saved properly, and then fail to serve up the info or zip that had failed upload, meaning the cache was in an unhealable broklen state, requiring a manual intervention.

## How is the fix applied?

To fix this, without breaking the singleflight behavior, introduce a metadata key that is set on the mod file during its initial upload, indicating that a `Stash` is still in progress on subsequent files, which gets removed once all three files are uploaded successfully, which can be checked if it it is determined that the mod file already exists. That way we can return a `errors.KindAlreadyExists` if a `Stash` is in progress, but also properly return it when a `Stash` is *not* currently in progress if and only if all three files exist on GCS, which prevents the cache from becoming permanently poisoned.

One note is that it is possible the GCS call to remove the metadata key fails, which would mean it is left on the mod object forever. To avoid this, consider it stale after 2 minutes.

N/A

# Extra Notes

This is my first time touching the Athens codebase, so apologies for any faux pas. We hit this at `$dayjob`, so thought I would take a stab at fixing it.

It is a more complex than I would like, so comments very welcome.

This also contains a small fix I found when writing the test: if `io.Copy` failed during an upload to GCS, it would call `Close()` on the writer, which would leave a partial/broken file. The `storage` docs say to cancel the context to prevent this.

### DIFF
--- a/pkg/storage/gcp/saver.go
+++ b/pkg/storage/gcp/saver.go
@@ -133,9 +133,12 @@ func (s *Storage) upload(ctx context.Context, path string, stream io.Reader, fir
 	const op errors.Op = "gcp.upload"
 	ctx, span := observ.StartSpan(ctx, op.String())
 	defer span.End()
+	cancelCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
 	wc := s.bucket.Object(path).If(storage.Conditions{
 		DoesNotExist: true,
-	}).NewWriter(ctx)
+	}).NewWriter(cancelCtx)
 
 	// We set this metadata only for the first of the three files uploaded,
 	// for use as a singleflight lock.
@@ -148,7 +151,7 @@ func (s *Storage) upload(ctx context.Context, path string, stream io.Reader, fir
 	// Once we support private storage buckets this may need refactoring
 	// unless there is a way to set the default perms in the project.
 	if _, err := io.Copy(wc, stream); err != nil {
-		_ = wc.Close()
+		// Purposely do not close it to avoid creating a partial file.
 		return err
 	}
 
--- a/pkg/storage/gcp/saver.go
+++ b/pkg/storage/gcp/saver.go
@@ -13,9 +13,9 @@ import (
 	googleapi "google.golang.org/api/googleapi"
 )
 
-// After how long we consider an "in_progress" metadata key stale,
+// Fallback for how long we consider an "in_progress" metadata key stale,
 // due to failure to remove it.
-const inProgressStaleThreshold = 2 * time.Minute
+const fallbackInProgressStaleThreshold = 2 * time.Minute
 
 // Save uploads the module's .mod, .zip and .info files for a given version
 // It expects a context, which can be provided using context.Background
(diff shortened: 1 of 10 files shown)


matt0x6F:
@dwbuiten thanks for your contribution to Athens! I'm doing a release today, but I'll get to reviewing this and we'll see if we can get it into one of the fix releases.

Just off the bat, one thing that'd help show what this is fixing is if you're able to write tests to it in E2E or unit tests.

dwbuiten:
@matt0x6F Sure - adding a unit test to `with_gcs_test.go`. And it is a good thing too... I noticed the test there does not run by default (without GCP creds in the env), and when I ran the existing test, I realized this fix actually breaks singleflight, while fixing the poison cache problem. I need to ponder a bit on the subject, but will update this PR with a different fix + test after.

dwbuiten:
@matt0x6F It ended up being more complex than I thought (or would like it to be), but old and new tests all pass now. I updated the PR description to reflect that.

CCing @dfinkel who helped me think this through.

pkg/storage/gcp/saver.go
```diff
@@ -12,6 +13,10 @@ import (
 	googleapi "google.golang.org/api/googleapi"
 )
 
+// After how long we consider an "in_progress" metadata key stale,
+// due to failure to remove it.
+const inProgressStaleThreshold = 2 * time.Minute
```
matt0x6F, line 18: Would a user ever want to configure this threshold?

matt0x6F (changes requested):
Looks good! I have an outstanding question regarding user configurability, but overall I think the PR looks great. I'm also glad we wrote some tests out. It makes me wonder if the GCP Storage ever _reliably_ worked without these changes.

pkg/storage/gcp/saver.go
dwbuiten: That's a good question. I could perhaps see it mattering on extremely slow networks, where it could take more than 2 minutes to upload the zip? That seems a little tenuous, though, I suppose.

I don't think I have a strong opinion one way or another.

matt0x6F: For Athens that's actually not an unreasonable usecase.

dwbuiten: I have added a singleflight option for it and updated the PR, but the way it works is kinda ugly, stemming from the fact that the config option is a singleflight backend option that needs to be accessed in the storage backend in `Save()`.

Happy to rework/remove/separate out as necessay, given feedback.

dwbuiten: Also, please review the config changes closely, in case I added it incorrectly.

matt0x6F (approved):
Looks great! Thanks for adding the configuration. We'll get this out in our next feature release.

matt0x6F (approved):

Issue #1961, opened 2024-05-09
Athens is creating unsigned requests to S3 buckets causing 403 errors


midestefanis:
**Describe the bug**
I'm running athenas in AWS ECS Fargate with S3 as the storage. I've followed this documentation to configure everything -> https://docs.gomods.io/install/install-on-aws-ecs-fargate/

This are the ENVs in the athenas container:

```
/ # env
ATHENS_STORAGE_TYPE=s3
HOSTNAME=********.us-west-2.compute.internal
SHLVL=1
HOME=/root
AWS_CONTAINER_CREDENTIALS_RELATIVE_URI=/v2/credentials/*********
AWS_EXECUTION_ENV=AWS_ECS_FARGATE
ECS_AGENT_URI=http://*******/api/eab1a45ef8034d1b9ef8fcb2b96523ad-********
GO111MODULE=on
AWS_DEFAULT_REGION=us-west-2
ECS_CONTAINER_METADATA_URI_V4=http://********/v4/eab1a45ef8034d1b9ef8fcb2b96523ad-****
TERM=xterm-256color
ECS_CONTAINER_METADATA_URI=http://*********/v3/eab1a45ef8034d1b9ef8fcb2b96523ad-*****
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
ATHENS_LOG_LEVEL=debug
AWS_USE_DEFAULT_CONFIGURATION=true
LANG=C.UTF-8
AWS_REGION=us-west-2
ATHENS_S3_BUCKET_NAME=athens-proxy-poc
```

**Error Message**

While testing the [walktrhough](https://docs.gomods.io/walkthrough/#using-the-athens-proxy) example i've found this problem: 

```
export GOPROXY=http://athens-*******.elb.us-west-2.amazonaws.com,direct
export GO111MODULE=on

go run .
go: github.com/athens-artifacts/samplelib@v1.0.0: reading http://athens-*************.elb.us-west-2.amazonaws.com/github.com/athens-artifacts/samplelib/@v/v1.0.0.mod: 500 Internal Server Error
```

This is what athens is logging when i get locally that 500 error:

```
[...]

PR #1963, opened 2024-05-13
Fix aws default credentials

yongzhang:

Unable to access s3 with default config in v0.14.0, but v0.13.x works very well.

## How is the fix applied?

Fixes https://github.com/gomods/athens/issues/1961

### DIFF
--- a/pkg/storage/s3/s3.go
+++ b/pkg/storage/s3/s3.go
@@ -40,10 +40,6 @@ func New(s3Conf *config.S3Config, timeout time.Duration, options ...func(*aws.Co
 		return nil, errors.E(op, err)
 	}
 
-	// Remove anonymous credentials from the default config so that
-	// session.NewSession can auto-resolve credentials from role, profile, env etc.
-	awsConfig.Credentials = nil
-
 	for _, o := range options {
 		o(&awsConfig)
 	}


pkg/storage/s3/s3.go
```diff
@@ -40,10 +40,6 @@ func New(s3Conf *config.S3Config, timeout time.Duration, options ...func(*aws.Co
 		return nil, errors.E(op, err)
 	}
 
-	// Remove anonymous credentials from the default config so that
-	// session.NewSession can auto-resolve credentials from role, profile, env etc.
-	awsConfig.Credentials = nil
```
matt0x6F, line 45: Ah, this is interesting. I wonder if roles, profiles, and ENV settings will break now?

matt0x6F: Okay, I think I understand the problem. Line 38 should be using [this](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws/external#LoadDefaultAWSConfig)

matt0x6F: Okay, nevermind - [this](https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/#specifying-credentials) clearly says that roles, profiles, env etc should be covered by `LoadDefaultConfig`

matt0x6F (approved):

matt0x6F:
@yongzhang have you tested the build of this branch in your own infrastructure to verify it works?

yongzhang:
> @yongzhang have you tested the build of this branch in your own infrastructure to verify it works?

yea I tested and confirmed it works, but actually I didn't get a chance to test [this part](https://github.com/gomods/athens/pull/1963/files#diff-15ce5177313138e8d4c811ae00f029e62e02c05d1c4a8ba8df349dee84826febL51-L60).

PR #1974, opened 2024-07-09
Compatibility with proxy.golang.org

sanyatuning:

Empty response body when major version not found.
The official proxy responds with a non-empty body.
(And some library depends on this)
```
$ curl 'https://proxy.golang.org/github.com/!ak-!army/xlog/v2/@v/list'
not found: module github.com/Ak-Army/xlog/v2: no matching versions for query "latest"
```

## How is the fix applied?

I created a simplified solution for my case.
Can you update it to fit your conventions?

Fixes https://github.com/icholy/gomajor/pull/34


matt0x6F:
@sanyatuning Thanks for your contribution to Athens! Do you mind writing a test that ensures this behaivor?

matt0x6F (approved):

PR #1978, opened 2024-08-14
Update protocol.md

berlintay:
the get requests were not working
```bash
GET $GOPROXY:3000/?=github.com/acidburn
```
is what worked for me
Ubuntu WSL2 , might differ elsewheres but just a suggestion :-)

The functionality of the commands for examples,
urls were not working without `?=` after the `$HOST$PORT/`
just wanted to clarify
Screenshot 2024-08-14 025019

## How is the fix applied?

Fixes #


matt0x6F (changes requested):
The proxy should not require query parameters.

PR #1981, opened 2024-08-27
Fix helm install command in docs

sb464f:

Helm install command in the doc was wrong

## How is the fix applied?

Update the flags for Name


DrPsychick (approved):
Thank you for your contribution!

PR #1982, opened 2024-09-03
Document log level format

raxod502-plaid:
Documentation mentioned that log level could be customized via an environment variable, but didn't explain what values could be assigned to that environment variable. I looked at the source code to find out, and copied that information into the documentation.


docs/content/configuration/logging.md
```diff
@@ -10,6 +10,8 @@ Athens is designed to support a myriad of logging scenarios.
 
 The standard structured logger can be configured in `plain` or `json` formatting via `LogFormat` or `ATHENS_LOG_FORMAT`. Additionally, verbosity can be controlled by setting `LogLevel` or `ATHENS_LOG_LEVEL`. In order for the standard structured logger to work, `CloudRuntime` and `ATHENS_CLOUD_RUNTIME` should not be set to a valid value.
 
+The logging is via [Logrus](https://github.com/sirupsen/logrus), so the allowed values for logging config options are determined by that project. For example, `ATHENS_LOG_LEVEL` can be `debug`, `info`, `warn`/`warning`, `error`, etc.
```
Ullaakut, line 13: Personally I don't think it's relevant to mention what logger is used, and if we want to specify the accepted values, which IMO is obvious enough, I'd simplify it to:

```suggestion
Supported log levels are `trace`, `debug`, `info`, `warn`, `error`, `fatal` and `panic`.
```

But again, personally I don't think it's relevant to list that here, since it's universally the same log levels in pretty much any logger in Go, besides maybe trace and panic which might be slightly unusual but I would argue no one would likely use to run Athens unless they know what they are doing in which case they don't really need this extra line in the documentation.

Ullaakut (changes requested):
Thank you @raxod502-plaid for your contribution!

Unfortunately like I explain in my comment, I would rather not merge this, but if someone thinks it's relevant, I wouldn't oppose it. If no one reaches out within a few days I'll close the PR.

docs/content/configuration/logging.md
nrwiersma: @matt0x6F Do you have an opinion on this?

raxod502-plaid: > unless they know what they are doing

Why would we assume that all people who deploy Athens are actively configuring low-level logging libraries for Golang in their regular work? Those seem like kind of unrelated work areas to me?

And why would we not want to clarify explicitly what standard Athens is following here, as opposed to leaving the user to guess?

matt0x6F: Athens' users have, historically, been people that work with Go but that could be changing. The change looks fine enough.

matt0x6F (approved):

PR #1987, opened 2024-09-28
Upgrade to Go 1.22 toolchain

matt0x6F:

Upgrades the Go toolchain to 1.22 which fixes several CVEs

## How is the fix applied?

Updates to the various files that set the Go version

N/a


.github/workflows/goreleaser.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: setup-go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.20"
+          go-version: "1.22"
```
nrwiersma, line 17: This will use the `go.mod` version, and save updating it all the time.
```suggestion
          go-version-file: 'go.mod'
```

nrwiersma (approved):
LGTM

.github/workflows/goreleaser.yml
matt0x6F: Good idea!

PR #1988, opened 2024-10-01
feat: support GitHub App authentication

nrwiersma:

This adds support for GitHub App authentication using Git Credential Helpers.

## How is the fix applied?

The GHA Git Credential Helper is installed in the Docker image by default, and the process is documented.


docs/content/configuration/authentication.md
```diff
@@ -308,3 +308,32 @@ $ docker run --rm -d \
...
+    useHttpPath = true
+
+[credential "https://github.com"]
+    helper = "cache --timeout=3500"
+
+[url "https://github.com"]
+    insteadOf = ssh://git@github.com
+```
+
+This instructs Git to authenticate with the GitHub App and cache the results for 3500s (the authentication token is valid for 1 hour).
```
Ullaakut, line 336:
```suggestion
This instructs Git to authenticate with the GitHub App and cache the results for 3600s (the authentication token is valid for 1 hour).
```

Ullaakut (approved):
I assume 3500 is a typo? Only thing that seems wrong

matt0x6F (changes requested):
So, this only works for github.com. The application is leveraging [bradleyfalzon/ghinstallation](https://github.com/bradleyfalzon/ghinstallation/) which has a clear path for supporting both GHE and github.com. Can we PR, fork, or import in order to support both instead?

nrwiersma:
I am wondering if GHE support should be a second step for this? I would prefer to try upstream the GHE support first, before running a separate fork, and this can take time. 

Would glad make an Issue to keep track and a note in the docs about this.

Ullaakut (approved):

matt0x6F (approved):

PR #1989, opened 2024-10-01
fix: go in docker failing

nrwiersma:

In v0.15.2, an error occurs when trying to use the proxy:
```
INFO[5:50PM]: exit status 2: go: cannot find GOROOT directory: 'go' binary is trimmed and GOROOT is not set
	http-method=GET http-path=/github.com/hamba/avro/@v/list kind=Not Found module= operation=download.ListHandler ops=[download.ListHandler pool.List protocol.List vcsLister.List] request-id=ba64734c-a3b6-4cf2-a0ac-54199c55153a version=
```

It seems in Go 1.21, so defaults are no longer baked into the binary, causing issues when running Go.

My first attempt at a fix was to set `GOROOT` in the Dockerfile, but Athens does not pass all env vars to the Go binary when executing it. I did consider passing `GOROOT` by default to the Go binary, but this seems wrong for people not running the docker image. There are also multiple go call sites, each using their own env calling conventions.

The downside to this approach is the image grows from `170MB` to `379MB`.

## How is the fix applied?

To get around the complexity, the root of the Dockerfile is set to `golang:<ver>-alpine`. This added the least complexity.

### DIFF
--- a/cmd/proxy/Dockerfile
+++ b/cmd/proxy/Dockerfile
@@ -47,7 +47,7 @@ RUN apk add --update git git-lfs mercurial openssh-client subversion procps foss
 ARG USER=athens
 RUN adduser -D -h /home/$USER $USER
 
-CMD mkdir /go
+RUN mkdir /go
 
 EXPOSE 3000
 
--- a/cmd/proxy/Dockerfile
+++ b/cmd/proxy/Dockerfile
@@ -27,7 +27,7 @@ RUN DATE="$(date -u +%Y-%m-%d-%H:%M:%S-%Z)" && \
     -ldflags "-X github.com/gomods/athens/pkg/build.version=$VERSION -X github.com/gomods/athens/pkg/build.buildDate=$DATE -s -w" \
     -o /bin/athens-proxy ./cmd/proxy
 
-FROM golang:alpine
+FROM alpine:${ALPINE_VERSION}
 
 ENV GOPROXY="https://proxy.golang.org,direct" \
     GOSUMDB="sum.golang.org" \
@@ -36,11 +36,13 @@ ENV GOPROXY="https://proxy.golang.org,direct" \
 
 COPY --from=builder /bin/athens-proxy /bin/athens-proxy
 COPY --from=builder /go/src/github.com/gomods/athens/config.dev.toml /config/config.toml
+COPY --from=builder /usr/local/go/bin/go /bin/go
 
 RUN chmod 644 /config/config.toml
 
 # Add tini, see https://github.com/gomods/athens/issues/1155 for details.
-RUN apk add --update git git-lfs mercurial openssh-client subversion procps fossil tini
+RUN apk add --update git git-lfs mercurial openssh-client subversion procps fossil tini && \
+    mkdir -p /usr/local/go
 
 ARG USER=athens
 RUN adduser -D -h /home/$USER $USER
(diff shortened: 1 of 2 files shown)


Ullaakut (approved):

cmd/proxy/Dockerfile
```diff
@@ -27,19 +27,17 @@ RUN DATE="$(date -u +%Y-%m-%d-%H:%M:%S-%Z)" && \
     -ldflags "-X github.com/gomods/athens/pkg/build.version=$VERSION -X github.com/gomods/athens/pkg/build.buildDate=$DATE -s -w" \
     -o /bin/athens-proxy ./cmd/proxy
 
-FROM alpine:${ALPINE_VERSION}
+FROM golang:${GOLANG_VERSION}-alpine
```
matt0x6F, line 30: So, I think if we just set the values for GOROOT=/go, GOSUMDB=off, and GOPROXY=direct then it'll fix this issue.

matt0x6F (changes requested):

cmd/proxy/Dockerfile
matt0x6F: The only thing I'm not super sure of is if our users would find turning the sumdb off and proxy to direct a sane default. Maybe we can figure out what it used to default to?

matt0x6F (approved):

matt0x6F:
For thoroughness, I pulled your branch, built the image locally, and ran it:

```
❯ docker run -it athens-test /bin/sh                        
/ # go mod download -json github.com/golang/go@latest
{
        "Path": "github.com/golang/go",
        "Version": "v0.0.0-20241002005801-630d4fb600ae",
        "Query": "latest",
        "Info": "/root/go/pkg/mod/cache/download/github.com/golang/go/@v/v0.0.0-20241002005801-630d4fb600ae.info",
        "GoMod": "/root/go/pkg/mod/cache/download/github.com/golang/go/@v/v0.0.0-20241002005801-630d4fb600ae.mod",
        "Zip": "/root/go/pkg/mod/cache/download/github.com/golang/go/@v/v0.0.0-20241002005801-630d4fb600ae.zip",
        "Dir": "/root/go/pkg/mod/github.com/golang/go@v0.0.0-20241002005801-630d4fb600ae",
        "Sum": "h1:j5XR87AKVbqM/5KRzpMFSxhMzMfXeLGM4nvLQZCcp9Y=",
        "GoModSum": "h1:VnTjtYw+XLkxokOYpCb9NBW3cOTFO8+uqxF7o10XJQk="
}
/ # 
```

matt0x6F:
and just to pin down the source: https://github.com/golang/go/blame/630d4fb600ae309476c87a4ca0e29b60425bc0c1/src/cmd/go/main.go#L127-L130

cmd/proxy/Dockerfile
nrwiersma: This does not work, you must move the GOROOT through the code, otherwise you get the same error.

I will make the needed changes, so we have a full comparison, but my concerns are as follows:
1. It is not obvious that GOROOT needs to move through the code, and will likely be missed down the line
2. I doubt this is the last rodio we will have with moving the Go bin around like this, and breaks are a little untestable.

nrwiersma:
This version uses the original Alpine base. The env control was more centralised than I had originally thought, so it is less ugly than I had imagined.

matt0x6F (approved):

PR #1994, opened 2024-10-16
Support using redis urls to construct the redis client

opalmer:
Currently Athens only supports connecting to Redis using a `hostname:port` combination in addition to a password. While this works in most cases it also means that if you have other options you wish to supply Athens has to be updated to support them. As a basic example Redis clusters that require TLS currently are not supported by Athens but with this change you can simply supply a [redis url](https://github.com/redis/redis-specifications/blob/master/uri/redis.txt) to connect over TLS. It also makes it easy to override the password, set a username and more all from a single configuration option:

`rediss://username:password@redis.example.com:6379/1?protocol=3`

The redis cluster I'm using has a specific username, password and requires TLS to connect to it. Athens doesn't appear to support these options today (other than password).

## How is the fix applied?

Updated the code constructing the redis client to support using a url, example here: https://github.com/redis/go-redis?tab=readme-ov-file#connecting-via-a-redis-url.

This change also ensures that if you are also setting a password but start using the new url form, and the two don't agree, an error will be returned. Didn't want someone to end up in a situation where they defined a password in the configuration but not in the redis url and expected the two to be combined.


scripts/build-image/Dockerfile
```diff
@@ -4,12 +4,12 @@ WORKDIR /tmp
 
 # Install Helm
 ENV HELM_VERSION=2.13.0
-RUN curl -sLO https://kubernetes-helm.storage.googleapis.com/helm-v${HELM_VERSION}-linux-amd64.tar.gz && \
+RUN curl -sLO https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz && \
```
matt0x6F, line 7: Why is this change required?

matt0x6F:
Overall your solution looks good, and thanks for the contribution!

scripts/build-image/Dockerfile
opalmer: I was getting a 403 for that url when trying to run the build locally. Not sure if this file is used in the build or not, if it is maybe cloud infrastructure is whitelisted. Happy to back it out if you'd like or only use that url if the first one fails.

PR #2010, opened 2024-12-02
replacing logrus with slog v1

### LINKED ISSUE #1890: Replace logrus with log/slog
Given the release of [log/slog](https://pkg.go.dev/golang.org/x/exp/slog), it would be nice to replace logrus with it. There isn't really a great standard for logging in this repository, so this could also be considered a feature request for better logging in general.

I found it hard to add new log lines when working on https://github.com/gomods/athens/pull/1887.

JodhwaniMadhur:

Removing logrus and implementing slog

## How is the fix applied?

Replaced logrus.

Implemented functions that are missing in slog but were present in slog

Fixes #1890


JodhwaniMadhur:
Ok, so this actually is not an MR at the moment but I needed some help and so wanted to ask about how can i resolve these issues and also get a review at my code so that I can understand if I am in the right direction or not.

image

image

pkg/log/format.go
```diff
@@ -1,73 +1,75 @@
...
 	"github.com/fatih/color"
-	"github.com/sirupsen/logrus"
 )
 
-func getGCPFormatter() logrus.Formatter {
-	return &logrus.JSONFormatter{
-		FieldMap: logrus.FieldMap{
-			logrus.FieldKeyLevel: "severity",
-			logrus.FieldKeyMsg:   "message",
-			logrus.FieldKeyTime:  "timestamp",
```
marwan-at-work, line 19: We need to also replace attrs for these as well as the level in order for the logs to be respected in GCP environments.

JodhwaniMadhur:
I also see a ticket open for switching to opentelemetry. since it also supports logging. please lmk and i can then work on the same and we can close this issue

pkg/log/format.go
JodhwaniMadhur: ok, but anything more? I would love a detailed review.

Ullaakut (changes requested):
Superficial review

Ullaakut (changes requested):

pkg/log/log_test.go
```diff
@@ -7,16 +7,16 @@ import (
 	"testing"
 	"time"
 
-	"github.com/sirupsen/logrus"
+	"log/slog"
```
nrwiersma, line 10: The imports are not formatted correctly.

nrwiersma (changes requested):

marwan-at-work:
@JodhwaniMadhur the idea here is correct: we should replace logrus with log/slog. However, the approach should be as follows: 

1. Let's only change the athens/log package. So that importers of the athens/log package don't need to change at all. 
2. This will fix some of your build errors above, and it will also divide up the PR and make it a smaller change rather than a large one. 
3. If we ever felt that we needed to change athens/log we can do so in a separate/follow up discussion.

pkg/log/entry.go
```diff
@@ -16,46 +18,52 @@ type Entry interface {
...
-	*logrus.Entry
+	*slog.Logger
 }
 
 func (e *entry) WithFields(fields map[string]any) Entry {
-	ent := e.Entry.WithFields(fields)
-	return &entry{ent}
+	ent := e.WithFields(fields)
+	return ent
 }
```
Ullaakut, line 44: Isn't this straight up an infinite loop?

pkg/log/log.go
```diff
@@ -1,47 +1,70 @@
...
-	e := &entry{Entry: logrus.NewEntry(l.Logger)}
+	e := &entry{Logger: l.Logger}
 	e.SystemErr(err)
 }
 
 // WithFields Entry implementation.
 func (l *Logger) WithFields(fields map[string]any) Entry {
-	e := l.Logger.WithFields(fields)
+	return l.WithFields(fields)
+}
```
Ullaakut, line 43: Infinite loop?

Ullaakut (changes requested):
I assume this doesn't compile/doesn't work as intended at the moment does it?

cmd/proxy/main.go
```diff
@@ -37,26 +37,17 @@ func main() {
...
-			logger.WithError(err).Warn("Could not close logrus writer pipe")
-		}
-	}()
-	stdlog.SetOutput(logrusErrorWriter)
-	stdlog.SetFlags(stdlog.Flags() &^ (stdlog.Ldate | stdlog.Ltime))
-
 	handler, err := actions.App(logger, conf)
 	if err != nil {
-		logger.WithError(err).Fatal("Could not create App")
+		logger.With(err.Error()).Error("Could not create App")
```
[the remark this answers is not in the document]
Ullaakut: It's exactly the reason why I hate loose coupling like `slog` does. They have stuff like `slog.Str` but you don't _have_ to use it.

```go
logger.Info(
  "incoming request",
  "method", "GET",
  slog.Int("time_taken_ms", 158),
  slog.String("path", "/hello/world?q=search"),
  "status", 200,
  slog.String(
    "user_agent",
    "Googlebot/2.1 (+http://www.google.com/bot.html)",
  ),
)
```

This is apparently valid

pkg/log/entry.go
JodhwaniMadhur: will take care of it! I am noob😅

JodhwaniMadhur:
> I assume this doesn't compile/doesn't work as intended at the moment does it?

Nope, it doesn't. But will take care of everything. 

 thanks for all the suggestions.

JodhwaniMadhur:
> @JodhwaniMadhur the idea here is correct: we should replace logrus with log/slog. However, the approach should be as follows:
> 
> 1. Let's only change the athens/log package. So that importers of the athens/log package don't need to change at all. We simply implement a slog version of the Entry interface.
> 2. This will fix some of your build errors above, and it will also divide up the PR and make it a smaller change rather than a large one.
> 3. If we ever felt that we needed to change athens/log we can do so in a separate/follow up discussion.
But then in the entry we have Severity coming in as logrus.Level which indeed should be then a generic levels decided by us or we switch to slog.Level, isn't it?

pkg/middleware/log_entry.go
```diff
@@ -15,11 +14,11 @@ func LogEntryMiddleware(lggr *log.Logger) mux.MiddlewareFunc {
...
-			ent := lggr.WithFields(logrus.Fields{
-				"http-method": r.Method,
-				"http-path":   r.URL.Path,
-				"request-id":  requestid.FromContext(ctx),
-			})
+			ent := lggr.With(
+				"http-method", r.Method,
+				"http-path", r.URL.Path,
+				"request-id", requestid.FromContext(ctx),
+			)
```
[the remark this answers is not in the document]
JodhwaniMadhur: If I replace this then I start getting errors in SetEntryInContext as ent is an entry and using with gives *logger so instead used a map in here just like logrus internally does.

LMK if I missed something here

pkg/log/format.go
JodhwaniMadhur: will work on this

JodhwaniMadhur: have added code for this please check and let me know if this is good enough

pkg/log/log_test.go
JodhwaniMadhur: please can you explain again what you meant here, can't understand as to how I should format it. 

I mean is it the line spacing in between?

pkg/log/log.go
```diff
@@ -1,47 +1,70 @@
...
+	"os"
 )
 
 // Logger is the main struct that any athens
 // internal service should use to communicate things.
 type Logger struct {
-	*logrus.Logger
+	*slog.Logger
+
+	Out *bytes.Buffer
```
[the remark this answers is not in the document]
JodhwaniMadhur: where should this be then?

cmd/proxy/main.go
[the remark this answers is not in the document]
JodhwaniMadhur: Hope I fixed this?

JodhwaniMadhur:
This time I have kept in mind to just change the interface and not the function calls of it that are being used from it. only 2 errors remain now which I wants some guidance about how to resolve.

image

I am guessing that left one would require a Fatal implementation?

Not sure of the right one though.

pkg/log/log.go
```diff
@@ -1,47 +1,90 @@
...
+func (l *Logger) WithContext(ctx context.Context) Entry {
+	keys := map[string]any{
+		"context": ctx,
+	}
+	return l.WithFields(keys)
+}
 
-	return &entry{e}
+// Define WriterLevel
+func (l *Logger) WriterLevel(level slog.Level) *io.PipeWriter {
```
JodhwaniMadhur, line 73: Please let me know of this function implementation?

had to take help of AI here.

pkg/log/log_test.go
nrwiersma: Import should be formatted:
```
import (
  <std imports>

  <external imports>
)
```

pkg/log/entry.go
```diff
@@ -1,66 +1,181 @@
...
+func (e *entry) Warn(args ...interface{}) {
+	e.logger.Warn(fmt.Sprint(args...))
+}
+
+func (e *entry) Error(args ...interface{}) {
+	e.logger.Error(fmt.Sprint(args...))
+}
+
+func (e *entry) Fatal(args ...interface{}) {
+	e.logger.Error(fmt.Sprint(args...)) // slog doesn't have Fatal, using Error
```
nrwiersma, line 101: This must call `os.Exit(1)`. That is what makes it Fatal

pkg/log/log.go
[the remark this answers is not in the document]
Ullaakut: It should not exist, unless there is a good reason for it which I can't see off the top of my head 🤔

JodhwaniMadhur:
can anyone guide me as to how I should modify the Severity function? as it was designed for logrus but for slog the TestSeverity fails

for logrus, the levels are
Panic = 0
Fatal = 1
err = 2
warn = 3
Info = 4
Debug = 5
Trace = 6

but for slog the levels are:
err = 8
warn = 4
info = 0
debug = -4

@marwan-at-work @nrwiersma @Ullaakut

pkg/log/entry.go
```diff
@@ -1,66 +1,243 @@
...
-	WithFields(fields map[string]any) Entry
-
-	// SystemErr is a method that disects the error
-	// and logs the appropriate level and fields for it.
-	SystemErr(err error)
+	LogOps
+	FormattedLogOps
+	ContextualLogOps
+	SystemLogger
 }
```
marwan-at-work, line 96: We should keep this interface exactly the same and just implement it in terms of slog, why create all of the new methods above if they were not being used?

marwan-at-work (commented):
As to your question, if you look at the entry interface, we weren't using all of logrus's levels, we also were just using debug/info/warn/error, so let's just keep those.

pkg/log/entry.go
JodhwaniMadhur: Even if I  remove the rest, I will have to keep them in a separate interface and take those interfaces in entry since if I write all functions inside entry, the linter in the pipeline gives error for interfacebloat

JodhwaniMadhur: As for the extra methods, I can remove those but the problem for logrus vs slog levels and problems in Severity due to this still remains

marwan-at-work: > As for the extra methods, I can remove those but the problem for logrus vs slog levels and problems in Severity due to this still remains

I'm not sure how they still remain, all you need to do is create a separate file that implements Entry in terms of log/slog?

marwan-at-work: Here's an example: 

```
type slogger struct {
	l *slog.Logger
}

// Debugf implements Entry.
func (s *slogger) Debugf(format string, args ...any) {
	s.l.Debug(fmt.Sprintf(format, args...))
}

// Errorf implements Entry.
func (s *slogger) Errorf(format string, args ...any) {
	s.l.Error(fmt.Sprintf(format, args...))
}
```

For severity in the `errors` package, you can continue to use logrus levels, _or_ you can change it to slog, or better yet create an abstraction in the athens log package because the error package shouldn't import a third party lib

matt0x6F: An abstraction is preferred, imo.

matt0x6F:
I would expect to see a go.mod/go.sum change that shows we removed Logrus as a dependency

pkg/log/entry.go
JodhwaniMadhur: ok, got the abstraction part but still a bit confused, so you mean to say that i should only have the function declarations in entry and then implement those in log.go like it was done before my changes for SystemErr implementation or like for WithFields implementation?

JodhwaniMadhur: ??

marwan-at-work: I mean that you keep the interface exactly the same and just create a log/slog v1 implementation of that interface. Therefore, nothing changes outside of the log package (and the errors/severity, due to the abstraction we talked about above)

matt0x6F:
Rerunning these tests

matt0x6F:
Done with #2127

PR #2021, opened 2025-01-15
Add configurable Stash timeout

### LINKED ISSUE #1964: Make Stash timeout configurable

#1690 raised an issue where timeouts occur after 10 minutes due to a [hardcoded timeout in Stash](https://github.com/gomods/athens/blob/87ebcbeb92de0f9fd6bfa9596efec3fbbff1a5e6/pkg/stash/stasher.go#L55). 

Make the Stash timeout configurable

Do nothing

**Additional context**
N/A

vlerdman:

Currently stash has hardcoded 10 minutes timeout which may leads to timeout errors.

## How is the fix applied?

Add `timeout` field to `stasher` for using in `Stash` call. It's configurable via `ATHENS_TIMEOUT` var.

Fixes #1964
Fixes #1960

### DIFF
--- a/pkg/stash/stasher_test.go
+++ b/pkg/stash/stasher_test.go
@@ -5,10 +5,10 @@ import (
 	"io"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/gomods/athens/pkg/index/nop"
 	"github.com/gomods/athens/pkg/storage"
-	"time"
 )
 
 type stashTest struct {


cmd/proxy/actions/app_proxy.go
```diff
@@ -105,7 +105,7 @@ func addProxyRoutes(
 	if err != nil {
 		return err
 	}
-	st := stash.New(mf, s, indexer, stash.WithPool(c.GoGetWorkers), withSingleFlight)
+	st := stash.New(mf, s, indexer, c.TimeoutDuration(), stash.WithPool(c.GoGetWorkers), withSingleFlight)
```
Ullaakut, line 108: If you are reusing the ATHENS_TIMEOUT variable, I'd suggest documenting that it now also serves this purpose.

Also, I don't know if there is a direct correlation between the timeout users might want to use for the VCS lister and for the stash. I'll let the others decide on whether it should be separate timeouts or not.

vlerdman: Thanks for comment! Also want to mention, that stasher uses storage backend for [uploading](https://github.com/gomods/athens/blob/main/pkg/stash/stasher.go#L71) files, and for some [providers](https://github.com/gomods/athens/blob/main/cmd/proxy/actions/app.go#L119) (for example s3) it uses ATHENS_TIMEOUT as timeout for upload operations. It was a reason for reusing ATHENS_TIMEOUT instead of adding separete var, as for me it would be more clear behaviour.

Ullaakut:
Closed in favor of #2097

PR #2022, opened 2025-01-15
Add miss chinese translation of all section

day253:
How is the fix applied?

Add miss chinese translation of all section


day253:
> Can't attest to the quality of the translation, I assume it is AI-generated? Content seems fine, snippets/commands are untouched (besides some comments). No code is changed, everything is correctly suffixed `.zh.md`. LGTM. Will let someone else take the decision on merging or not though.

I apologize, but this is still in draft form and requires further revisions. The AI only helped me generate the basic structure, and I will carefully refine it later.

Ullaakut:
Actually I just realized that there is only one other localized document in here, in Spanish, which seems not to have been updated in some time (6 years, against 9 months for the original English document) which leads me to wonder whether it's worth it to localize at all.

I'd say it's only worth it if you can guarantee that you or someone else will maintain the Chinese localization up-to-date with the English version to the best of your ability. If this is a one-time contribution, it is not worth it as it will only end up worsening the quality/reliability of the documentation over time.

Also I am not convinced that _all_ of the markdown documents within the repository need localization. I would assume only the main ones need to. But I'll let others share their opinion on the matter.

day253:
I very much agree with your point. I also think that the maintenance cost is too high. In a project I participated in before, vcpkg has a complete and independent pipeline to solve this problem. The reason why I submitted this merge request is that if the language of this document page is switched, the complete directory structure cannot be seen even if there is still an English part. Nowadays, translation is so convenient, so it is better to only support English. Although I am willing to follow up and maintain this translation, I cannot track the submission of every document. Unless translation is a strict part of the document submission check, all languages need to be modified simultaneously for each modification. In addition, I found that the document links on this homepage are not completely  #maintained. Some links are invalid, and the homepages in three languages are not synchronized. My more important need is to add more usage instructions in the document or wiki, so translation is not the core issue.

> Actually I just realized that there is only one other localized document in here, in Spanish, which seems not to have been updated in some time (6 years, against 9 months for the original English document) which leads me to wonder whether it's worth it to localize at all.
> 
> I'd say it's only worth it if you can guarantee that you or someone else will maintain the Chinese localization up-to-date with the English version to the best of your ability. If this is a one-time contribution, it is not worth it as it will only end up worsening the quality/reliability of the documentation over time.
> 
> Also I am not convinced that _all_ of the markdown documents within the repository need localization. I would assume only the main ones need to. But I'll let others share their opinion on the matter.

PR #2033, opened 2025-02-18
use https if proxied

swills:

I have athens deployed in k8s with the service on http and an ingress on https. The home page is displaying an http URL, which is incorrect. Not a huge deal, but I *think* this will fix it. Haven't tested yet, wanted to run this by you and see if it makes sense before going to all the trouble of setting up the builds and testing, since it's a simple one liner.

## How is the fix applied?

I think (hope) all that's needed is this one line change.

I didn't make an issue for this, but I could if you want.


matt0x6F:
The intent behind this feature was to completely replace the template for the home page, although, I kind of think we shouldn't have included a default because it can be confusing.

PR #2043, opened 2025-04-04
Add handling for SIGCHLD to cleanup child processes

jorng:

Athens is constantly spawning child processes to run the go toolchain for module management and downloading. The go toolchain also spawns child processes for git, which spawns child processes to use ssh, when appropriate.

Occasionally, things go wrong with these child processes, which can then end up as zombie processes. Currently, these zombies are not being reaped, causing significant resource leakage over extended times. This is especially problematic in shared resource environments, like Kubernetes.

## How is the fix applied?

This change adds handling for the `SIGCHLD` signal, waiting on the child processes to cleanup their resources. It also updates the shutdown signal handling to make use of NotifyContext.

None

### DIFF
--- a/internal/shutdown/signals.go
+++ b/internal/shutdown/signals.go
@@ -3,8 +3,13 @@
 package shutdown
 
 import (
+	"context"
+	"errors"
 	"os"
+	"os/signal"
 	"syscall"
+
+	"github.com/sirupsen/logrus"
 )
 
 // GetSignals returns the appropriate signals to catch for a clean shutdown, dependent on the OS.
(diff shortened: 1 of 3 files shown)


internal/shutdown/signals.go
```diff
@@ -13,3 +18,42 @@ import (
...
+	done, cancel := context.WithCancel(context.WithoutCancel(ctx))
+	go func() {
+		defer cancel()
+		for {
+			select {
+			case <-ctx.Done():
+				reap(logger)
+				return
+			case <-sigChld:
+				reap(logger)
```
nrwiersma, line 40: Nitpick: Would be good to add an info log line here. If there are no processes to reap, there is no sign that the reaping happened.

nrwiersma (commented):
1 minor nitpick, otherwise looks good

internal/shutdown/signals.go
jorng: Logging happens in the `reap` function for every process reaped. Not sure what else you are looking for.

nrwiersma: I am aware, but if you signal the process and no children are reaped, there is no indication anything happened.

jorng: We could add a log item, but it would just be adding **a lot** of noise for no purpose. This is not a signal one normally sends manually to the process. The kernel sends this to the process any time a child (or sub-child) process state change, so there will be a **huge** number of these signals that were likely already handled properly, and just filling the logs with noise for no purpose.

If you want assurance that it is running, you'll probably have to run long enough and with enough traffic that some zombies will appear. If it's of any reassurance, I'm already running this patch at my company (with **huge** amounts of traffic) and it works great, and has solved some significant resource issues (which is how I discovered this needed to be added).

If you really want me to add an additional log line, I can get my legal process started to approve the change. It may be a while.

nrwiersma (approved):

PR #2050, opened 2025-06-20
Move Single-flight from Fetcher to Stasher

south-mer:

This issue addresses some cache corruption issue within Athens when there are multiple requests in-flight to a package that is not yet cached.

When multiple requests are sent to the same package, each request calls `stasher.Stash`, and each call `fetcher.Fetch`.

The implementation of `goGetFetcher.Fetch` uses [singleflight.Group](https://pkg.go.dev/golang.org/x/sync/singleflight#Group) to prevent fetching the same package multiple times. As such, it returns a shared `storage.Version` for each `stasher.Stash`, which goes on to call `storage.Save`.

The problem here, is that the `Zip io.ReadCloser` within `storage.Version` is shared between the 2 calls to `storage.Save`. When the storage implementation tries to use the `io.ReadCloser` through e.g. `io.Copy`, the buffer is being consumed by both `storage.Save` calls and results in corrupt cache.

## How is the fix applied?

Move the single-flight mechanism to `stasher.Stash`: it returns only the semver which is safe to share, and the files can be safely concurrently read from the storage provider.

No such issues; we have encountered corrupted cache while running in production.

### DIFF
--- a/pkg/module/go_get_fetcher.go
+++ b/pkg/module/go_get_fetcher.go
@@ -14,7 +14,6 @@ import (
 	"github.com/gomods/athens/pkg/observ"
 	"github.com/gomods/athens/pkg/storage"
 	"github.com/spf13/afero"
-	"golang.org/x/sync/singleflight"
 )
 
 type goGetFetcher struct {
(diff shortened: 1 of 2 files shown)


nrwiersma (approved):
LGTM

Ullaakut (approved):

south-mer:
Thank you for the reviews. Could you let me know what the next step would be here?

PR #2053, opened 2025-06-23
chore: update workflows for consistency

nrwiersma:

This updates the GitHub Workflows to be consistent and take over from AppVoyer, which seems to be down and not coming back.

## How is the fix applied?

Update the workflow files, added a build step in testing.


DrPsychick (approved):
LGTM 🎉

Ullaakut (approved):

PR #2108, opened 2026-04-17
fix: use default config for config files

nrwiersma:

This uses the default config as the basis for config files, so new options being added get auto-defaulted in existing installations.

## How is the fix applied?

Start with a default, go from there.

Fixes #


Ullaakut (approved):
Code SGTM, didnt read config file in testdata though

PR #2109, opened 2026-04-17
feat: refactor config package

nrwiersma:

The `config` package is inconsistent and difficult to parse. This PR attempts to refactor it to be more consistent and idiomatic.

## How is the fix applied?

The general sense of the change is:
* Config structs are grouped by purpose
* Config structs are renamed to avoid conflicts and make meaning more apparent
* Functions and structs are grouped more idiomatically


pkg/config/config_test.go
```diff
@@ -14,58 +14,7 @@ import (
...
-	err := envOverride(conf)
-	if err != nil {
-		t.Fatalf("Env override failed: %v", err)
-	}
-	expPort := ":3000"
-	if conf.Port != expPort {
-		t.Errorf("Port was incorrect. Got: %s, want: %s", conf.Port, expPort)
-	}
-}
+var testFile = filepath.Join("testdata", "config.dev.toml")
```
Ullaakut, line 17: Why use filepath.Join for this instead of just making this a const defined as `testdata/config.dev.toml` ?

pkg/config/storage.go
```diff
@@ -2,11 +2,71 @@ package config
...
+}
+
+// MongoStorage specifies the properties required to use MongoDB as the storage backend.
+type MongoStorage struct {
+	URL                   string `envconfig:"ATHENS_MONGO_STORAGE_URL" validate:"required"`
+	DefaultDBName         string `default:"athens"                     envconfig:"ATHENS_MONGO_DEFAULT_DATABASE"`
+	DefaultCollectionName string `default:"modules"                    envconfig:"ATHENS_MONGO_DEFAULT_COLLECTION"`
+	CertPath              string `envconfig:"ATHENS_MONGO_CERT_PATH"`
+	InsecureConn          bool   `envconfig:"ATHENS_MONGO_INSECURE"`
+}
```
Ullaakut, line 44: Wouldnt it be more readable to have the defaults second so that the envconfig tags align? Is it because of a linter or smth that they are ordered they way they are now?

pkg/config/storage.go
```diff
@@ -2,11 +2,71 @@ package config
...
+	Key                                string `envconfig:"AWS_ACCESS_KEY_ID"`
+	Secret                             string `envconfig:"AWS_SECRET_ACCESS_KEY"`
+	Token                              string `envconfig:"AWS_SESSION_TOKEN"`
+	Bucket                             string `envconfig:"ATHENS_S3_BUCKET_NAME"                  validate:"required"`
+	UseDefaultConfiguration            bool   `envconfig:"AWS_USE_DEFAULT_CONFIGURATION"`
+	ForcePathStyle                     bool   `envconfig:"AWS_FORCE_PATH_STYLE"`
+	CredentialsEndpoint                string `envconfig:"AWS_CREDENTIALS_ENDPOINT"`
+	AwsContainerCredentialsRelativeURI string `envconfig:"AWS_CONTAINER_CREDENTIALS_RELATIVE_URI"`
+	Endpoint                           string `envconfig:"AWS_ENDPOINT"`
+}
```
Ullaakut, line 58: Sometimes I like `tagalign`, here I hate it lmao

pkg/storage/minio/minio_test.go
```diff
@@ -29,12 +30,12 @@ func TestNewStorageExists(t *testing.T) {
...
 
 	for _, test := range tests {
-		backend, err := NewStorage(&config.MinioConfig{
+		backend, err := NewStorage(&config.MinioStorage{
 			Endpoint: url,
 			Key:      "minio",
 			Secret:   "minio123",
 			Bucket:   test.name,
-		}, config.GetTimeoutDuration(300))
+		}, 300*time.Second)
```
Ullaakut, line 38: `5 * time.Minute` seems clearer to me since on a quick read this could look like 30s, but no big deal.

Ullaakut (approved):
Nitpicks. So much insane stuff left in those packages though, no idea how much work it would be to bring everything up to standard.

PR #2111, opened 2026-04-19
add chinese translations for docs

sufu777:

chineses translations for docs

## How is the fix applied?

Translated by minimax v2.7 and manually proofread.


matt0x6F (changes requested):
Hello! Thank you for contribution to Athens. We are not a CNCF project so we'd need to fix that before continuing with this change.

PR #2120, opened 2026-05-12
fix: support external_account workload identity, add tests

### LINKED ISSUE #2119: `athens@0.17.0` breaks support of workload identity for GCP
**Describe the bug**
`athens@0.17.0` breaks support of workload identity for GCP.

[The change](https://github.com/gomods/athens/pull/2105/changes#diff-06a9b71403d0771f1b627dbf1ac83575d74afea43fecb614b399809a409d4d8a) was well hidden in Golang Update PR #2105 without any proper release note.

**Error Message**
```
INFO[10:33AM]: Exporter not specified. Traces won't be exported
FATAL[10:33AM]: Could not create App	error=getting storage configuration: could not get GCS credentials: google: expected credential type "service_account", found "external_account"
```

**Expected behavior**
Athens supporting workload identity as it was before.
Additionally, please avoid breaking changes without release notes.

DrPsychick:

Resolve #2119

## How is the fix applied?

Support all credential types, cover `service_account` and `external_account` with tests.

Fixes #2119

### DIFF
--- a/pkg/storage/gcp/gcp.go
+++ b/pkg/storage/gcp/gcp.go
@@ -59,22 +59,13 @@ func newClient(ctx context.Context, gcpConf *config.GCPConfig, timeout time.Dura
 		}
 
 		var keyType struct {
-			Type string `json:"type"`
+			Type google.CredentialsType `json:"type"`
 		}
 		if err := json.Unmarshal(key, &keyType); err != nil {
 			return nil, errors.E(op, fmt.Errorf("failed to parse JSON key: %w", err))
 		}
-		var credType google.CredentialsType
-		switch keyType.Type {
-		case "service_account":
-			credType = google.ServiceAccount
-		case "external_account":
-			credType = google.ExternalAccount
-		default:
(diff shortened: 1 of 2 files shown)


pkg/storage/gcp/gcp.go
```diff
@@ -56,7 +57,24 @@ func newClient(ctx context.Context, gcpConf *config.GCPConfig, timeout time.Dura
...
+
+		var keyType struct {
+			Type string `json:"type"`
+		}
+		if err := json.Unmarshal(key, &keyType); err != nil {
+			return nil, errors.E(op, fmt.Errorf("failed to parse JSON key: %w", err))
+		}
+		var credType google.CredentialsType
+		switch keyType.Type {
+		case "service_account":
```
matt0x6F: Should we use the actual type from the package here?

matt0x6F (approved):
Gave a nit-pick, otherwise looks good

pkg/storage/gcp/gcp.go
DrPsychick: good point. I gave it another shot, now supporting all types that would work.

Issue #2121, opened 2026-05-13
Add cluster-mode client support for Redis singleflight


leerieo:
**Is your feature request related to a problem? Please describe.**
We are currently seeing recurring 500 Internal Server Error from Athens on the testing cluster's cold-cache module fetches by switching its singleflight Redis from Cluster Mode Enabled to Cluster Mode Disabled. Athens v0.17.0 singleflight uses a plain redis.NewClient ([github.com/go-redis/redis](https://github.com/redis/go-redis)) that can't follow MOVED redirects from a Cluster-Mode-Enabled ElastiCache.

```
ERROR[8:48PM]: MOVED 5288 athens-redis-testing-testing-rg-0001-001.athens-redis-testing-testing-rg.lv1j0z.use1.cache.amazonaws.com:6379
  http-path=/github.com/teambition/rrule-go/@v/v1.8.2.info
  kind=Internal Server Error
  operation=download.InfoHandler
  ops=[download.InfoHandler pool.Info protocol.Info protocol.processDownload redis.Stash]
```

**Describe the solution you'd like**
We would like Athens single flight to support Redis Cluster mode. 

**Describe alternatives you've considered**
We could disable redis cluster mode for our cache but this will worsen our reliability since single node means if the node goes down we have failing wfs until a new node spins up 

**Additional context**

leerieo:
Have a draft pr im working on for this https://github.com/gomods/athens/pull/2122

leerieo:
PR is ready for review https://github.com/gomods/athens/pull/2122

PR #2122, opened 2026-05-13
stash/redis: add cluster-mode client support for singleflight

leerieo:

Athens v0.17.0 singleflight uses redis.NewClient, which can't follow MOVED/ASK redirects. When the configured endpoint points at a Cluster Mode Enabled Redis (e.g. AWS ElastiCache's `clustercfg.*` configuration endpoint), every Stash op that happens to land on a non-slot-owning node returns MOVED, which the client surfaces as 500 to the caller — even when the cluster has only one shard, because Cluster Mode replies MOVED any time the connected node doesn't own the slot for the key.

## How is the fix applied?

Add a `Cluster` bool to the SingleFlight.Redis config (env override: ATHENS_REDIS_CLUSTER). When set, WithRedisLock constructs a redis.NewClusterClient and stores it in the redisLock as a redis.UniversalClient — both go-redis client types satisfy the bsm/redislock RedisClient interface, so the Stash/Release call sites need no further change. The endpoint may be a single host:port, redis[s]:// URL, or a comma-separated list of seed addresses; TLS, username, and embedded passwords are honored the same way the single-node path honors them.

Backwards compatible: existing configs default to Cluster = false and keep using redis.NewClient.

Tests:
- new env-gated TestWithRedisClusterLock (REDIS_CLUSTER_TEST_ENDPOINT) exercising the cluster-client code path against a real cluster
- new Test_getRedisClusterClientOptions table-test covering host:port, multi-seed, rediss://, URL-embedded password validation
- existing single-node tests pass through `false` for the new arg

Stood up grokzen/redis-cluster:6.2.0 locally (6 nodes, ports 7000-7005), confirmed cluster formation with cluster_state:ok and slots 16384/16384 ok

```
docker run -d --name redis-cluster  -e IP=0.0.0.0  -p 7000-7005:7000-7005 grokzen/redis-cluster:6.2.0  
docker exec redis-cluster redis-cli -p 7000 cluster info | head -3
# expect: cluster_state:ok

REDIS_CLUSTER_TEST_ENDPOINT=localhost:7000 go test -v -run TestWithRedisClusterLock ./pkg/stash/...
REDIS_CLUSTER_TEST_ENDPOINT=localhost:7000 go test -v -run TestWithRedisLock ./pkg/stash/...

```

Fixes #

#2121

### DIFF
--- a/cmd/proxy/actions/app_proxy.go
+++ b/cmd/proxy/actions/app_proxy.go
@@ -155,6 +155,7 @@ func getSingleFlight(l *log.Logger, c *config.Config, s storage.Backend, checker
 			&athensLoggerForRedis{logger: l},
 			c.SingleFlight.Redis.Endpoint,
 			c.SingleFlight.Redis.Password,
+			c.SingleFlight.Redis.Cluster,
 			checker,
 			c.SingleFlight.Redis.LockConfig)
 	case "redis-sentinel":
--- a/config.dev.toml
+++ b/config.dev.toml
@@ -344,12 +344,23 @@ ShutdownTimeout = 60
         # Endpoint is the redis endpoint for a SingleFlight lock. Should be either a host:port
         # pair or redis url such as:
         #    redis[s]://user:password@127.0.0.1:6379/0?protocol=3
-        # TODO(marwan): enable multiple endpoints for redis clusters.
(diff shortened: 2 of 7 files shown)


DrPsychick (approved):
Thank you for your contribution, LGTM!

PR #2127, opened 2026-06-13
Replace logrus with log/slog

### LINKED ISSUE #1890: Replace logrus with log/slog

matt0x6F:

#1890: replace logrus with the standard library's `log/slog`.

This picks up where #2010 left off. That PR tried to convert the whole repo at once and stalled on the level/severity mapping. Rather than revive it, I went with @marwan-at-work's suggestion from that review: only change `pkg/log` and keep the `Entry` interface identical, so importers don't have to change.

## How is the fix applied?

- `Entry` is unchanged. The slog implementation lives in a new `entry_slog.go`, and the printf methods just wrap slog with `fmt.Sprintf`, so the existing call sites compile as-is.
- `pkg/errors` `Severity` moves from `logrus.Level` to `slog.Level`. slog's zero value is `Info` (a real level), so instead of relying on the zero value the way the logrus code did, there's an explicit "unset" sentinel. This is the bit that tripped up #2010.
- The colored dev output is kept as a custom slog handler. GCP still emits `severity`/`message`/`timestamp`, so Cloud Logging routing is unaffected.
- `ParseLevel` maps the legacy logrus level names (`trace`, `warning`, `fatal`, `panic`) onto slog levels so existing config keeps working.
- logrus drops from a direct dependency to indirect. It's still pulled in transitively by etcd's embedded server, so it can't leave `go.mod` entirely without dropping etcd support.

The config knobs (`ATHENS_LOG_LEVEL`, `ATHENS_LOG_FORMAT`, `ATHENS_CLOUD_RUNTIME`) behave the same and GCP routing is intact. The non-GCP JSON keys now follow slog's native shape rather than logrus's.

Fixes #1890

### DIFF
--- a/cmd/proxy/actions/app.go
+++ b/cmd/proxy/actions/app.go
@@ -82,7 +82,7 @@ func App(logger *log.Logger, conf *config.Config) (http.Handler, func(), error) 
 		conf.GoEnv,
 	)
 	if err != nil {
-		logger.Info(err)
+		logger.Infof("%v", err)
 	} else {
 		cleanupTraces = flushTraces
 	}
@@ -93,7 +93,7 @@ func App(logger *log.Logger, conf *config.Config) (http.Handler, func(), error) 
 	cleanupStats := noop
 	flushStats, err := observ.RegisterStatsExporter(r, conf.StatsExporter, Service)
 	if err != nil {
-		logger.Info(err)
+		logger.Infof("%v", err)
 	} else {
 		cleanupStats = flushStats
 	}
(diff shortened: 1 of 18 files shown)


matt0x6F:
This was written with Claude and supervised by me. I did my best to address everyone's critiques from the previous iteration of this PR. I also managed to follow a line of implementation that fully extracts logrus from the SBOM.

nrwiersma (approved):

Issue #1881, opened 2023-09-05
Athens does not calculate the correct checksum


idefixcert:
**Describe the bug**
Maybe this is not the right forum, but I am not sure where to ask else.

I use a docker athene container and another fresh golang container.
In the golang container I setup athene container as goproxy.
when I run `go install github.com/mgechev/revive@latest` I see an error message
```
ERROR[9:17AM]: github.com/mgechev/revive@v1.3.3: verifying module: checksum mismatch
	downloaded: h1:H5KpQYVmht69pg47kbIg/E4S1jA1yfFBEwy7Zv3uDX8=
	sum.golang.org: h1:GUWzV3g185agbHN4ZdaQvR6zrLVYTUSA2ktvIinivK0=
```
but if I do not use athene as a proxy I can't see this error.

**Error Message**
If applicable add error message to help explain your problem or console output enclose in triple back-ticks eg,

<pre>
```console
INFO[9:17AM]: incoming request	http-method=GET http-path=/github.com/mgechev/@v/list http-status=404 request-id=5a1e4cb1-5b83-4a29-ba1e-bf97198a25c1
INFO[9:17AM]: incoming request	http-method=GET http-path=/github.com/mgechev/revive/@v/list http-status=200 request-id=468b8f6c-4062-4b86-87d1-6422e58b9936
ERROR[9:17AM]: github.com/mgechev/revive@v1.3.3: verifying module: checksum mismatch
	downloaded: h1:H5KpQYVmht69pg47kbIg/E4S1jA1yfFBEwy7Zv3uDX8=
	sum.golang.org: h1:GUWzV3g185agbHN4ZdaQvR6zrLVYTUSA2ktvIinivK0=

SECURITY ERROR
This download does NOT match the one reported by the checksum server.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.
```
</pre>

**To Reproduce**
```
docker run --rm -p 3000:3000 gomods/athens:v0.12.1

docker run --rm -it golang:latest
export GOPROXY="http://172.17.0.1:3000/";
go install github.com/mgechev/revive@latest
```

**Expected behavior**
Same behavior with and without proxy.

**Environment (please complete the following information):**
 - OS: ubuntu with the docker containers
 - Go version : 1.21
 - Proxy version : v0.12.1

bozaro:
Same issue with github.com/homeport/dyff@v1.5.7 package (without proxy module downloaded succesfully):
```
[...]

Issue #1470, opened 2019-11-22
sumdb problem: different behavior with athens:0.7 and proxy.golang.org downloading module


pjediny:
Here is what I did,

clean local cache
tried to get the fabric module via athens proxy (fail)
tried to get the fabric module via https://proxy.golang.org (ok)
tried to get the fabric module via athens proxy (ok, it's already stored locally)
```
+ go mod init test
go mod init: go.mod already exists
+ go clean -modcache
+ GOPROXY=http://localhost:3333
+ go get github.com/hyperledger/fabric@v1.4.4
go: github.com/hyperledger/fabric@v1.4.4: reading http://localhost:3333/github.com/hyperledger/fabric/@v/v1.4.4.mod: 500 Internal Server Error
+ GOPROXY=https://proxy.golang.org
+ go get github.com/hyperledger/fabric@v1.4.4
go: finding github.com/hyperledger/fabric v1.4.4
go: downloading github.com/hyperledger/fabric v1.4.4
go: extracting github.com/hyperledger/fabric v1.4.4
+ GOPROXY=http://localhost:3333
+ go get github.com/hyperledger/fabric@v1.4.4
```

the Athens proxy logs this error:
```
time="2019-11-22T16:29:02Z" level=error msg="exit status 1: go: finding github.com/hyperledger/fabric v1.4.4\nverifying github.com/hyperledger/fabric@v1.4.4: checksum mismatch\n\tdownloaded: h1:2m+tDXB5QWyZL1UTyYs3fxUhkgvKD88yVE6sIwYf1KM=\n\tsum.golang.org: h1:Joa6eO9HEGnzcuZF5RD+dZBPeYqxGF+ehYb7OSs3glY=\n\nSECURITY ERROR\nThis download does NOT match the one reported by the checksum server.\nThe bits may have been replaced on the origin server, or an attacker may\nhave intercepted the download attempt.\n\nFor more information, see 'go help module-auth'.\n" http-method=GET http-path=/github.com/hyperledger/fabric/@v/v1.4.4.info http-url=/github.com/hyperledger/fabric/@v/v1.4.4.info kind="Internal Server Error" module=github.com/hyperledger/fabric operation=download.InfoHandler ops="[download.InfoHandler pool.Info protocol.Info protocol.processDownload stash.Pool stasher.Stash stasher.fetchModule goGetFetcher.Fetch module.downloadModule]" version=v1.4.4
```

[...]

Issue #1080, opened 2019-02-21
Incompatible checksums on Athens v0.2.0 docker with go.sum generated by Go >= 1.11.4


machiel:
**Describe the bug**
My go.sum is generated using Go 1.11.5, however the Athens v0.2.0 image I pull from docker is running on Go 1.11.2. There has been a fix in 1.11.4 affecting the way checksums are calculated for modules, so now when pulling my dependencies from Athens there's a checksum mismatch, since Athens still uses the 1.11.2 way.

More about this here: https://github.com/golang/go/issues/27925

**Error Message**
```
go: verifying github.com/kylelemons/go-gypsy@v0.0.0-20160905020020-08cad365cd28: checksum mismatch
     downloaded: h1:SzHLOWB819X9aUXh2ct6krDIoIujoTsj5J1rDMjkRyo=
     go.sum:     h1:mkl3tvPHIuPaWsLtmHTybJeoVEW7cbePK73Ir8VtruA=
```

**To Reproduce**
Add for example `kylelemons/go-gypsy` to your go.mod file using go 1.11.5, then try building your application using the Athens v0.2.0 docker image running go 1.11.2.

When rebuilding the image with go 1.11.5 everything works as expected.

**Expected behavior**
To work with go 1.11.4 and higher generated go.sum's. A rebuild of the current docker file should be enough as it relies on 1.11, which is currently an alias for go 1.11.5

**Environment (please complete the following information):**
 - OS: [e.g. Linux 64bit] Linux 64bit
 - Go version : 1.11.2
 - Buffalo Version : -
 - Proxy version : v0.2.0
 - Storage (fs/mongodb/s3 etc.) : fs

marwan-at-work:
@machiel thanks for reporting this. The `canary` version of the docker image uses `go1.12-rc` which should be compatible with 1.11.5. Feel free to give this a try and let us know if the issue still persists. The next releas `v0.3.0` will have 1.12 on by default. 

Also, don't forget to delete that module from your storage backend as it was compiled with pre1.11.5 and will need to be re-compiled with the new `go mod download` command. 

Thanks!

machiel:
Perfect, thanks! I will await the `0.3.0` release, and in the mean time use the canary version.

Thanks!

arschles:
@machiel just FYI, the [v0.3.1](https://github.com/gomods/athens/releases/tag/v0.3.1) release is out. I recommend using that instead of v0.3.0 because it has a fix for the minio storage driver in it

[...]

PR #2143, opened 2026-06-30
fix(config): default GoBinaryEnvVars to proxy.golang.org, not direct-only

karlkfi:
## What

Change the default `GoBinaryEnvVars` from `["GOPROXY=direct"]` to `["GOPROXY=https://proxy.golang.org,direct"]`.

## Why

With the current default, Athens' internal `go` subprocess fetches each module from its origin VCS and **rebuilds the module zip locally**. When the Go toolchain bundled in the Athens image differs from the one that produced a module's notarized zip, the rebuilt zip hashes differently, and Athens rejects the **valid** module:

```
github.com/onsi/ginkgo/v2@v2.32.0: verifying module: checksum mismatch
	downloaded: h1:OPshXBha3Pavij2LzHWXX30meRBGIucLwh97bHLjLpk=
	sum.golang.org: h1:Hw7s2pVrQo/8Yz5N77qdnpHaoc+c6cC9WIV1Jce+J6E=
SECURITY ERROR
This download does NOT match the one reported by the checksum server.
```

The client just sees a `404`. This is a recurring, hard-to-diagnose failure for ordinary public modules — e.g. a module declaring a `go` directive newer than the image's bundled Go. Prior reports: #1881, #1080, #1470.

Concretely reproduced on `gomods/athens:v0.15.1` (bundles Go 1.20.14) fetching `github.com/onsi/ginkgo/v2@v2.32.0` (declares `go 1.25`). Cross-checked that `proxy.golang.org`, `sum.golang.org`, a clean direct-VCS fetch with a current Go, and the module's `go.sum` **all agree** on `h1:Hw7s2pVrQo...` — only Athens' local rebuild diverged. The module is not tampered; the rebuild is.

## The change

Defaulting to `GOPROXY=https://proxy.golang.org,direct` makes Athens fetch the **immutable, notarized** zip from the public proxy, whose hash matches `sum.golang.org`, so the checksum check passes. The `,direct` fallback still resolves modules the proxy doesn't mirror. This matches Go's own default `GOPROXY` and eliminates the whole class of toolchain-skew checksum failures.

Operators who specifically need origin-only fetches (air-gapped / compliance) can still opt back in explicitly:

```toml
GoBinaryEnvVars = ["GOPROXY=direct"]
```

## Testing

- Updated `config.dev.toml`, the `pkg/config/testdata/config.dev.toml` fixture, and the `TestParseExampleConfig` assertion in lockstep so the `defaultConfig()` ⇄ example-config consistency test still holds.
- `go test ./pkg/config/... ./pkg/middleware/... ./pkg/module/... ./pkg/download/...` all pass.

### DIFF
--- a/config.dev.toml
+++ b/config.dev.toml
@@ -43,7 +43,7 @@ GoEnv = "development"
 # ATHENS_GO_BINARY_ENV_VARS='GODEBUG=true'
 # Then the final value that the Go binary will receive is ["GODEBUG=true"] and NOT ["GOPROXY=direct", "GODEBUG=true"]
 # Therefore, whether you use the config file or the env var, make sure you have all the values you need there.
-GoBinaryEnvVars = ["GOPROXY=direct"]
+GoBinaryEnvVars = ["GOPROXY=https://proxy.golang.org,direct"]
 
 # GoGetWorkers specifies how many times you can concurrently
 # go mod download, this is so that low performance instances
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -149,7 +149,7 @@ func Load(configFile string) (*Config, error) {
 func defaultConfig() *Config {
 	return &Config{
 		GoBinary:              "go",
-		GoBinaryEnvVars:       EnvList{"GOPROXY=direct"},
+		GoBinaryEnvVars:       EnvList{"GOPROXY=https://proxy.golang.org,direct"},
 		GoEnv:                 "development",
 		GoGetWorkers:          10,
 		ProtocolWorkers:       30,
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -303,7 +303,7 @@ func TestParseExampleConfig(t *testing.T) {
 		TraceSamplingFraction: 1.0,
 		StatsExporter:         "prometheus",
 		SingleFlightType:      "memory",
-		GoBinaryEnvVars:       []string{"GOPROXY=direct"},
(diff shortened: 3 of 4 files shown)


matt0x6F:
I think this is probably the wrong way to solve this particular problem.

1. a number of Athens' users can't access the public module proxy
2. this is an abrupt breaking change
3. the hash calculation is public and reproducible

We should probably seek to fix whatever is wrong with the hash calculation.

#1470 is where someone double published a tag. A lot of maintainers don't know that the proxy (and sumdb) see their version on release and that they can't simply delete published versions to issue a fix.

#1080 was a change in the way _Go_ calculated checksums, not Athens, and would exist with your change.

#1881 is the exact same issue as #1470.

In the future we recommend starting with an issue so that we can guide you in the right direction. Second, I'd greatly appreciate doing some research — to include scanning maintainer comments of the issues you cited.

karlkfi:
My mistake. I didn't realize `,` only tolerated HTTP 404 or 410. I guess we would need to use `|` to tolerate any error and fall through to direct in air-gapped or firewalled environments.

But I can also see how you might not want to add a failed GET to existing user's flows, even if it did fail gracefully.

That said, your responses to https://github.com/gomods/athens/issues/1881 point at a re-pushed tag being the smoking gun, which doesn't seem to be the case in the ginkgo problem I encountered here. This seems like a different root cause entirely.

After further analysis, and referencing some more recent reports, here is a more in-depth bug report: https://github.com/gomods/athens/issues/2145

PR #2154, opened 2026-08-25
Make Cache-Control on the info, mod and zip endpoints configurable

### LINKED ISSUE #1183: Allow cache configurability on the mod, zip and info files

These 3 endpoints are technically cachable forever because Athens is supposed to be immutable. Currently the athens.azurefd.net proxy has a caching proxy in front of it that honors the no-cache headers that Athens sends back for these endpoints.

In the past, Athens has failed to download the entire zip file, or failed to transmit it to the proxy for one reason or another. I haven't been able to fully investigate the issue so don't have intimate details on what happened.

Additionally, when the upgrade from go 1.11.x to 1.12.x happened, checksums changed. Even if I deleted all the modules in Athens's storage, they would still be cached in the proxy forever, so that action would have been worthless.

The end result is that the proxy serves broken or old data forever, unless I manually go purge it. I currently do this if I notice a problem happening over and over. When I do purge the cache, Athens experiences a huge spike in load.

Rather than do the manual purge and experience the resulting surge, I'd like to be able to configure Athens to return timeouts from the `Cache-Control` header that it sends back to the proxy. By doing this, mod, zip and info endpoints for a given module would periodically be hit to refresh the proxy's cache.

I could write a script to periodically purge module data in the proxy's cache.

**Additional context**

None

ChrisJr404:
Fixes #1183.

The info, mod and zip endpoints serve immutable module content but Athens sends no Cache-Control header on them today, so a caching proxy out front either caches them forever or not at all. As #1183 describes, caching forever means a broken or stale zip can get pinned in the proxy until it's manually purged, and purging causes a load spike.

This adds a `ContentCacheControl` config option (env `ATHENS_CONTENT_CACHE_CONTROL`) that sets the Cache-Control header on those three endpoints. Setting something like `public, max-age=300` lets the proxy cache them and refresh periodically instead. It's off by default (empty value, no header) so existing setups don't change.

I reused the existing `middleware.CacheControl` that already powers the no-cache header on the list/latest endpoints, so it takes a raw header value. Added a test that checks the header shows up on info/mod/zip when configured and stays off when it isn't, plus the config wiring and docs in config.dev.toml.

### DIFF
--- a/cmd/proxy/actions/app_proxy.go
+++ b/cmd/proxy/actions/app_proxy.go
@@ -122,7 +122,7 @@ func addProxyRoutes(
 
 	dp := download.New(dpOpts, addons.WithPool(c.ProtocolWorkers))
 
-	handlerOpts := &download.HandlerOpts{Protocol: dp, Logger: l, DownloadFile: df}
+	handlerOpts := &download.HandlerOpts{Protocol: dp, Logger: l, DownloadFile: df, ContentCacheControl: c.ContentCacheControl}
 	download.RegisterHandlers(r, handlerOpts)
 
 	return nil
--- a/config.dev.toml
+++ b/config.dev.toml
@@ -341,6 +341,14 @@ IndexType = "none"
 # Env override: ATHENS_SHUTDOWN_TIMEOUT
 ShutdownTimeout = 60
 
+# ContentCacheControl is the Cache-Control header sent on the info, mod and zip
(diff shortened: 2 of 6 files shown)


ChrisJr404:
Closing this one. I opened #2153 for the same issue at almost the same time by mistake, so that PR already covers it. Sorry for the noise.

PR #2162, opened 2026-09-15
pkg/stash: add S3 singleflight lock backed by conditional writes

royreznik:
## What

Adds an `s3` `SingleFlightType`, so multiple Athens instances sharing one S3 bucket can coordinate module saves without running redis or etcd.

## How

- While a module version is being saved, the instance holds a small object at `lock/<module>@<version>` in the storage bucket, created with `If-None-Match: *`. Exactly one writer wins; the others poll (1s interval, bounded by `Timeout`/`MaxRetries`), then find the module already in storage via the `Checker` and serve it. Same shape as the redis and azureblob wrappers: acquire → `Exists` → inner `Stash` → release.
- A lock older than `TTL` is treated as left behind by a crashed instance and reclaimed with `If-Match: <observed ETag>`, so at most one waiter takes it over. Each lock body is a fresh UUID so every acquisition has a distinct ETag.
- Release is a `DeleteObject` with `If-Match` on our own ETag, on a context that ignores the caller's cancellation so aborted requests still free their lock. A lock reclaimed by someone else is left alone.
- `storage/s3.New` is split so `NewClient` can be reused by the lock; behaviour of the storage backend is unchanged.
- Config under `[SingleFlight.S3]` (`TTL`, `Timeout`, `MaxRetries`, same defaults as redis) with env overrides `ATHENS_S3_LOCK_TTL`, `ATHENS_S3_LOCK_TIMEOUT`, `ATHENS_S3_LOCK_MAX_RETRIES`. `config.dev.toml`, testdata and `docs/content/configuration/storage.md` updated.
- Lock keys never end in `.info` and never sit under `<module>/@v`, so `Catalog` and `List` ignore them.

## Tests

`pkg/stash/with_s3_test.go` runs against the MinIO service the `storage/s3` tests already use (`ATHENS_MINIO_ENDPOINT`), in its own bucket so the storage tests wiping theirs cannot race it:

- `TestWithS3Lock` – 5 concurrent stashes, only the first hits the inner stasher, lock object is gone afterwards.
- `TestWithS3LockHeld` – a live foreign lock blocks the stash and the inner stasher never runs.
- `TestWithS3LockStale` – a foreign lock older than `TTL` is reclaimed and released.

Verified locally against MinIO `RELEASE.2025-09-07` with `-race -count=5`; `golangci-lint v2.11.4` reports 0 issues.

## Notes

- Requires a store that honours `If-None-Match` on `PutObject` (Amazon S3, recent MinIO). Docs say so.
- MinIO currently ignores `If-Match` on `DeleteObject` (verified: delete with a wrong ETag succeeds). On MinIO the release is therefore an unconditional delete; on Amazon S3 it is conditional as intended. Only affects the narrow case of a holder outliving its own `TTL`.
- Follows the redis lock semantics: no lock renewal, so `TTL` must exceed the longest expected module save (default 900s vs `StashTimeout` default 600s).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


royreznik:
The `test` job was failing on `main` before this PR: `docker pull minio/minio:edge-cicd` is denied since MinIO images left Docker Hub. The fix is #2163; its commit is cherry-picked here so CI can run. I will drop it from this branch once #2163 lands.

matt0x6F:
Rebase once #2169 merges

matt0x6F (approved):
Looks good. The S3 wrapper follows the existing locking and configuration patterns. Lock writes now honor the configured encryption settings for both initial acquisition and stale-lock takeover, with HTTP-header regression coverage and tests for S3 lock configuration overrides.

PR #2163, opened 2026-09-15
ci: pull MinIO from quay.io and start it as a step

royreznik:
## What

`main` CI is red: the `test` job fails in "Initialize containers" with

```
Error response from daemon: pull access denied for minio/minio, repository does not exist or may require 'docker login'
```

MinIO images are no longer published on Docker Hub. This moves CI and `docker-compose.yml` to the official `quay.io/minio/minio` image.

## How

- The quay.io images have no default command (`Cmd` is just `minio`, the entrypoint only prepends the binary), and there is no `edge-cicd` tag there; `latest-cicd` behaves the same. A GitHub Actions `service` cannot pass a command, so MinIO now starts as a plain `docker run -d ... server /data` step with a health-check wait before the tests.
- Pinned to `RELEASE.2025-09-07T16-13-09Z` rather than `latest`.
- `docker-compose.yml` uses the same pinned image (it already passed `command: server /data`).
- Both switch from the legacy `MINIO_ACCESS_KEY` / `MINIO_SECRET_KEY` to `MINIO_ROOT_USER` / `MINIO_ROOT_PASSWORD`.

## Verification

Ran the exact `docker run` from the step locally and executed `go test -race ./pkg/storage/s3/ ./pkg/storage/minio/` against it: both pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/ci.yml
```diff
@@ -126,6 +118,21 @@ jobs:
...
+          docker run -d --name minio -p 9000:9000 \
+            -e MINIO_ROOT_USER=minio -e MINIO_ROOT_PASSWORD=minio123 \
+            quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z server /data
+          for _ in $(seq 1 30); do
+            curl -sf http://localhost:9000/minio/health/live && exit 0
+            sleep 1
+          done
+          echo "MinIO did not become healthy" >&2
+          docker logs minio >&2
+          exit 1
```
matt0x6F, line 134: There really needs to be a better way to do this, or we can use something other than minio since it's no longer FOSS

matt0x6F (changes requested):

matt0x6F:
I ended up replacing Minio with RustFS in #2169

